PR #17, opened 2019-07-16
Minor test and docs improvements for main OpenTelemetry module

indrekj:
View by commit.


fbogsany:
> It'll be easier with classes when we'll develop SDK part and subclass the default (no-op) tracer/meter, I suppose.

I don't think the SDK will subclass the default tracer/meter. I think we'll just rely on 🦆 typing and a decent conformance test suite. I'm not convinced about the change to use classes, except that it is a little less magical to Ruby newcomers (evidenced by the fact that I got it wrong when I wrote the initial version).

PR #18, opened 2019-07-16
DistributedContext API

fbogsany:
This PR provides the [`DistributedContext` API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/distributedcontext-api.md) and the [`BinaryFormat`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/propagators-api.md#binary-format) part of the `Propagators` API.

Note: `DistributedContext::Manager` is part of https://github.com/open-telemetry/opentelemetry-ruby/pull/17

### DIFF
--- a/api/lib/opentelemetry/distributed_context/distributed_context.rb
+++ b/api/lib/opentelemetry/distributed_context/distributed_context.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module DistributedContext
+    # An immutable implementation of the DistributedContext that does not contain any entries.
+    class DistributedContext
+      def entries
+        []
+      end
+
+      def [](_key)
+        nil
+      end
+    end
+  end
+end
--- a/api/lib/opentelemetry/distributed_context/distributed_context.rb
+++ b/api/lib/opentelemetry/distributed_context/distributed_context.rb
@@ -8,8 +8,12 @@ module OpenTelemetry
   module DistributedContext
     # An immutable implementation of the DistributedContext that does not contain any entries.
     class DistributedContext
+      EMPTY_ENTRIES = [].freeze
+
+      private_constant(:EMPTY_ENTRIES)
+
       def entries
-        []
+        EMPTY_ENTRIES
       end
 
       def [](_key)
--- a/api/lib/opentelemetry/distributed_context/entry.rb
+++ b/api/lib/opentelemetry/distributed_context/entry.rb
@@ -1,0 +1,68 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module DistributedContext
+    # An Entry consists of Entry::Metadata, Entry::Key, and Entry::Value.
+    class Entry
+      attr_reader :metadata, :key, :value
+
+      # Entry::Key is the name of the Entry. Entry::Key along with Entry::Value can be used to aggregate and group stats,
+      # annotate traces and logs, etc.
+      #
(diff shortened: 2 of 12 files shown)


api/lib/opentelemetry/distributed_context/entry.rb
```diff
@@ -0,0 +1,68 @@
...
+      # - Must have length greater than zero and less than 256.
+      # - Must not be empty.
+      class Key
+        attr_reader :name
+
+        def initialize(name)
+          # TODO: name should be printable ascii (32..126), not just ascii.
+          raise ArgumentError unless name&.ascii_compatible? && (1..255).include?(name.length)
+
+          @name = -name
```
indrekj: Why `-name`?

api/lib/opentelemetry/distributed_context/entry.rb
```diff
@@ -0,0 +1,68 @@
...
+        # An @see Entry with NO_PROPAGATION is considered to have local scope and is used within the process
+        # where it is created.
+        NO_PROPAGATION = 0
+
+        # An @see Entry with UNLIMITED_PROPAGATION can propagate unlimited hops. However, it is still subject
+        # to outgoing and incoming (on remote side) filter criteria.
+        UNLIMITED_PROPAGATION = -1
+
+        def initialize(entry_ttl)
+          raise ArgumentError unless entry_ttl&.integer?
```
indrekj: If we only want to allow integers and disallow strings for example then `unless entry_ttl.is_a?(Integer)` should work better. `#integer?` is not defined in non numeric classes.

api/lib/opentelemetry/distributed_context/entry.rb
fbogsany: We want this to be a frozen copy. The spec requires it to be immutable.

indrekj: I was just surprised because I haven't seen this syntax before and cannot find and documentation about it.

api/lib/opentelemetry/distributed_context/distributed_context.rb
```diff
@@ -0,0 +1,20 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module DistributedContext
+    # An immutable implementation of the DistributedContext that does not contain any entries.
+    class DistributedContext
+      def entries
+        []
```
mwear: Should this be a constant so it doesn't allocate an array every time it's called?

api/lib/opentelemetry/distributed_context/entry.rb
fbogsany: https://ruby-doc.org/core-2.4.0/String.html#method-i-2D-40

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -0,0 +1,24 @@
...
+      # process boundaries.
+      #
+      # Encoding is expected to conform to the HTTP Header Field semantics. Values are often encoded as RPC/HTTP request
+      # headers.
+      #
+      # The carrier of propagated data on both the client (injector) and server (extractor) side is usually an http request.
+      # Propagation is usually implemented via library-specific request interceptors, where the client-side injects values
+      # and the server-side extracts them.
+      class HTTPTextFormat
+        # TODO
```
mwear: Is the implementation for `HTTPTextFormat` going to be part of this PR?

fbogsany: That was my intent, although it's tempting to punt that to another PR so we can 🚢 this and regain some momentum.

mwear: I'd be fine with that. I'm adding my approval. We can talk about this at the meeting and if nobody has any objections, we should merge it.

PR #23, opened 2019-07-18
Improvements to Span, SpanContext and introduction of SpanId, TraceId

indrekj:


api/lib/opentelemetry/trace/span.rb
```diff
@@ -6,42 +6,143 @@
...
+      #
+      # @param [String, Symbol] key
+      # @param [String, Symbol, Boolean, Numeric] value
+      #
+      # @return [self] returns itself
       def set_attribute(key, value)
         raise ArgumentError unless valid_key?(key)
         raise ArgumentError unless valid_value?(value)
+
+        self
```
fbogsany: Indentation is wrong here.

api/lib/opentelemetry/trace/span.rb
```diff
@@ -6,42 +6,143 @@
...
       def add_event(name, **attrs)
         raise ArgumentError if name.nil?
+
+        self
       end
 
+      # Adds a link to another Span from this Span. Linked Span can be from the
+      # same or different trace. See Links description.
+      #
+      #
```
fbogsany: Unnecessary blank line.

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
...
+    class SpanId
+      INVALID_ID = 0
+      private_constant :INVALID_ID
+
+      MAX_64BIT_UNSIGNED_INT = (1 << 64) - 1
+      private_constant :MAX_64BIT_UNSIGNED_INT
+
+      def self.generate
+        id = rand(MAX_64BIT_UNSIGNED_INT)
+        return generate if id == INVALID_ID
```
fbogsany: No need for this to be recursive. You can use a range with `rand` to avoid generating an invalid id.

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    # A class that represents a trace identifier. A valid span identifier is an
+    # 8-byte array with at least one non-zero byte.
```
fbogsany: This is technically incorrect. The identifier here is a integer, not an 8-byte array.

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def initialize(id)
+        @id = id
+      end
+
+      # Checks if SpanId is valid
+      #
+      # A valid span identifier is an 8-byte array with at least one non-zero
+      # byte.
```
fbogsany: This is also incorrect.

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
...
+    # 8-byte array with at least one non-zero byte.
+    class SpanId
+      INVALID_ID = 0
+      private_constant :INVALID_ID
+
+      MAX_64BIT_UNSIGNED_INT = (1 << 64) - 1
+      private_constant :MAX_64BIT_UNSIGNED_INT
+
+      def self.generate
+        id = rand(MAX_64BIT_UNSIGNED_INT)
```
fbogsany: In Shopify’s tracing gem, we limited the span IDs to the range `1..(2**62 - 1)`, sacrificing 2 bits to ensure we only generated Fixnums (avoiding extra allocations for Bignums).

api/lib/opentelemetry/trace/trace_id.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    # A class that represents a span identifier. A valid span identifier is a
+    # 16-byte array with at least one non-zero byte.
```
fbogsany: Also technically incorrect.

api/lib/opentelemetry/trace/trace_id.rb
```diff
@@ -0,0 +1,50 @@
...
+    class TraceId
+      INVALID_ID = 0
+      private_constant :INVALID_ID
+
+      MAX_128BIT_UNSIGNED_INT = (1 << 128) - 1
+      private_constant :MAX_128BIT_UNSIGNED_INT
+
+      def self.generate
+        id = rand(MAX_128BIT_UNSIGNED_INT)
+        return generate if id == INVALID_ID
```
fbogsany: Same thing here: use a range instead of recursion.

api/lib/opentelemetry/trace/trace_id.rb
```diff
@@ -0,0 +1,50 @@
...
+      end
+
+      def initialize(id)
+        @id = id
+      end
+
+      # Checks if TraceId is valid
+      #
+      # A valid trace identifier is a 16-byte array with at least one non-zero
+      # byte.
```
fbogsany: Also incorrect.

api/lib/opentelemetry/trace/trace_id.rb
```diff
@@ -0,0 +1,50 @@
...
+        @id != INVALID_ID
+      end
+
+      def ==(other)
+        @id == other.id
+      end
+
+      protected
+
+      attr_reader :id
```
fbogsany: Why does this need to be protected?

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
...
+        @id != INVALID_ID
+      end
+
+      def ==(other)
+        @id == other.id
+      end
+
+      protected
+
+      attr_reader :id
```
fbogsany: Why does this need to be protected?

api/lib/opentelemetry/trace/span_id.rb
```diff
@@ -0,0 +1,50 @@
...
+
+      INVALID = new(INVALID_ID)
```
fbogsany: IIRC this constant won’t be protected, despite its location under that keyword. My Ruby-fu is probably lacking, tho.

api/lib/opentelemetry/trace/span.rb
indrekj: Are you sure, seems correct to me

api/lib/opentelemetry/trace/span_id.rb
indrekj: Didn't want it to be protected. But it has to be defined after the constructor, otherwise it will fail. Moved it out from protected section to avoid confusion

api/lib/opentelemetry/trace/span_id.rb
indrekj: I don't think it's a problem since ruby 2.4 when Bignum and Fixnum were merged into Integer.
```
0.size # => 8
((1 << 64) - 1).size # => 8
```

api/lib/opentelemetry/trace/trace_id.rb
indrekj: In java implementation they implement the id as an integer but expose only the conversion methods (e.g. toLowerBase16): https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/SpanId.java . I currently did the same thing here. I used protected to be still able to compare two Ids in `#==`. If we expose the integer representation of the id then it probably could use a better name (e.g. to_i)

indrekj:
> Does the abstraction offer significant benefits? Does it impose unnecessary restrictions on vendor implementations?

It allows easily to change Id internals. e.g. we could represent trace id with two integers high & low like https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/TraceId.java#L57 and still pass along one object.

I think the main idea is that different actors expect different representations of the Ids. Vendors usually expect ints, but users expect base16.

I as a end-user would expect to easily be able to get the base16 representation (e.g. to use it in my logs). `span.context.span_id.to_lower_base16` seems quite convenient.

So far it looks like java and dotnet implementations are doing the wrapping. Dotnet has `ToHexString` and java has `toLowerBase16()`.

I do see you're point though that this will cause an extra object allocation. I'm open to ideas.

api/lib/opentelemetry/trace/span.rb
fbogsany: Strange - this rendered incorrectly on an iPad, but looks correct when viewed on my laptop 🤷‍♂

mwear:
> > Does the abstraction offer significant benefits? Does it impose unnecessary restrictions on vendor implementations?
> 
> It allows easily to change Id internals. e.g. we could represent trace id with two integers high & low like https://github.com/open-telemetry/opentelemetry-java/blob/master/api/src/main/java/io/opentelemetry/trace/TraceId.java#L57 and still pass along one object.
> 
> I think the main idea is that different actors expect different representations of the Ids. Vendors usually expect ints, but users expect base16.
> 
> I as a end-user would expect to easily be able to get the base16 representation (e.g. to use it in my logs). `span.context.span_id.to_lower_base16` seems quite convenient.
> 
> So far it looks like java and dotnet implementations are doing the wrapping. Dotnet has `ToHexString` and java has `toLowerBase16()`.
> 
> I do see you're point though that this will cause an extra object allocation. I'm open to ideas.

## Research from other OTel Projects

I took at look at opentelemetry-js and opentelemetry-python to see how they handle ID generation. Here's what I found:

opentelemetry-js generates IDs using a function called `crypto.randomBytes`. The Ruby equivalent would be [`SecureRandom.hex`](https://ruby-doc.org/stdlib-2.6.3/libdoc/securerandom/rdoc/Random/Formatter.html#method-i-hex), which depending on our needs, could be something worth considering. That is, if `rand()` is not going to be random enough for some situations. You can see  [id.ts here](https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-core/src/platform/node/id.ts#L21-L35). I'll paste a portion below as well:

```typescript
/**
 * Returns a random 16-byte trace ID formatted/encoded as a 32 lowercase hex
 * characters corresponding to 128 bits.
 */
export function randomTraceId(): string {
  return randomSpanId() + randomSpanId();
}

/**
 * Returns a random 8-byte span ID formatted/encoded as a 16 lowercase hex
 * characters corresponding to 64 bits.
 */
export function randomSpanId(): string {
  return crypto.randomBytes(SPAN_ID_BYTES).toString('hex');
}
```

For Python, I searched through the repo and open PRs, but surprisingly couldn't find id generation code. I don't think I missed anything; I think that code has yet to be written. I can keep an eye out for when it appears for another data point though.

## Actual Feedback

At the expense of an object allocation per Span, I think arguments can be made for or against the `SpanId` and `TraceId` classes. On one hand, if users regularly use the conversion methods, it makes sense to include this abstraction. On the other hand, if your average user doesn't use them, we'd be forcing an allocation for each span on them with no benefit.

I don't have a super strong opinion on this, but in general, I would try to limit per span allocations as much as possible. In terms of introducing the `SpanId` and `TraceId` abstraction, I would generally go with a minimal implementation first and see how that works out. If it becomes clear there is a need for additional functionality, add it then.

indrekj:
Closing this as it is not relevant atm. Can be reopen in future if needed.

PR #24, opened 2019-08-02
Improve Resource

mwear:
This PR makes some improvements to the `Resource` class. It implements `merge` according to the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-resources.md) and also adds tests and documentation. While working on this, I discovered that the original `merge` implementation wasn't working. This was because `initialize` expects labels as kwargs, but merge was passing them in as a `Hash<String, String>`. Since the labels are stored as a `Hash<String, String>` in resource, I updated `initialize` to expect a hash instead. See the comment and code in [this commit](https://github.com/open-telemetry/opentelemetry-ruby/commit/ecdc1e4897602fdb640abdacc1b8159536423b2b) for more context.

### DIFF
--- a/api/lib/opentelemetry/resources/resource.rb
+++ b/api/lib/opentelemetry/resources/resource.rb
@@ -9,8 +9,16 @@ module OpenTelemetry
     # Resource represents a resource, which captures identifying information about the entities
     # for which telemetry (metrics or traces) is reported.
     class Resource
+      # Returns the labels for this {Resource}
+      #
+      # @return [Hash<String, String>]
       attr_reader :labels
 
+      # Returns a newly created {Resource} with the specified labels
+      #
+      # @param [Hash<String, String>] kvs Hash of key-value pairs to be used
+      #   as labels for this resource
+      # @return [Resource]
       def initialize(kvs)
         # TODO: how defensive should we be here?
         @labels = kvs.each_with_object({}) do |(k, v), memo|
@@ -18,6 +26,13 @@ module OpenTelemetry
         end
       end
 
+      # Returns a new, merged {Resource} by merging the current {Resource} with
+      # the other {Resource}. In case of a collision, the current {Resource}
+      # takes precedence
+      #
+      # @param [Resource] other The other resource to merge
+      # @return [Resource] A new resource formed by merging the current resource
+      #   with other
       def merge(other)
         raise ArgumentError unless other.is_a?(Resource)
 
--- a/api/lib/opentelemetry/resources/resource.rb
+++ b/api/lib/opentelemetry/resources/resource.rb
@@ -18,11 +18,17 @@ module OpenTelemetry
         end
       end
 
-      # TODO: Already set labels MUST NOT be overwritten unless they are empty string.
       def merge(other)
         raise ArgumentError unless other.is_a?(Resource)
 
-        self.class.new(labels.merge(other.labels))
+        merged_labels = \
+          other.labels.each_with_object(labels.dup) do |(k, v), memo|
+            next if (current = memo[k]) && !current.empty?
+
+            memo[k] = v
+          end
+
+        self.class.new(merged_labels)
       end
     end
   end
--- a/api/lib/opentelemetry/resources/resource.rb
+++ b/api/lib/opentelemetry/resources/resource.rb
@@ -19,11 +19,8 @@ module OpenTelemetry
       # @param [Hash<String, String>] kvs Hash of key-value pairs to be used
       #   as labels for this resource
       # @return [Resource]
-      def initialize(kvs = {})
-        # TODO: how defensive should we be here?
-        @labels = kvs.each_with_object({}) do |(k, v), memo|
-          memo[-k] = -v
-        end
+      def initialize(raw_labels = {})
+        @labels = check_and_freeze_labels(raw_labels)
       end
 
       # Returns a new, merged {Resource} by merging the current {Resource} with
@@ -44,6 +41,16 @@ module OpenTelemetry
           end
 
         self.class.new(merged_labels)
+      end
+
+      private
+
+      def check_and_freeze_labels(raw_labels)
+        raw_labels.each_with_object({}) do |(k, v), memo|
+          raise ArgumentError, 'label keys and values must be strings' unless k.is_a?(String) && v.is_a?(String)
+
+          memo[-k] = -v
+        end.freeze
       end
     end
   end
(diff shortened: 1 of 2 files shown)


api/lib/opentelemetry/resources/resource.rb
```diff
@@ -9,18 +9,41 @@ module Resources
...
+      #
+      # @param [Hash<String, String>] kvs Hash of key-value pairs to be used
+      #   as labels for this resource
+      # @return [Resource]
+      def initialize(kvs = {})
         # TODO: how defensive should we be here?
-        @labels = kvs.to_h { |k, v| [k.to_s.clone.freeze, v.to_s.clone.freeze] }.freeze
+        @labels = kvs.each_with_object({}) do |(k, v), memo|
+          memo[-k] = -v
+        end
```
fbogsany: Should `.freeze` `@labels` after initialization. The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-resources.md#retrieve-labels) requires `labels` to be exposed as

> a read only collection of labels

api/lib/opentelemetry/resources/resource.rb
```diff
@@ -9,18 +9,41 @@ module Resources
     # Resource represents a resource, which captures identifying information about the entities
     # for which telemetry (metrics or traces) is reported.
     class Resource
+      # Returns the labels for this {Resource}
+      #
+      # @return [Hash<String, String>]
       attr_reader :labels
```
fbogsany: The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-resources.md#retrieve-labels) says:

> The most common operation when retrieving labels is to enumerate over them. As such, it is recommended to optimize the resulting collection for fast enumeration over other considerations such as a way to quickly retrieve a value for a label with a specific key.

Should we expose an enumerator instead of a frozen hash?

api/lib/opentelemetry/resources/resource.rb
```diff
@@ -9,18 +9,41 @@ module Resources
...
 
-      def initialize(**kvs)
+      # Returns a newly created {Resource} with the specified labels
+      #
+      # @param [Hash<String, String>] kvs Hash of key-value pairs to be used
+      #   as labels for this resource
+      # @return [Resource]
+      def initialize(kvs = {})
         # TODO: how defensive should we be here?
```
fbogsany: The spec requires the params to be:

> a collection of name/value labels, where name and value are both strings

Should we type-check the keys and values provided?

api/lib/opentelemetry/resources/resource.rb
```diff
@@ -9,18 +9,41 @@ module Resources
...
       def merge(other)
         raise ArgumentError unless other.is_a?(Resource)
 
-        self.class.new(labels.merge(other.labels))
+        merged_labels = \
+          other.labels.each_with_object(labels.dup) do |(k, v), memo|
+            next if (current = memo[k]) && !current.empty?
+
+            memo[k] = v
+          end
```
fbogsany: Would it be cleaner to use `merge` with a block? I.e.

```ruby
merged_labels = labels.merge(other.labels) { |_, old_v, new_v| old_v.empty? ? new_v : old_v }
```

api/lib/opentelemetry/resources/resource.rb
```diff
@@ -9,18 +9,41 @@ module Resources
...
+
+        self.class.new(merged_labels)
```
fbogsany: It's unfortunate that we dup the hash here. If we adopted the factory method approach (#25), we could make the constructor private and more flexible.

api/lib/opentelemetry/resources/resource.rb
mwear: Just to clarify, as you suggesting that we return `@labels.to_enum`? If so, would we still call the method `labels`?

mwear: I added a method `label_enumerator` that returns an enumerator, and kept the `labels` accessor, but changed its visibility to `protected` so that we can use it when merging.

api/lib/opentelemetry/resources/resource.rb
mwear: I went the factory method route which spares us the additional hash allocation and type checking. Users will instantiate a resource with `Resource.create` which will type check and freeze labels on the incoming hash. The constructor is private and expects that it receives a frozen hash with frozen string key-value pairs.

mwear: In regards to #25, we can talk about caching, what that would look like, and whether or not it would be useful.

api/lib/opentelemetry/resources/resource.rb
indrekj: I'm not sure what exactly are we gaining here with `label_enumerator`. If I need to go over all the labels (which is the common case reading the spec) then `resource.label_enumerator.each {.` seems weird. I'd rather have `resource.labels.each` or just `resource.each` though the latter also seems a bit odd.

`@labels` is a hash which already is an enumerable. If the sdk implementor wants to create a different data structure for faster access then they just have to return a different enumerable from `def labels`.

I might be missing something here though

fbogsany: The idea is to reduce the API surface area. Returning a frozen hash from `resource.labels` in the API sets the expectation that all implementations will have to return a frozen hash. The spec is more constrained - the expectation is that we support enumeration of labels but not indexing.

I agree that `resource.label_enumerator.each {` is weird. Renaming the method to `labels` and doing something else with the `attr_reader` is better. Or `resource.each_label {`?

mwear: I'm open to any of these suggestions. The spec does allow us to:

> ...provide a way to retrieve a read only collection of labels associated with a resource

While returning a frozen hash of labels would force other SDKs to return something with the same interface, I'm not sure what other data structure they would use, and if it'd ever be a pain point to return a frozen hash.

From the user perspective, I could see wanting access to the underlying hash. For example, they may want to merge metadata together and the functions available on hash would be helpful.

If we think that would be a use case and that we'd ultimately end up exposing the hash, we should probably just expose the hash, and not the enumerator. On the other hand, if we think returning a hash would be a burden on SDK implementations, we may want to avoid it.

We can try to make a decision here in this PR, or we can defer that decision and open an issue. Thoughts @indrekj and @fbogsany?

fbogsany: I'd prefer to defer the decision and open an issue.

mwear: I opened #43. Should we merge this for the time being?

fbogsany:
Fixes #26.
Fixes #25.

PR #42, opened 2019-08-07
Add docs to Span

indrekj:
Based on https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/tracing-api.md

Extracted span docs commit from #23

### DIFF
--- a/api/lib/opentelemetry/trace/span.rb
+++ b/api/lib/opentelemetry/trace/span.rb
@@ -6,42 +6,141 @@
 
 module OpenTelemetry
   module Trace
-    # Span is a mutable object storing information about the current operation execution.
+    # Span represents a single operation within a trace. Spans can be nested to
+    # form a trace tree. Often, a trace contains a root span that describes the
+    # end-to-end latency and, optionally, one or more sub-spans for its
+    # sub-operations.
+    #
+    # Once Span {Tracer#start_span is created} - Span operations can be used to
+    # add additional properties to it like attributes, links, events, name and
+    # resulting status. Span cannot be used to retrieve these properties. This
+    # prevents the mis-use of spans as an in-process information propagation
+    # mechanism.
+    #
+    # {Span} must be ended by calling {#finish}.
     class Span
+      # Retrieve the spans SpanContext
+      #
+      # The returned value may be used even after the Span is finished.
+      #
+      # @return [SpanContext]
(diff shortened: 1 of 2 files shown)


api/lib/opentelemetry/trace/span.rb
```diff
@@ -6,42 +6,142 @@
 
 module OpenTelemetry
   module Trace
-    # Span is a mutable object storing information about the current operation execution.
+    # Span represents a single operation within a trace. Spans can be nested to
+    # form a trace tree. Often, a trace contains a root span that describes the
+    # end-to-end latency and, optionally, one or more sub-spans for its
+    # sub-operations.
+    #
+    # Once Span {Tracer#start_span is created} - Span operations can be used to
```
fbogsany: Should this be `{Tracer#start_span} is created, Span` rather than `{Tracer#start_span is created} - Span`?

api/lib/opentelemetry/trace/span.rb
```diff
@@ -6,42 +6,142 @@
...
+      #
+      # @return [void]
       def name=(new_name)
         raise ArgumentError if new_name.nil?
       end
 
-      def finish; end
+      # Finishes the Span
+      #
+      # This call will take the current timestamp to set as Span's end time.
```
fbogsany: This sentence is ambiguous (in the spec too). It **will** be correct once https://github.com/open-telemetry/rfcs/pull/8 propagates to the spec, but for now it suggests the method takes the current timestamp as an argument when in fact it has no parameters.

api/lib/opentelemetry/trace/span.rb
mwear: This is a thing I just learned about yard, but that syntax will give you a link with the text `is created` that goes to `Tracer#start_span`.

indrekj: @mwear is correct, it's basically `{href link text}`

api/lib/opentelemetry/trace/span.rb
indrekj: removed this sentence for now

api/lib/opentelemetry/trace/span.rb
fbogsany: TIL. 👍

PR #51, opened 2019-08-08
Implement Status

### LINKED ISSUE #33: API: implement Status
[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#status)

fbogsany:
Fixes #33

### DIFF
--- a/api/lib/opentelemetry/trace/status.rb
+++ b/api/lib/opentelemetry/trace/status.rb
@@ -1,0 +1,112 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    # Status represents the status of a finished {Span}. It is composed of a
+    # canonical code in conjunction with an optional descriptive message.
+    class Status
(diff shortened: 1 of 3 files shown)


api/lib/opentelemetry/trace/status.rb
```diff
@@ -0,0 +1,112 @@
...
+      # Retrieve the description of this Status.
+      #
+      # @return [String]
+      attr_reader :description
+
+      # Initialize a Status.
+      #
+      # @param [Integer] canonical_code One of the standard GRPC codes: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+      # @param [String] description
+      def initialize(canonical_code, description: '')
```
mwear: It seems like it could be useful to have a default description for a code. I searched through the spec and Java repo and didn't find a default description either place. So, I guess an empty string as good as anything for now (we could add a default later if we decide to).

However, if we do go with an empty string, should we default it to a frozen string constant?

fbogsany: (Discussed IRL on the SIG call today): we have the `frozen_string_literal: true` comment at the top of the file, so the current code is equivalent to defaulting to a frozen string.

Issue #38, opened 2019-08-07
API: implement Sampling API


fbogsany:
[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-sampling.md)

PR #52, opened 2019-08-14
Sampling API

mwear:
This PR implements the [sampling API](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-sampling.md) and addresses #38. 

There are [ongoing discussions](https://github.com/open-telemetry/opentelemetry-specification/issues?q=is%3Aopen+is%3Aissue+label%3Asampling) around sampling in the specification repo, so I would not be surprised if changes come in this area. The API is implemented per the current spec for now.


api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
```diff
@@ -0,0 +1,46 @@
...
+      # The {AlwaysSampleSampler} always returns a true sampling decision.
+      class AlwaysSampleSampler
+        ALWAYS_SAMPLE_DECISION = BasicDecision.new(decision: true)
+
+        # Returns the sampling {Decision} for a {Span} to be created
+        #
+        # @param [SpanContext] span_context The {SpanContext} of a parent span,
+        #   typically extracted from the wire. Can be nil.
+        # @param [Boolean] extracted_context True if span_context was extracted
+        #   from the wire. Can be nil.
```
fbogsany: It is not clear from the specification that this can be `nil`.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
```diff
@@ -0,0 +1,46 @@
...
+        # @param [Integer] trace_id The trace_id of the {Span} to be created
+        # @param [Integer] span_id The span_id of the {Span} to be created
+        # @param [String] span_name Name of the {Span} to be created
+        # @param [Array<Links>] links A collection of links to be associated
```
fbogsany: Do we want to be this specific with the type? I.e. `Array<...>`

api/lib/opentelemetry/trace/samplers/basic_decision.rb
```diff
@@ -0,0 +1,39 @@
...
+        # @return [Boolean] sampling decision
+        def sampled?
+          @decision
+        end
+
+        # Always returns an empty set of attributes
+        #
+        # @return [Hash<String, Object>]
+        def attributes
+          {}
```
fbogsany: I *think* the attributes hash should be frozen. The spec isn't clear on this, but it implies the attributes are captured by the Decision rather than being modified afterwards.

api/lib/opentelemetry/trace/samplers/decision.rb
```diff
@@ -0,0 +1,38 @@
...
+      class Decision < BasicDecision
+        # Returns a new decision with the specified sampling decision and
+        # attributes.
+        #
+        # @param [Boolean] decision Whether or not a span should be sampled
+        # @param [Hash<String, Object>] attributes Attributes to be attached
+        #   to a root span
+        def initialize(decision:, attributes: nil)
+          super(decision: decision)
+          @attributes = attributes
```
fbogsany: Is there a significant benefit to splitting the Decision functionality across `BasicDecision` and `Decision`? If we adopt a frozen constant empty Hash as the default attributes, there shouldn't be any performance difference. E.g.:
```ruby
class Decision
  EMPTY_ATTRIBUTES = {}.freeze

  def initialize(decision:, attributes: nil)
    @decision = decision
    @attributes = attributes || EMPTY_ATTRIBUTES
  end
end
```

api/lib/opentelemetry/trace/samplers/never_sample_sampler.rb
```diff
@@ -0,0 +1,46 @@
...
+        #   typically extracted from the wire. Can be nil.
+        # @param [Boolean] extracted_context True if span_context was extracted
+        #   from the wire. Can be nil.
+        # @param [Integer] trace_id The trace_id of the {Span} to be created
+        # @param [Integer] span_id The span_id of the {Span} to be created
+        # @param [String] span_name Name of the {Span} to be created
+        # @param [Array<Links>] links A collection of links to be associated
+        #   with the {Span} to be created
+        # @return [Decision] The sampling decision
+        def should_sample(span_context: nil,
```
fbogsany: This feels "almost" like a predicate, which leads to questions like: is `should_` a common prefix for a predicate in Ruby? Should this end with `?`? It's a little annoying, and I wish we could rename the method to be more descriptive of the return value. It *would* be a boolean if we didn't also want to provide attributes describing the decision.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
```diff
@@ -0,0 +1,46 @@
...
+        #   with the {Span} to be created
+        # @return [Decision] The sampling decision
+        def should_sample(span_context: nil,
+                          extracted_context: nil,
+                          trace_id:,
+                          span_id:,
+                          span_name:,
+                          links:)
+          ALWAYS_SAMPLE_DECISION
```
fbogsany: Should we type-check the arguments here? On one hand, this might well be used in production, and we shouldn't add overhead there. On the other hand, we want to be able to catch errors in use.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
fbogsany: It feels like `nil` should be allowed here. We often won't have any Links, and we shouldn't require the allocation of an empty array just to call this method.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
mwear: The [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-sampling.md#shouldsample) says:

> SpanContext of a parent Span. Typically extracted from the wire. Can be null.

mwear: Oh, you're asking about the `extracted_context`. I suppose that is ambiguous. I cross referenced the Java Decision Interface, where it is also `nullable`: 

https://github.com/open-telemetry/opentelemetry-java/blob/6d1b31bfa639af653649d694ea5660fb7c3fd5ce/api/src/main/java/io/opentelemetry/trace/util/Samplers.java#L100-L108

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
mwear: This is also unclear from the spec. Cross referencing the Java Implementation, `parentLinks` are not nullable on the AlwaysSampleSampler or NeverSampleSampler. 
AlwaysSampleSampler:
https://github.com/open-telemetry/opentelemetry-java/blob/6d1b31bfa639af653649d694ea5660fb7c3fd5ce/api/src/main/java/io/opentelemetry/trace/util/Samplers.java#L71-L81

NeverSampleSampler
https://github.com/open-telemetry/opentelemetry-java/blob/6d1b31bfa639af653649d694ea5660fb7c3fd5ce/api/src/main/java/io/opentelemetry/trace/util/Samplers.java#L100-L108

But they are on the ProbabilitySampler:

https://github.com/open-telemetry/opentelemetry-java/blob/6d1b31bfa639af653649d694ea5660fb7c3fd5ce/sdk/src/main/java/io/opentelemetry/sdk/trace/samplers/ProbabilitySampler.java#L76-L83

api/lib/opentelemetry/trace/samplers/basic_decision.rb
mwear: That was my instinct, but I couldn't find anything official or unofficial pointing towards an immutable data structure.

The obligatory cross reference to java shows `Map<String, AttributeValue>` return type: https://github.com/open-telemetry/opentelemetry-java/blob/6d1b31bfa639af653649d694ea5660fb7c3fd5ce/api/src/main/java/io/opentelemetry/trace/Sampler.java#L79-L86

api/lib/opentelemetry/trace/samplers/decision.rb
mwear: If attributes are immutable, there would not be a benefit to have multiple Decision classes. The only reason for this division is the uncertainty about the immutability of attributes.

api/lib/opentelemetry/trace/samplers/never_sample_sampler.rb
mwear: Originally, I was treating this as a predicate method and had the `?`, however, since it doesn't return a boolean value, it's really not a predicate, so I decided remove it. Since it returns a `Decision` instance, it really can't be used like a predicate either. You really need to assign to the return value, and then check the return value. Or chain through to it: `sampler.should_sample(...).sampled?`. I'm definitely not loving the API overall.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
mwear: We can. In that case I would make a Sampler abstract class with template methods and the type checking logic.

mwear:
@fbogsany: I agree with all of these comments and really had the same questions as I was implementing this. There is ambiguity in the spec and the Java implementation as to what arguments are required for `should_sample` and whether or not `Decision#attributes` is immutable. Both of those have resulting impacts on the implementation.

I will open an issue on the Spec repo to get some clarification. Once we can definitively answer some of these questions, cleaning this will be pretty straightforward.

api/lib/opentelemetry/trace/samplers/basic_decision.rb
mwear: Reading deeper into the Javadocs, `Collections.emptyMap` is immutable. So, it might just be a matter of clarifying some of these details on the spec.

mwear:
I opened a spec issue here: https://github.com/open-telemetry/opentelemetry-specification/issues/218. I'll start making some changes here based on the feedback, and speculation that the proposed clarifications can be made.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
mwear: I changed this to an `Enumerable<Link>`, that can be nil. The nil part is subject to the clarification on the spec. Does the `Enumerable<Link>` work for you?

api/lib/opentelemetry/trace/samplers/basic_decision.rb
mwear: Assuming that `Decision#attributes` can be immutable, and there are hints from the Java implementation that is possible, this is done.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
mwear: I implemented type checking in the abstract `Sampler` class. It's a bit cumbersome and did not go over especially with RuboCop. I'm open to suggestions here.

api/lib/opentelemetry/trace/samplers/never_sample_sampler.rb
mwear: I would be open to other names that stick to the spirit of the spec, but are more idiomatic. Possibly calling it `sampling_decision(...)` or `sampling_decision_for(...)`would be slight improvements. I'm open to other ideas though.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
fbogsany: 👍 would be good to add the comment from there: `nil if this is a root span`.

mwear: I went ahead and qualified the nilness of `span_context` and `extracted_context`. `extracted_context` might get added to span context: https://github.com/open-telemetry/opentelemetry-specification/pull/216, which would clean this up further.

api/lib/opentelemetry/trace/samplers/never_sample_sampler.rb
fbogsany: Either of `decision(...)` or `decision_for(...)` would be 👍 for me. The tests don't read well at the moment:
```ruby
sampler.should_sample(...).sampled?
```
vs.
```ruby
sampler.sampling_decision(...).sampled?
```
vs.
```ruby
sampler.sampling_decision_for(...).sampled?
```
vs.
```ruby
sampler.decision(...).sampled?
```
vs.
```ruby
sampler.decision_for(...).sampled?
```

In most cases, the sampler will be called something like `sampler`, so the shorter method name reduces stuttering a little.

api/lib/opentelemetry/trace/samplers/always_sample_sampler.rb
fbogsany: I don't think it'd help with RuboCop, but would a simple `super` be sufficient here, with type checking implemented in `Sampler#should_sample`?

mwear: We could do that. Would you have any qualms about `Sampler#should_sample` only doing type checking and not having a `Decision` return value?

fbogsany: No qualms. Returning `nil` from `Sampler#should_sample` seems OK, even if we document it to return a `Decision`. An alternative would be a type-checking mixin, but I'm not sure that's any better.

mwear:
This is hot off the presses: https://github.com/open-telemetry/rfcs/pull/24, so I'm not sure what that means for this work for now.

mwear:
Does it make sense to try to get this in a mergeable state and make an issue to update the sampling API based on the proposed RFC once it has had sufficient discussion and approval?

PR #53, opened 2019-08-20
Update the span creation API

fbogsany:
This update the span creation API to reflect the [latest version of the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span-creation) and the [`SpanData` removal PR](https://github.com/open-telemetry/opentelemetry-specification/pull/215).

### DIFF
--- a/api/lib/opentelemetry/trace/span.rb
+++ b/api/lib/opentelemetry/trace/span.rb
@@ -140,6 +140,8 @@ module OpenTelemetry
       #
       # This API MUST be non-blocking.
       #
+      # @param [Time] end_timestamp optional end timestamp for the span.
+      #
       # @return [self] returns itself
       def finish(end_timestamp: nil)
         self
--- a/api/lib/opentelemetry/trace/span.rb
+++ b/api/lib/opentelemetry/trace/span.rb
@@ -72,9 +72,10 @@ module OpenTelemetry
       # @param [Hash<String, Object>] attrs One or more key:value pairs, where
       #   the keys must be strings and the values may be string, boolean or
       #   numeric type.
+      # @param [Time] timestamp optional timestamp for the event.
       #
       # @return [self] returns itself
-      def add_event(name, attrs = nil)
+      def add_event(name, attrs = nil, timestamp: nil)
         raise ArgumentError if name.nil?
         raise ArgumentError unless valid_attributes?(attrs)
 
@@ -140,7 +141,7 @@ module OpenTelemetry
       # This API MUST be non-blocking.
       #
       # @return [self] returns itself
-      def finish
+      def finish(end_timestamp: nil)
         self
       end
 
--- a/api/lib/opentelemetry/trace/tracer.rb
+++ b/api/lib/opentelemetry/trace/tracer.rb
@@ -31,8 +31,8 @@ module OpenTelemetry
       #
       # With this helper:
       # OpenTelemetry.tracer.in_span('do-the-thing') do ... end
-      def in_span(name, sampler: nil, links: nil, recording_events: nil, kind: nil)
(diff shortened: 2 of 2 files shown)


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -31,22 +31,22 @@ def current_span
       #
       # With this helper:
       # OpenTelemetry.tracer.in_span('do-the-thing') do ... end
-      def in_span(name, sampler: nil, links: nil, recording_events: nil, kind: nil)
-        span = start_span(name, sampler: sampler, links: links, recording_events: recording_events, kind: kind)
+      def in_span(name, attributes: nil, links: nil, events: nil, resource: nil, span_id: nil, start_timestamp: nil)
```
mwear: Should we document the parameters to the API methods in Tracer? Or should we handle this separately?

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -31,22 +31,22 @@ def current_span
...
 
-      def start_root_span(name, sampler: nil, links: nil, recording_events: nil, kind: nil)
+      def start_root_span(name, attributes: nil, links: nil, events: nil, resource: nil, span_id: nil, start_timestamp: nil)
         raise ArgumentError if name.nil?
 
         Span.create_random
       end
 
-      def start_span(name, with_parent: nil, with_parent_context: nil, sampler: nil, links: nil, recording_events: nil, kind: nil)
+      def start_span(name, with_parent: nil, with_parent_context: nil, attributes: nil, links: nil, events: nil, resource: nil, span_id: nil, start_timestamp: nil)
```
mwear: It seems like we might need a `end_timestamp` according to this comment: https://github.com/open-telemetry/opentelemetry-specification/pull/215#discussion_r315437851. This would apply to `in_span` and `start_root_span` as well.

fbogsany:
Looks like removing `kind` is premature: https://github.com/open-telemetry/opentelemetry-specification/pull/226

api/lib/opentelemetry/trace/span.rb
```diff
@@ -140,7 +141,7 @@ def name=(new_name)
       # This API MUST be non-blocking.
       #
       # @return [self] returns itself
-      def finish
+      def finish(end_timestamp: nil)
```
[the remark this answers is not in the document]
indrekj: should it also default to `Time.now`?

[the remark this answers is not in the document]
fbogsany: In normal use, this argument shouldn't be provided. For the SDK, defaulting to `Time.now` is a reasonable choice. For the API, which provides a minimal implementation, strives to impose low overhead, and which doesn't need to evaluate the argument, I think `nil` is a better choice.

api/lib/opentelemetry/trace/tracer.rb
fbogsany: We should document them all.

api/lib/opentelemetry/trace/tracer.rb
fbogsany: I think that comment misinterprets the RFC. The [relevant text](https://github.com/open-telemetry/rfcs/blob/8dd6e21/text/0002-remove-spandata.md#internal-details) only has the `end_timestamp` passed to `span.end()` (`span.finish` in our implementation).

fbogsany:
This is ready for review. It anticipates https://github.com/open-telemetry/oteps/pull/26. It isn't clear to me whether that also requires removal of the optional timestamps. I'd vote for 🚢 this now, and removing the optional timestamps once the spec is updated.

Issue #37, opened 2019-08-07
API: implement Span.add_lazy_link


fbogsany:
[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#add-links). Need to decide whether to provide a `Link` interface or a concrete `Link` class and a `LinkFormatter`.

mwear:
The specification doesn't define the API for a `LinkFormatter`, so I'm not sure what a concrete `Link` class with a `LinkFormatter` would look like. I opened this issue for clarification: https://github.com/open-telemetry/opentelemetry-specification/issues/222.

PR #54, opened 2019-08-21
Implement Span#add_lazy_link

mwear:
This PR adds `Span#add_lazy_link`. After some clarification on the specification, a concrete Link class is an implementation option, and the one that I decided to go with as it seems to make the most sense for Ruby. 

This fixes: #37.


mwear:
We talked about this briefly in the SIG meeting, but I wanted to capture that conversation here. This implementation follows the Java implementation pretty closely, which, does not appear to be very lazy. The properties of the `Link` are lazily accessed, but the `Link` itself is eagerly initialized.

See:

* [Link Interface](https://github.com/open-telemetry/opentelemetry-java/blob/66a15d3bd5ab8e43b1787fb6e444f2b17cb485b3/api/src/main/java/io/opentelemetry/trace/Link.java#L31-L47)
* [Span#add_link](https://github.com/open-telemetry/opentelemetry-java/blob/66a15d3bd5ab8e43b1787fb6e444f2b17cb485b3/api/src/main/java/io/opentelemetry/trace/Span.java#L178-L187)
* [SimpleLink (implements Link)](https://github.com/open-telemetry/opentelemetry-java/blob/66a15d3bd5ab8e43b1787fb6e444f2b17cb485b3/api/src/main/java/io/opentelemetry/trace/util/Links.java#L34-L71)

We discussed the option of providing a callable that returns a link as a way to make this more lazy. This would be in line with the `LinkFormatter` and concrete `Link` class option. The downside is that a callable will require an object allocation for the callable, plus an additional allocation for the Link, when called. If attributes are created using static data, this would be additional overhead and possible not very desirable. However, if attributes are generated dynamically in way that is potentially expensive, or would result in additional allocations, this could be preferable.

I have a few ideas of how we can handle this:

### Option 1 - Not very lazy

Leave things as is (a Link is an already realized instance, whose properties are accessed lazily, by the tracer when needed).

### Option 2 - Lazy Link by callable

We could support both use cases by allowing `add_lazy_link` to accept either a `Link` instance, or a callable that returns a `Link`. The SDK would have to handle the conversion at the point in the time the link is accessed, which might lead to some complications.

### Option 3 - Lazy attributes on Link

We could possibly get the best of both worlds by making the `Link` class its self optionally lazy. Assuming that attributes are the the potentially expensive thing to make (and that the span_context is essentially free) we could change `Link#initialize` to accept a span_context and attributes, or a span_context and a block that returns a frozen hash of attributes, where the block would be executed lazily when attributes are first accessed. This option would give us lazy attributes, and the SDK would only handle `Link` instances and not have to worry about managing callables in addition, as it would with **Option 2**.

I think legitimate use cases will help to drive the best decision. Whatever we choose, we should make sure that the approach will adaptable for `add_lazy_event`, where a lazy event has a `name` and `attributes`. If a name is cheap, but attributes are expensive, then the lazy attribute approach would apply there as well.

api/lib/opentelemetry/trace/span.rb
```diff
@@ -81,25 +81,37 @@ def add_event(name, attrs = nil)
...
+      #
+      # @param [Link] link A link to another span whose attributes are lazily
+      #   accessed
+      #
+      # @return [self] returns itself
+      def add_lazy_link(link)
+        raise ArgumentError if link.nil? || !link.is_a?(Link)
+
+        self
+      end
```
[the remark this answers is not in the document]
mwear: The spec is pretty clear that we should have an API method that takes a `span_context` and attributes, and a method that takes a `Link` object.

> - An API to record a single Link where the Link properties are passed as arguments. This MAY be called AddLink.
> - An API to record a single lazily initialized Link. This can be implemented by providing a Link interface or a concrete Link definition and a LinkFormatter. If the language supports overloads then this MAY be called AddLink otherwise AddLazyLink MAY be consider.

I went ahead and combined `Span#add_link` and `Span#add_lazy_link`. The user can pass in a `span_context` with `attributes`, or an `attribute_formatter`, or just a link instance.

api/lib/opentelemetry/internal.rb
```diff
@@ -20,5 +20,17 @@ def printable_ascii?(string)
...
+
+    def valid_key?(key)
+      key.instance_of?(String)
+    end
+
+    def valid_value?(value)
+      value.instance_of?(String) || value == false || value == true || value.is_a?(Numeric)
+    end
+
+    def valid_attributes?(attrs)
```
mwear: I moved `valid_attributes?` and friends to the `Internal` module. I'm not sure that this is the best place for these methods, or if they should keep the same names if it is. We could alternatively extract this concern into its own module if it makes sense.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -0,0 +1,69 @@
...
+      # was provided when creating the link, it will be called lazily when
+      # attributes are first accessed.
+      #
+      # @return [Hash<String, Object>]
+      def attributes
+        if @attributes
+          @attributes
+        else
+          attributes = @attribute_formatter.call
+          raise 'Attribute formatter returned invalid attributes' unless Internal.valid_attributes?(attributes)
```
mwear: I'm a little uneasy about raising an exception here. We could default to `EMPTY_ATTRIBUTES` if the `attribute_formatter` returns something invalid. Thoughts?

api/lib/opentelemetry/trace/span.rb
```diff
@@ -76,29 +76,38 @@ def set_attribute(key, value)
...
       # @return [self] returns itself
-      def add_link(span_context_or_link, attrs = nil)
-        raise ArgumentError if span_context_or_link.nil?
-        raise ArgumentError unless span_context_or_link.instance_of?(SpanContext) || attrs.nil? || attrs.empty?
-        raise ArgumentError unless valid_attributes?(attrs)
+      #
+      # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      def add_link(span_context_or_link, attrs = nil, &attribute_formatter)
+        raise ArgumentError unless span_context_or_link.instance_of?(SpanContext) || span_context_or_link.is_a?(Link)
+        raise ArgumentError if span_context_or_link.instance_of?(SpanContext) && attrs.nil? && attribute_formatter.nil?
```
fbogsany: I don't think this check is correct. It's OK to have a `SpanContext` without any attributes.

api/lib/opentelemetry/trace/span.rb
```diff
@@ -76,29 +76,38 @@ def set_attribute(key, value)
...
+        raise ArgumentError if attrs && !Internal.valid_attributes?(attrs)
```
fbogsany: I think this can just be `raise ArgumentError unless Internal.valid_attributes?(attrs)`, since `valid_attributes?` allows `nil` attributes.

api/lib/opentelemetry/internal.rb
fbogsany: 👍 for leaving it here for now.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -0,0 +1,69 @@
...
+      #   callable that returns a hash of attributes for this link. Will be
+      #   called lazily and its return value will be frozen when attributes are
+      #   first accessed. If both attributes and a link_formatter are provided,
+      #   an ArgumentError will be raised.
+      # @return [Link]
+      # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+      def initialize(span_context:, attributes: nil, &attribute_formatter)
+        raise ArgumentError unless span_context.instance_of?(SpanContext)
+        raise ArgumentError unless attributes.nil? || attributes.is_a?(Hash)
+        raise ArgumentError unless attributes.nil? || Internal.valid_attributes?(attributes)
```
fbogsany: Maybe move the `attributes.is_a?(Hash)` into `valid_attributes` and then replace these two lines with:
```ruby
raise ArgumentError unless Internal.valid_attributes?(attributes)
```
(you already have the nil check there).

api/lib/opentelemetry/trace/link.rb
```diff
@@ -0,0 +1,60 @@
...
+        raise ArgumentError unless span_context.instance_of?(SpanContext)
+        raise ArgumentError unless attributes.nil? || attributes.is_a?(Hash)
+        raise ArgumentError, 'Expected attributes or block, but received both' if attributes && attribute_formatter
+
+        @context = span_context
+
+        if attribute_formatter
+          @attribute_formatter = attribute_formatter
+        else
+          @attributes = attributes.freeze || EMPTY_ATTRIBUTES
```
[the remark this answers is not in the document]
fbogsany: I'm wrong - this is already done above.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -0,0 +1,69 @@
...
+      def initialize(span_context:, attributes: nil, &attribute_formatter)
+        raise ArgumentError unless span_context.instance_of?(SpanContext)
+        raise ArgumentError unless attributes.nil? || attributes.is_a?(Hash)
+        raise ArgumentError unless attributes.nil? || Internal.valid_attributes?(attributes)
+        raise ArgumentError, 'Expected attributes or block, but received both' if attributes && attribute_formatter
+
+        @context = span_context
+
+        if attribute_formatter
+          @attributes = nil
```
fbogsany: Do you need to do this explicitly? The field should be implicitly `nil`.

api/lib/opentelemetry/trace/link.rb
fbogsany: Log error and default, IMO. A caller won't know to anticipate an error here.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -0,0 +1,69 @@
...
+      #
+      # @return [Hash<String, Object>]
+      def attributes
+        if @attributes
+          @attributes
+        else
+          attributes = @attribute_formatter.call
+          raise 'Attribute formatter returned invalid attributes' unless Internal.valid_attributes?(attributes)
+
+          @attributes = attributes.freeze
```
fbogsany: From https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/concurrency.md
> **SpanData** - SpanData is immutable and is safe to be used concurrently.
>
> **Link** - same as SpanData.

This is problematic with lazy links. What are the expectations around calling the `attribute_formatter` once only? Do we need to protect this method with a mutex?

api/lib/opentelemetry/trace/span.rb
```diff
@@ -76,29 +76,43 @@ def set_attribute(key, value)
...
-        raise ArgumentError if span_context_or_link.nil?
-        raise ArgumentError unless span_context_or_link.instance_of?(SpanContext) || attrs.nil? || attrs.empty?
-        raise ArgumentError unless valid_attributes?(attrs)
+      #
+      # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/AbcSize
+      def add_link(span_context_or_link, attrs = nil, &attribute_formatter)
+        raise ArgumentError unless span_context_or_link.instance_of?(SpanContext) || span_context_or_link.is_a?(Link)
+        raise ArgumentError unless Internal.valid_attributes?(attrs)
+        raise ArgumentError unless attrs.nil? || attribute_formatter.nil?
+        raise ArgumentError if span_context_or_link.is_a?(Link) && (!attrs.nil? || !attribute_formatter.nil?)
```
fbogsany: I found myself thinking through this way too much. Might be clearer if we express it as:
```ruby
raise ArgumentError if span_context_or_link.is_a?(Link) && !(attrs.nil? && attribute_formatter.nil?)
```

api/test/opentelemetry/trace/link_test.rb
```diff
@@ -0,0 +1,61 @@
...
+  end
+  describe '.attributes' do
+    it 'returns an empty hash by default' do
+      link = OpenTelemetry::Trace::Link.new(span_context: span_context)
+      link.attributes.must_equal({})
+    end
+    it 'returns and freezes attributes passed in' do
+      attributes = { 'foo' => 'bar', 'bar' => 'baz' }.freeze
+      link = OpenTelemetry::Trace::Link.new(span_context: span_context,
+                                            attributes: attributes)
```
fbogsany: Shouldn't we pass in non-frozen attributes here and check that they're frozen below?

api/test/opentelemetry/trace/link_test.rb
```diff
@@ -0,0 +1,61 @@
...
+        end
+      end.must_raise(ArgumentError)
+    end
+    it 'raises an exception if formatter returns invalid attributes' do
+      proc do
+        link = OpenTelemetry::Trace::Link.new(span_context: span_context) do
+          { 'bar' => Object.new }
+        end
+        link.attributes
+      end.must_raise(RuntimeError)
```
fbogsany: `wont_raise` and change the description if we decide to log and default rather than raising.

api/lib/opentelemetry/trace/link.rb
mwear: Would it possibly be a better design for us to not allow lazy attributes on a Link and instead change the API to accept a LinkFormatter (proc) that returns an immutable Link instance? It would be a slight change from what we have, but it would spare us from having to protect attributes with a mutex. This would essentially be [Option 2](https://github.com/open-telemetry/opentelemetry-ruby/pull/54#issuecomment-523646334) from my comment futher up this PR.

fbogsany: It'd be interesting to see what that option looks like. That'd simplify `Link`, while pushing the complexity to `Span`. I think that's fine. `Span` is going to have to deal with mutation anyway.

api/lib/opentelemetry/trace/link.rb
mwear: It was needed, otherwise we'd get a warning about referencing an uninitialized ivar, that code is gone now though.

mwear:
I went ahead and implemented this using [Option 2](https://github.com/open-telemetry/opentelemetry-ruby/pull/54#issuecomment-523646334) - lazy link by callable instead of [Option 3](https://github.com/open-telemetry/opentelemetry-ruby/pull/54#issuecomment-523646334) lazy attributes. This pushes the complexity to Span, but makes the implementation for Link trivial as it allows us to skirt concurrency issues around an `attribute_formatter` being executed more than once. Span might still have to deal with this situation, but could do so using a single lock when evaluating links, rather than a lock per link.

PR #55, opened 2019-08-29
http_text_format + w3c trace parent

ibawt:
## w3c Trace Parent support
- omitting trace state as it's being removed ( at least as of now )
- the actual integration points aren't well defined right now, but the TraceParent thing should be alright
- I went through the w3c SHOULD's and MUST's and did most of them, at least in terms for the actual parsing

## questions
- pass through services?  apparently we shouldn't do anything with it, not sure how to represent in the current api ( what defines a pass through service? why would anyone write a proxy in ruby anyhoo )
- are trace-ids every not 32 bytes?  there's language in the w3c standard indicating we should fill them with random values and not zeros.  However it also says the trace-id is exactly 32 bytes long. 
- is this too many objects etc?

https://www.w3.org/TR/trace-context/


ibawt:
@fbogsany 👀 lets start the bike shed of the actual http_text_format api.

api/test/opentelemetry/distributed_context/propagation/trace_parent_test.rb
```diff
@@ -0,0 +1,94 @@
...
+    it 'must not accept a span-id of all zeros' do
+      value = '00-0000000000000000000000000000000a-0000000000000000-01'
+
+      proc {
+        TraceParent.from_string(value)
+      }.must_raise TraceParent::InvalidSpanIDError
+    end
+
+    it 'must parse a higher version header according to the w3c standard' do
+      value = '00-0000000000000000000000000000000a-000000000000000a-ff'
```
ibawt: whoops need to fix the version numbers in the test.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,108 @@
...
+      @sampled ? "01" : "00"
+    end
+  end
+
+  def sampled?
+    flags.sampled?
+  end
+
+  def to_s
+    sprintf("00-%032x-%016x-%s", trace_id, span_id, flags.to_s)
```
ibawt: the text seems to indicate you would parse the version, but then output as the IMPLEMENTATION version

ibawt:
@fbogsany we could also just not parse the headers at and leave teh trace-ids as hex encoded strings?  it would require 2x the memory,  but would omit serialization and deserializations.  Kind of feel like it should be a Number. maybe we should profile first.

fbogsany:
> we could also just not parse the headers at and leave teh trace-ids as hex encoded strings?

See discussion in #57. A similar point was raised there. I don't mind leaving them as hex strings, but we should still validate them.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,108 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+class OpenTelemetry::DistributedContext::Propagation::TraceParent
```
fbogsany: We've generally marched to the right:
```ruby
module OpenTelemetry
  module DistributedContext
    module Propagation
      class TraceParent
```

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,108 @@
...
+  TRACE_PARENT_HEADER = "traceparent"
+  SUPPORTED_VERSION = 0
+  MAX_VERSION = 254
+
+  REGEXP = /^([A-Fa-f0-9]{2})-([A-Fa-f0-9]{32})-([A-Fa-f0-9]{16})-([A-Fa-f0-9]{2})(-.*)?$/.freeze
+  private_constant :REGEXP
+
+  attr_reader :version, :trace_id, :span_id, :flags
+
+  class Flags
```
fbogsany: #57 adds `TraceFlags`. Maybe use that instead?

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,108 @@
...
+
+  def sampled?
+    flags.sampled?
+  end
+
+  def to_s
+    sprintf("00-%032x-%016x-%s", trace_id, span_id, flags.to_s)
+  end
+
+  def initialize(trace_id: nil, span_id: nil, version: 0, flags: Flags.new)
```
fbogsany: `... version: SUPPORTED_VERSION, ...` ?

fbogsany: And maybe `..., flags: TraceFlags::DEFAULT)` ?

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +18,21 @@ module Propagation
       # Propagation is usually implemented via library-specific request interceptors, where the client-side injects values
       # and the server-side extracts them.
       class HTTPTextFormat
-        # TODO
+        def extract(carrier, getter)
+          tp = TraceParent.from_string(getter.get(carrier, TraceParent::TRACE_PARENT_HEADER))
+
+          SpanContext.new(trace_id: tp.trace_id, span_id: tp.span_id)
```
fbogsany: This should pass the `trace_flags` as well.

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -3,6 +3,7 @@
 # Copyright 2019 OpenTelemetry Authors
 #
 # SPDX-License-Identifier: Apache-2.0
+require_relative './trace_parent'
```
fbogsany: We've preferred to add `require`s at a higher level, rather than using `require_relative`. I don't know if there's any justification for that, though I recall reading somewhere that `require_relative` was "bad" in some way (somewhere on the internet, so it must be true). In this case, I'd add it to `api/lib/opentelemetry/distributed_context/propagation.rb`.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: sure

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: I can rebase off it

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: sure

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
ibawt: sounds good

ibawt:
@fbogsany comments have been addressed

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        def parse_span_id(string)
+          raise InvalidSpanIDError, string if string == OpenTelemetry::Trace::INVALID_SPAN_ID
+
+          string.downcase
+        end
+
+        def parse_flags(string)
+          OpenTelemetry::Trace::TraceFlags.from_byte(string.to_i(16))
```
fbogsany: D'you think we should add a `from_string` to `TraceFlags` and push the responsibility there?

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,101 @@
...
+
+          @version = parse_version(matches[1])
+          raise InvalidFormatError if version > SUPPORTED_VERSION && string.length < 55
+
+          @trace_id = parse_trace_id(matches[2])
+          @span_id = parse_span_id(matches[3])
+          @flags = parse_flags(matches[4])
+
+          self
+        end
```
fbogsany: Can we do all the parsing in `from_string`, call the constructor with all the parsed things there, and remove this method?

api/test/opentelemetry/distributed_context/propagation/trace_parent_test.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+require 'test_helper'
+
+describe OpenTelemetry::DistributedContext::Propagation::TraceParent do
+  TraceParent = OpenTelemetry::DistributedContext::Propagation::TraceParent
```
fbogsany: Sneaky 👍

api/lib/opentelemetry/distributed_context/propagation/binary_format.rb
```diff
@@ -22,7 +22,7 @@ def to_bytes(span_context)
         def from_bytes(bytes)
           raise ArgumentError if bytes.nil?
 
-          SpanContext.INVALID
+          OpenTelemetry::Trace::SpanContext.invalid
```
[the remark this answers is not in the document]
ibawt: I think that was from your pr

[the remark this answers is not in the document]
ibawt: but I'll go around and see what prefixes I can drop.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: does the opentelemetry spec list the number of bytes and the hex format?  ie 2 digit hex.  
if it lives in the w3c but not in the OT spec I would say that would govern where it should live.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        def sampled?
+          flags.sampled?
+        end
+
+        def to_s
+          "00-#{trace_id}-#{span_id}-#{flags}"
+        end
+
+        def initialize(trace_id: nil, span_id: nil, version: SUPPORTED_VERSION, flags: OpenTelemetry::Trace::TraceFlags::DEFAULT)
```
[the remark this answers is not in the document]
ibawt: yah that sounds good.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: sure

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +16,22 @@ module Propagation
       # Propagation is usually implemented via library-specific request interceptors, where the client-side injects values
       # and the server-side extracts them.
       class HTTPTextFormat
-        # TODO
+        def extract(carrier, &block)
```
mwear: Would an implicit block be better since we opt for yielding and not using the explicit `&block` param.

In either case can we document the parameters to this method as well?

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +16,22 @@ module Propagation
...
+          header = yield carrier, TraceParent::TRACE_PARENT_HEADER
+          tp = TraceParent.from_string(header)
+
+          SpanContext.new(trace_id: tp.trace_id, span_id: tp.span_id, flags: tp.flags)
+        rescue OpenTelemetry::Error
+          SpanContext.new
+        end
+
+        def inject(context, carrier, &block)
```
mwear: Same comments as for `extract`.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,102 @@
...
+        InvalidTraceIDError = Class.new(Error)
+        InvalidSpanIDError = Class.new(Error)
+
+        TRACE_PARENT_HEADER = 'traceparent'
+        SUPPORTED_VERSION = 0
+        private_constant :SUPPORTED_VERSION
+        MAX_VERSION = 254
+        private_constant :MAX_VERSION
+
+        REGEXP = /^([A-Fa-f0-9]{2})-([A-Fa-f0-9]{32})-([A-Fa-f0-9]{16})-([A-Fa-f0-9]{2})(-.*)?$/.freeze
```
mwear: Would this (and the parsing) be more clear if we used named captures?

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,102 @@
...
+
+            raise InvalidVersionError, v if v > MAX_VERSION
+
+            v
+          end
+
+          def parse_trace_id(string)
+            raise InvalidTraceIDError, string if string == OpenTelemetry::Trace::INVALID_TRACE_ID
+
+            string.downcase
```
mwear: Can we use the mutating `downcase!` here? It would save us a string allocation. If so, we can do the same in `parse_span_id`.

ibawt: Sure

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
ibawt: imo with an implicit block you have to read the documentation/implementation rather than the function signature.   If you feel strongly I don't mind.

documentation for sure.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: sure it's easy to add.

ibawt:
re: rebase, now you know why I was poking at getting the other merged :D

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
ibawt: I did the changes, have a look.  Not so sure about my documentation though.

api/lib/opentelemetry/trace/trace_flags.rb
```diff
@@ -25,6 +25,10 @@ def from_byte(flags)
         end
       end
 
+      def to_s
```
fbogsany: Add a test?

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,104 @@
...
+          end
+
+          def parse_version(string)
+            v = string.to_i(16)
+            raise InvalidFormatError, string unless v
+
+            raise InvalidVersionError, v if v > MAX_VERSION
+
+            v
+          end
```
fbogsany: The empty lines in this method weird me out.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,104 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+module OpenTelemetry
+  module DistributedContext
+    module Propagation
+      # A TraceParent is an implemenation of the W3C trace context specification
```
fbogsany: s/implemenation/implementation/

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,104 @@
...
+
+        def to_s
+          "00-#{trace_id}-#{span_id}-#{flags}"
+        end
+
+        def self.from_context(ctx)
+          new(trace_id: ctx.trace_id, span_id: ctx.span_id, flags: ctx.trace_flags)
+        end
+
+        def self.from_string(string)
```
fbogsany: Would be good to document all the public methods here. Especially the ones that `raise`.

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
mwear: Looks good. The alternative would have been to use the block param (block.call) instead of using `yield`, but its slightly faster to use yield with an implicit block.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
fbogsany: The OT spec says TraceFlags is a single byte of boolean flags. Otherwise it defers to the W3C spec.

ibawt: then it should live here, as this is using a hex encoding of a byte.  Therefore the at the boundary it should be a native byte.

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
ibawt: So I looked through the rails source, and I think a good rule is, only name the block if we pass it down to another method.  Then it's clear the yield is getting "passed" as it were.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
ibawt: it's the void sucking you in...there's a lovecraftian horror in there.

api/lib/opentelemetry/trace/trace_flags.rb
ibawt: so I moved this into the trace parent, as the trace parent defines the flags it can support due to it's version.  It also says that we can only send flags we know about re: version.

ibawt:
RE: http_text_format class, I'm not too sold on the "getter" and "setter" as blocks.  But I'd like to see how this will integrate with the rest of the code before bike shedding on it too much.  Hence most of the `meat` is in the trace_parent class.

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +16,35 @@ module Propagation
...
+
+          SpanContext.new(trace_id: tp.trace_id, span_id: tp.span_id, flags: tp.flags)
+        rescue OpenTelemetry::Error
+          SpanContext.new
+        end
+
+        # inject will set the span context on the supplied carrier
+        # @param [Context] the carrier
+        # @yield [Carrier, String, String] carrier, header key, header value
+        # {SpanContext}
```
fbogsany: Don't need this. We don't return the `SpanContext` (at least not intentionally - maybe we should explicitly return `self`).

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +16,35 @@ module Propagation
       # Propagation is usually implemented via library-specific request interceptors, where the client-side injects values
       # and the server-side extracts them.
       class HTTPTextFormat
-        # TODO
+        # extract will return a SpanContext from the supplied carrier
+        # invalid headers will result in a new SpanContext
+        # @param [Carrier] the carrier to get the header from
+        # @yield [Carrier, String] the header key
+        # {SpanContext}
```
fbogsany: `@return [SpanContext]`

api/lib/opentelemetry/distributed_context/propagation/http_text_format.rb
```diff
@@ -17,7 +16,35 @@ module Propagation
...
+        def extract(carrier)
+          raise Error, 'block must be supplied' unless block_given?
```
fbogsany: I think this should be an ordinary `ArgumentError` rather than the OT `Error`. Same thing below in `inject`.

api/lib/opentelemetry/distributed_context/propagation/trace_parent.rb
```diff
@@ -0,0 +1,124 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+module OpenTelemetry
+  module DistributedContext
+    module Propagation
+      # A TraceParent is an implementation of the W3C trace context specification
+      # https://www.w3.org/TR/trace-context/
+      # {SpanContext}
+      class TraceParent
```
mwear: The layout of this class is not what I am used to seeing. I'm not sure if we have a convention, or if we should have one, but typically I'm used to seeing something like:

```ruby
class MyClass
  #constants
  MY_CONST = 1
  class << self
     #public class methods
     def public_class_method; end
     
     private
     
    #private class methods
     def private_class_method; end
  end
  
  #public instance methods

  def public_instance_method; end

  private
  
  #private instance methods
  def private_instance_method; end
end
```

Would there be any objections to adopting a similar format for OTel Ruby (and this file)?

ibawt: so you want the "static" methods up top?  That doesn't bother me at all, I'll move it up.

mwear: Yeah, there are some that are in the `class << self` block and a few that are `self.` prefixed. If you wouldn't mind moving them into one `class << self` block at the top that'd be awesome!

mwear: Looks great!

PR #60, opened 2019-09-06
Implement ProbabilitySampler

fbogsany:
This implements ProbabilitySampler based on the [Java implementation](https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/samplers/ProbabilitySampler.java).


sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb
```diff
@@ -8,8 +8,87 @@ module OpenTelemetry
...
+
+            # If the parent is sampled keep the sampling decision.
+            if span_context&.trace_flags&.sampled?
+              SAMPLE_DECISION
+            elsif links&.any? { |link| link.context.trace_flags.sampled? }
+              # If any parent link is sampled keep the sampling decision.
+              SAMPLE_DECISION
+            else
+              # TODO: what should we assume about the distribution of trace_id
+              # bits? How best to compare against @probability?
```
fbogsany: Need to figure out the "best" way to compute the sampling decision. I don't think the Java implementation is ideal here. On the other hand, the distribution of `trace_id.hash` probably isn't great either - we have fixed-length IDs with a limited range of characters. I may be totally overthinking that, though. The harder part of this is the scaled comparison - we don't know the range of `#hash` values. In practice, it is truncated to a `Fixnum`, but that detail, and the range of `Fixnum` is obscured in MRI 2.5+.

The OpenTelemetry Service has a probability sampler that computes the murmur3 hash of the trace ID. A Ruby implementation of that won't be efficient on MRI, and we don't want to introduce C extensions in the SDK.

mwear: It might not be necessary to use the trace_id when computing the probability at all. The javascript implementation basis its decision off `Math.random()`, see: https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-core/src/trace/sampler/ProbabilitySampler.ts

fbogsany: 👍 that's exactly what we do in our internal tracing gem at Shopify. Happy to stick with that.

fbogsany:
This is ready for re-review.

sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb
```diff
@@ -8,8 +8,87 @@ module OpenTelemetry
...
+            # @raise [ArgumentError] if {probability} is out of range
+            # @return [OpenTelemetry::Trace::Samplers::Sampler]
+            def create(probability)
+              raise ArgumentError, 'probability must be in range [0.0, 1.0]' unless (0.0..1.0).include?(probability)
+
+              return OpenTelemetry::Trace::Samplers::NeverSampleSampler.new if probability == 0.0
+              return OpenTelemetry::Trace::Samplers::AlwaysSampleSampler.new if probability == 1.0
+
+              new(probability)
+              # idUpperBound = (long) (probability * Long.MAX_VALUE);
```
mwear: We might not need this commented out Java code.

sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb
```diff
@@ -8,8 +8,88 @@ module OpenTelemetry
...
+            # probability.
+            #
+            # @param [Numeric] probability The desired probability of sampling.
+            #   Must be within [0.0, 1.0].
+            # @raise [ArgumentError] if probability is out of range
+            # @return [OpenTelemetry::Trace::Samplers::Sampler]
+            def create(probability)
+              raise ArgumentError, 'probability must be in range [0.0, 1.0]' unless (0.0..1.0).include?(probability)
+
+              return OpenTelemetry::Trace::Samplers::NeverSampleSampler.new if probability == 0.0
```
fbogsany: This interferes with the logic in `decision`. Specifically, it does not respect sampling decisions made by the parent span or links. Which raises a couple of interesting questions:
1. Should `probability == 0` be equivalent to "never sample", even in the face of upstream decisions?
2. Should `NeverSampleSampler` respect upstream decisions?

The spec currently says nothing about `ProbabilitySampler`. It *is* specific about [`NeverSampleSampler`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-sampling.md):

> * Never sample. `Sampler` returns `Decision` with `IsSampled=false` and empty arguments collection. Description MUST be `NeverSampleSampler`.

mwear: These are good questions. I'll see if I can get some clarification at the spec meeting tomorrow.

mwear: The [sampling OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0006-sampling.md#default-samplers) has some additional information. From my read, the NeverSampleSampler (ALWAYS_OFF) sampler should not respect upstream decisions. It also says the probability sampler can be configured to ignore the parent sampling decision. So, it would be equivalent to the NeverSampleSampler when probability is 0.0 and its configured to ignore the parent sampling decision.

It still doesn't tell us exactly what we should do for the "classic" sampling specification, which is what we're currently working with. Since we don't have the ability to configure whether or not to ignore the parent sampling decision, we should probably respect it. Alternatively, we could probably just go with what we have if it's going to be replaced by the OTEP anyway.

fbogsany: I removed it. We can figure out the right thing to do when we implement the OTEP.

sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb
```diff
@@ -8,8 +8,87 @@ module OpenTelemetry
...
+            else
+              DONT_SAMPLE_DECISION
+            end
+          end
+
+          # Returns a description of the sampler
+          #
+          # @return [String]
+          def description
+            format('ProbabilitySampler{%.6f}', @probability)
```
mwear: Should we memoize the description, or is that a premature optimization?

fbogsany: I'd say premature.

PR #61, opened 2019-09-06
add a code coverage tool to the test suite

ibawt:
Adding a code coverage thingy so hopefully we won't put more untested stuff in.

re: full test coverage


ibawt:
accepting bike sheds on the min. coverage value.  It's ~86% now, so I rounded down.

fbogsany:
LGTM, but please add to the SDK side as well.

sdk/test/test_helper.rb
```diff
@@ -0,0 +1,12 @@
...
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'simplecov'
+SimpleCov.start
+SimpleCov.minimum_coverage 85
+
+require 'opentelemetry'
```
fbogsany: Should be `opentelemetry/sdk`.

This is going to be a merge conflict with #59 but I'll deal with that.

api/test/test_helper.rb
```diff
@@ -4,5 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'opentelemetry'
+require 'simplecov'
+SimpleCov.start
+
+require 'opentelemetry/sdk'
```
fbogsany: Should be just `opentelemetry` - this is the API test helper, not the SDK.

sdk/test/test_helper.rb
ibawt: whoops I thought I added that.

api/test/test_helper.rb
ibawt: oh I see what happened....I should pay more attention to what *path* that file was in :|

PR #62, opened 2019-09-06
add a logger

ibawt:
Given the design of having no runtime impact, I thnk that error conditions will end up being eaten and not reported.  While I don't think the user of the library will need to handle the cases directly it'd be good know if for instance they screwed up the header for propagation etc etc.

@fbogsany


api/lib/opentelemetry/logger.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+require 'logger'
+
+module OpenTelemetry
+  class << self
+    attr_accessor :logger
```
indrekj: would it make more sense to put it to https://github.com/open-telemetry/opentelemetry-ruby/blob/2a22a85e09a2c4823a804d71481c3e4cc606f5fd/api/lib/opentelemetry.rb which already includes similar things and it's on top level also ?

ibawt: we could for sure.  I started this thinking we would subclass `Logger` but I didn't see any reason at the moment. 

any opinions @fbogsany ?

fbogsany: Nope, no opinions.

ibawt: lets see what everyone thinks.  I'm ok either way.

ibawt:
👀

api/lib/opentelemetry/logger.rb
ibawt: I moved it in

PR #69, opened 2019-09-09
Implement lazy events for Span

### LINKED ISSUE #36: API: implement Span.add_lazy_event
[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#add-events). Need to decide whether to provide an `Event` interface or a concrete `Event` class and an `EventFormatter`.

mwear:
This fixes #36.

This PR modifies `Span#add_event` to accept an EventFormatter to facilitate the creation of lazy events. The approach is similar to what was done for lazy links (#54).


api/lib/opentelemetry/trace/event.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    # A text annotation with a set of attributes.
+    class Event
+      EMPTY_ATTRIBUTES = {}.freeze
```
fbogsany: Does this need to be public?

mwear: It definitely doesn't. We had the same in the `Link` class. I went ahead and changed the visibility for both of them.

PR #72, opened 2019-09-10
SDK implementation of Span and TraceConfig

fbogsany:
This adds the Span and TraceConfig implementations to the SDK. It is a fairly major overhaul of the API. It moves Link and Event to the SDK, removing their definitions from the API. Instead, it adds factory methods to the Tracer API. This lets us provide simple concrete implementations in the SDK, and avoids exposing the types in the API, providing more freedom for alternative implementations.

Note that this kind of reorganization is necessary because the SDK needs to be able to impose limits on Link attributes, in particular.


fbogsany:
@ibawt I'd love some 👀 on this. It still needs tests, but I think everything else is ready for review.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -1,43 +0,0 @@
-# frozen_string_literal: true
-
-# Copyright 2019 OpenTelemetry Authors
-#
-# SPDX-License-Identifier: Apache-2.0
-
-module OpenTelemetry
```
mwear: Doesn't this need to part of the API? Presumably a user could have code such as:

```Ruby
span.add_link(-> { OpenTelemetry::Trace::Link.new(ctx) }) 
```
and this should "work" with the minimal implementation. Events will be similar.

mwear: I suppose that would "work" so long as the proc isn't executed.

fbogsany: With this API, that turns into:
```ruby
span.add_link(-> { tracer.create_link(ctx) })
```

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,214 @@
 module OpenTelemetry
   module SDK
     module Trace
+      # Implementation of {OpenTelemetry::Trace::Span} that records trace events.
+      #
+      # rubocop:disable Metrics/ClassLength
       class Span < OpenTelemetry::Trace::Span
-        # TODO
+        # TODO: does this need synchronization? I don't think so...
+        attr_reader :name
```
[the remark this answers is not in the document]
fbogsany: I believe Ruby's memory model prevents this. Access to a single variable is effectively atomic, whereas a Mutex is required for atomic access to multiple variables. It *is* possible to read "old" values, though probably not in MRI: https://docs.google.com/document/d/1pVzU8w_QF44YzUCCab990Q_WZOdhpKolCIHaiXG-sPw/pub

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,214 @@
...
+        #   only be used when passing in a SpanContext, not a LinkFormatter.
+        #
+        # @return [self] returns itself
+        def add_link(span_context_or_link_formatter, attrs = nil)
+          super
+          link =
+            if span_context_or_link_formatter.instance_of?(SpanContext)
+              OpenTelemetry::Trace::Link.new(span_context_or_link_formatter, attrs)
+            else
+              span_context_or_link_formatter.call
```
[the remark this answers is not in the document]
fbogsany: The benefit comes about because a decision **not** to sample results in a no-op Span, which won't call the callable. For positive sampling decision, the Span would eventually be finished anyway, so deferring the call of the callable isn't terribly useful.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,236 @@
...
+        #
+        # @param [String] key
+        # @param [String, Boolean, Numeric] value
+        #
+        # @return [self] returns itself
+        def set_attribute(key, value)
+          super
+          @mutex.synchronize do
+            if @ended
+              logger.debug('Calling set_attribute on an ended Span.')
```
ibawt: this should probably be a warn

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,236 @@
...
+        # @param [Time] timestamp optional timestamp for the event.
+        #
+        # @return [self] returns itself
+        def add_event(name_or_event_formatter, attrs = nil, timestamp: nil)
+          super
+          timed_event =
+            if name_or_event_formatter.instance_of?(String)
+              TimedEvent.new(name_or_event_formatter, attrs, timestamp || Time.now)
+            else
+              name_or_event_formatter.call
```
ibawt: in this case should you pass the attrs and timestamp to the .call ?

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,236 @@
...
+        # Finishes the Span
+        #
+        # Implementations MUST ignore all subsequent calls to {#finish} (there
+        # might be exceptions when Tracer is streaming event and has no mutable
+        # state associated with the Span).
+        #
+        # Call to {#finish} MUST not have any effects on child spans. Those may
+        # still be running and can be ended later.
+        #
+        # This API MUST be non-blocking.
```
ibawt: pedantically this method will block under contention :P

sdk/lib/opentelemetry/sdk/trace/config/trace_config.rb
```diff
@@ -0,0 +1,93 @@
...
+          #
+          # @return [TraceConfig] with the desired values.
+          # @raise [ArgumentError] if any of the max numbers are not positive.
+          def initialize(sampler: DEFAULT_SAMPLER,
+                         max_attributes_count: DEFAULT_MAX_ATTRIBUTES_COUNT,
+                         max_events_count: DEFAULT_MAX_EVENTS_COUNT,
+                         max_links_count: DEFAULT_MAX_LINKS_COUNT,
+                         max_attributes_per_event: DEFAULT_MAX_ATTRIBUTES_PER_EVENT,
+                         max_attributes_per_link: DEFAULT_MAX_ATTRIBUTES_PER_LINK)
+            raise ArgumentError, 'max_attributes_count' unless max_attributes_count.positive?
```
ibawt: should the message have `must be positive` or something

sdk/lib/opentelemetry/sdk/trace/span.rb
fbogsany: 👍 adding clarification.

sdk/lib/opentelemetry/sdk/trace/span.rb
fbogsany: No. The attrs and timestamp are only supplied for "eagerly evaluated" Events, not for lazily-evaluated ones.

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -43,20 +44,64 @@ def with_span(span)
...
+      # @param [optional Hash<String, Object>] attrs One or more key:value
+      #   pairs, where the keys must be strings and the values may be string,
+      #   boolean or numeric type.
+      # @param [optional Time] timestamp Optional timestamp for the event.
+      # @return a new Event.
+      def create_event(name, attrs = nil, timestamp: nil)
+        raise ArgumentError unless name.is_a?(String)
+        raise ArgumentError unless Internal.valid_attributes?(attrs)
+
+        EVENT_OR_LINK
```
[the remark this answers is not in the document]
mwear: If there is not an interface for a Link or Event, then there's no issue. Returning `EVENT_OR_LINK` makes complete sense!

sdk/lib/opentelemetry/sdk/trace/span.rb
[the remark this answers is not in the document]
mwear: I was thinking about the case where a link isn't used because max_links was reached, but this is probably not a huge concern and would fall into the premature optimization category.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -7,9 +7,241 @@
...
+        #   or an EventFormatter, a lazily evaluated callable that returns an
+        #   Event instance.
+        # @param [optional Hash<String, Object>] attrs One or more key:value pairs, where
+        #   the keys must be strings and the values may be string, boolean or
+        #   numeric type. This argument should only be used when passing in a
+        #   name, not an EventFormatter.
+        # @param [optional Time] timestamp Timestamp for the event.
+        #
+        # @return [self] returns itself
+        def add_event(name_or_event_formatter, attrs = nil, timestamp: nil)
```
rdooley: @sveiss and I found that this has some unfortunate side effects (further context here)
https://bugs.ruby-lang.org/issues/12717 

One suggestion might be changing these 3 to all be pos arguments
```ruby
def add_event(name_or_event_formatter, attrs = nil, timestamp = nil)
```

fbogsany: @ibawt and I had concerns about the ergonomics of this API anyway. If we leave it as is, I'd prefer to go the other way with keyword args for both optionals:
```ruby
def add_event(name_or_event_formatter, attributes: nil, timestamp: nil)
```
However, the "lazy" form of this requires a lambda instead of the more natural (IMO) block. I'd prefer either splitting the API into `add_event(name, attributes: nil, timestamp: nil)` and `add_lazy_event(&block)` **or** taking an implicit block and making all other args optional:
```ruby
def add_event(name: nil, attributes: nil, timestamp: nil)
  raise ArgumentError unless block_given? == (name.nil? && attributes.nil? && timestamp.nil?)
  raise ArgumentError unless Internal.valid_attributes?(attrs)
  # ^ in the API, v in the SDK
  event = block_given? ? yield : Event.new(name: name, attributes: attributes, timestamp: timestamp || Time.now)
  @mutex.synchronize do ... end
end
```

sdk/lib/opentelemetry/sdk/trace/config/trace_config.rb
```diff
@@ -0,0 +1,109 @@
...
+          end
+
+          # @api private
+          # Removes oldest entries from an Array whose size exceeds a limit.
+          #
+          # @param [Array] arr Array of links or events that is modified in-place.
+          # @param [Symbol, Integer] limit The maximum number of entries
+          #   allowed in the array. May be expressed as a number or a symbol
+          #   representing one of the limits in this class.
+          def trim_links_or_events(arr, limit)
```
fbogsany: I'm going to walk back my suggestions from Slack. I don't think the metaprogramming is warranted here. It's convenient for `trim_attributes` because we have 3 different attribute limits, but here both the API and the implementation are simpler with separate `trim_links` and `trim_events`.

Along those lines, I might replace the metaprogramming in `trim_attributes` with simple wrapper methods for each case - that's easier to understand and probably more performant.

fbogsany: Done

sdk/lib/opentelemetry/sdk/trace/span.rb
fbogsany: I went with the second approach.

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -7,4 +7,220 @@
...
+      span.finish
+      span.instance_variable_get(:@ended).must_equal(true)
+    end
+
+    it 'does not allow ending more than once' do
+      span.finish
+      span.instance_variable_get(:@ended).must_equal(true)
+      ts = span.instance_variable_get(:@end_timestamp)
+      span.finish
+      span.instance_variable_get(:@end_timestamp).must_equal(ts)
```
fbogsany: I'd prefer to remove most/all of the `instance_variable_get` from these tests. I've done that in many cases above by using `to_proto` (which is really `.to_h`, but will eventually be a real thing).

rdooley: I was briefly using `to_proto` but had switched to `instance_variable_get` to avoid the tests depending on the in effect `to_h` method.
No personal preference, I'll switch em back

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -7,4 +7,220 @@
...
+      span = Span.new(context, 'name', SpanKind::INTERNAL, nil, trace_config,
+                      mock_span_processor, nil, nil, nil, Time.now)
+      mock_span_processor.expect :on_end, nil, [span]
+      span.finish
+      mock_span_processor.verify
+    end
+
+    it 'marks the span as ended' do
+      span.finish
+      span.instance_variable_get(:@ended).must_equal(true)
```
rdooley: Rather than add `ended` to the `to_proto` export this is left as `instance_variable_get`

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -7,4 +7,220 @@
...
+                      span_processor, attributes, nil, nil, Time.now)
+      span.to_proto[:total_recorded_attributes].must_equal(2)
+      span.to_proto[:attributes].length.must_equal(1)
+    end
+
+    it 'counts attributes' do
+      attributes = { 'foo': 'bar', 'other': 'attr' }
+      span = Span.new(context, 'name', SpanKind::INTERNAL, nil, trace_config,
+                      span_processor, attributes, nil, nil, Time.now)
+      span.instance_variable_get(:@total_recorded_attributes).must_equal(2)
```
rdooley: `to_proto` will freeze attributes so I left this as an `instance_variable_get` rather than create the span twice in the test to test both ways attributes can be added

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -7,4 +7,220 @@
...
+      span.instance_variable_get(:@total_recorded_events).must_equal(2)
+      span.add_event(name: 'added')
+      span.to_proto[:total_recorded_events].must_equal(3)
+    end
+
+    it 'trims excess events' do
+      events = %w[foo bar]
+      span = Span.new(context, 'name', SpanKind::INTERNAL, nil, trace_config,
+                      span_processor, nil, nil, events, Time.now)
+      span.instance_variable_get(:@events).size.must_equal(1)
```
rdooley: similar re: events getting frozen

sdk/test/opentelemetry/sdk/trace/span_test.rb
fbogsany: I replaced the `to_proto` with `to_span_data`, which returns a `Struct` called `SpanData`. I *think* this is how we want to package spans for export, but that's by no means final.

sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb
```diff
@@ -29,8 +29,7 @@ def export(spans)
               begin
                 merge_result_code(result_code, span_exporter.export(spans))
               rescue => e # rubocop:disable Style/RescueStandardError
-                # TODO: log instead of puts
-                puts "WARNING: exception thrown by export - #{e}"
+                logger.warn("exception raised by export - #{e}")
```
sveiss: @fbogsany is this meant to be `OpenTelemetry.logger`? 

I've just rebased #98 against master with this PR merged, and this is now failing with "undefined local variable or method `logger'" in my newly added tests.

fbogsany: Yes, it should be. Is there a reason why `OpenTelemetry.logger` wouldn’t be accessible as `logger` here? CI passed for this branch pre-merge.

PR #74, opened 2019-09-10
Batch span processor

ibawt:
For context, the (draft) spec is here: https://github.com/open-telemetry/opentelemetry-specification/pull/205/files


ibawt:
@fbogsany want to have a look?  Looking for inspiration for non racy more thready tests.

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +22,92 @@ module Export
...
+          attr_reader :spans, :max_queue_size, :batch_size
+
+          def work
+            loop do
+              keep_running = nil
+              batch = nil
+              lock do
+                if  spans.size < max_queue_size
+                  loop do
+                    break if !spans.empty? || !@keep_running
```
fbogsany:
```ruby
until !spans.empty? || !@keep_running
```

fbogsany: Or, clearer IMO:
```ruby
while spans.empty? && @keep_running
```

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +22,92 @@ module Export
...
+                    @condition.wait(@mutex, @delay)
+                  end
+                end
+                keep_running = @keep_running
```
fbogsany: Why not `break unless @keep_running` here and avoid the local var? Then rely on `flush` in the `shutdown` to do the final exports?

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +22,92 @@ module Export
...
+            @thread = Thread.new { work }
+          end
+
+          # does nothing for this processor
+          def on_start(span)
+            # noop
+          end
+
+          # adds a span to the batcher, threadsafe may block on lock
+          def on_end(span)
```
mwear: Whose responsibility is it to check if the span was sampled? Should that happen here, or further up?

rdooley:
I signed it

rdooley:
I signed it

rdooley:
I signed it

rdooley:
I signed it

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +21,90 @@ module Export
...
+                if spans.size < max_queue_size
+                  @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
+                  keep_running = @keep_running
+                end
+                batch = fetch_batch
+              end
+              # this is done outside the lock to unblock the producers
+              @exporter.export(batch)
+              break unless keep_running
+            end
```
fbogsany: Simpler:
```ruby
loop do
  batch = lock do
    if spans.size < max_queue_size
      @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
    end
    break unless @keep_running
    fetch_batch
  end
  # this is done outside the lock to unblock the producers
  @exporter.export(batch)
end
flush
```

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +21,90 @@ module Export
...
+            flush
+          end
+
+          def flush
+            @exporter.export(fetch_batch) until spans.empty?
```
fbogsany: This needs to lock while calling `fetch_batch`. It is also possible that the producers will keep producing indefinitely, outpacing the `flush`. It is preferable to take a snapshot of the `spans` and flush those, while ignoring any spans added after the snapshot.

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
```diff
@@ -21,7 +21,90 @@ module Export
...
+
+          def fetch_batch
+            batch = []
+            loop do
+              break if batch.size >= @batch_size || spans.empty?
+
+              batch << spans.shift.to_span_proto
+            end
+            batch
+          end
```
fbogsany: `shift(n)` will do the same thing more efficiently:
```ruby
spans.shift(@batch_size).map!(&:to_proto)
```

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
ibawt: oh I like using block's return value. very functional

ibawt:
@rdooley did you want to finish this off or are you busy?

rdooley:
@ibawt sure thing, I'll get right on this after getting the CI running for sdk as well as api

sdk/lib/opentelemetry/sdk/trace/export/batch_sampled_span_processor.rb
rdooley: @fbogsany `BatchRecordingEventsSpanProcessor` ?

fbogsany: I'm proposing just calling it the `BatchSpanProcessor`.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+          def on_start(span)
+            # noop
+          end
+
+          # adds a span to the batcher, threadsafe may block on lock
+          def on_end(span)
+            return unless span.recording_events?
+
+            lock do
+              spans.shift if spans.size >= max_queue_size
```
fbogsany: If we think it is possible for `spans.size > max_queue_size`, then we should use `spans.shift(n)` here, where `n = spans.size - max_queue_size - 1`. Otherwise we'll have a slow leak.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+
+          attr_reader :spans, :max_queue_size, :batch_size
+
+          def work
+            loop do
+              keep_running = nil
+              batch = lock do
+                if spans.size < max_queue_size
+                  @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
+                  keep_running = @keep_running
```
fbogsany:
```ruby
break unless @keep_running
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+                end
+                fetch_batch
+              end
+              # this is done outside the lock to unblock the producers
+              @exporter.export(batch)
+              break unless keep_running
```
fbogsany: Remove this `break`

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+            @thread.join
+          end
+
+          private
+
+          attr_reader :spans, :max_queue_size, :batch_size
+
+          def work
+            loop do
+              keep_running = nil
```
fbogsany: Remove this local var.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+              end
+              # this is done outside the lock to unblock the producers
+              @exporter.export(batch)
+              break unless keep_running
+            end
+            flush
+          end
+
+          def flush
+            until spans.empty?
```
fbogsany: This should grab a snapshot of `spans` rather than continuing to flush until the global queue is empty. The concern is that other threads could continue to enqueue spans, and we will end up flushing "forever".

I think it'd be cleaner to:
```ruby
def flush
  snapshot = lock { spans.shift(spans.size) }
  @exporter.export(snapshot.shift(@batch_size).map!(&:to_span_data)) until snapshot.empty?
end
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+              batch = lock { fetch_batch }
+              @exporter.export(batch)
+            end
+          end
+
+          def fetch_batch
+            spans.shift(@batch_size).map!(&:to_span_proto)
```
fbogsany: Should be `:to_span_data` instead of `:to_span_proto`.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+  end
+
+  describe 'batching' do
+    it 'should batch up to but not over the max_batch' do
+      te = TestExporter.new
+
+      bsp = BatchSpanProcessor.new(exporter: te, max_queue_size: 6, max_export_batch_size: 3)
+
+      tss = [[TestSpan.new, TestSpan.new, TestSpan.new, TestSpan.new]]
```
fbogsany: Why the nested array here?

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+      bsp.on_end(tss[0][0])
+      bsp.on_end(tss[0][1])
+      bsp.on_end(tss[0][2])
+      bsp.on_end(tss[0][3])
```
fbogsany: `tss.each { |ts| bsp.on_end(ts) }`

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+    end
+
+    it 'should batch only recording_events samples' do
+      te = TestExporter.new
+
+      bsp = BatchSpanProcessor.new(exporter: te, max_queue_size: 6, max_export_batch_size: 3)
+
+      tss = [[TestSpan.new, TestSpan.new(nil, false)]]
+      bsp.on_end(tss[0][0])
+      bsp.on_end(tss[0][1])
```
fbogsany: As above, you don't need the nested array, and you can just iterate over the array rather than adding each element explicitly.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+      te.batches[0].size.must_equal(1)
+    end
+  end
+
+  describe 'stress test' do
+    it 'shouldnt blow up with a lot of things' do
+      te = TestExporter.new
+
+      bsp = BatchSpanProcessor.new(exporter: te)
+      producers = 0.upto(9).map do |i|
```
fbogsany: `10.times do |i|`

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+        Thread.new do
+          x = i * 10
+          0.upto(9) do |j|
```
fbogsany: `10.times do |i|`

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,112 @@
...
+          end
+          sleep(rand(0.01))
+        end
+      end
+      producers.each(&:join)
+      bsp.shutdown
+
+      out = te.batches.flatten.map(&:id).sort
+
+      expected = 0.upto(99).map { |i| i }
```
rdooley: `100.times.map...`

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,108 @@
...
+              keep_running = nil
+              batch = lock do
+                if spans.size < max_queue_size
+                  @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
+                  keep_running = @keep_running
+                end
+                fetch_batch
+              end
+              # this is done outside the lock to unblock the producers
+              @exporter.export(batch)
```
fbogsany: We're not paying attention to the return value here, but `export` should return `FAILED_RETRYABLE`, `FAILED_NOT_RETRYABLE` or `SUCCESS`. For `FAILED_RETRYABLE`, we should consider re-enqueuing the spans.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
rdooley: Switching to only breaking here means we will never break if `spans.size` never hits `max_queue_size` like in the lifecycle test where we start and then shutdown a bsp https://github.com/open-telemetry/opentelemetry-ruby/pull/74/files/6f4bc137866bf0d52e2078c448945475840233b5#diff-52e55bcc9048d069724858fb66e63b9eR40

I think we need both the `break unless @keep_running` here and on ln 83

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
rdooley: I will add this and corresponding tests in a subsequent commit.

rdooley: oh we need this in flush as well

fbogsany: `flush` probably should not care - it is very much best-effort at that point, since we're shutting down

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,110 @@
...
+
+          # shuts the consumer thread down and flushes the current accumulated buffer
+          # will block until the thread is finished
+          def shutdown
+            lock do
+              @keep_running = false
+              @condition.signal
+            end
+
+            @thread.join
```
fbogsany: This needs to call `@exporter.shutdown`. It is important that we call it **after** `flush`, which means after `@thread.join`. The spec requires the exporter to return `FAILED_NOT_RETRYABLE` from `export` after `shutdown` is called.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -0,0 +1,176 @@
...
+  FAILED_RETRYABLE = OpenTelemetry::SDK::Trace::Export::FAILED_RETRYABLE
+  FAILED_NOT_RETRYABLE = OpenTelemetry::SDK::Trace::Export::FAILED_NOT_RETRYABLE
+
+  class TestExporter
+    def initialize(status_codes: nil)
+      @status_codes = if status_codes.nil?
+                        []
+                      else
+                        status_codes
+                      end
```
fbogsany: Typical Ruby practice would be:
```ruby
@status_codes = status_codes || []
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,110 @@
...
+                  @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
+                  break unless @keep_running
+                end
+                fetch_batch
+              end
+              break unless @keep_running
+
+              # this is done outside the lock to unblock the producers
+              result_code = FAILED_RETRYABLE
+              result_code = @exporter.export(batch) while result_code == FAILED_RETRYABLE
```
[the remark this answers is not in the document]
rdooley: I think logging the error seems reasonable for now.
Lemme do some research on the java and go versions re spinning vs re-enqueue

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,110 @@
...
+          def work
+            loop do
+              batch = lock do
+                if spans.size < max_queue_size
+                  @condition.wait(@mutex, @delay) while spans.empty? && @keep_running
+                  break unless @keep_running
+                end
+                fetch_batch
+              end
+              break unless @keep_running
```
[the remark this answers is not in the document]
rdooley: Doesn't this break break us out of the loop started on ln 75 with closing `end` on ln 88 meaning we'll hit flush either way?

sdk/.rubocop.yml
```diff
@@ -1,8 +1,12 @@
 AllCops:
   TargetRubyVersion: '2.4.0'
 
+Metrics/CyclomaticComplexity:
+  Max: 7
 Metrics/AbcSize:
-  Max: 18
+  Max: 20
+Metrics/BlockLength:
+  Enabled: false
```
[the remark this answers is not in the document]
rdooley: I swear it was more than this, but the only one i'm getting with these reverted is
```
lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:73:11: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for work is too high. [7/6]
          def work ...
          ^^^^^^^^
```

I can local disable that and revert these

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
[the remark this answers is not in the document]
fbogsany: Yes, but if we `break` between `fetch_batch` and `export` then we drop whatever was in the batch.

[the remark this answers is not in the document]
rdooley: After some more playing around in the debugger, this break was only breaking us out of the `batch = lock do` block and not the loop started on 75, so we need another `break unless @keep_running` on ln 88 after the export call (so tests pass/my laptop does not turn into a jet)

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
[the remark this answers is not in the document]
rdooley: go: unimplemented
java: https://github.com/open-telemetry/opentelemetry-java/blob/master/sdk/src/main/java/io/opentelemetry/sdk/trace/export/BatchSampledSpansProcessor.java#L280 so just logging in case of exception here.

Gut says that if a batch is failed to export, the batch is still useful so backoff and spin rather than requeue and risk dropping some of the batch (in favor of I guess letting the queue get too large and thereby letting elements of the next batch be shifted off the queue by `on_end` )

fbogsany:
I pushed up a change that I think cleans up the code a little, and fixes a couple bugs. Namely, the timeout argument to `@condition.wait` should be in seconds, and we treated it as milliseconds. That's what turned your laptop into a jet. And that meant the second `break` wasn't necessary. I switched the `break` for a `return` (TIL about `break`, or relearn, I guess) and moved the `flush` call to the `shutdown` method.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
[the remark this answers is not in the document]
fbogsany: See https://github.com/open-telemetry/opentelemetry-ruby/pull/74#issuecomment-534766181

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,130 @@
...
+              return result_code unless result_code == FAILED_RETRYABLE
+
+              sleep(0.1 * retries)
+            end
+            FAILED_RETRYABLE
+          end
+
+          def flush
+            snapshot = lock { spans.shift(spans.size) }
+            # TODO: should this call export_batch or just blindly attempt to export and ignore failures?
```
fbogsany: Open question. Not sure we need to resolve it in this PR. Can maybe open an issue and revisit this later?

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,130 @@
...
+            loop do
+              batch = lock do
+                @condition.wait(@mutex, @delay_seconds) if spans.size < batch_size && @keep_running
+                @condition.wait(@mutex, @delay_seconds) while spans.empty? && @keep_running
+                return unless @keep_running
+
+                fetch_batch
+              end
+
+              export_batch(batch) # TODO: log or count errors
```
fbogsany: I'd punt this TODO to later, after some discussion with other implementers.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -84,33 +87,30 @@ def shutdown
...
             end
-            flush
           end
 
           def export_batch(batch)
-            retries = 1
-            result_code = @exporter.export(batch)
-            while result_code == FAILED_RETRYABLE && retries < @export_retry_attempts
-              sleep(0.1 * retries)
+            @export_attempts.times do |retries|
```
rdooley: slick

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,130 @@
...
+            @thread = Thread.new { work }
+          end
+
+          # does nothing for this processor
+          def on_start(span)
+            # noop
+          end
+
+          # adds a span to the batcher, threadsafe may block on lock
+          def on_end(span)
```
mwear: We are calling our `Span#end` method `Span#finish` due to `end` being a keyword in ruby. Do we want that to spill over into our processors and use `on_finish`, or should we go with the conventions other languages are using everywhere else.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,130 @@
...
+              export_batch(batch) # TODO: log or count errors
+            end
+          end
+
+          def export_batch(batch)
+            @export_attempts.times do |retries|
+              result_code = @exporter.export(batch)
+              return result_code unless result_code == FAILED_RETRYABLE
+
+              sleep(0.1 * retries)
```
mwear: Is the wait between retries something that should be configurable? Should it be a parameter to the constructor like `wait_time_millis`and friends?

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -0,0 +1,130 @@
...
+        # export fails with {FAILED_RETRYABLE}, backing off linearly in 100ms
+        # increments.
+        class BatchSpanProcessor
+          SCHEDULE_DELAY_MILLIS = 5
+          MAX_QUEUE_SIZE = 2048
+          MAX_EXPORT_BATCH_SIZE = 512
+          MAX_EXPORT_ATTEMPTS = 5
+          private_constant(:SCHEDULE_DELAY_MILLIS, :MAX_QUEUE_SIZE, :MAX_EXPORT_BATCH_SIZE, :MAX_EXPORT_ATTEMPTS)
+
+          def initialize(exporter:,
```
mwear: Do we expect these settings to come from configuration utlimately? If so, does it make sense to have the constants (above) and default configuration values in addition, or do we only need the defaults in one place?

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
mwear: My tentative answer is, yes, it should probably blindly attempt, ignore and log and failures.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: I could go either way on this. In general, I favour internal consistency, which suggests `on_finish`. I'll make that change.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: The spec says nothing about limiting retries, or how backoff should be handled. I'm not a huge fan of adding too many knobs here. Is there a way we could punt all of the backoff & retry logic to the user, with a reasonable default? I.e. allow them to provide a proc.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: Defaults should live in one place. How do you see configuration working for this?

fbogsany:
I renamed `on_end` to `on_finish` and added some error logging at export. I'm not sure how configuration should work. If you have ideas @mwear, let me know and we can try to implement them in this PR, or we can punt it for now and deal with configuration in a separate PR.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
mwear: I think we can punt for now and consider allowing a user specified proc in the future.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
mwear: I was wondering if the user would be instantiating the BatchSpanProcessor themselves, or if the SDK would instantiate it on their behalf. If it's the user, then I think things are fine as they are. If it's the SDK, then perhaps we should add options to TraceConfig that default to the values we have in the constants.

PR #75, opened 2019-09-11
Use string comparisons for ProbabilitySampler

fbogsany:
This switches comparison in the ProbabilitySampler to be closer to the Java implementation, but compare in the ASCII codespace rather than in Integer space. May need some tweaks if we want an exact match, but I think this is more in the spirit of the "spec" than the `rand` approach.


fbogsany:
@ibawt this needs more testing, but WDYT?

fbogsany:
My rationale for this is a slightly vague notion of the distribution of sampling decisions. With the `rand` approach, sampling decisions are made entirely locally **iff** no sampling decision has been made by a parent span or a link. In a collection of services that are all traced with sample rate 0.5, A -> B <- C, B will have approximately half its requests from A already sampled along with half the requests from C. It will **additionally** sample half of the unsampled requests, for a combined sample rate of 0.75.

The trace-id comparison will instead sample the global trace-id space, so if all services are traced at sample rate 0.5, the effective sample rate for B in the example above will be the expected 0.5.

PR #79, opened 2019-09-13
Implement Sampling OTEP

fbogsany:
Implements the changes from https://github.com/open-telemetry/oteps/blob/master/text/0006-sampling.md


api/lib/opentelemetry/trace/sampling_hint.rb
```diff
@@ -0,0 +1,22 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    # Hints to influence sampling decisions. The default option for span
+    # creation is to not provide any suggestion.
+    module SamplingHint
+      # Suggest to not record events and not propagate.
+      NOT_RECORD = :__not_record__
```
ibawt: Is there no better idiom for an enum here?  I feel like I would just used the :blah everywhere.

fbogsany: Google for `enums in ruby` for many many answers. Seems like everyone's favourite project is implementing enums in Ruby. We could pick one and pull it in, but I feel like it isn't worth the extra dependency.

sdk/lib/opentelemetry/sdk/trace/samplers/decision.rb
```diff
@@ -0,0 +1,27 @@
...
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Trace
+      module Samplers
+        # The Decision class represents an arbitrary sampling decision. It can
```
mwear: This comment is no longer accurate. `Decision` is now a module an enumeration of constants.

sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb
```diff
@@ -8,86 +8,70 @@ module OpenTelemetry
...
-            #   Must be within [0.0, 1.0].
-            # @raise [ArgumentError] if probability is out of range
-            # @return [OpenTelemetry::Trace::Samplers::Sampler]
-            def create(probability)
-              raise ArgumentError, 'probability must be in range [0.0, 1.0]' unless (0.0..1.0).include?(probability)
+          # @api private
+          #
+          # Callable interface for probability sampler. See {Samplers}.
+          def call(trace_id:, span_id:, parent_context:, hint:, links:, name:, kind:, attributes:)
+            take_hint(hint) ||
```
mwear: This is pretty gnarly logic, but nicely organized into understandable, well-named methods.

fbogsany:
> Should we leave this open until then, or do the integration in a different PR?

I kinda want to merge one of them, and then finish up the wiring in the other. I don't much mind which goes first. (I haven't read your feedback on #72 yet.)

fbogsany:
I have a few more tests to add, but I'm fairly happy with the implementation. I'd appreciate more 👀 on this.

sdk/lib/opentelemetry/sdk/trace/samplers/result.rb
```diff
@@ -0,0 +1,58 @@
...
+        # The Result class represents an arbitrary sampling result. It has
+        # boolean values for the sampling decision and whether to record
+        # events, and a collection of attributes to be attached to a sampled
+        # root span.
+        class Result
+          EMPTY_HASH = {}.freeze
+          DECISIONS = [Decision::RECORD, Decision::NOT_RECORD, Decision::RECORD_AND_PROPAGATE].freeze
+          private_constant(:EMPTY_HASH, :DECISIONS)
+
+          # Returns a frozen hash of attributes to be attached span. These
```
mwear: I do not see the same vocabulary used in OTEP as was used in the initial spec around attributes being added to root spans, or when the sampling decision changes. It doesn't mean it doesn't apply though. I do see the OTEP saying the following for [SamplingResult](https://github.com/open-telemetry/oteps/blob/master/text/0006-sampling.md#sampler-interface):


* A set of span Attributes that will also be added to the Span.
  * These attributes will be added after the initial set of Attributes.

So, based on the OTEP, the sampler will receive an initial set of attributes that presumably get added to the Span. The result will return an additional set of attributes that will be added to the span later. Perhaps we should indicate that in this comment here?

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -122,6 +119,18 @@ def create_link(span_context, attrs = nil)
...
           Link.new(span_context: span_context, attributes: attrs)
         end
+
+        private
+
+        def internal_create_span(result, name, kind, trace_id, span_id, parent_span_id, attributes, links, events, start_timestamp)
+          if result.record_events? && !@stopped
+            trace_flags = result.sampled? ? OpenTelemetry::Trace::TraceFlags::SAMPLED : OpenTelemetry::Trace::TraceFlags::DEFAULT
+            context = OpenTelemetry::Trace::SpanContext.new(trace_id: trace_id, trace_flags: trace_flags)
+            Span.new(context, name, kind, parent_span_id, @active_trace_config, @active_span_processor, attributes, links, events, start_timestamp || Time.now)
```
mwear: My interpretation of the sampling [OTEP](https://github.com/open-telemetry/oteps/blob/master/text/0006-sampling.md#sampler-interface) is that the initial set of attributes are the ones passed to start span, and the additional attributes, to be added later, are those returned on the sampling result.

If that interpretation is correct, then we probably need to merge the attributes from the `SamplingResult` with the initial attributes here.

fbogsany: Good catch, thanks! I’ll fix that up tomorrow and add some tests.

fbogsany: Fixed

sdk/lib/opentelemetry/sdk/trace/samplers/result.rb
fbogsany: Fixed the comment. Thanks for catching that!

PR #98, opened 2019-09-16
SDK: add tests for exporters and processors

### LINKED ISSUE #84: SDK: test exporters and processors
Empty tests:
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/noop_span_processor_test.rb
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/multi_span_processor.rb (file misnamed)
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/export/multi_span_exporter_test.rb
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/export/noop_span_exporter_test.rb
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/export/simple_sampled_span_processor_test.rb

Note that https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/test/opentelemetry/sdk/trace/export/batch_sampled_span_processor_test.rb will be implemented in #74

sveiss:
Fixes #84


sveiss:
cc @rdooley as well, as we were looking at the tests in this repo together yesterday.

sdk/test/opentelemetry/sdk/trace/export/simple_sampled_span_processor_test.rb
```diff
@@ -7,4 +7,95 @@
...
+  let :mock_span_exporter do
+    mock = Minitest::Mock.new
+    def mock.nil?
+      false
+    end
+
+    mock
+  end
+
+  let :stub_span_unsampled do
```
mwear: Is there a reason why an `OpenStruct` is better than a concrete `Span` and `SpanContext`?

sveiss: Ah -- the reason seems to be me not paying quite enough attention!

I saw the TODO comment and empty class body in https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/lib/opentelemetry/sdk/trace/span.rb#L11, and so I thought this was not yet implemented and stubbed out just enough to make the tests pass.

Looking again, I missed how that inherits from the pieces in /api/, which do have implementations. I'll swap this out for concrete Span and SpanContexts tomorrow.

rdooley:
These tests look very reasonable to me

sdk/lib/opentelemetry/sdk/trace/export/in_memory_span_exporter.rb
```diff
@@ -66,6 +66,7 @@ def export(spans)
             @mutex.synchronize do
               return FAILED_NOT_RETRYABLE if @stopped
 
+              spans = spans.entries unless spans.is_a? Array
```
fbogsany: I had to look this up. I feel like an unconditional `@finished_spans.concat(spans.to_a)` below is clearer. It'll do the right thing for both Array and non-Array Enumerables.

sveiss:
@fbogsany with #72 merged, I think the `Span#to_span_proto` call [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/d3fb7fbf15db898356943de396a28496bd6a8351/sdk/lib/opentelemetry/sdk/trace/export/simple_sampled_span_processor.rb#L52) should now be `to_span_data`. Is that correct?

sdk/lib/opentelemetry/sdk/trace/export/in_memory_span_exporter.rb
sveiss: Updated

sveiss:
I've rebased this PR and updated it to account for the changes from #72, and to address the reviewer comments.

sdk/test/opentelemetry/sdk/trace/export/simple_sampled_span_processor_test.rb
sveiss: I've updated this in the new version of the PR.

sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb
```diff
@@ -29,7 +29,7 @@ def export(spans)
               begin
                 merge_result_code(result_code, span_exporter.export(spans))
               rescue => e # rubocop:disable Style/RescueStandardError
-                logger.warn("exception raised by export - #{e}")
+                OpenTelemetry.logger.warn("exception raised by export - #{e}")
```
fbogsany: Do you know why `logger` doesn’t work here? It seems awkward to “namespace” as `OpenTelemetry.logger`. Is this a `require` order issue, or do I misunderstand the way the method lookup works?

fbogsany:
> @fbogsany with #72 merged, I think the `Span#to_span_proto` call [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/d3fb7fbf15db898356943de396a28496bd6a8351/sdk/lib/opentelemetry/sdk/trace/export/simple_sampled_span_processor.rb#L52) should now be `to_span_data`. Is that correct?
> 
> EDIT: looking at the other `SpanExporter#export` call [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb#L27), I'm wondering if the original `@span_exporter.export([span.to_span_proto]) ` was wrong and it should just be `@span_exporter.export([span])`. It seems we're using the call inconsistently across the two places.

Thanks for pointing out the inconsistency. I’ll fix this in a PR once I’m back in the office tomorrow. The pipeline should be: `span.finish -> span_processor.on_end(span) -> exporter.export([span.to_span_data])`. SpanProcessors have access to the Span (I think we need to have some discussion about how “final” the Span should be at that point - I believe SpanProcessors are supposed to be able to modify the Span, which is not possible with what we have right now - we `.freeze` attributes and events before passing the Span to the SpanProcessor, and even if we didn’t, we protect against further writes to the Span once it has ended). SpanExporters receive the Span in a finalized form, i.e. SpanData. In Java, this is actually the protobuf structure for the OT protobuf format, but I think we want to avoid taking a dependency on protobuf in the SDK and API gems.

sdk/lib/opentelemetry/sdk/trace/export/simple_sampled_span_processor.rb
```diff
@@ -49,9 +49,9 @@ def on_start(span)
           def on_end(span)
             return unless span.context.trace_flags.sampled?
 
-            @span_exporter.export([span.to_span_proto])
+            @span_exporter.export([span.to_span_data])
```
fbogsany: 👍

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
```diff
@@ -7,4 +7,67 @@
 require 'test_helper'
 
 describe OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter do
+  export = OpenTelemetry::SDK::Trace::Export
```
fbogsany: Two questions:
1. Do we need to do this? We only access this in one place AFAICT (`let(:subject) { ... }`), so it doesn’t seem worthwhile.
1. In other places, we’ve done this “trick” to simplify access to the class under test, which in this case would look like:
```ruby
InMemorySpanExporter = OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter
```

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
```diff
@@ -7,4 +7,67 @@
...
+
+  let(:span1)   { OpenTelemetry::Trace::Span.new }
+  let(:span2)   { OpenTelemetry::Trace::Span.new }
+
+  let(:subject) { export::InMemorySpanExporter.new }
+
+  it 'accepts an Array of Spans as argument to #export' do
+    subject.export([span1, span2])
```
fbogsany: While this *does* work, thanks to the wonders of dynamic typing, it’s technically incorrect. We should be exporting `SpanData`:
```ruby
subject.export([span1.to_span_data, span2.to_span_data])
```
Of course, that makes the expectations a little more difficult to express...

fbogsany: Same concern throughout these tests.

sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb
fbogsany: OK, so yeah, I misunderstood the lookup here. Constant lookup follows module nesting (except when it doesn't). Method lookup only follows the inheritance chain, and module nesting is irrelevant.

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
sveiss: We're also using it to shorten the error constants ```OpenTelemetry::SDK::Trace::Export::SUCCESS``` and ```OpenTelemetry::SDK::Trace::Export::FAILED_NOT_RETRYABLE```.

I'm avoiding using Constants for the alias names, because we end up with multiple definitions warnings if multiple test files take the same approach.

I'm not wedded to the short names -- I was just trying to keep line length under control. I can swap them back out for the fully qualified versions if you prefer.

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
sveiss: I was working from the doc comments on the method under test:

```          # Called to export sampled {Span}s.
          #
          # @param [Enumerable<Span>] spans the list of sampled {Span}s to be
          #   exported.
          # @return [Integer] the result of the export, SUCCESS or
          #   FAILED_NOT_RETRYABLE
          def export(spans)
```

https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/lib/opentelemetry/sdk/trace/export/in_memory_span_exporter.rb#L59

So these comments need switching to refer to `SpanData`s not `Span`s too?

fbogsany: Yep - comments are wrong.

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
fbogsany: Ah, missed the error constants. I don't have a strong opinion.

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
```diff
@@ -7,4 +7,67 @@
...
+    finished_spans[1].must_equal span2
+  end
+
+  it 'accepts an Enumerable of Spans as argument to #export' do
+    # An anonymous Struct serves as a handy implementor of Enumerable
+    enumerable = Struct.new(:span1, :span2).new
+    enumerable.span1 = span1
+    enumerable.span2 = span2
+
+    subject.export(enumerable)
```
indrekj: minor style comment: In rspec it's usually good practice to use named subjects. in this case I'd just rename subject to exporter so it's easier to read it

rdooley:
Talked w/ @sveiss and I'm going to attempt to finish this up

rdooley:
@indrekj and @fbogsany this should be ready for re-review

sdk/lib/opentelemetry/sdk/trace/export/in_memory_span_exporter.rb
```diff
@@ -37,10 +37,10 @@ def initialize
             @mutex = Mutex.new
           end
 
-          # Returns a frozen array of the finished {Span}s, represented by
+          # Returns a frozen array of the finished {SpanData}s, represented by
           # {io.opentelemetry.proto.trace.v1.Span}.
           #
-          # @return [Array<Span>] a frozen array of the finished {Span}s.
+          # @return [Array<SpanData>] a frozen array of the finished {SpanData}s.
           def finished_spans
```
rdooley: One active question here, should this be `def finished_span_datas` ?

fbogsany: It's a good question. Let's leave it like this for now, and we can adjust when the SDK spec is fleshed out.

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
```diff
@@ -7,4 +7,67 @@
...
+    exporter.export(enumerable)
+
+    finished_spans = exporter.finished_spans
+    finished_spans[0].must_equal span_data1
+    finished_spans[1].must_equal span_data2
+  end
+
+  it 'freezes the return of #finished_spans' do
+    exporter.export([span_data1])
+    exporter.finished_spans.frozen?.must_equal true
```
fbogsany: `exporter.finished_spans.must_be :frozen?`

sdk/test/opentelemetry/sdk/trace/export/multi_span_exporter_test.rb
```diff
@@ -7,4 +7,121 @@
...
+    # retryable error first
+    mock_span_exporter.expect :export, export::FAILED_RETRYABLE, [Object]
+    mock_span_exporter2.expect :export, export::FAILED_NOT_RETRYABLE, [Object]
+
+    exporter_multi.export(spans).must_equal export::FAILED_NOT_RETRYABLE
+    mock_span_exporter.verify
+    mock_span_exporter2.verify
+  end
+
+  it 'returns the worst error from #export (reveresed order)' do
```
fbogsany: s/reveresed/reversed/

sdk/test/opentelemetry/sdk/trace/export/noop_span_exporter_test.rb
```diff
@@ -7,4 +7,25 @@
...
+    # An anonymous Struct serves as a handy implementor of Enumerable
+    enumerable = Struct.new(:span0, :span1).new
+    enumerable.span0 = spans[0]
+    enumerable.span1 = spans[1]
+
+    exporter.export(enumerable).must_equal export::SUCCESS
+  end
+
+  it 'accepts calls to #shutdown' do
+    exporter.shutdown
```
fbogsany: Interesting to note that the no-op exporter doesn't return `FAILED_NOT_RETRYABLE` from `export` after `shutdown` has been called, which kinda violates the expectations for an Exporter. Maybe we should open an issue for that?

rdooley: https://github.com/open-telemetry/opentelemetry-ruby/issues/110

rdooley: I can handle that in a subsequent PR this (pacific) afternoon

rdooley:
Sure thing

sdk/test/opentelemetry/sdk/trace/export/noop_span_exporter_test.rb
rdooley: Hmm perhaps monday

sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb
```diff
@@ -0,0 +1,103 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+def stub_span_builder(sampled: false)
```
fbogsany: Minor quibble: replace "sampled" with "recording" here and throughout this file.

sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb
```diff
@@ -0,0 +1,103 @@
...
+  ctx = OpenTelemetry::Trace::SpanContext.new
+  span = OpenTelemetry::Trace::Span.new(span_context: ctx)
+  def span.to_span_data; end
+  if sampled
+    def span.recording_events?
+      true
+    end
+  end
```
fbogsany: Why this vs.
```ruby
def span.recording_events?
  recording
end
```

sdk/test/opentelemetry/sdk/trace/export/noop_span_exporter_test.rb
rdooley: Monday monday https://github.com/open-telemetry/opentelemetry-ruby/pull/98/commits/e85a425c0226ae38ed5245f462b7e4fb5233a119

sdk/lib/opentelemetry/sdk/trace/export/noop_span_exporter.rb
```diff
@@ -13,20 +13,33 @@ module Export
         # recorded data for sampled spans in their own format.
         #
         # To export data an exporter MUST be registered to the {Tracer} using
-        # a {SimpleSampledSpanProcessor} or a {BatchSpanProcessor}.
+        # a {SimpleSpanProcessor} or a {BatchSpanProcessor}.
         class NoopSpanExporter
+          def initialize
+            @stopped = false
+            @mutex = Mutex.new
```
fbogsany: We don't need this `Mutex`. The Ruby memory model essentially provides atomic updates of instance variables. Had we been updating or reading multiple dependent variables, a `Mutex` would have been necessary, but for this simple boolean, it isn't.

sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb
fbogsany: I'm not sure I understand what this is trying to do. Don't we just want to return the `recording` value here?

sveiss: Yes, I was following the test code too literally when I wrote this, and clearly didn't take a second glance for sanity. Just returning the value of `recording` is fine.

sdk/lib/opentelemetry/sdk/trace/export/noop_span_exporter.rb
rdooley: Thank you for the missing context!
This is done now.

sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb
fbogsany: @rdooley ^

rdooley: It's doing that, but because the comment is on the shared end line the comment pane on github isn't getting updated i think 
https://github.com/open-telemetry/opentelemetry-ruby/pull/98/files#diff-664f94ed9cc21f7f4fe36d14d27ddc07R14
which was introduced in https://github.com/open-telemetry/opentelemetry-ruby/pull/98/commits/e990865865b3a0b0afe9c69117f2ce445c8d516c

fbogsany: opentelemetry-ruby 2019-10-01 14-11-03

^ is what I see. This should be:
```ruby
def span.recording_events?
  recording
end
```

rdooley: If I do not wrap in the lambda I get two test errors
```
Error:
OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor#test_0003_forwards recorded spans from #on_finish:
MockExpectationError: expected export(Array) => 0
    PATH_REDACTED/opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb:52:in `block (2 levels) in <top (required)>'
```
and
```
Error:
OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor#test_0006_catches and logs exporter exceptions in #on_finish:
MockExpectationError: mocked method :error called with unexpected arguments ["unexpected error in span.on_finish - undefined local variable or method `recording' for #<OpenTelemetry::Trace::Span:0x00007fd08a961ee8>\nDid you mean?  recording_events?"]
    PATH_REDACTED/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/trace/export/simple_span_processor.rb:54:in `rescue in on_finish'
    PATH_REDACTED/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/trace/export/simple_span_processor.rb:49:in `on_finish'
    PATH_REDACTED/opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb:91:in `block (3 levels) in <top (required)>'
    PATH_REDACTED/opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/export/simple_span_processor_test.rb:90:in `block (2 levels) in <top (required)>'
```

fbogsany: I played around with this. You need to define the method differently, in order to access `recording`:
```ruby
span.define_singleton_method(:recording_events?) { recording }
```

PR #101, opened 2019-09-19
Split out test dirs in travis.yml for api and sdk

rdooley:
Note: The `2.4` and `truffleruby` sdk builds are failing on Travis, but with this PR they are actually being run.

cc @fbogsany


fbogsany:
This is cool! Thanks! Good to see the failures. I'd personally prefer to merge this and then open issues to fix the two failures. For folks who can't see the failures, the 2.4 one is due to the use of a 2.5 (I think) enhancement to `Struct.new` that allows requiring keyword args:
```
/home/travis/build/Shopify/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/trace/span_data.rb:11:in `new': {:keyword_init=>true} is not a symbol (TypeError)
```
And the truffleruby failure is actually a bug in truffleruby (which I'll fix next week):
```
TypeError: can't clone NilClass
    /home/travis/build/Shopify/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/trace/span.rb:247:in `clone'
```
But we can work around that easily in the meantime.

PR #102, opened 2019-09-23
Update README

elskwid:
# Overview 

Updates the README with badges, contributing information, initial
release schedule, and useful links.

## Details

* Takes a little from [otel-python](https://github.com/open-telemetry/opentelemetry-python/blob/master/README.md) and [otel-js](https://github.com/open-telemetry/opentelemetry-js/blob/master/README.md)
* Leaves out details on the packages/gems
* Current schedule is `Unknown`

@fbogsany and/or @mwear feel free to chime in with a good initial schedule and I'll update this PR before it's merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,2 +1,54 @@
-[![Join the chat at https://gitter.im/open-telemetry/opentelemetry-ruby](https://badges.gitter.im/open-telemetry/opentelemetry-ruby.svg)](https://gitter.im/open-telemetry/opentelemetry-ruby?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge)
-[![Build Status](https://travis-ci.org/open-telemetry/opentelemetry-ruby.svg?branch=master)](https://travis-ci.org/open-telemetry/opentelemetry-ruby)
+# OpenTelemetry Ruby
+
+[![Gitter chat][gitter-image]][gitter-url]
+[![Travis CI][ci-image]][ci-url]
+[![Apache License][license-image]][license-image]
+
+The Ruby [OpenTelemetry](https://opentelemetry.io/) client.
+
+## Contributing
+
+We'd love your help! Use tags [good first issue][issues-good-first-issue] and
+[help wanted][issues-help-wanted] to get started with the project.
+
+The Ruby special interest group (SIG) meets regularly. See the OpenTelemetry
+[community page][ruby-sig] repo for information on this and other language SIGs.
+
+
+## Release Schedule
+
+OpenTelemetry Ruby is under active development. Our goal is to release an
+_alpha_ version of the library at the end of September 2019. This release isn't
+guaranteed to conform to a specific version of the specification, and future
+releases will not attempt to maintain backwards compatibility with the alpha
+release.
+
+| Component                   | Version | Target Date       |
+| --------------------------- | ------- | ----------------- |
+| Tracing API                 | Alpha   | September 30 2019 |
+| Tracing SDK                 | Alpha   | September 30 2019 |
+| Metrics API                 | Alpha   | Unknown           |
+| Metrics SDK                 | Alpha   | Unknown           |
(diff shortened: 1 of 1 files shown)


elskwid:
I signed it

README.md
```diff
@@ -1,2 +1,55 @@
...
+OpenTelemetry Ruby is under active development. Our goal is to release an
+_alpha_ version of the library at the end of September 2019. This release isn't
+guaranteed to conform to a specific version of the specification, and future
+releases will not attempt to maintain backwards compatibility with the alpha
+release.
+
+| Component                   | Version | Target Date       |
+| --------------------------- | ------- | ----------------- |
+| Tracing API                 | Alpha   | Unknown           |
+| Tracing SDK                 | Alpha   | Unknown           |
```
fbogsany: Both of ^ should be Sept 30, 2019.

README.md
```diff
@@ -1,2 +1,55 @@
...
+| Metrics API                 | Alpha   | Unknown           |
+| Metrics SDK                 | Alpha   | Unknown           |
+| Jaeger Trace Exporter       | Alpha   | Unknown           |
+| Zipkin Trace Exporter       | Alpha   | Unknown           |
```
fbogsany: Zipkin exporter isn't a requirement for Alpha, I think?

README.md
```diff
@@ -1,2 +1,55 @@
...
+guaranteed to conform to a specific version of the specification, and future
+releases will not attempt to maintain backwards compatibility with the alpha
+release.
+
+| Component                   | Version | Target Date       |
+| --------------------------- | ------- | ----------------- |
+| Tracing API                 | Alpha   | Unknown           |
+| Tracing SDK                 | Alpha   | Unknown           |
+| Metrics API                 | Alpha   | Unknown           |
+| Metrics SDK                 | Alpha   | Unknown           |
```
fbogsany: I'd leave these as Unknown until the metrics API solidifies.

README.md
```diff
@@ -1,2 +1,55 @@
...
+| Jaeger Trace Exporter       | Alpha   | Unknown           |
```
fbogsany: Leave this as Unknown for now. I started scoping the work today, but I'm not sure exactly what's expected (UDP export to the agent only? Thrift? Protobuf? grpc? HTTP?). AFAICT we don't have an existing Jaeger client for Ruby to piggyback on.

README.md
```diff
@@ -1,2 +1,55 @@
...
+| Zipkin Trace Exporter       | Alpha   | Unknown           |
+| Prometheus Metrics Exporter | Alpha   | Unknown           |
+| OpenTracing Bridge          | Alpha   | Unknown           |
+| OpenCensus Bridge           | Alpha   | Unknown           |
```
fbogsany: Leave these as Unknown for now.

README.md
elskwid: FYI: @fbogsany, I mentioned this in a team meeting and it seems the other libs are implementing metrics as it stands but I can see the reasons to hold off. 😁

README.md
elskwid: Perhaps the [jaeger-client-ruby](https://github.com/salemove/jaeger-client-ruby) lib linked in [this issue](https://github.com/jaegertracing/jaeger/issues/268) has something we can use. From my quick glance this looks more like an OpenTracing Tracer than a Jaeger client library.

README.md
elskwid: @fbogsany It's not technically part of the alpha requirements - seems that folks are putting that in because they already have it done. I've removed it from our schedule.

elskwid:
@fbogsany I've updated the schedule with the dates we have now and removed the Zipkin exporter from the list. Thanks for the review!

README.md
fbogsany: Thanks for the link! We can definitely use that, either as a starting point or we can wrap it. I’ll dig into it some more.

PR #108, opened 2019-09-25
Opentracing bridge

rdooley:
I believe this is ready for review again (and probably could do with me squashing work)


.travis.yml
```diff
@@ -13,4 +14,4 @@ rvm:
   - truffleruby
 
 install: ( cd api && bundle install --jobs=3 --retry=3 )
-script: ( cd $TEST_DIR && rake )
+script: ( if [ $TEST_DIR=='opentracingshim' ]; then gem install opentracing -v 0.5.0; fi && cd $TEST_DIR && rake )
```
rdooley: This is not good, but gets around the we aren't really installing `sdk` or `opentracingshim` issue (since opentracingshim depends on opentracing)

fbogsany:
I'd remove the "shim" suffix everywhere.

fbogsany:
Also, the suggested terminology here is "bridges"

rdooley:
@fbogsany so something like `opentracingbridge/lib/opentelemetry/opentracingbridge/span.rb` ?

fbogsany:
Either of the following:
```
opentelemetry-ruby/opencensus/lib/opentelemetry/opencensus/span.rb
opentelemetry-ruby/opencensus/opentelemetry-opencensus.gemspec
opentelemetry-ruby/opentracing/lib/opentelemetry/opentracing/span.rb
opentelemetry-ruby/opentracing/lib/opentelemetry-opentracing.gemspec
```
or
```
opentelemetry-ruby/bridges/opentracing/lib/opentelemetry/bridge/opentracing/span.rb
opentelemetry-ruby/bridges/opentracing/opentelemetry-bridge-opentracing.gemspec
...
```
I kinda like the second. I'm a little 🤷‍♂ about pluralizing the top-level directory, but leaning toward the plural. So:
```
api
sdk
bridges
exporters
adapters
```
at the top-level.

rdooley:
The plural makes sense to me, on it

bridges/opentracing/lib/opentelemetry/bridge/opentracing.rb
```diff
@@ -0,0 +1,27 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentracing'
+require 'opentelemetry'
+
+module OpenTelemetry
+  module Bridge
+    # OpenTracing Bridge provides a means of converting
+    # from OpenTelemetry to OpenTracing objects
+    OT = OpenTracing
```
fbogsany: This is a bit of a weird way to access the "other" `OpenTracing` namespace. You can instead (below):
```ruby
FORMAT_TEXT_MAP = ::OpenTracing::FORMAT_TEXT_MAP
FORMAT_RACK = ::OpenTracing::FORMAT_RACK
```

bridges/opentracing/lib/opentelemetry/bridge/opentracing.rb
```diff
@@ -0,0 +1,27 @@
...
+    module OpenTracing
+      FORMAT_TEXT_MAP = OT::FORMAT_TEXT_MAP
+      FORMAT_RACK = OT::FORMAT_RACK
```
fbogsany: Do you need to define these constants? Clients should just use the same constants from the `::OpenTracing` namespace.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,81 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Bridge
+    module OpenTracing
+      # Tracer provides a means of referencing
+      # an OpenTelemetry::Tracer as a OpenTracing::Tracer
+      class Tracer < OT::Tracer
```
fbogsany: You don't need to subclass. You're overriding everything, so subclassing isn't useful here.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,81 @@
...
+        HTTP_TEXT_FORMAT = OpenTelemetry::DistributedContext::Propagation::HTTPTextFormat.new
+        BINARY_FORMAT = OpenTelemetry::DistributedContext::Propagation::BinaryFormat.new
+        attr_reader :scope_manager
+
+        def initialize(tracer)
+          @tracer = tracer
```
fbogsany: I'm not sure I'd configure this with an `OpenTelemetry::Trace::Tracer` instance here. You can instead access the global tracer with `OpenTelemetry.tracer` whenever you need it.

bridges/opentracing/lib/opentelemetry/bridge/opentracing.rb
rdooley: Ah ok thanks

bridges/opentracing/lib/opentelemetry/bridge/opentracing.rb
rdooley: Unneeded and removed

.travis.yml
```diff
@@ -13,4 +14,4 @@ rvm:
   - truffleruby
 
 install: ( cd api && bundle install --jobs=3 --retry=3 )
-script: ( cd $TEST_DIR && rake )
+script: ( if [ $TEST_DIR=='bridges/opentracing' ]; then gem install opentracing -v 0.5.0; fi && cd $TEST_DIR && rake )
```
fbogsany: Why is this necessary? Is this because we can't run `bundle install` in the `bridges/opentracing` directory?

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope_manager.rb
```diff
@@ -0,0 +1,34 @@
...
+      # A ScopeManager provides an API for interfacing with
+      # OpenTelemetry Tracers and Spans as OpenTracing objects
+      class ScopeManager
+        def initialize(tracer)
+          @tracer = tracer
+        end
+
+        # Activate the given span
+        #
+        # @param [Span] span An OpenTelemetry Span
```
fbogsany: This should take an `OpenTracing::Span` rather than an `OpenTelemetry::Span`.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope_manager.rb
```diff
@@ -0,0 +1,34 @@
...
+        # @return [Span]
```
fbogsany: This should return a `Scope`.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope_manager.rb
```diff
@@ -0,0 +1,34 @@
...
+        def activate(span, finish_on_close: true)
+          Span.new(@tracer.with_span(span))
```
fbogsany: This should:
1. Create a `Scope` with the given `span` and whether to `finish_on_close`.
1. Track the current active `Scope` in an instance variable.
1. Return the `Scope`.

After 👀 some more, it looks like the current `Context` related APIs in OpenTelemetry are slightly incompatible with the `Scope` API from OpenTracing - notably, OpenTelemetry `Context` is block-structured. It may be worth discussing whether that's the right thing to do. For now, though, I'd implement OpenTracing's `Scope` independently of OpenTelemetry's `Context`.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope_manager.rb
```diff
@@ -0,0 +1,34 @@
...
+        end
+
+        # Get the active span as a OpenTracingBridge::Span
+        #
+        # @return [Span]
+        def active
+          Span.new(@tracer.current_span)
```
fbogsany: This should return the active `Scope` from an instance variable.

fbogsany: Reading some more, it looks like the current Scope should be tracked in a thread-local variable rather than an instance variable. This is similar to what we do in `OpenTelemetry::Context`.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope.rb
```diff
@@ -0,0 +1,23 @@
...
+      # an OTracing scope
+      class Scope
+        def span
+          OpenTelemetry.tracer.current_span
+        end
+
+        def close
+          OpenTelemetry.tracer.current_span.finish
+        end
+      end
```
fbogsany: I *think* what we need here is management of a stack of `Scope`s, so:
```ruby
class Scope
  attr_reader :span

  def initialize(manager, span, finish_on_close)
    @manager = manager
    @parent = manager.active
    @span = span
    @finish_on_close = finish_on_close
  end

  def close
    @span.finish if @finish_on_close
    @manager.active = @parent
  end
end
```

Where:
```ruby
class ScopeManager
  KEY = :__opentelemetry_opentracing_scope__
  private_constant :KEY

  def active
    Thread.current[KEY]
  end

  def active=(scope)
    Thread.current[KEY] = scope
  end

  def activate(span, finish_on_close: true)
    self.active = Scope.new(self, span, finish_on_close)
  end
end
```

.travis.yml
rdooley: 💯

fbogsany:
Can we remove the WIP and ask @indrekj for a 👀? He and @mwear have the most context on this.

fbogsany:
Also note that @elskwid is working on a change to the `in_span` API that'll simplify the mapping considerably.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/span.rb
```diff
@@ -0,0 +1,64 @@
...
+          nil
+        end
+
+        def log(event: nil, timestamp: Time.now, **fields)
+          nil
+        end
+
+        def log_kv(timestamp: Time.now, **fields)
+          nil
+        end
```
mwear: It should be possible to implement `log` and `log_kv` as events in OpenTelemetry. 

See the java reference implementation [here](https://github.com/open-telemetry/opentelemetry-java/blob/master/opentracing_shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java#L111-L133).

bridges/opentracing/lib/opentelemetry/bridge/opentracing/span.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        # The following have no OpenTelemetry Equivalent and are left noop
+        def set_baggage_item(key, value)
+          self
+        end
+
+        def get_baggage_item(key)
+          nil
+        end
```
mwear: It should be possible to implement baggage using DistributedContext. 

See the java reference implementation [here](https://github.com/open-telemetry/opentelemetry-java/blob/master/opentracing_shim/src/main/java/io/opentelemetry/opentracingshim/SpanShim.java#L135-L154).

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,75 @@
...
+      # an OpenTelemetry::Tracer as a OpenTracing::Tracer
+      class Tracer
+        HTTP_TEXT_FORMAT = OpenTelemetry::DistributedContext::Propagation::HTTPTextFormat.new
+        BINARY_FORMAT = OpenTelemetry::DistributedContext::Propagation::BinaryFormat.new
+
+        def active_span
+          OpenTelemetry.tracer.current_span
+        end
+
+        def start_active_span(operation_name,
```
mwear: This should optionally take a block and evaluate the block within the context of the newly created span. Also, this method should return a scope.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,75 @@
...
+                              finish_on_close: true)
+          span = OpenTelemetry.tracer.start_span(operation_name,
+                                                 with_parent: child_of,
+                                                 attributes: tags,
+                                                 links: references,
+                                                 start_timestamp: start_time)
+          OpenTelemetry.tracer.with_span(span)
+        end
+
+        def start_span(operation_name,
```
mwear: Like `start_active_span`, this method should optionally take a block and evaluate the block within the context of the newly created span. It, however, returns a span.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,75 @@
...
+                       tags: nil,
+                       ignore_active_scope: false)
+          OpenTelemetry.tracer.start_span(operation_name,
+                                          with_parent: child_of,
+                                          attributes: tags,
+                                          links: references,
+                                          start_timestamp: start_time)
+        end
+
+        def inject(span_context, format, carrier, &block)
```
mwear: `inject` does not take a block in OpenTracing

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,75 @@
...
+            HTTP_TEXT_FORMAT.inject(context, carrier, &block)
+          when ::OpenTracing::FORMAT_BINARY
+            # TODO: I don't think this is right
+            yield carrier, TraceParent::TRACE_PARENT_HEADER, BINARY_FORMAT.to_bytes(span_context)
+          else
+            warn 'Unknown inject format'
+          end
+        end
+
+        def extract(format, carrier, &block)
```
mwear: Extract does not take a block in OpenTracing. One thing to keep in mind for `FORMAT_RACK` is that it extracts headers that have been normalized by rack. So, if you are looking for a header named `traceparent` the rack-normalized version will be `HTTP_TRACEPARENT`.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+                              tags: nil,
+                              ignore_active_scope: false,
+                              finish_on_close: true,
+                              &block)
+          span = OpenTelemetry.tracer.start_span(operation_name,
+                                                 with_parent: child_of,
+                                                 attributes: tags,
+                                                 links: references,
+                                                 start_timestamp: start_time)
+          OpenTelemetry.tracer.with_span(span, &block) if block_given?
```
mwear: When the optional block is passed in, `start_active_span` will yield the scope and return the block's return value. See this issue: https://github.com/opentracing/opentracing-ruby/issues/41.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+                       start_time: Time.now,
+                       tags: nil,
+                       ignore_active_scope: false,
+                       &block)
+          span = OpenTelemetry.tracer.start_span(operation_name,
+                                                 with_parent: child_of,
+                                                 attributes: tags,
+                                                 links: references,
+                                                 start_timestamp: start_time)
+          OpenTelemetry.tracer.with_span(span, &block) if block_given?
```
mwear: Similar to `start_active_span`, when a block is passed to `start_span` it will yield the span and return the block's return value.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+          span
+        end
+
+        def inject(span_context, format, carrier)
```
mwear: It looks like `inject` and `extract` are still works in progress?

rdooley: Yes

elskwid:
@rdooley I believe the `in_span` change to which @fbogsany was referring is in #118.

rdooley:
@mwear I believe this is ready for review now. I'm unsure why circleci jruby is failing on the jaeger exporter but will be looking into that today

rdooley:
I can just hold off until things are settled and update this PR then. Thank you for the update.

mwear:
It might make more sense to address everything except for baggage in this PR. We can revisit baggage once context propagation has settled. What do think about that @rdooley?

rdooley:
Works for me!

bridges/opentracing/Gemfile
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+gem 'opentelemetry-api', '~> 0.0', path: '../../api'
+gem 'opentelemetry-sdk', '~> 0.0', path: '../../sdk'
```
fbogsany: Do you need the version constraint here? We don't have it in the Jaeger exporter or HTTP example Gemfiles.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,95 @@
...
+        attr_reader :tracer
+
+        def initialize(tracer)
+          @tracer = tracer
+          @tracer_factory = OpenTelemetry::Trace::TracerFactory.new
+        end
+
+        def active_span
+          @tracer.current_span
+        end
```
fbogsany: This should match the reference implementation, returning the OpenTracing span from the active scope:
```ruby
    def active_span
      scope = scope_manager.active
      scope.span if scope
    end
```

This also requires an implementation of `scope_manager` that returns a `ScopeManager` instance.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,95 @@
...
+                              tags: nil,
+                              ignore_active_scope: false,
+                              finish_on_close: true,
+                              &block)
+          span = @tracer.start_span(operation_name,
+                                    with_parent: child_of,
+                                    attributes: tags,
+                                    links: references,
+                                    start_timestamp: start_time)
+          scope = Scope.new(ScopeManager.new, span, finish_on_close)
```
fbogsany: This needs more wrapping & unwrapping. `child_of` could be an OpenTracing Span or a SpanContext, and we need to distinguish the two. We then need to unwrap the OpenTelemetry context before passing it to `@tracer.start_span(... with_parent:`. We need an OpenTracing `Reference` implementation that wraps an OpenTelemetry `Link` or allows "coercing" to a `Link`. The `span` needs to be wrapped in an OpenTracing `Span`.

We should be using the `scope_manager` from this OpenTracing Tracer instance rather than creating a new `ScopeManager` here.

.circleci/config.yml
```diff
@@ -48,6 +48,12 @@ commands:
       - run:
           name: CI (Jaeger)
           command: "cd exporters/jaeger && bundle exec rake"
+      - run:
+          name: Bundle (OpenTracing Bridge)
+          command: "cd bridges/opentracing && gem install --no-document bundler && bundle install --jobs=3 --retry=3"
```
rdooley: Very weird jruby thing where if you bundle OTrace bridge before CI jaeger (or vise versa) whichever you bundled before the other one, fails its CI step with e.g.
```
File does not exist: thrift
```

api/lib/opentelemetry/distributed_context/distributed_context.rb
```diff
@@ -19,6 +19,8 @@ def entries
       def [](_key)
         nil
       end
+
+      def []=(_key, _value) end
```
fbogsany: @mwear should this be part of the API? I wonder if we should gate this PR on #147 and build on the APIs defined there?

rdooley: @fbogsany since #147 is now merged I'll work on building off the APIs defined there tomorrow

bridges/opentracing/lib/opentelemetry/bridge/opentracing/reference.rb
```diff
@@ -0,0 +1,35 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Bridge
+    module OpenTracing
+      # Reference provides a means of treating
+      # an OpenTelemetry::Link as an OpenTracing::Reference
+      class Reference
```
fbogsany: This won't work because `OpenTracing::Reference` is effectively a concrete class -- users of OpenTracing will create instances directly rather than creating instances of our implementation.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/scope_manager.rb
```diff
@@ -0,0 +1,41 @@
...
+    module OpenTracing
+      # A ScopeManager provides an API for interfacing with
+      # OpenTelemetry Tracers and Spans as OpenTracing objects
+      class ScopeManager
+        SCOPE_KEY = :__opentelemetry_opentracing_scope__
+        MANAGER_KEY = :__opentelemetry_opentracing_scope_manager__
+        private_constant :SCOPE_KEY
+        private_constant :MANAGER_KEY
+
+        def self.current
```
fbogsany: I don't think we need this at all? I think we just need `Tracer#scope_manager` to return the singleton instance.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/reference.rb
rdooley: I switched from this to convenience methods `from_link` and `to_link`

mwear:
:wave: @rdooley I definitely owe you a review, which I'll try to get in tomorrow. Thanks for your patience 🙏!

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,105 @@
...
+          wrapped
+        end
+
+        def inject(span_context, format, carrier)
+          context = OpenTelemetry::Context.current
+          span = OpenTelemetry::Trace::Span.new(span_context: span_context.context)
+          context = context.set_value(OpenTelemetry::Trace::Propagation::ContextKeys.current_span_key, span)
+          case format
+          when ::OpenTracing::FORMAT_TEXT_MAP
+            OpenTelemetry::Trace::Propagation.http_trace_context_injector.inject(context, carrier) do |c, k, v|
```
mwear: If we want to interop with the OpenTelemetry Tracer, both inject and extract should use the global propagators, i.e. OpenTelemetry.propagation.inject and OpenTelemetry.propagation.extract.

bridges/opentracing/lib/opentelemetry/bridge/opentracing/tracer.rb
```diff
@@ -0,0 +1,105 @@
...
+                              finish_on_close: true,
+                              &block)
+          child_of = child_of.span unless child_of.instance_of? OpenTelemetry::Trace::Span
+          span = @tracer.start_span(operation_name,
+                                    with_parent: child_of,
+                                    attributes: tags,
+                                    links: references,
+                                    start_timestamp: start_time)
+          wrapped = Span.new(span, dist_context: span.context)
+          scope = Scope.new(ScopeManager.instance, wrapped, finish_on_close)
```
mwear: Scope activation / deactivation (closing) needs to correspond with an OpenTelemetry::Context containing the active span being attached and detached. Attach makes that the current context, whereas detach restores the previous context. Manual Context#attach and Context#detach are somewhat dangerous, as each call to attach and detach need to match, however, I think the scope concept from OT should make this work out.

fbogsany:
This is out of date and the spec has tightened up around OT compatibility for Otel. Closing this in favour of a new implementation.

PR #112, opened 2019-10-01
Add a Jaeger trace exporter

fbogsany:
This targets Thrift Compact protocol over UDP - essentially, the agent.

It is heavily "inspired" by the code from @indrekj here: https://github.com/salemove/jaeger-client-ruby, especially around the `Transport` and the use of a dummy transport to calculate sizes. @indrekj do you have any concerns about licensing?

This requires a little more testing, ideally including against the actual Jaeger agent.


indrekj:
> @indrekj do you have any concerns about licensing?

No, feel free to copy or do whatever.

exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
```diff
@@ -0,0 +1,174 @@
...
+              'refType' => Thrift::SpanRefType::CHILD_OF,
+              'traceIdLow' => int64(link.context.trace_id[16, 16]),
+              'traceIdHigh' => int64(link.context.trace_id[0, 16]),
+              'spanId' => int64(link.context.span_id)
+            )
+          end
+        end
+
+        def encode_tags(attributes)
+          @type_map ||= {
```
indrekj: you could also do something like this https://github.com/salemove/jaeger-client-ruby/blob/6ecc35fcc79153625fd9a8d6e563f30d5a4f594b/lib/jaeger/thrift_tag_builder.rb#L4 to get some compile time checking of keys and to avoid using strings directly

exporters/jaeger/Gemfile.lock
```diff
@@ -0,0 +1,72 @@
+PATH
```
rdooley: I notice we explicitly ignore the `api/Gemfile.lock` in the gitignore.

Should we make this more general and ignore this lock file as well?

.gitignore
```diff
@@ -1,4 +1,4 @@
-api/Gemfile.lock
+**/Gemfile.lock
```
rdooley: 👍

exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
fbogsany: Addressed in https://github.com/open-telemetry/opentelemetry-ruby/pull/112/commits/500c49065de98f0818c476102e8e9acbaba5c503

exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
```diff
@@ -0,0 +1,225 @@
...
+          @transport ||= SizingTransport.new
+          @protocol ||= ::Thrift::CompactProtocol.new(@transport)
+          @transport.flush
+          encoded_span.write(@protocol)
+          @transport.size
+        end
+
+        def encoded_process
+          @encoded_process ||= begin
+            tags = OpenTelemetry.tracer.resource.label_enumerator.map do |key, value|
```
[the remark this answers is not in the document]
fbogsany: Jaeger really expects `process` to be global, but we have a batch of spans here each with their own `library_resource`. I'm not sure what to put in the global `process`, if anything.

exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
```diff
@@ -0,0 +1,225 @@
...
+              Thrift::Tag.new('key' => key, 'vType' => Thrift::TagType::STRING, 'vStr' => value)
+            end
+            Thrift::Process.new('serviceName' => 'TODO: extract from resource?', 'tags' => tags)
```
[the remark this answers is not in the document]
fbogsany: As mentioned above, we don't have a global `resource` anymore - only a per-spandata resource (really per-Tracer, but at this point we only have the spandatas). Adding it as an option when initializing the exporter is probably the only valid choice here.

fbogsany:
> One issue that is present on master that I also had to work around was adding require 'singleton' to the top of noop_span_processor.rb. We should fix that at some point.

I've fixed that here.

exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
```diff
@@ -0,0 +1,226 @@
...
+          encoded_span.write(@protocol)
+          @transport.size
+        end
+
+        def encoded_process
+          @encoded_process ||= begin
+            tags = [] # TODO: figure this out.
+            # tags = OpenTelemetry.tracer.resource.label_enumerator.map do |key, value|
+            #   Thrift::Tag.new('key' => key, 'vType' => Thrift::TagType::STRING, 'vStr' => value)
+            # end
```
fbogsany: I commented this out until we can figure out what to do here.

mwear:
Screen Shot 2019-10-29 at 8 38 54 PM

Currently traces look like this ☝️. It seems like we should have a resource that represents the service as a whole, but it appears that we lost that in the fall out from named tracers. 

I'd rather merge this sooner than later. Not having a service name doesn't result in a very good experience with Jaeger. We can merge this as is, or we can add a `service_name` argument to the exporter's constructor. Any thoughts?

PR #115, opened 2019-10-09
Try removing runtime checks

fbogsany:
This is an experimental PR to see how the code looks & feels with most runtime argument checking removed. It provides defaults in some cases. Configuration arguments are still checked.


fbogsany:
I’ll work up a PR that moves the checks out to an optional “strict” mode too, to see if we prefer that.

api/lib/opentelemetry/trace/trace_flags.rb
```diff
@@ -16,10 +16,9 @@ class << self
         # Returns a newly created {TraceFlags} with the specified flags.
         #
         # @param [Integer] flags 8-bit byte of bit flags
-        # @raise [ArgumentError] If flags is not an 8-bit byte
         # @return [TraceFlags]
         def from_byte(flags)
-          raise ArgumentError, 'flags must be an 8-bit byte' unless flags & ~0xFF == 0 # rubocop:disable Style/NumericPredicate
+          flags = 0 unless flags & ~0xFF == 0 # rubocop:disable Style/NumericPredicate
```
fbogsany: This will blow up with `NoMethodError` if passed, e.g., a String. Are we OK with that? Technically, while this is a public API, I expect it to only be called by instrumentation (specifically things extracting trace context in various formats from the wire), most of which will live in the SDK & API.

api/lib/opentelemetry/trace/link.rb
```diff
@@ -32,8 +32,7 @@ class Link
       #   this link. Attributes will be frozen during Link initialization.
       # @return [Link]
       def initialize(span_context, attributes = nil)
-        raise ArgumentError unless span_context.instance_of?(SpanContext)
-        raise ArgumentError unless Internal.valid_attributes?(attributes)
+        attributes = nil unless Internal.valid_attributes?(attributes)
```
fbogsany: This check has the same overhead as the raising version. Does it buy us much? Should we keep the "check with default" or handle the failing cases later, during serialization?

api/lib/opentelemetry/trace/event.rb
```diff
@@ -36,8 +36,7 @@ class Event
       #   Defaults to Time.now.
       # @return [Event]
       def initialize(name:, attributes: nil, timestamp: nil)
-        raise ArgumentError unless name.is_a?(String)
-        raise ArgumentError unless Internal.valid_attributes?(attributes)
+        attributes = nil unless Internal.valid_attributes?(attributes)
```
fbogsany: Same question here. Do we need/want this check?

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -21,7 +21,7 @@ class << self
           # @return [Resource]
           def create(labels = {})
             frozen_labels = labels.each_with_object({}) do |(k, v), memo|
-              raise ArgumentError, 'label keys and values must be strings' unless k.is_a?(String) && v.is_a?(String)
+              return new({}.freeze) unless k.is_a?(String) && v.is_a?(String)
```
fbogsany: Same question here. This is done as configuration, though, so technically it's OK to keep the stricter raising check rather than the defaulting check. I guess this could blow up below if someone passed `false` as a value, for example. 🤷‍♂ opinions?

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -57,7 +57,7 @@ def label_enumerator
         # @return [Resource] A new resource formed by merging the current resource
         #   with other
         def merge(other)
-          raise ArgumentError unless other.is_a?(Resource)
+          return self unless other.is_a?(Resource)
```
fbogsany: I think this is totally reasonable, although I wonder, in Ruby 🦆 typing fashion, whether we might want to allow passing in a Hash or some other enumerable of pairs?

Meta question: what configuration should ultimately be supported for Resource? This seems to be an open question for the spec, not just for Ruby.

fbogsany:
@mwear @bai you both approved an earlier version of this, but I have a couple more questions (and made some more minor changes).

fbogsany:
> I’ll work up a PR that moves the checks out to an optional “strict” mode too, to see if we prefer that.

I'm actually not going to do this. Re-reading this PR, I'm relatively happy with the result. If we feel that a strict version of the API is a useful thing (e.g. for instrumentation developers), we can add that later, referring to the checks removed by this PR.

elskwid:
@fbogsany following on from our SIG call today:

We discussed taking this changeset and turning it into a `strict` mode implementation before we lose the value of this diff. Do you and/or @mwear have any ideas on the API? Would `OpenTelementry.strict` be good enough for now?

sdk/lib/opentelemetry/sdk/resources/resource.rb
fbogsany: Given that this *is* called at configuration time, and the error handling spec suggests raising for configuration errors, I think I should revert this change and stick with eagerly raising. Would you be OK with that @mwear ?

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -264,39 +264,46 @@ def trim_span_attributes(attrs)
           nil
         end
 
-        def trim_links(links, max_links_count, max_attributes_per_link) # rubocop:disable Metrics/AbcSize
+        def trim_links(links, max_links_count, max_attributes_per_link) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
```
fbogsany: Deferring checks to the SDK seems like the right choice, but Rubocop agrees that the resulting code *is* complex.

fbogsany:
> Do we want to hold off with these changes and wait for strict mode, or do we want to go forward with these and replace with strict mode when it's ready?

I think we should move ahead with these and get the strict mode in ASAP in another PR, before the codebase diverges significantly from this diff.

> We discussed taking this changeset and turning it into a strict mode implementation before we lose the value of this diff. Do you and/or @mwear have any ideas on the API? Would OpenTelementry.strict be good enough for now?

Options for implementation:
1. Configuration flag (e.g. on `TraceConfig`) tested at runtime around the current set of checks. This incurs overhead for checking the flag in every public API method, and increases the complexity of those methods.
1. Retain current set of checks and delegate `raise`/`log`/default decision to a central method (possibly on `TraceConfig`). This retains the overhead of checks always.
1. Create a `strict_sdk` parallel SDK implementation containing the strict checks. This seems pretty heavyweight in terms of code size and presents change synchronization problems (because tests for the strict SDK live in a different gem from the main SDK).
1. Move checks to modules within the SDK (e.g. `module StrictStatus`) and add a configuration method that `prepend`s the modules to the appropriate classes.

I like the last approach the best. It will invalidate method caches, but this is pretty common in Ruby applications during startup. It seems the cleanest in terms of code maintenance. Its major downside is that it isn't an easily reversible action.

sdk/lib/opentelemetry/sdk/trace/span.rb
mwear: I also agree on both points!

mwear:
> I think we should move ahead with these and get the strict mode in ASAP in another PR, before the codebase diverges significantly from this diff.

👍 The latest round of changes look good

I am also +1 on option 4 for strict mode.

PR #116, opened 2019-10-10
Add basic benchmarks for tracer and span

mwear:
I was looking over #115 and thought it would be useful if we had some simple benchmarks as a point of comparison. This PR adds simple benchmarks for tracer and span. They can be run from the `/api` folder by typing: `bundle exec ruby benchmarks/tracer_bench.rb` or `bundle exec ruby benchmarks/span_bench.rb`. If we like these benchmarks, we can look into adding some niceties around running them.

### DIFF
--- a/api/benchmarks/span_bench.rb
+++ b/api/benchmarks/span_bench.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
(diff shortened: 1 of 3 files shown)


mwear:
We can do that. The structure of the benchmarks and the independent repos make that a little bit of a challenge, although, not insurmountable. Would you like to see that as part of this PR, or in a followup?

mwear:
On second thought. I'll address sharing in a followup PR. I'll need to noodle a bit on the best approach.

PR #119, opened 2019-10-13
Update error messages in Span#set_attribute

elskwid:
# Overview

Updates the error messages in `Span#set_attribute` to provide more information about the argument error and adds tests for the improvements.

## Details

I found I needed a little more information from `#set_attribute` while working on some examples. This just tells the caller if the `key` or the `value` is the reason they're receiving the error.


fbogsany:
#115 is about to merge, which invalidates this. We should incorporate this change in the strict mode, though.

elskwid:
@mwear and @fbogsany thanks, this will come back in with strict mode.

PR #121, opened 2019-10-16
Recording vs sampled

fbogsany:
Includes #120 

https://github.com/open-telemetry/opentelemetry-specification/pull/291 states:
> Span Processors will receive all spans with this flag _(`IsRecording`)_ set. However, Span Exporter will not receive them unless the `Sampled` flag was set.

https://github.com/open-telemetry/opentelemetry-specification/pull/296 replaced `RECORD_AND_PROPAGATE` with `RECORD_AND_SAMPLED`.

### DIFF
--- a/api/lib/opentelemetry/trace/sampling_hint.rb
+++ b/api/lib/opentelemetry/trace/sampling_hint.rb
@@ -9,14 +9,14 @@ module OpenTelemetry
     # Hints to influence sampling decisions. The default option for span
     # creation is to not provide any suggestion.
     module SamplingHint
-      # Suggest to not record events and not propagate.
+      # Suggest to not record events and not sample.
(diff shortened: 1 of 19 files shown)


api/lib/opentelemetry/trace/sampling_hint.rb
```diff
@@ -16,7 +16,7 @@ module SamplingHint
       RECORD = :__record__
 
       # Suggest to record events and propagate.
-      RECORD_AND_PROPAGATE = :__record_and_propagate__
+      RECORD_AND_SAMPLED = :__record_and_propagate__
```
mwear: Is there any reason not to make the value `:__record_and_sampled__` to match the constant name?

PR #122, opened 2019-10-16
Initialize Tracer resource and pass to Spans

fbogsany:
https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/sdk-tracing.md#tracer-creation
> New Tracer instances are always created through a TracerFactory (see API). The name and version arguments supplied to the TracerFactory must be used to create a Resource instance which is stored on the created Tracer. The readable representations of all Span instances created by a Tracer must provide a getLibraryResource method that returns this Resource information held by the Tracer.


fbogsany:
The spec originally allowed providing a Resource with a Span, where a Resource could represent another service (e.g. an untraceable entity like Memcached, Redis or MySQL). This feature was removed after some debate, along with the removal of SpanData. Named Tracers re-adds multiple Resources. I’m not sure there is much practical difference, at least in terms of Resources, between your TraceProxy proposal above and the current implementation of Named Tracers. TraceProxy adds a level of indirection that isn’t required by Named Tracers, and removes the possibility of separate configuration (e.g. separate sampling policies). The latter could be seen as beneficial 😄.

I’m OK with leaving this PR hanging around for awhile, to facilitate debate. I don’t think TraceProxy is a significant improvement over Named Tracers, though.

One thing I’d like to discuss further is a comment from @elskwid in the weekly meeting about the word Factory giving him pause in Ruby code. If that’s a code smell to you, can you suggest what *would* be more idiomatic in Ruby?

fbogsany:
Ironically, https://github.com/open-telemetry/opentelemetry-specification/pull/313 moves everything except the Resource from Tracer to TracerFactory, leaving Tracer looking awfully like your TracerProxy. I've implemented that in my latest change. It feels really rough, and makes the "Factory" part of the name grate even more.

This was not fun to implement, and I've undoubtedly missed testing some pieces. 👀 appreciated!

api/lib/opentelemetry/trace/tracer_factory.rb
```diff
@@ -8,6 +8,10 @@ module OpenTelemetry
   module Trace
     # No-op implementation of a tracer factory.
     class TracerFactory
+      HTTP_TEXT_FORMAT = DistributedContext::Propagation::HTTPTextFormat.new
```
mwear: I see that https://github.com/open-telemetry/opentelemetry-specification/pull/313 moved span_processors. Does that necessarily mean text format and binary format need to move here as well?

fbogsany: > All configuration objects (SDK specific) and extension points (span processors,
**propagators**) must be provided to the `TracerFactory`.

The text mentions moving propagators as well.

PR #125, opened 2019-10-18
Update metrics API

fbogsany:
Update the metrics API to reflect the latest version:
https://github.com/open-telemetry/opentelemetry-specification/blob/a131ae1ccabee8cc8f3b8cf109b81be378006fea/specification/api-metrics.md
and
https://github.com/open-telemetry/opentelemetry-specification/blob/5802f6e33644c5996c2fb46ef9f444f81ac2b158/specification/api-metrics-user.md


api/lib/opentelemetry/metrics/instruments.rb
```diff
@@ -0,0 +1,121 @@
...
+    # produced -- mathematical, statistical summaries of certain observable
+    # behavior in the program. "Instruments" are the devices used by the
+    # program to record observations about their behavior. Therefore, we use
+    # "metric instrument" to refer to a program object, allocated through the
+    # API, used for recording metrics. There are three distinct instruments in
+    # the Metrics API, commonly known as Counters, Gauges, and Measures.
+    module Instruments
+      # A float gauge instrument.
+      class FloatGauge
+        def set(value, label_set_or_label = nil, *labels); end
```
mwear: Are the `label_or_label_set` and `*labels` arguments here for overloading `set` to work with the [label set calling convention](https://github.com/open-telemetry/opentelemetry-specification/blob/5802f6e33644c5996c2fb46ef9f444f81ac2b158/specification/api-metrics-user.md#label-set-calling-convention) and the [option to bypass meter.Lables()`?

fbogsany: Yep. I'm open to splitting the methods for each calling convention.

mwear: I suppose we can see how the implementation looks. We can leave them as is for now and decide to change them if it makes sense.

fbogsany:
Remaining work here is to add tests. I'll get that done tonight/tomorrow morning, and aim to get this merged by tomorrow's SIG meeting.

api/lib/opentelemetry/metrics/instruments.rb
fbogsany: It turned out this was wrong. I've fixed in recent commits.

fbogsany:
Tests added. This is ready for review again.

Issue #130, opened 2019-10-23
Add simple HTTP client/server example


elskwid:
As part of #97, add an example in the same style as the `http` example in the [Go library](https://github.com/open-telemetry/opentelemetry-go/tree/master/example/http).

This should be a simple client/server all-in-one example with a corresponding docker-compose service to ease startup and usage.

elskwid:
I'm getting ready to submit a PR for this! 🆘

fbogsany:
Done in #131

PR #131, opened 2019-10-23
Add http client/server example

elskwid:
# Overview

Adds a simple http client/server example using a [Sinatra](http://sinatrarb.com/) server and [Faraday](https://lostisland.github.io/faraday/) client. The example is modeled after the [Go http example](https://github.com/open-telemetry/opentelemetry-go/tree/master/example/http).

**Note: the console exporter won't be called until we merge in the `in_span` fix from #118**

## Details

* Adds the `examples` top-level directory
* Adds an `http` example directory that contains the client, server, and a small helper script
* Adds the `ex-http` docker-compose server to simplify setup/running of the example
* Adds a README to walk a person through running the example

## Depends on:

This PR depends on #144 to function but it can be merged in any order.

## Finishes

#130 - Add simple HTTP client/server example

### DIFF
--- a/examples/http/client.rb
+++ b/examples/http/client.rb
@@ -1,0 +1,49 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+require 'rubygems'
+require 'bundler/setup'
+require 'faraday'
+# Require otel-ruby
+require 'opentelemetry/sdk'
+
+# Allow setting the host from the ENV
+host = ENV.fetch('HTTP_EXAMPLE_HOST', '0.0.0.0')
+
+SDK = OpenTelemetry::SDK
+OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
+
+# Configure tracer
+exporter = SDK::Trace::Export::ConsoleSpanExporter.new
+processor = SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+OpenTelemetry.tracer_factory.add_span_processor(processor)
+tracer = OpenTelemetry.tracer_factory.tracer('faraday', 'semver:1.0')
+formatter = OpenTelemetry.tracer_factory.http_text_format
+
+connection = Faraday.new("http://#{host}:4567")
+url = '/hello'
+
+# For attribute naming, see:
+# https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-client
+
+# Span name should be set to URI path value:
+tracer.in_span(
+  url,
+  attributes: {
+    'component' => 'http',
+    'http.method' => 'GET',
+  },
+  kind: :client
+) do |span|
+  response = connection.get(url) do |request|
+    # Inject context into request headers
+    formatter.inject(span.context, request.headers)
+  end
+
+  span.set_attribute('http.url', response.env.url.to_s)
+  span.set_attribute('http.status_code', response.status)
+  span.set_attribute('http.status_text', response.reason_phrase)
+end
--- a/examples/http/server.rb
+++ b/examples/http/server.rb
@@ -1,0 +1,73 @@
+#!/usr/bin/env ruby
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'rubygems'
+require 'bundler/setup'
+require 'sinatra/base'
+# Require otel-ruby
+require 'opentelemetry/sdk'
+
+SDK = OpenTelemetry::SDK
+OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
+
+exporter = SDK::Trace::Export::ConsoleSpanExporter.new
+processor = SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+OpenTelemetry.tracer_factory.add_span_processor(processor)
+
+# Rack middleware to extract span context, create child span, and add
+# attributes/events to the span
+class OpenTelemetryMiddleware
+  def initialize(app)
+    @app = app
+    @formatter = OpenTelemetry.tracer_factory.http_text_format
+    @tracer = OpenTelemetry.tracer_factory.tracer('sinatra', 'semver:1.0')
+  end
+
+  def call(env)
+    # Extract context from request headers
+    context = @formatter.extract(env)
+
+    status, headers, response_body = 200, {}, ''
+
+    # Span name SHOULD be set to route:
+    span_name = env['PATH_INFO']
+
+    # For attribute naming, see
+    # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-server
+
+    # Span kind MUST be `:server` for a HTTP server span
+    @tracer.in_span(
+      span_name,
+      attributes: {
+        'component' => 'http',
+        'http.method' => env['REQUEST_METHOD'],
+        'http.route' => env['PATH_INFO'],
+        'http.url' => env['REQUEST_URI'],
+      },
+      kind: :server,
+      with_parent_context: context
+    ) do |span|
+      # Run application stack
+      status, headers, response_body = @app.call(env)
+
+      span.set_attribute('http.status_code', status)
+    end
+
+    [status, headers, response_body]
+  end
+end
+
+class App < Sinatra::Base
+  set :bind, '0.0.0.0'
+  use OpenTelemetryMiddleware
+
+  get '/hello' do
+    'Hello World!'
+  end
+
+  run! if app_file == $0
+end
--- a/sdk/lib/opentelemetry/sdk/trace/tracer_factory.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/tracer_factory.rb
@@ -8,7 +8,7 @@ module OpenTelemetry
   module SDK
     module Trace
       # {TracerFactory} is the SDK implementation of {OpenTelemetry::Trace::TracerFactory}.
-      class TracerFactory
+      class TracerFactory < OpenTelemetry::Trace::TracerFactory
         Key = Struct.new(:name, :version)
         private_constant(:Key)
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,6 +24,11 @@ services:
       context: .
     working_dir: /app
 
+  ex-http:
+    <<: *base
+    command: ./start_server.sh
+    working_dir: /app/examples/http
+
   sdk:
     <<: *base
     working_dir: /app/sdk
--- a/examples/http/Gemfile
+++ b/examples/http/Gemfile
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "faraday", "~> 0.16.1"
+gem "opentelemetry-api", path: "../../api"
+gem "opentelemetry-sdk", path: "../../sdk"
+gem "sinatra", "~> 2.0"
--- a/examples/http/Gemfile.lock
+++ b/examples/http/Gemfile.lock
@@ -1,0 +1,39 @@
+PATH
+  remote: ../../api
+  specs:
+    opentelemetry-api (0.0.0)
+
+PATH
+  remote: ../../sdk
+  specs:
+    opentelemetry-sdk (0.0.0)
+      opentelemetry-api (~> 0.0)
+
+GEM
+  remote: https://rubygems.org/
+  specs:
+    faraday (0.16.2)
+      multipart-post (>= 1.2, < 3)
+    multipart-post (2.1.1)
+    mustermann (1.0.3)
+    rack (2.0.7)
+    rack-protection (2.0.7)
+      rack
+    sinatra (2.0.7)
+      mustermann (~> 1.0)
+      rack (~> 2.0)
+      rack-protection (= 2.0.7)
+      tilt (~> 2.0)
+    tilt (2.0.10)
+
+PLATFORMS
+  ruby
+
+DEPENDENCIES
+  faraday (~> 0.16.1)
+  opentelemetry-api!
+  opentelemetry-sdk!
+  sinatra (~> 2.0)
+
+BUNDLED WITH
+   2.0.2
--- a/examples/http/README.md
+++ b/examples/http/README.md
@@ -1,0 +1,21 @@
+# OpenTelemetry Ruby Example
+
+## HTTP
+
+This is a simple example that demonstrates tracing an HTTP request from client to server. The example shows several aspects of tracing, such as:
+
+* Using the `TracerFactory`
+* Span Attributes
+* Using the console exporter
+
+### Running the example
+
+The example uses Docker Compose to make it a bit easier to get things up and running.
+
+1. Follow the `Developer Setup` instructions in [the main README](../../README.md)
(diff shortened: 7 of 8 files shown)


elskwid:
I added a note that this requires the fix from #118, which I'm working on next.

examples/http/client.rb
```diff
@@ -0,0 +1,45 @@
...
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+require 'rubygems'
+require 'bundler/setup'
+require 'faraday'
+# Require otel-ruby
+require 'opentelemetry'
```
fbogsany: You shouldn't need to `require` this and the SDK -- just the SDK.

examples/http/client.rb
```diff
@@ -0,0 +1,45 @@
...
+require 'opentelemetry/sdk'
+
+require_relative '../exporters/console'
+
+# Allow setting the host from the ENV
+host = ENV.fetch('HTTP_EXAMPLE_HOST', '0.0.0.0')
+
+SDK = OpenTelemetry::SDK
+OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
```
fbogsany: This is a little annoying - we should really auto-register the SDK pieces if you've `require`d them.

examples/http/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+# Configure tracer
+exporter = Examples::Exporters::Console.new
+processor = SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+tracer = OpenTelemetry.tracer_factory.tracer('http.client', 'semver:1.0')
+tracer.add_span_processor(processor)
```
fbogsany: So much wiring :-(

examples/http/client.rb
```diff
@@ -0,0 +1,45 @@
...
+
+connection = Faraday.new("http://#{host}:4567")
+url = '/hello'
+
+# For attribute naming, see:
+# https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-client
+
+# Span name should be set to URI path value:
+tracer.in_span('/hello', kind: 'Client') do |span|
```
fbogsany: `kind:` should be `OpenTelemetry::Trace::SpanKind::CLIENT`, which really rolls off the tongue 😞. We should consider whether we just want to support symbols directly here, like:
```ruby
:client
:server
:internal
:producer
:consumer
```

examples/http/server.rb
```diff
@@ -0,0 +1,54 @@
...
+    set :tracer, tracer
+  end
+
+  get '/hello' do
+    # If route is available, use it:
+    span_name = env['sinatra.route'].split.last
+    # Otherwise, use RFC 3986 URI path value:
+    # span_name = request.path
+
+    settings.tracer.in_span(span_name, kind: 'Server') do |span|
```
fbogsany: `OpenTelemetry::Trace::SpanKind::SERVER`

examples/http/server.rb
```diff
@@ -0,0 +1,54 @@
...
+      # For attribute naming, see
+      # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-server
+      span.set_attribute('component', 'http')
+      span.set_attribute('http.url', env['REQUEST_URI'])
+      span.set_attribute('http.route', env['sinatra.route'].split.last)
+      span.set_attribute('http.method', request.request_method)
+      span.set_attribute('http.status_code', response.status)
```
fbogsany: Does `response.status` default to 200 in Sinatra?

examples/http/client.rb
elskwid: @fbogsany okay, force of habit!

examples/http/server.rb
```diff
@@ -0,0 +1,54 @@
...
+      span.add_event(name: 'handle http.server.get')
+    end
+
+    'Hello World!'
```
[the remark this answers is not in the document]
elskwid: @fbogsany I tried that and wasn't getting responses. I'll try again and see if I can outline the problem.

mwear:
It's possible that we could de-grapify the example from: https://github.com/open-telemetry/opentelemetry-ruby/pull/132 and use it here.

fbogsany:
I completely agree with Matt. A simple example with Rack middleware to extract + creation of a child span in the server, and root span + Faraday + inject in the client is 💯

elskwid:
@mwear @fbogsany thanks for the feedback. I'll get the direct code feedback in here today/this weekend. As far as the context propagation - I mentioned that in the PR description, I would LOVE to have that in there but it isn't really clear how ours works. I guess its the 🐔 + 🥚 problem of having no examples while writing the examples. 😄 Any chance I could get some guidance on propagation?

I think there's some usefulness in having really really simple examples. I'm not sure if that means _this_ example should use Rack middleware or not. I realize that most people looking at our examples may be more than passingly familiar with tracing concepts but as a complete newbie I found the Go example to be very clear. Having said all that, I'm happy to take this in the direction you think it should go.

examples/http/client.rb
elskwid: @fbogsany can do! I didn't realize Span kind was limited to a list of values. Funny timing on [this spec issue](https://github.com/open-telemetry/opentelemetry-specification/pull/337)

elskwid:
@mwear @fbogsany just to add a little bit of background to my question about context propagation:

The docs for `inject` say that the method "sets" the context on the carrier but the method really looks to just yield `carrier`, `header`, and `context` which leads me to believe the setting needs to be part of that block for the specific carrier. In my case, I need to write the headers to the request on the client-side (and do the actual extraction on the server-side). 

It looks like other libraries have propagators that know how to interact with the carrier. Does that sound right? Are planning to create propagators for typical use cases?

mwear:
@elskwid Currently there is an HTTPTextFormat propagator. The API is based off of this [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-propagators.md#http-text-format). The blocks play the role of the `getter` and `setter` in that document.

This is a diversion from OpenTracing, where inject and extract know how to get or set the context in the carrier. Your examples have shown us where our APIs could use some improvement previously, I'm curious if you can point out the differences you see with other languages here.

As far as distributed context goes, we don't yet have an implementation for the SDK. This is something that is very likely to change in light of context propagation OTEP.

examples/http/client.rb
elskwid: _Note: we can use the symbols after #134_

examples/http/server.rb
elskwid: _Note: we can use the symbol after #134_

examples/http/client.rb
```diff
@@ -0,0 +1,45 @@
...
+# Allow setting the host from the ENV
+host = ENV.fetch('HTTP_EXAMPLE_HOST', '0.0.0.0')
+
+SDK = OpenTelemetry::SDK
+OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
+
+# Configure tracer
+exporter = Examples::Exporters::Console.new
+processor = SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+tracer = OpenTelemetry.tracer_factory.tracer('http.client', 'semver:1.0')
```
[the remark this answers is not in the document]
elskwid: I went with `faraday` for now. Thanks!

examples/http/server.rb
```diff
@@ -0,0 +1,54 @@
...
+OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
+
+class App < Sinatra::Base
+  set :bind, '0.0.0.0'
+
+  configure do
+    # Configure tracer
+    exporter = Examples::Exporters::Console.new
+    processor = SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+    tracer = OpenTelemetry.tracer_factory.tracer('http.server', 'semver:1.0')
```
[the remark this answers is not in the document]
elskwid: I've updated to `sinatra`.

sdk/lib/opentelemetry/sdk/trace/tracer_factory.rb
```diff
@@ -8,7 +8,7 @@ module OpenTelemetry
   module SDK
     module Trace
       # {TracerFactory} is the SDK implementation of {OpenTelemetry::Trace::TracerFactory}.
-      class TracerFactory
+      class TracerFactory < OpenTelemetry::Trace::TracerFactory
```
fbogsany: For the formatters? 😞 

I was kinda hoping to rely on 🦆 typing here. I suspect we may want the formatters to be configurable, at which point we can remove the subclassing, but for now this is 👍

sdk/lib/opentelemetry/sdk/trace/noop_span_processor.rb
```diff
@@ -3,6 +3,7 @@
 # Copyright 2019 OpenTelemetry Authors
 #
 # SPDX-License-Identifier: Apache-2.0
+require 'singleton'
```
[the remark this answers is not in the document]
elskwid: @fbogsany great catch, I had it in an old commit. Rebased away 👋

sdk/lib/opentelemetry/sdk/trace/tracer_factory.rb
elskwid: @fbogsany yes, I wanted to be able to use the `http_text_format` method. I think it would be nice to be able to do:

```ruby
# name is just an example
OpenTelemetry::Propagators::HTTPTextHeader.inject(span.context, request.headers)
```
versus the wiring I need to do now:
```ruby
formatter = OpenTelemetry.tracer_factory.http_text_format
# or
# formatted = OpenTelemetry::DistributedContext::Propagation::HTTPTextFormat.new

formatter.inject(span.context, request.headers) do |carrier, key, value|
  carrier[key] = value
end
```
I know #140 (default getter/setter) will help with this but I was wondering if maybe moving that to a `Propagator` would allow the formatter to stay simple. (cc: @mwear)

examples/http/server.rb
```diff
@@ -0,0 +1,70 @@
...
+class OpenTelemetryMiddleware
+  def initialize(app)
+    @app = app
+    @formatter = formatter = OpenTelemetry.tracer_factory.http_text_format
+    @tracer = OpenTelemetry.tracer_factory.tracer('sinatra', 'semver:1.0')
+  end
+
+  def call(env)
+    # Extract context from request headers
+    context = @formatter.extract(env) { |carrier, key| carrier[key] }
```
elskwid: @fbogsany @mwear I have the `context`, now what do I do with it? 😆 I didn't see any tests or examples that made this clear in how I can pass this context to the span.

mwear: We talked about this briefly, but ultimately we want to use the extracted context to parent the span we're starting on the server. `in_span` doesn't appear to have a `with_parent_context` argument. Is that intentional @fbogsany? If not, then we should probably add options for `with_parent_context` and `with_parent`.

elskwid: @mwear I had a spike on this branch that added `with_parent` and `with_parent_context` to `in_span` and it seemed to work. If we want to go that way I'll add some tests and make sure it works as advertised.

fbogsany: Not intentional. It should have the same signature as `start_span`.

elskwid: @mwear @fbogsany I've turned this into [a small PR](https://github.com/open-telemetry/opentelemetry-ruby/pull/144) so we don't clutter this one up any more than it is.

examples/http/server.rb
elskwid: @fbogsany one of my recent round of changes fixes this.

elskwid:
@mwear @fbogsany I believe this is good to go based on your feedback. @mwear can you take a look at my usage of `context` in the server and make sure it tracks with your thinking? This won't _work_ until #144 is merged but code-wise, it's there.

examples/http/server.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    status, headers, response_body = 200, {}, ''
+
+    # Span name SHOULD be set to route:
+    span_name = env['PATH_INFO']
+
+    # Span kind MUST be `:server` for a HTTP server span
+    @tracer.in_span(span_name, kind: :server, with_parent_context: context) do |span|
```
fbogsany: 👍

examples/http/server.rb
```diff
@@ -0,0 +1,70 @@
...
+      # For attribute naming, see
+      # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-server
+      span.set_attribute('component', 'http')
+      span.set_attribute('http.method', env['REQUEST_METHOD'])
+      span.set_attribute('http.route', env['PATH_INFO'])
+      span.set_attribute('http.url', env['REQUEST_URI'])
+      span.add_event(name: 'handle http.server.get')
```
fbogsany: This is a slightly odd example. A more realistic one might be the size of the response body (after the `@app.call`).

elskwid: @fbogsany as an `event`?  Okay, I was reading the name event literally as in, this thing happened at this time.

examples/http/client.rb
```diff
@@ -0,0 +1,48 @@
...
+
+# Span name should be set to URI path value:
+tracer.in_span(url, kind: :client) do |span|
+  span.set_attribute('component', 'http')
+  span.set_attribute('http.method', 'GET')
+  span.add_event(name: 'request http.client.get')
+
+  response = connection.get(url) do |request|
+    # Inject context into request headers
+    formatter.inject(span.context, request.headers) do |carrier, key, value|
```
mwear: You can omit the block now that #140 has merged.

examples/http/server.rb
```diff
@@ -0,0 +1,70 @@
...
+    span_name = env['PATH_INFO']
+
+    # Span kind MUST be `:server` for a HTTP server span
+    @tracer.in_span(span_name, kind: :server, with_parent_context: context) do |span|
+      # For attribute naming, see
+      # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md#http-server
+      span.set_attribute('component', 'http')
+      span.set_attribute('http.method', env['REQUEST_METHOD'])
+      span.set_attribute('http.route', env['PATH_INFO'])
+      span.set_attribute('http.url', env['REQUEST_URI'])
```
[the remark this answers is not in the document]
elskwid: @fbogsany done and done. Although, the `in_span` call is now large and in charge. 😆

examples/http/server.rb
elskwid: @fbogsany I have removed the explicit block now.

examples/http/client.rb
elskwid: @mwear yep, thank you, it's done!

elskwid:
@mwear @fbogsany I've updated after the latest round of feedback. Thank you!

examples/http/server.rb
fbogsany: That's true, but for this particular operation, that's implicitly the start of the span, so the event isn't that interesting.

I like this blog post as an example of events in a trace (look at the pictures rather than the text): https://medium.com/google-cloud/tracing-google-cloud-8f0c8ba8181c. There, the events are informative little snippets about things that occurred during the span. Some of them may be numeric values (like `146 bytes sent` and `40 bytes received`) while others are things like `DNSStart` and `DNSDone` to record when DNS lookup happened. The latter could conceivably be a span itself, but doesn't have to be.

fbogsany: I guess my larger point here is that events should typically be used to add colour to the span, and this specific one doesn't really do that.

PR #132, opened 2019-10-23
Code examples: HTTP server (grape) instrumentation example

### LINKED ISSUE #97: Code examples: HTTP client and server instrumentation examples

duonoid:
# Overview

This is a standalone http server example using `grape`.  It helps toward #97 and should work in conjunction with #130/#131.

The intention is to demonstrate (as correctly as possible), the usage of the api/sdk as middleware.

## Details

* Adds `examples/http/server/grape/`, to demonstrate 'http server' wiring
* Adds 'ex-grape' docker-compose service, for running in isolated environment
* Example runs on port 4568, so it can run alongside another example server

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -24,6 +24,17 @@ services:
       context: .
     working_dir: /app
 
+  # docker-compose run --rm ex-grape bundle install
+  # docker-compose run --rm --service-ports ex-grape
+  ex-grape:
+    <<: *base
+    command: bundle exec rackup --host 0.0.0.0 --port 4568 --debug
+    environment:
+      BUNDLE_GEMFILE: Gemfile
+    ports:
+      - 4568:4568
+    working_dir: /app/examples/http/server/grape
+
   sdk:
     <<: *base
     working_dir: /app/sdk
--- a/examples/exporters/console.rb
+++ b/examples/exporters/console.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry/sdk'
+require 'pp'
+
+module Examples
+  module Exporters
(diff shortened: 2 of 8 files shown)


mwear:
This got me thinking about our examples a little more and I made this comment: https://github.com/open-telemetry/opentelemetry-ruby/pull/131#pullrequestreview-306856148. 

I think the middleware from this example could be converted to a generic rack middleware and used in the server example. While it's true that there will also be Rack auto-instrumentation, I think a Rack middleware is an ideal place to demonstrate how to extract context from an incoming request and use it to parent spans in-process. 

Again, I'm open to any other thoughts or ideas.

elskwid:
@mwear Do you think there is any value in having examples that show the auto-instrumentation versus just having people go look at the actual implementation? 

@duonoid and I used the DataDog code as a way to understand how tracing is actually being used at the code level and we felt like it would be helpful to contribute that code back. For example, I have a version of my http client/server example (#131) that uses middleware and Sinatra hooks for the server and middleware for the Faraday client - exactly as the DataDog library does.

Would it better if we just chalked those up to learning and good reference? It's fine either way since there was very real value in coding up the examples. I totally understand not all example code belongs in the `examples` for the project. 😁

duonoid:
It seems like there are a couple of ways forward:
* extract a more generic middleware implementation, for use in #131
* change 'grape' example to more closely resemble the current datadog implementation

mwear:
I don't think I should have the last say in this, but here are my thoughts. 

If #131 has a Rack Middleware example, then there is not much to be gained from also having this example in the repo. The generic examples will help folks as a starting point for instrumenting web frameworks and http clients that aren't auto-instrumented.

I also don't think there is much value in another grape example that is more similar to the grape auto-instrumentation as we could just point to the grape auto-instrumentation as the example.

mwear:
We can try to reuse some of this work in #131. Closing this as we discussed at the SIG meeting.

PR #133, opened 2019-10-24
Add exportTimeoutMillis parameter to BatchSpanProcessor

### LINKED ISSUE #129: Add exportTimeoutMillis parameter to BatchSpanProcessor
See https://github.com/open-telemetry/opentelemetry-specification/pull/326. Note that the Exporter interface doesn't include a timeout, so this probably requires the use of `Timeout::timeout`.

duonoid:
# Overview
Adds `export_timeout_millis` initialization parameter to `BatchSpanProcessor` as well as an implementation using `Timeout`.

## Details
* Upstream spec uses `exporterTimeoutMillis` (slight naming difference)
* Upstream spec comment suggests returning 'failedNoRetry'
* Upstream spec default timeout value is 30 seconds

## Concerns
* Using `Timeout` is known to cause some subtle system instability issues -- probably better to pass finer-grained timeout values to exporter, but the interface does not currently exist

## Fixes
* Fixes #129


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -25,20 +27,23 @@ module Export
         # export fails with {FAILED_RETRYABLE}, backing off linearly in 100ms
         # increments.
         class BatchSpanProcessor
+          EXPORT_TIMEOUT_MILLIS = 30_000
           SCHEDULE_DELAY_MILLIS = 5
```
fbogsany: While you're in here, this should be `5_000` instead of `5`.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -25,20 +27,23 @@ module Export
         # export fails with {FAILED_RETRYABLE}, backing off linearly in 100ms
         # increments.
         class BatchSpanProcessor
+          EXPORT_TIMEOUT_MILLIS = 30_000
```
fbogsany: Transcription error, maybe (by me)? We should stick close to the spec here, so `EXPORTER_TIMEOUT_MILLIS`.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -82,7 +87,7 @@ def shutdown
 
           private
 
-          attr_reader :spans, :max_queue_size, :batch_size
+          attr_reader :spans, :max_queue_size, :batch_size, :export_timeout_seconds
```
fbogsany: I don't think we need to expose this (other than for testing, and I'd rather do that another way).

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -101,14 +106,21 @@ def work
...
 
               sleep(0.1 * attempts)
             end
             report_result(result_code, batch)
           end
 
+          def export_with_timeout(batch)
+            Timeout.timeout(export_timeout_seconds) { @exporter.export(batch) }
+          rescue Timeout::Error
+            # TODO: tell exporter that it timed out? (e.g., cleanup network connections)
```
fbogsany: The exporter should be able to `rescue Timeout::Error` and do its own cleanup. We don't need an extra cleanup API.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -37,6 +37,12 @@ def export(batch)
     def shutdown; end
   end
 
+  class TestTimeoutExporter < TestExporter
+    def export(batch)
+      raise Timeout::Error
```
fbogsany: This isn't really testing the functionality (that the processor enforces the timeout). I'd rather see a test that sets the timeout quite low, triggers the export and ensures that a) the exporter.export was called and b) it was interrupted by a timeout.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
duonoid: :heavy_check_mark:

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
duonoid: :heavy_check_mark:

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
duonoid: I've removed the reader, as well as rewritten the test. :heavy_check_mark:

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
duonoid: I've removed the comment. :heavy_check_mark:

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
duonoid: I've rewritten the test, demonstrating that the exporter was `:called` and does not get to the `:not_interrupted` state.

duonoid: :heavy_check_mark:

duonoid:
@fbogsany I think I have addressed your points so far (great feedback, thank you!), and is ready for another look :eyes:

PR #140, opened 2019-11-03
Default getter / setter for carrier; extract Rack normalized keys

mwear:
This PR improves `HTTPTextFormat#extract` and `HTTPTextFormat#inject` by providing default getter and setters for carriers that implement `[]` and `[]=`. Users can still provide their own getters and setter for carriers that have different needs.

Here's what changed:

```
#extract before (passing a custom getter is still possible)
context = formatter.extract(carrier) { |c, k| c[k] }

#extract now
context = formatter.extract(carrier)

#inject before (passing a custom setter is still possible)
formatter.inject(span_context, carrier) { |c, k, v| c[k] = v }

#inject now
formatter.inject(span_context, carrier)
```

While working on these changes, I noticed that `extract` was looking up keys as `traceparent` and `tracestate` rather than the Rack normalized versions `HTTP_TRACEPARENT` and `HTTP_TRACESTATE`. I changed extract to lookup the Rack normalized keys. We can discuss whether this is the correct thing to do (I can remove that commit for now if it's controversial), but being able to extract span context from a Rack env is going to be a common use case. In OpenTracing this was less of an issue as extract took a third argument for `format`. We had `FORMAT_TEXTMAP` and `FORMAT_RACK` for the lowercase and Rack normalized keys. 

There are two other options we can consider for handling this situation:

**Option 1:**
Have extract lookup the rack normalized key and the unnormalized version. It would use the first one it finds. 

**Option 2:**
We could have a different propagator for Rack normalized keys. We could introduce a RackHTTPTextFormat propagator that subclasses the HTTPTextFormat propagator and overrides extract.


fbogsany:
I'm leaning toward option 2. My current concerns:
1. `HTTPTextFormat#fields` doesn't correctly represent the fields that we lookup. (I don't really understand the rationale for `fields`, but it should at least be consistent.)
1. `inject` and `extract` are inconsistent - `extract` looks for the Rack normalized keys, whereas `inject` sets the unnormalized keys.

mwear:
I'd be happy to work up Option 2. Any preference as to whether I should tack it onto this PR or do that separately @fbogsany? If we do that work separately, I can back out the last commit making this PR deal with adding the default getter / setter only.

fbogsany:
I’d be happy to see it tacked onto this PR, but if you’d rather get this merged and do it separately, that’s fine too.

mwear:
I'll tack it on. I'm not in a hurry to get this merged. The consideration of a separate PR was more to keep the changeset reasonable for reviewers. I think it will still be in the realm of reasonability with the additions, so it's not a big deal.

mwear:
I extracted a common superclass for `HTTPTextFormat` and `RackHTTPTextFormat`. The only implementation that the subclasses need to provide are constants for `TRACE_PARENT_HEADER_KEY`, `TRACE_STATE_HEADER_KEY`, and `FIELDS`. I also added an additional `rack_http_text_format` accessor to TracerFactory. Sharing test cases with Minitest Spec proved to be somewhat of a challenge. I'll point that out in another comment.

api/test/opentelemetry/distributed_context/propagation/text_formatter_shared_test_cases.rb
```diff
@@ -0,0 +1,112 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module TextFormatterSharedTestCases
+  def self.included(base)
+    base.class_eval(&TESTS)
```
mwear: With Minitest (classic) sharing test cases is straight forward and is as simple as including a module with the shared test cases. Sharing tests with Minitest Spec using module inclusion proved to be much more of a challenge. It works if test cases are all defined within `describe` blocks, but anything more advanced becomes progressively more difficult. `let` cannot exist outside of `describe` and sharing state, such as variables defined with `let`, from the including class are issues that I was running into. This approach, while ugly, solves them. If anyone knows of a better way to share test cases and state from the including class with Minitest Spec, let me know.

api/lib/opentelemetry/trace/tracer_factory.rb
```diff
@@ -9,6 +9,7 @@ module Trace
     # No-op implementation of a tracer factory.
     class TracerFactory
       HTTP_TEXT_FORMAT = DistributedContext::Propagation::HTTPTextFormat.new
+      RACK_HTTP_TEXT_FORMAT = DistributedContext::Propagation::RackHTTPTextFormat.new
```
fbogsany: Make this a private constant?

api/lib/opentelemetry/distributed_context/propagation/rack_http_text_format.rb
```diff
@@ -0,0 +1,28 @@
...
+      # Encoding is expected to conform to the HTTP Header Field semantics. Values are often encoded as RPC/HTTP request
+      # headers.
+      #
+      # The carrier of propagated data on both the client (injector) and server (extractor) side is usually an http request.
+      # Propagation is usually implemented via library-specific request interceptors, where the client-side injects values
+      # and the server-side extracts them.
+      class RackHTTPTextFormat < TextFormatBase
+        TRACE_PARENT_HEADER_KEY = 'HTTP_TRACEPARENT'
+        TRACE_STATE_HEADER_KEY = 'HTTP_TRACESTATE'
+        FIELDS = [TRACE_PARENT_HEADER_KEY, TRACE_STATE_HEADER_KEY].freeze
```
fbogsany: Would these still be accessible from the superclass if they were private constants?

api/lib/opentelemetry/trace/tracer_factory.rb
mwear: Can do!

api/lib/opentelemetry/distributed_context/propagation/rack_http_text_format.rb
mwear: They need to be exposed for the superclass to access them.

api/lib/opentelemetry/trace/tracer_factory.rb
```diff
@@ -29,6 +30,10 @@ def binary_format
       def http_text_format
         HTTP_TEXT_FORMAT
       end
+
+      def rack_http_text_format
+        RACK_HTTP_TEXT_FORMAT
+      end
```
fbogsany: Given that we instantiate each of the formats once, and they’re only accessible through these methods, would it be cleaner to have a single TextFormat class and to pass the header keys to the constructor? I.e.:
```ruby
def rack_http_text_format
  @rack_http_text_format ||= TextFormat.new(tracestate_header_key: 'HTTP_TRACESTATE', traceparent_header_key: 'HTTP_TRACEPARENT')
end
```

mwear: It would clean up a couple of things. The testing woes and having to nudge constant lookup by prefixing with `self.class::` are two things I don't particularly like.

elskwid: That would be nice! I was also [kicking around some ideas](https://github.com/open-telemetry/opentelemetry-ruby/pull/131#discussion_r342726625) on how we could make the formats easy to use and I wonder if it would also have the side effect of making the format(ters) much simpler since we wouldn't need to do the default getter/setter.

From the comment linked above - slightly changes given this conversation...

I think it would be nice to be able to do:

```ruby
# name is just an example
OpenTelemetry::Propagators::RackHTTPTextHeader.inject(span.context, request.headers)
```
versus the wiring I need to do now:
```ruby
formatter = OpenTelemetry.tracer_factory.http_text_format
# or
# formatted = OpenTelemetry::DistributedContext::Propagation::TextFormat.new

formatter.inject(span.context, request.headers) do |carrier, key, value|
  carrier[key] = value
end
```
I know #140 (default getter/setter) will help with this but I was wondering if maybe moving that to a `Propagator` would allow the formatter to stay simple and maybe we wouldn't need to expose the `format` on the factory which feels super strange to me anyway.

mwear:
Thanks! Your suggestions were super helpful!

PR #143, opened 2019-11-05
SDK: integration tests

### LINKED ISSUE #82: SDK: integration tests
We need integration tests in the SDK gem to exercise the integration between the API and the SDK. A good structure for this is probably a method that, for example, traces a bunch of operations in different ways (root spans, child-of-remote spans, child-of-local spans, with and without links, etc.), and then a bunch of tests that configure the global tracer in different ways, call the tracing method, and asserts the results using an `InMemorySpanExporter`.

duonoid:
Fixes #82


duonoid:
NOTE: the failing tests may be due to the test's misuse of the API (false positive?), or they may be exposing some actual underlying issues that we would want to address.

sdk/test/integration/api_trace_test.rb
```diff
@@ -0,0 +1,109 @@
...
+      _(@root_span.name).must_equal 'root'
+    end
+
+    it 'root has a child' do
+      _(@child_of_root.to_span_data.parent_span_id).must_equal @root_span.to_span_data.span_id
+    end
+
+    it 'root has accurate child_count' do
+      # TODO: is child_count maintained by an exporter or processor?
+      _(@root_span.to_span_data.child_count).must_equal 1
```
fbogsany: Bug. Will fix. This should be maintained by the Span in coordination with the Tracer. Currently, Tracer doesn't inform Span of child creation events.

fbogsany: See #150 for discussion of this.

sdk/test/integration/api_trace_test.rb
```diff
@@ -0,0 +1,109 @@
...
+
+    it "doesn't have links" do
+      assert_nil @child_of_root.links
+    end
+  end
+
+  describe 'tracing child-of-remote spans' do
+    before do
+      @remote_span = tracer.start_span('remote', with_parent_context: remote_span_context)
+      @child_of_remote = tracer.start_span('child1', with_parent: @remote_span)
```
fbogsany: Generally speaking, `child_of_remote` should be created using `with_parent_context: @remote_span.context` because we won't have a direct reference to the remote parent (the parent Span exists as an object in the remote process, but in the local process it only exists as a SpanContext).

sdk/test/integration/api_trace_test.rb
```diff
@@ -0,0 +1,109 @@
...
+      tracer.in_span('root') do |s|
+        tracer.in_span('child_with_links', links: links) do |span|
+          @child_with_links = span.to_span_data
+        end
+      end
+    end
+
+    it 'span has links' do
+      # TODO: is there a way to retrieve an arbitrary span, by name?
+      #       e.g., tracer.get_span(name)
```
fbogsany: Nope, there isn't. It wouldn't make much sense anyway, since we might have an arbitrary number of 'open' spans with the same name in the current thread).

In the `shopify-tracing` gem, we *do* have a `TestProducer` that implements `find_by(name)` or `select_by(name)` functionality. It is useful for testing, and we might want to implement something similar in the SDK.

sdk/test/integration/api_trace_test.rb
```diff
@@ -0,0 +1,109 @@
...
+      _(@child_with_links.links.size).must_equal number_of_links
```
fbogsany: It is preferable to do this with `@child_with_links.to_span_data`, since `#links` is not considered part of the public API for `Span`.

sdk/test/integration/api_trace_test.rb
duonoid: Changed :heavy_check_mark:

sdk/test/integration/api_trace_test.rb
duonoid: Thank you for the confirmation.  I was hoping there was a more elegant way to access a span than storing in an instance variable (i.e., be able to use `let(:some_span)`).

duonoid: :heavy_check_mark:

sdk/test/integration/api_trace_test.rb
duonoid: I've clarified the testing syntax for consistency to use `#to_span_data`.  Thank you! :heavy_check_mark:

duonoid:
I've addressed the feedback from @fbogsany (thanks! :smile:) thus far, and will watch #150 for changes that should help with that one last failing test.

mwear:
It looks like #150 is slightly complicated. Should we consider commenting out (or not including) the child count test for the time being?

duonoid:
> consider commenting out (or not including) the child count test for the time being

I'd be ok with that.

PR #144, opened 2019-11-06
Add with_parent and with_parent_context to Tracer#in_span

elskwid:
# Overview

Adds `with_parent` and `with_parent_context` arguments to `Tracer#in_span`, which then passes them down to `#start_span`. This allows us to set the parent or parent context as expected.

## Details

I noticed that there wasn't any way to set the parent or parent context when using `#in_span` while working on the http client/server example in #130. See [this PR conversation](https://github.com/open-telemetry/opentelemetry-ruby/pull/131#discussion_r342727161).


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -26,8 +26,8 @@ def current_span
       #   OpenTelemetry.tracer.with_span(OpenTelemetry.tracer.start_span('do-the-thing')) do ... end
       #
       # On exit, the Span that was active before calling this method will be reactivated.
-      def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, sampling_hint: nil)
-        span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, sampling_hint: sampling_hint)
+      def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, sampling_hint: nil, with_parent: nil, with_parent_context: nil)
+        span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, sampling_hint: sampling_hint, with_parent: with_parent, with_parent_context: with_parent_context)
```
fbogsany: I'd prefer the parameters to be ordered the same way in the declarations of `in_span` and `start_span`.

fbogsany:
Let's resolve #145 in another PR. I'll update the branch & merge this as is.

api/lib/opentelemetry/trace/tracer.rb
elskwid: @fbogsany I worried about this more than I care to admit.

PR #146, opened 2019-11-07
Context propagation prototype

fbogsany:
This is an attempt to translate the context propagation prototype from https://github.com/open-telemetry/opentelemetry-java/pull/655

### DIFF
--- a/api/lib/opentelemetry.rb
+++ b/api/lib/opentelemetry.rb
@@ -6,6 +6,7 @@
 require 'logger'
 
 require 'opentelemetry/error'
+require 'opentelemetry/baggage'
 require 'opentelemetry/context'
 require 'opentelemetry/distributed_context'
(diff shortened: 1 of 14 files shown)


mwear:
Nice @fbogsany. I have been working on something similar. My work should move to SDK as it's actual implementation. Perhaps we can compare notes and try to synthesize the two? Here is my branch: https://github.com/mwear/opentelemetry-ruby/tree/context_prop_spike.

The Java proto currently has a mix between the explicit context passing an implicit. I think the goal it to use all explicit or all implicit (with an option to get a handle on context for advanced usage).

mwear:
I went ahead and combined this work, with the context prop work I have been doing. The result is here: https://github.com/open-telemetry/opentelemetry-ruby/pull/147.

fbogsany:
Closing in favour of #147

PR #147, opened 2019-11-08
Context prop combined

mwear:
This is a context propagation prototype based on https://github.com/open-telemetry/oteps/pull/66. 

## All in one - Example

The example below is an all-in-one example that shows configuration, handling an HTTP request, and making a call to an external service. In the all-in-one example there is deeper nesting than you will typically see, due to the collocation of the code. In the real world, different components would handle different parts of the request in their own separate locations.

```ruby
# Set global injector / extractors (accepts an array)
OpenTelemetry.propagation.http_injectors = [OpenTelemetry::Trace::Propagation.http_trace_context_injector, OpenTelemetry::CorrelationContext::Propagation.http_injector]
OpenTelemetry.propagation.http_extractors = [OpenTelemetry::Trace::Propagation.rack_http_trace_context_extractor, OpenTelemetry::CorrelationContext::Propagation.rack_http_extractor]
OpenTelemetry.correlations = OpenTelemetry::SDK::CorrelationContext::Manager.new
tracer = OpenTelemetry.tracer_factory.tracer('myapp', '1.0.0')

# One time setup above ^^^^, subsequent usage below:

# Extract context from an inbound request
extracted_context = OpenTelemetry.propagation.extract(rack_env)

# Add correlations to the extracted context
context = OpenTelemetry.correlations.set_value('k1', 'v1', context: extracted_context)

# Set current context
OpenTelemetry::Context.with_current(context) do
  # SpanContext from the extracted context will be an implicit parent for `service-span`
  tracer.in_span('service-span') do
    tracer.in_span('external-req-span') do
      headers = {'Accept' => 'application/json'}
      # Inject context into an outbound request
      headers = OpenTelemetry.propagation.inject(headers)
      Net::HTTP.get(URI('http://site.com'), headers)
    end
  end
end

# Inject and Extract can take explicit context, injectors and extractors

context = OpenTelemetry.propagation.extract(carrier, context: some_context, http_extractors: [ex1, ex2, ex3])

headers = OpenTelemetry.propagation.inject(headers, context: some_context, http_injectors: [in1, in2, in3])
```

## OTEP Examples

The following examples are Ruby versions of the examples in https://github.com/open-telemetry/oteps/pull/66.

### Global initialization

```ruby
OpenTelemetry.propagation.http_injectors = [OpenTelemetry::Trace::Propagation.http_trace_context_injector, OpenTelemetry::CorrelationContext::Propagation.http_injector]
OpenTelemetry.propagation.http_extractors = [OpenTelemetry::Trace::Propagation.rack_http_trace_context_extractor, OpenTelemetry::CorrelationContext::Propagation.rack_http_extractor]
OpenTelemetry.correlations = OpenTelemetry::SDK::CorrelationContext::Manager.new
tracer = OpenTelemetry.tracer_factory.tracer('myapp', '1.0.0')
```

### Inject & Extract with Implicit Context

```ruby
def handle_request(headers)
  context = OpenTelemetry.propagation.extract(headers)

  OpenTelemetry::Context.with_current(context)
    tracer.in_span('span-name') do
      version = OpenTelemetry.correlations.value('client-version')

      case(version)
      when 'v1.0'
        data = fetch_data_from_service_b
      when 'v2.0'
        data = fetch_data_from_service_c
      end

      send_response(data)
    end
  end
end
```

```ruby
def fetch_data_from_service_b
  headers = {}
  headers = OpenTelemetry.propagation.inject(headers)

  # make an http request
  uri = URI('http://site.com')
  http.get(('/', headers)
end
```

### Inject and Extract with Explicit Context

```ruby
def handle_request(headers)
  extractors = OpenTelemetry.propagation.http_extractors
  context = OpenTelemetry.propagation.extract(headers, context: context, http_extractors: extractors)

  span = tracer.start_span('span-name', with_context: context)

  version = OpenTelemetry.correlations.value('client-version', context: context)
  tracer.with_span(span) do |span|
    case(version)
    when 'v1.0'
      data = fetch_data_from_service_b(Context.current)
    when 'v2.0'
      data = fetch_data_from_service_c(Context.current)
    end
  end

  send_response(data)
  span.finish
end
```

```ruby
def fetch_data_from_service_b(context)
  headers = {}

  injectors = OpenTelemetry.propagation.http_injectors
  headers = OpenTelemetry.propagation.inject(headers, context: context, http_injectors: injectors)

  # make an http request
  uri = URI('https://site.com')
  http.get(('/', headers)
end
```

### Complex Modifications to Correlation Context

The `CorrelationContext::Manager` provides methods to make single modifications and multi-step modifications to correlation context as easy and as efficient as possible.

#### Single Line Modifications

```ruby
context = OpenTelemetry.correlations.set_value('k1', 'v1)
```

#### Multi-line Modifications

```ruby
# this creates a single Context instance with correlation context modified by 
# multiple operations

context = OpenTelemetry.correlations.build_context do |correlations|
  correlations.remove_value('k1')  #remove k1 from the implicit parent correlation context
  correlations.set_value('k2', 'v2')
  correlations.set_value('k3', 'v3') 
end

# which is easier and more efficient than the alternative, which creates three 
# context instances, each with a single modification of correlation context

ctx1 = OpenTelemetry.correlations.remove_value('k1')
ctx2 = OpenTelemetry.correlations.set_value('k2', 'v2', context: ctx1)
ctx3 = OpenTelemetry.correlations.set_value('k3', 'v3', context: ctx2)
```

Some may object to the use of the builder pattern in Ruby, but I think it makes for a better API. The alternative would be a method with kwargs and the following signature: `build_context(values: {}, remove_keys: [], clear: false, context: Context.current)`. The method implementation and usage is not exactly straightforward, and with the place holder `{} `and `[]` there will be more object allocations than with the builder. 

Giving the user a builder when they need it or single line modifications when they don't give users a good option for all scenarios.

### Realistic Scenario Extracting and Setting Correlations

Below is a likely realistic scenario where context is extracted off the wire and correlations setup at a request boundary.

```ruby
# extract context off the wire, includes parent span information and correlations from the caller  
extracted_context = OpenTelemetry.propagation.extract(headers)

# modify correlations from the inbound service
context = OpenTelemetry.correlations.build_context(context: extracted_context) do |correlations|
  correlations.remove_value('k1')  #remove k1 from the extracted correlation context
  correlations.set_value('k2', 'v2')
  correlations.set_value('k3', 'v3') 
end

# Set context as current
OpenTelemetry::Context.with_current(context) do
  # code that executes in the block will execute in this context
end
```

### Context considerations

During this PR I've created two different context implementations. [This commit](https://github.com/open-telemetry/opentelemetry-ruby/pull/147/commits/387f813f929ba802a5055ab58be3fd881f64456e) shows the difference.

The initial implementation maintains a reference to a parent context and a hash of entries. New contexts duplicate their parent's entries and add their own additional entries. Since contexts are immutable, they are safe to share between threads without additional synchronization.

The second, and current implementation takes a linked list approach. Each a context has a parent, key and value. Keys are looked up by traversing the parent reference until one is found. 

The hash based context is optimized for quick lookups, while the linked-list implementation is optimized (possibly) for more efficient memory usage. Both implementations expose the same API. 

I've preemptively gone with the linked list implementation, but other implementations are still on the table. We should craft some benchmarks to determine which implementation is going to be best for most use cases. When benchmarking we should look at deeply nested traces, moderately nested, and shallow traces to get a sense of the performance characteristics.

### Next steps

This PR is pretty sizable. There are a few reasons for this. One, is that it the implementation was following an in progress OTEP. Another reason, is that context propagation touches a large surface area, so some of this is inevitable. Nevertheless, there is a lot to digest in this PR and we should figure out the best way to evaluate and integrate this work. I expect I'll probably do a walk through at a future SIG meeting. I'm interested in any feedback people have about the approach and any suggestions to improve it.


mwear:
:wave: Unfortunately I cannot put this back into (GH doesn't have this ability yet), but there are going to be some fairly significant changes to this work, so I wouldn't put any effort into reviewing this at the moment. I'll update the title and let folks know when it's really ready. Sorry for the false alarm.

mwear:
I'm putting this back into a ready for review state. I expect there to be some suggestions, changes, and improvements. I think it's in a reasonable state to start getting some :eyes: on it.

api/lib/opentelemetry/context/propagation/propagation.rb
```diff
@@ -0,0 +1,154 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  class Context
+    module Propagation
+      # The Propagation class provides methods to inject and extract context
+      # to pass across process boundaries
+      class Propagation
+        HTTP_TRACE_CONTEXT_EXTRACTOR = Trace::Propagation::HttpTraceContextExtractor.new
```
benedictfischer09: A small twist on this design from the otep was that propagators were modularized and returned their extractor/injector when invoked
```
  bagExtract, bagInject = Correlations::HTTPPropagator()
  traceExtract, traceInject = Tracer::B3Propagator()
```
It might read a little more elegantly if we don't have a single module for everything propagation

api/lib/opentelemetry/context.rb
```diff
@@ -4,28 +4,146 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/context/propagation'
```
benedictfischer09: I'm still trying to wrap my head around this whole thing but I'm a little surprised to see this dep because the otep author seems to indicate
```
The Propagator API knows about the Context API, but the Context API does not need to know about Propagation.
```

https://github.com/open-telemetry/oteps/pull/66/files#r359638018

api/lib/opentelemetry/correlation_context/propagation/http_correlation_context_extractor.rb
```diff
@@ -0,0 +1,58 @@
...
+          getter ||= default_getter
+          header = getter.call(carrier, @correlation_context_key)
+
+          entries = header.gsub(/\s/, '').split(',')
+
+          correlations = entries.each_with_object({}) do |entry, memo|
+            # The ignored variable below holds properties as per the W3C spec.
+            # OTel is not using them currently, but might they may be used for
+            # metadata in the future
+            kv, = entry.split(';')
```
fbogsany: My reading of the spec is that the properties are also delimited by `;`, so `entry.split(';', 2)` could limit the number of allocations.

sdk/lib/opentelemetry/sdk/correlation_context/builder.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module CorrelationContext
+      # No op implementation of CorrelationContext::Builder
```
fbogsany: This isn't a no-op implementation.

sdk/lib/opentelemetry/sdk/correlation_context/builder.rb
```diff
@@ -0,0 +1,40 @@
...
+        def initialize(entries)
+          @entries = entries
+        end
+
+        # Set key-value in the to-be-created correlation context
+        #
+        # @param [String] key The key to store this value under
+        # @param [String] value String value to be stored under key
+        def set_value(key, value)
+          @entries[key] = value.to_s
```
fbogsany: Do we need the `.to_s` if we expect `value` to be a `String`?

api/lib/opentelemetry/trace/propagation/http_trace_context_injector.rb
```diff
@@ -0,0 +1,53 @@
...
+          setter.call(carrier, @tracestate_header_key, span_context.tracestate) unless span_context.tracestate.nil?
+
+          carrier
+        end
+
+        private
+
+        def span_context_from(context)
+          context[ContextKeys.current_span_key]&.context ||
+            context[ContextKeys.extracted_span_context_key]
```
fbogsany: Way too much `context` here, and yet I still feel like I'm missing context 😞

api/lib/opentelemetry/trace/propagation/http_trace_context_extractor.rb
```diff
@@ -0,0 +1,56 @@
...
+        # @param [String] tracestate_header_key The tracestate header key used in the carrier
+        # @return [HttpTraceContextExtractor]
+        def initialize(traceparent_header_key: 'traceparent',
+                       tracestate_header_key: 'tracestate')
+          @traceparent_header_key = traceparent_header_key
+          @tracestate_header_key = tracestate_header_key
+        end
+
+        # Extract a remote {Trace::SpanContext} from the supplied carrier.
+        # Invalid headers will result in a new, valid, non-remote {Trace::SpanContext}.
```
fbogsany: This feels wrong to me. If I fail to extract the headers for some reason, I most likely want to create a new root span for a trace. This makes it harder to distinguish that case - I have to check if the extracted `SpanContext` is remote, and ignore it if it is non-remote.

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -8,11 +8,29 @@ module OpenTelemetry
...
+      # Returns the the active span context from the given {Context}, or current
+      # if one is not explicitly passed in. The active span context may refer to
+      # a {SpanContext} that has been extracted. If both a current {Span} and an
+      # extracted, {SpanContext} exist, the context of the current {Span} will be
+      # returned.
+      #
+      # @param [optional Context] context The context to lookup the active
+      #   {SpanContext} from.
+      #
+      def active_span_context(context = nil)
```
fbogsany: I'm a little confused about when I should use `active_span_context` vs. `current_span&.context`.

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -54,10 +72,12 @@ def start_root_span(name, attributes: nil, links: nil, start_timestamp: nil, kin
       #   +with_parent_context+.
       # @param [optional SpanContext] with_parent_context Explicitly managed. Overridden by
       #   +with_parent+.
+      ## @param [optional Context] with_context Explicitly managed context. Overridden by
+      #   +with_parent_context+ or +with_parent+
       #
       # @return [Span]
-      def start_span(name, with_parent: nil, with_parent_context: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil, sampling_hint: nil)
-        span_context = with_parent&.context || with_parent_context || current_span.context
+      def start_span(name, with_parent: nil, with_parent_context: nil, with_context: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil, sampling_hint: nil)
```
fbogsany: The new parameter name is **really** confusing. We now have 3 parameters to specify the parent span context, and they're confusingly similarly named while accepting completely different argument types:
- `with_parent`: `Span`
- `with_parent_context`: `SpanContext`
- `with_context`: `Context`

The latter expands to 5 different options:
- implicit context current span's context
- implicit context extracted span context
- explicit context current span's context
- explicit context extracted span context
- invalid context

api/lib/opentelemetry/correlation_context/propagation/context_keys.rb
```diff
@@ -0,0 +1,24 @@
...
+    module Propagation
+      # The ContextKeys module contains the keys used to store correlations
+      # in a {Context} instance
+      module ContextKeys
+        extend self
+
+        CORRELATION_CONTEXT_KEY = 'correlation-context'
+        private_constant(:CORRELATION_CONTEXT_KEY)
+
+        def span_context_key
```
fbogsany: I find it slightly confusing to call this `span_context_key` when it is actually the correlation context key.

api/lib/opentelemetry/context.rb
```diff
@@ -4,28 +4,146 @@
...
+    # context is not the current context, a warning will be logged, but
+    # ctx_to_attach will stil be made the current context.
+    #
+    # @param ctx_to_attach The ctx to be attached when this context is detached
+    def detach(ctx_to_attach = nil)
+      OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if self.class.current != self
+
+      ctx_to_attach ||= @parent || ROOT
+      ctx_to_attach.attach
+    end
```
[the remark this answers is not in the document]
mwear: These are all valid concerns and I share most of them. 

I don't think that `attach` and `detach` should be exposed if we can avoid it. I'll double check and see, but we might be able to remove them. Other options would be to put an `@api private` comment, or if we must expose them for edge cases, we can put a nasty gram in the comment to use at your own risk. They were inspired by this prior art: https://github.com/grpc/grpc-java/blob/master/context/src/main/java/io/grpc/Context.java#L405-L449.

I borrowed the single valued contexts idea from go. Take a look at line 480 until the end of the file: https://golang.org/src/context/context.go. You can also take a look at the hash based context I had earlier in this PR: https://github.com/open-telemetry/opentelemetry-ruby/pull/147/commits/a22646947971d3ce048b52626023d355acd09227. Let me know if you'd be more comfortable with one over the other.

api/lib/opentelemetry/trace/propagation/http_trace_context_extractor.rb
mwear: I agree this looks a little 🐟 y, but I didn't add this behavior in this PR. This is how it's implemented on master. I'll do some research to figure out the backstory.

api/lib/opentelemetry/trace/tracer.rb
mwear: You should always use always use `active_span_context`. This because of complications around implicit parenting and whether you have a current span in the context, or an extracted span context (or both). 

The easiest way to deal with this complication is to establish the precedence that the current span's context will be used as an implicit parent if one exists, otherwise the extracted span context will be used as an implicit parent. 

I tried other approaches and this one worked the best. It also was viewed favorably by other SIGs as a good approach. It was recommended that I should draft some text to add as guidance for SDKs, which I did, but I haven't followed through on publishing it anywhere yet. I should followup on that.

This complication sparked a bit of interest in: https://github.com/open-telemetry/oteps/issues/68. No spans, no problem!

api/lib/opentelemetry/trace/tracer.rb
mwear: We could probably get rid of `with_parent_context`. I don't see a use case where you'd have a span context without a `Context` or a `Span` any longer.

api/lib/opentelemetry/context/propagation/propagation.rb
mwear: This definitely needs some clean up. I have some ideas.

api/lib/opentelemetry/context.rb
mwear: The `Context` API is actually separate. That require is for namespacing reasons, so that all propagation code nests under `OpenTelemetry::Context::Propagation`. Since `OpenTelemetry::Context` doesn't actually depend on anything in `OpenTelemetry::Context::Propagation`, we could consider moving propagation up to the top level, `OpenTelemetry::Propagation`. However, propagation does depend on context, so there is a relationship there.

api/lib/opentelemetry/trace/propagation/http_trace_context_injector.rb
mwear: The context behind this is the same as: https://github.com/open-telemetry/opentelemetry-ruby/pull/147#discussion_r364892343

api/lib/opentelemetry/context/propagation/propagation.rb
mwear: This should be a little better now.

api/lib/opentelemetry/context.rb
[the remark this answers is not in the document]
mwear: I went ahead and removed `Context#attach` and `Context#detach`. If they aren't necessary for the block form, and if we find they become necessary, we can bring them back later with caveats and warnings.

api/lib/opentelemetry/correlation_context/propagation/http_correlation_context_extractor.rb
mwear: Good idea!

sdk/lib/opentelemetry/sdk/correlation_context/builder.rb
mwear: I was being extra safe, but I'm with you; we shouldn't enable poor API use.

api/lib/opentelemetry/trace/tracer.rb
mwear: I got rid of `with_context` and repurposed `with_parent_context` to take a `Context` instead of a `SpanContext`.

api/lib/opentelemetry/trace/tracer.rb
mwear: This has been added to the OTEP now: https://github.com/open-telemetry/oteps/pull/66/commits/286fb0a8e6c6d488ca4a6a4c997a7d43524d20b2#diff-129ca9568557ecf192a3d13a0398c809R514-R534

api/lib/opentelemetry/trace/propagation/http_trace_context_extractor.rb
mwear: I'm not exactly sure how we ended up with this implementation, but I definitely think it is wrong. I think we should return `SpanContext::INVALID` when parsing fails and everything else should work out as expected. Do you agree?

mwear: Now that extract returns a context, I think it makes sense to return the original context when parsing fails. I've gone with that in the current round of changes.

mwear:
I've addressed everything except the single-valued contexts question as I think that is still up to debate. This is ready for more feedback when folks have the time.

api/lib/opentelemetry/context.rb
```diff
@@ -4,28 +4,126 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/context/propagation'
+
 module OpenTelemetry
-  # The Context module provides per-thread storage.
-  module Context
-    extend self
+  # Manages context on a per-thread basis
```
fbogsany: Technically, per-fiber.

sdk/lib/opentelemetry/sdk/correlation_context/builder.rb
```diff
@@ -0,0 +1,40 @@
...
+        end
+
+        # Removes key from the to-be-created correlation context
+        #
+        # @param [String] key The key to remove
+        def remove_value(key)
+          @entries.delete(key)
+        end
+
+        # Clears all correlations from the parent context
```
fbogsany: s/parent/to-be-created correlation/ ?

sdk/lib/opentelemetry/sdk/correlation_context/manager.rb
```diff
@@ -0,0 +1,86 @@
...
+        #   to +Context.current+
+        # @return [Context]
+        def clear(context: Context.current)
+          context.set_value(CORRELATION_CONTEXT_KEY, EMPTY_CORRELATION_CONTEXT)
+        end
+
+        # Returns the corresponding correlation value (or nil) for key
+        #
+        # @param [String] key The lookup key
+        # @param [optional Context] context The context use to retrieve key.
```
fbogsany: s/use/from which/

sdk/lib/opentelemetry/sdk/correlation_context/manager.rb
```diff
@@ -0,0 +1,86 @@
...
+
+        # Returns a new context with value at key removed
+        #
+        # @param [String] key The key to remove
+        # @param [optional Context] context The context to remove correlation
+        #   from. Defaults to +Context.current+
+        # @return [Context]
+        def remove_value(key, context: Context.current)
+          correlations = correlations_for(context)
+          return context unless correlations.key?(key)
```
fbogsany: I'm still wrapping my head around how this all hangs together... Does the early return of the existing context (instead of always returning a new context) create any surprises for the caller?

sdk/test/opentelemetry/sdk/correlation_context/manager_test.rb
```diff
@@ -0,0 +1,190 @@
...
+    describe 'explicit context' do
+      it 'returns context with key removed from correlation context' do
+        ctx = manager.set_value('foo', 'bar', context: Context.empty)
+        _(manager.value('foo', context: ctx)).must_equal('bar')
+
+        ctx2 = manager.remove_value('foo', context: ctx)
+        _(manager.value('foo', context: ctx2)).must_be_nil
+      end
+
+      it 'returns same context if key does not exist' do
```
fbogsany: Is this required behaviour or an implementation detail? Should callers care?

sdk/test/opentelemetry/sdk/correlation_context/manager_test.rb
```diff
@@ -0,0 +1,190 @@
...
+      it 'returns context with key removed from correlation context' do
+        Context.with_current(manager.set_value('foo', 'bar')) do
+          _(manager.value('foo')).must_equal('bar')
+
+          ctx = manager.remove_value('foo')
+          _(manager.value('foo', context: ctx)).must_be_nil
+        end
+      end
+
+      it 'returns same context if key does not exist' do
```
fbogsany: Same question here.

api/lib/opentelemetry/context.rb
```diff
@@ -4,28 +4,126 @@
...
+      end
+
+      private
+
+      def attach(ctx, yield_value = nil, &blk)
+        prev = current
+        self.current = ctx
+        yield_value.nil? ? blk.call : blk.call(yield_value)
+      ensure
+        self.current = prev
```
fbogsany: This method is slightly less efficient than its predecessor `with`. The old method accessed the fiber-local variable once (returning a hash), and then manipulated the hash. The new version has the fiber-local variable access, and two fiber-local variable sets. I'm unsure of the added overhead, but it will add up as uses of context correlations and spans increase.

I'm not sure there's a clean and efficient alternative - really just raising the concern at this point.

api/lib/opentelemetry/correlation_context/propagation/http_extractor.rb
```diff
@@ -0,0 +1,58 @@
...
+        #
+        # @param [Context] context The context to be updated with extracted correlations
+        # @param [Carrier] carrier The carrier to get the header from
+        # @param [optional Callable] getter An optional callable that takes a carrier and a key and
+        #   returns the value associated with the key. If omitted the default getter will be used
+        #   which expects the carrier to respond to [] and []=.
+        # @yield [Carrier, String] if an optional getter is provided, extract will yield the carrier
+        #   and the header key to the getter.
+        # @return [Context] context updated with extracted correlations, or the original context
+        #   if extraction fails
```
fbogsany: It is not clear to me from reading the code that we ever return `the original context if extraction fails`.

api/lib/opentelemetry/context.rb
```diff
@@ -4,28 +4,126 @@
...
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/context/propagation'
+
 module OpenTelemetry
-  # The Context module provides per-thread storage.
-  module Context
-    extend self
+  # Manages context on a per-thread basis
+  class Context
```
fbogsany: Thinking some more about this, the [Context API](https://github.com/open-telemetry/oteps/blob/master/text/0066-separate-context-propagation.md#Context-API) from the OTEP differs in an important way from this implementation: it recommends a `CreateKey(name) -> key` function that returns a key for a name that is subsequently used to get, set and remove values.

I think the addition of this abstraction might allow a more efficient underlying implementation.

The OTEP describes two cross-cutting concerns (Observability and Correlations), while the Context API is generic enough to support additional concerns outside of the two specified. I think there is value in leveraging knowledge of the two specified concerns to improve lookup efficiency. For example, giving `Context` three fields:
```ruby
@observability = nil
@correlations = nil
@extensions = {}.freeze
```
and copying the fields in `attach` rather than creating a linked list.

My assumption here is that we can define a private class that contains all the relevant context for Observability and likewise for Correlations. For Correlations, that may well be just a frozen Hash.

sdk/lib/opentelemetry/sdk/correlation_context/manager.rb
mwear: Given that Contexts are immutable this should not be a surprise and it should be a safe optimization.

sdk/test/opentelemetry/sdk/correlation_context/manager_test.rb
mwear: It's an implementation detail (optimization).

api/lib/opentelemetry/context.rb
mwear: I returned to the previous implementation that had `Context#attach` and `Context#detach`. I just put an `@api private` tag on them.

api/lib/opentelemetry/context.rb
mwear: The `Context` object itself should be a general purpose context object that is not tied to observability or even OpenTelemetry. There has been some talk of extracting Context into a separate package at some point in the future. Whether or not that will happen remains to be seen, but we should design with that goal in mind.

i went ahead and reverted back to my original frozen hash design to alleviate concerns over the linked list approach. I also implemented `Context#create_key` and am using `Context::Key`s for indexing entries. I think this gives us what we need and keeps the context decoupled from observability.

api/lib/opentelemetry/correlation_context/propagation/http_extractor.rb
mwear: I added code to make the comment accurate 🙂.

mwear:
I think I have addressed the feedback from this most recent round. I also went ahead and did a rebase since this was getting a bit stale. It's ready for 👀 whenever folks have time.

fbogsany:
Follow-up from today's SIG meeting: this should be rebased and merged as-is. It's been hanging out a long time and is difficult to repeatedly review. We'll address any concerns with issues & PRs.

The primary contention at present is the implementation of the `Context`. The concerns are mostly related to efficiency, and we're better off getting some practical experience, benchmarks, and data to figure out if the concerns are warranted.

mwear:
Thanks for your reviews. As mentioned during our meeting, this is just a starting point and we'll continue to refine this work over time.

Issue #67, opened 2019-09-09
Instrumentation: port Datadog's auto-instrumentation libraries


fbogsany:
See https://github.com/open-telemetry/oteps/pull/41
Relevant code is here: https://github.com/DataDog/dd-trace-rb (instrumentation here: https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib).

elskwid:
@fbogsany I'm digging in to the DataDog Integration instrumentation libraries. Not sure how you go about "assigning" these issues but I'll take this one. I'll post my thoughts here after I've had a chance to read the docs and spend a little time with the way DD has set up the integrations.

fbogsany:
> Not sure how you go about "assigning" these issues but I'll take this one.

❤️ 

Looks like only codeowners can assign issues to themselves, which is a little annoying for everyone else. We'll have to do it informally like this for now.

fbogsany:
Oh, I'm wrong, I think - I assigned you.

fbogsany:
Experimentation suggests that if you comment on an issue, then it can be assigned to you.

elskwid:
@fbogsany @mwear here's what I have from the SIG meeting today. I can't seem to update the issue description. Let me know how this looks.

## Requirements

* Create gem per integration instrumentation
* Create top-level directory for instrumentations
    * We will also do this for exporters
* Create initial integration gems without auto-installation
    * This can be added later (See #19)
* Unify configuration where possible


## Preliminary plan

* Break out single instrumentation from `dd-trace-rb`
* Add enough configuration to support first instrumentation
* Use manual install method for instrumentation (versus auto-installation)
* Create simple app to wire up SDK for first instrumentation

**Next up**:
* Scope out auto-installation based on first instrumentation
* Add more instrumentations

**Notes**:
* Consider options for testing including: appraisals, ci configuration, and Docker Compose

fbogsany:
The spec refers to these things as "instrumentation adapters":
https://github.com/open-telemetry/opentelemetry-specification/blob/57714f7547fe4dcb342ad0ad10a80d86118431c7/specification/overview.md#instrumentation-adapters

I'm not sure we need to use both words in describing them. I'd be happy with "pick one" for the top-level directory name and gem names. I.e. either:
```
opentelemetry-ruby/adapters/redis/opentelemetry-adapter-redis.gemspec
```
or
```
[...]

PR #148, opened 2019-11-11
Add faraday instrumentation adapter #67

duonoid:
# Overview
Adds faraday instrumentation adapter.

## Details
* faraday adapter exists in its own gem
* Adds a runnable example for demonstration purposes and testing

## Related
* Effort toward resolving #67 
* #155


adapters/faraday/lib/opentelemetry/adapters/faraday.rb
```diff
@@ -0,0 +1,21 @@
...
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module_function
+
+      TRACER_NAME = 'faraday'
+      TRACER_VERSION = Gem.loaded_specs[TRACER_NAME].version.to_s
+
+      def install(config = {name: TRACER_NAME, version: TRACER_VERSION})
```
mwear: If at all possible, I'd like to see the auto-installer take care of binding the tracer to instrumentation. I think the install method should take a reference to the tracer and a config. We can talk through this more to see if there are any complications with this scheme.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,43 @@
...
+    module Faraday
+      module Middlewares
+        class TracerMiddleware < ::Faraday::Middleware
+          def call(env)
+            tracer.in_span(env.url.to_s,
+                          attributes: { 'component' => 'http',
+                                        'http.method' => env.method },
+                          kind: :client) do |span|
+              trace_request(span, env)
+
```
mwear: We're missing context propagation in the request.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,43 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module Middlewares
+        class TracerMiddleware < ::Faraday::Middleware
+          def call(env)
+            tracer.in_span(env.url.to_s,
+                          attributes: { 'component' => 'http',
```
mwear: We'll want to go through the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md) with a fine-toothed comb.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,43 @@
...
+          private
+
+          attr_reader :app
+
+          def tracer
+            Faraday::Adapter.tracer
+          end
+
+          def trace_request(span, env)
+            span.set_attribute('http.url', env.url.to_s)
```
fbogsany: I think this can be set as part of the `attributes:` to `in_span`.

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module_function
+
+      TRACER_NAME = 'faraday'
+      TRACER_VERSION = Gem.loaded_specs[TRACER_NAME].version.to_s
```
mwear: This won't work if Faraday isn't present, which is a use case we're going to want to accommodate.

adapters/faraday/opentelemetry-adapters-faraday.gemspec
```diff
@@ -0,0 +1,32 @@
...
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.4.0'
+
+  spec.add_dependency 'opentelemetry-api', '~> 0.0'
+  spec.add_dependency 'faraday', '~> 0.17.0'
```
mwear: We probably want to refrain from a hard dependency on Faraday. That will allow us to ship more instrumentation than users will need, but auto-install the libraries that are applicable for their system.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
duonoid: I've now added `propogate_context`, but it could use more verification.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
duonoid: It looks like for the client, that only `http.url` is required.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
duonoid: Changed :heavy_check_mark:

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
duonoid: I've changed the line to handle when faraday isn't explicitly loaded already. :heavy_check_mark:

adapters/faraday/opentelemetry-adapters-faraday.gemspec
duonoid: I've changed this to be a development dependency.  :heavy_check_mark:

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,49 @@
...
+            end
+          end
+
+          private
+
+          attr_reader :app
+
+          def propagate_context(span, env)
+            formatter.extract(env)
+            formatter.inject(span.context, env.request_headers)
```
fbogsany: We only want the `inject` here, not the `extract`. For outbound requests, we *inject* the current span context into the request headers. For an inbound request (e.g. in the Sinatra adapter), we *extract* the remote span context from the request headers and start a new span with the extracted context as the parent context (`with_parent_context`).

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
duonoid: I've made changes to allow for `config[:tracer]`, which overrides the `default_tracer` implementation (in `api/lib/opentelemetry/adapter`).

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
duonoid: :heavy_check_mark:

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
duonoid: Thanks for that guidance!  I've made the relevant changes for this PR, and will need to check #149 as well :heavy_check_mark:

duonoid:
I've extracted the API changes to #154.

duonoid:
> we'll need to consider testing and how that would look

I have noted this point in issue #155.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,49 @@
...
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module Middlewares
+        class TracerMiddleware < ::Faraday::Middleware
+          def call(env)
+            tracer.in_span(env.url.to_s,
+                          attributes: { 'component' => 'http',
+                                        'http.method' => env.method,
```
mwear: dd-trace-rb has the target host and port as attributes: https://github.com/DataDog/dd-trace-rb/blob/master/lib/ddtrace/contrib/faraday/middleware.rb#L45-L46. 

I think we should strive the keep the same attributes and see if there are analogs in the semantic conventions. If they are absent from the semantic conventions, perhaps we should consider opening a spec issue or PR to add them.

fbogsany: [Semantic conventions for HTTP client](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#http-client) state:

> One of the following sets of attributes is required (in order of usual preference unless for a particular web client/framework it is known that some other set is preferable for some reason; all strings must be non-empty):
>
> * http.url
> * http.scheme, http.host, http.target
> * http.scheme, peer.hostname, peer.port, http.target
> * http.scheme, peer.ip, peer.port, http.target

So I think just recording `http.url` is preferred.

mwear: WFM 👍

duonoid:
I have made a change (revertable, if needed) to remove the requirement to change the API (`OpenTelemetry::Adapter` #154) by isolating the "propagator/formatter" into the `opentelemetry/adapters/faraday/adapters.rb` file.

My hope is to be able to defer some of those "superclass/abstraction" decisions until we understand the shape of things (e.g., "propagators") better.

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module_function
+
+      TRACER_NAME = 'faraday'
+
+      def install(config = {name: TRACER_NAME,
+                            version: tracer_version})
```
fbogsany: Do the `name` and `version` need to be overridable by consumers of the adapter?

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
```diff
@@ -0,0 +1,26 @@
...
+        require_relative 'faraday/adapter'
+        Faraday::Adapter.install(config)
+      end
+
+      def tracer_version
+        Gem.loaded_specs[TRACER_NAME]&.version.to_s
```
fbogsany: If the `name` is overridable, the `version` is looked up, potentially, using a different name from the one provided. Is that the right thing to do? The `version` returned here would be the Faraday version rather than the adapter gem version. I'm not sure that's the intent -- I *think* the version is supposed to be the adapter version.

adapters/faraday/lib/opentelemetry/adapters/faraday/adapter.rb
```diff
@@ -0,0 +1,57 @@
...
+      class Adapter
+        class << self
+          attr_reader :config,
+                      :propagator,
+                      :tracer
+
+          def install(config = {})
+            @config = config
+            @propagator = OpenTelemetry.tracer_factory.rack_http_text_format
+            @tracer = config[:tracer] || default_tracer
```
fbogsany: I don't think the tracer should be overridable. We should just use the one we lookup from the tracer factory.

duonoid:
I have merged in the testing work mentioned in #155, now that the "shape" of the API is a little more solid (per sig feedback, etc.).

adapters/faraday/gemfiles/faraday_0.13.gemfile
```diff
@@ -0,0 +1,12 @@
+# This file was generated by Appraisal
```
fbogsany: I haven't used Appraisal before. Is it common practice to commit these generated files to source control?

duonoid: I'm not sure if it's common practice, but the Appraisal docs recommend it: https://github.com/thoughtbot/appraisal#version-control (commit gemfiles, exclude lock files).

mwear: You _could_ probably just run `bundle exec rake appraisal generate` before running tests, but this is probably a shortcut.

adapters/faraday/lib/opentelemetry/adapters/faraday/adapter.rb
```diff
@@ -0,0 +1,55 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'faraday'
+require 'opentelemetry'
+
+require_relative 'middlewares/tracer_middleware'
```
mwear: Do we have a preference for `require_relative` vs `require`? I have a mild preference for `require`, but don't want to force it on everyone unless there's more consensus.

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
elskwid: @fbogsany this was removed after our SIG work session.

adapters/faraday/lib/opentelemetry/adapters/faraday.rb
elskwid: We discussed this at our last SIG - the version should be the version of the adapter and we've removed it from the config and made it so it is not overridable.

adapters/faraday/lib/opentelemetry/adapters/faraday/adapter.rb
elskwid: This was moved to a helper method and removed from config after our SIG work session.

adapters/faraday/test/opentelemetry/adapters/faraday/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,53 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+# require Adapter so .install method is found:
+require_relative '../../../../../lib/opentelemetry/adapters/faraday/adapter'
+require_relative '../../../../../lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware'
+
+describe OpenTelemetry::Adapters::Faraday::Middlewares::TracerMiddleware do
```
mwear: Just comparing this test to the data dog version: https://github.com/DataDog/dd-trace-rb/blob/master/spec/ddtrace/contrib/faraday/middleware_spec.rb. and there's much less coverage here: 

I assume that we do not have all of the same functionality, so less tests are expected. Do we have a corresponding test for all common functionality?

adapters/faraday/lib/opentelemetry/adapters/faraday/adapter.rb
duonoid: I tend to use `require` for resources (gems) outside of the project, and `require_relative` for internal files.

mwear: That makes sense and appears to be an established convention. I redact my preference for require 🙂 .

adapters/faraday/test/opentelemetry/adapters/faraday/middlewares/tracer_middleware_test.rb
duonoid: Most of those tests are for configuration overrides, distributed context, overriding service name, opt-in/opt-out request header tracing... a lot of functionality that doesn't exist in this version.

I think we do have decent coverage for the functionality that does exist.

mwear: Thanks for sifting through those and double checking!

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Faraday
+      module Middlewares
+        class TracerMiddleware < ::Faraday::Middleware
+          def call(env)
+            tracer.in_span(env.url.to_s,
```
benedictfischer09: I don't know if we need to solve this as part of this PR, but what happens if the tracer uses Faraday as the transport library to report spans to a collector?  It would be up to the tracer to inspect the spans generated from that and discard them? Is there a need for this middleware to provide a hook that could be used to inspect the request and not generate a span for a particular request

fbogsany: I believe this can be worked around using the Faraday middleware mechanisms, but we struggled to understand that last time we looked at it. I believe it amounts to using, but disabling the `Middlewares::TracerMiddleware` at the point where we open the connection.

duonoid: I've added a commit that should address this "disable for span reporting connection" issue.

duonoid:
I saw that faraday-1.0.0 was released, and have added an Appraisal for it.

Also, there seems to be an issue with circleci + bundler-2.1.x (including 2.1.4) + appraisal (2.2.0), which is solved locally by using bundler-2.0.x and jruby-2.8.x (circleci config).

mwear:
:wave: @duonoid. I had to integrate the changes from master before merging this PR (which I was just about to do). We now test against Ruby 2.7 and those tests are failing due to not being able to find a compatible version of bundler: https://circleci.com/gh/mutations/opentelemetry-ruby/1498?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link. Could you take a look?

duonoid:
@mwear I had previously been able to work around the bundler+appraisal issues that I saw by uninstalling other versions of bundler and using, e.g., 2.0.x.  It seems that ruby-2.7 is now including bundler-2.1.2 as a default gem, which cannot be uninstalled because it is a default gem

https://github.com/thoughtbot/appraisal/issues/162 mentions a similar issue

For an immediate workaround, I could try removing Appraisal and just use `rake test`.

mwear:
Does this mean that we are only testing against one version of Faraday now? If so, I'm wondering how hard it would be to disable appraisal for Ruby 2.7 only. If it's more effort than it should be, then I'd settle for this and an issue to reinstate appraisal once a fix has been found.

duonoid:
@mwear  I have made some circle ci config changes to exclude 'appraisals'  from running under ruby-2.7 for the time-being.

mwear:
:tada: Thanks for all your hard work @duonoid!

PR #149, opened 2019-11-11
Add sinatra instrumentation adapter  #67

duonoid:
# Overview
Adds sinatra instrumentation adapter.

## Details
* Implements tracing using sinatra middleware and extension
* Adds runnable example for demonstration and testing

## Related
* Similar: #148
* Effort toward resolving #67


adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
```diff
@@ -0,0 +1,37 @@
...
+      module Extensions
+        module TracerExtension
+          # Sinatra hook after extension is registered
+          def self.registered(app)
+            # Create tracing `render` method
+            ::Sinatra::Base.module_eval do
+              def render(engine, data, *)
+                Sinatra::Adapter.tracer.in_span(
+                  'sinatra.render_template',
+                  kind: :server,
```
fbogsany: `kind:` shouldn't be specified here. It will default to `:internal`. This span isn't a server span - it's parent span is (the middleware span).

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
```diff
@@ -0,0 +1,37 @@
...
+                  with_parent: Sinatra::Adapter.tracer.current_span
```
fbogsany: This shouldn't be necessary. The `current_span` will be used as the parent by default.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
```diff
@@ -0,0 +1,37 @@
...
+                ) do |span|
+                  template_name = data.is_a?(Symbol) ? data : :literal
+                  span.set_attribute('sinatra.template_name', template_name.to_s)
+                  output = super
```
fbogsany: Do you need to assign to `output` or can you just call `super` here?

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+          def parent_context(env)
+            OpenTelemetry.tracer_factory.http_text_format.extract(env)
+          end
+
+          def tracer
+            OpenTelemetry::Adapters::Sinatra::Adapter.tracer
+          end
+
+          def trace_request(span, env)
+            span.set_attribute('http.url', env['PATH_INFO'])
```
fbogsany: Can this be passed as part of the `attributes:` hash to `in_span`?

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,57 @@
...
+          end
+
+          def trace_response(span, env, resp)
+            status, _headers, _response_body = resp
+
+            span.set_attribute('http.status_code', status)
```
fbogsany: We should set the `span.status` here as well.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
duonoid: fixed :heavy_check_mark: by removing `kind:`

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
duonoid: I've removed `with_parent:` :heavy_check_mark:

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
duonoid: I've addressed some other unused variables as well.  :heavy_check_mark:

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
duonoid: I've moved the call to `attributes:`, which I'm assuming is performance-related (due to the mutex in the sdk implementation of `set_attributes`?). :heavy_check_mark:

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
duonoid: Hopefully on the right track here... I had to add `Trace::Status.from_http_status(code) => Status`, which didn't _exactly_ map 1-to-1 http -> gRPC status.

duonoid:
This PR also uses #154.

duonoid:
I could extract the "gRPC status" changes to a separate PR, if needed.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+          def call(env)
+            tracer.in_span(
+              env['PATH_INFO'],
+              attributes: { 'component' => 'http',
+                            'http.method' => env['REQUEST_METHOD'],
+                            'http.url' => env['PATH_INFO'] },
+              kind: :server,
+              with_parent_context: parent_context(env)
+            ) do |span|
+              app.call(env).tap { |resp| trace_response(span, env, resp) }
```
mwear: dd-trace-rb has the ability to opt-in to collecting additional headers: https://github.com/DataDog/dd-trace-rb/blob/master/lib/ddtrace/contrib/faraday/middleware.rb#L45-L46.

Should we keep this functionality?

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+            OpenTelemetry.tracer_factory.http_text_format.extract(env)
+          end
+
+          def tracer
+            OpenTelemetry::Adapters::Sinatra::Adapter.tracer
+          end
+
+          def trace_response(span, env, resp)
+            status, _headers, _response_body = resp
+
```
mwear: It similarly has the ability to opt-in to collecting response_headers: https://github.com/DataDog/dd-trace-rb/blob/master/lib/ddtrace/contrib/sinatra/tracer_middleware.rb#L37-L39.

Should we preserve this behavior?

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
fbogsany: Didn't see the opt-in on the lines you pointed at, but the [semantic conventions for HTTP clients](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#http-client) include the option of `peer.hostname` and `peer.port` or `peer.ip` and `peer.port`, so I think it's worth considering their inclusion here. That said, the semantic conventions favour `http.url`:

> One of the following sets of attributes is required (in order of usual preference unless for a particular web client/framework it is known that some other set is preferable for some reason; all strings must be non-empty):
>
> * http.url
> * http.scheme, http.host, http.target
> * http.scheme, peer.hostname, peer.port, http.target
> * http.scheme, peer.ip, peer.port, http.target

fbogsany: I think you mean https://github.com/DataDog/dd-trace-rb/blob/ab135f65c4d761ddbac02d4d0f04e3c17382218f/lib/ddtrace/contrib/sinatra/tracer_middleware.rb#L30-L32

I'd lean toward "no" for the initial implementation, just to keep it simple and focussed on the current semantic specifications.

api/lib/opentelemetry/trace/status.rb
```diff
@@ -19,9 +19,56 @@ class Status
...
 
+      # Implemented according to
+      # https://cloud.google.com/apis/design/errors#handling_errors
+      #
+      # Note that some HTTP status do not map 1-to-1 to a gRPC status.
+      #
+      # @param code Numeric HTTP status
+      #
+      # @return Status
+      def self.from_http_status(code) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
```
fbogsany: This has been extracted to #157 and merged, so can be removed from here.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
mwear: lol... yes... that's the correct link

duonoid:
I have sync'd changes per recent discussions on #148 (API changes, tests, etc.)

adapters/sinatra/lib/opentelemetry/adapters/sinatra/adapter.rb
```diff
@@ -0,0 +1,67 @@
...
+          end
+        end
+
+        def install
+          register_tracer_extension
+        end
+
+        private
+
+        def register_tracer_extension
```
mwear: We are likely going to need an `InstrumentationRegistry` that is part of the OpenTelemetry::API to ensure that instrumentation is installed only once. Is the point of this code to fake that we have a registry so that it can be easily changed later?

If so, I'm wondering if it wouldn't be better for us to just have a class level `installed?` flag for now, and update it to use the actual registry later?

duonoid: I've switched to the `installed?` flag.

adapters/sinatra/lib/opentelemetry/adapters/sinatra.rb
```diff
@@ -0,0 +1,30 @@
...
+      module_function
+
+      def install(config = {})
+        require_relative 'sinatra/adapter'
+        Sinatra::Adapter.install(config)
+      end
+
+      # Convenience method to access the nested module name
+      def name
+        Module.nesting[0].to_s
```
fbogsany: TIL

adapters/sinatra/lib/opentelemetry/adapters/sinatra/adapter.rb
```diff
@@ -0,0 +1,63 @@
...
+              Sinatra.version
+            )
+          end
+
+          attr_accessor :installed
+          alias_method :installed?, :installed
+
+          private
+
+          def tracer_factory
```
fbogsany: This method seems unnecessary. It is only called in one place (`install`) and is *not* used in one other place it could be (`tracer`). I don't think it provides enough value.

adapters/sinatra/example/server.rb
```diff
@@ -0,0 +1,43 @@
...
+SDK = OpenTelemetry::SDK
+
+factory = OpenTelemetry.tracer_factory = SDK::Trace::TracerFactory.new
+factory.add_span_processor(
+  SDK::Trace::Export::SimpleSpanProcessor.new(
+    SDK::Trace::Export::ConsoleSpanExporter.new
+  )
+)
+
+OpenTelemetry::Adapters::Sinatra.install(name: 'sinatra-example', version: '1.0')
```
fbogsany: The `config` hash isn't used anywhere. It seems confusing to specify `name` and `version` here when they're not used in anyway and don't actually mean anything.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/adapter.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+        end
+
+        def install
+          return :already_installed if self.class.installed?
+          self.class.installed = true
+
+          register_tracer_extension
+
+          :installed
```
fbogsany: I'm probably missing a bunch of earlier discussion. What is the intent behind the `:installed` and `:already_installed` return values? I don't see the return value used in the example.

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
```diff
@@ -0,0 +1,36 @@
...
+          # Sinatra hook after extension is registered
+          def self.registered(app)
+            # Create tracing `render` method
+            ::Sinatra::Base.module_eval do
+              def render(_engine, data, *)
+                Sinatra::Adapter.tracer.in_span(
+                  'sinatra.render_template'
+                ) do |span|
+                  template_name = data.is_a?(Symbol) ? data : :literal
+                  span.set_attribute('sinatra.template_name', template_name.to_s)
```
fbogsany: `template_name` can be computed before the `span` is created. It will be more efficient to provide the attribute to `in_span`. I.e.:
```ruby
def render(_engine, data, *)
  template_name = data.is_a?(Symbol) ? data : :literal
  Sinatra::Adapter.tracer.in_span(
    'sinatra.render_template',
    attributes: {'sinatra.template_name' => template_name.to_s}
  ) do
    super
  end
end
```

adapters/sinatra/lib/opentelemetry/adapters/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,53 @@
...
+              app.call(env).tap { |resp| trace_response(span, env, resp) }
+            end
+          end
+
+          private
+
+          attr_reader :app
+
+          def parent_context(env)
+            OpenTelemetry::Adapters::Sinatra::Adapter.propagator.extract(env)
```
benedictfischer09: Is it the job of a propagator to inspect this payload and decide whether it is trustworthy? Can someone craft headers malicious headers that would cause confusion in tracing systems?  Maybe a simple flag to always start a new trace instead of accepting existing context.  Or, this might be a case where offering hooks into the instrumentation would enable more sophisticated behavior like client code looking at the request and deciding whether to allow the context it contains to be extracted

adapters/sinatra/lib/opentelemetry/adapters/sinatra/adapter.rb
duonoid: removed :heavy_check_mark:

adapters/sinatra/example/server.rb
duonoid: Updated to avoid passing `name` and `version` :heavy_check_mark:

adapters/sinatra/lib/opentelemetry/adapters/sinatra/extensions/tracer_extension.rb
duonoid: Updated to pass data when span is created (for performance). :heavy_check_mark:

duonoid:
I have addressed feedback to this point, as well as added some fixes/workarounds for a build issue regarding circleci + bundler-2.1.x (including 2.1.4) + appraisal (2.2.0).

adapters/sinatra/lib/opentelemetry/adapters/sinatra/adapter.rb
fbogsany: Discussed IRL: this is mostly documentation & makes testing more convenient.

mwear:
@duonoid Could you take a look at the failures on Ruby 2.7? It's same story as https://github.com/open-telemetry/opentelemetry-ruby/pull/148#issuecomment-572670786.

duonoid:
@mwear I have applied the same solution (remove appraisal from circleci, use `rake test`) as the one that fixed https://github.com/open-telemetry/opentelemetry-ruby/pull/148#issuecomment-572670786, as described here: https://github.com/open-telemetry/opentelemetry-ruby/pull/148#issuecomment-572692800

duonoid:
@mwear I've updated https://github.com/open-telemetry/opentelemetry-ruby/pull/148#issuecomment-572781259

mwear:
Would you mind resolving these conflicts and I'll get this merged!

duonoid:
conflicts: resolved :heavy_check_mark:

mwear:
Thanks @duonoid! 🚀

PR #152, opened 2019-11-13
Update versions for v0.2.0 release

mwear:
This PR updates the versions of the API, SDK, and Jaeger Exporter to v0.2.0  for our release. Do these look good to everybody?


elskwid:
I'm late to the party but I agree. Thanks for the update.

Issue #19, opened 2019-07-17
Instrumentation Packaging and Installation


mwear:
# Third Party Instrumentation Proposal for OpenTelemetry Ruby

## Overview

There is an [RFC for auto-instrumentation](https://github.com/open-telemetry/rfcs/blob/master/0002-telemetry-without-manual-instrumentation.md) and this document describes a possible approach that we could use for managing instrumentation for the OpenTelemetry Ruby Project. 

### Packaging

* Instrumentation should be packaged as gems
* Packages should list dependencies in their gemspec for
  * OpenTelemetry API Version
  * Framework version
* This will allow us to use Bundler to do the heavy lifting regarding versioning and compatibility of dependencies 

### Installation

* The tracer should be able to auto detect any instrumentation that is listed as a dependency in an application's Gemfile
* The time at which to install instrumentation in a Ruby application can be challenging to get right
  * To make this straightforward for the reference implementation, the SDK should have an API method that a user can call to install instrumentation at the right time for their application
  * Various SDK implementations can engage in whatever clever means they need to try to install instrumentation at the right time by calling this API method at the right time on behalf of the user

## Implementation

The implementation I am proposing is very similar to and inspired by [`Rails::Railtie`](https://api.rubyonrails.org/classes/Rails/Railtie.html).

OpenTelemetry Ruby will ship with a common base class, `OpenTelemetry::Instrumentation` that
* registers instrumentation with tracer (via the [`Class#inherited` callback](http://ruby-doc.org/core-2.6.3/Class.html#method-i-inherited))
* provides hooks for the instrumentation to be installed
* provides the ability to enable or disable instrumentation
  * by having an optional, explicit hook that can be defined for complex use cases
  * establishing a naming convention for environment variables users can specify to override whether or not instrumentation is enabled

### Example Instrumentation Package

A package should list its dependencies in its `gemspec`

```ruby
Gem::Specification.new do |spec|
  spec.add_dependency 'activerecord', '>= 5.0'
  spec.add_dependency 'opentelemetry-api', '~> 1.0'
end
```

[...]

PR #154, opened 2019-11-18
API: Add OpenTelemetry::Adapter interface

duonoid:
# Overview
Adds `OpenTelemetry::Adapter` API components that are used in #148, #149 (instrumentation adapters).

Provides a way for specific adapter implementations to interact with OpenTelemetry API configurations, tracers, formatters, etc.

## Details
* extracted from #148

## Related
* #67 
* #148 
* #149

### DIFF
--- a/api/lib/opentelemetry/adapter.rb
+++ b/api/lib/opentelemetry/adapter.rb
@@ -1,0 +1,23 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # The basic interface for Adapter objects
+  class Adapter
+    class << self
+      attr_reader :config
+
+      def install(config = {})
+        @config = config
+        new.install
+      end
+
+      def tracer
+        OpenTelemetry.tracer_factory.tracer(config[:name], config[:version])
+      end
+    end
+  end
+end
--- a/api/lib/opentelemetry/adapter.rb
+++ b/api/lib/opentelemetry/adapter.rb
@@ -15,6 +15,10 @@ module OpenTelemetry
         new.install
       end
 
+      def http_formatter
+        OpenTelemetry.tracer_factory.http_text_format
+      end
+
       def tracer
         OpenTelemetry.tracer_factory.tracer(config[:name], config[:version])
       end
--- a/api/lib/opentelemetry/adapter.rb
+++ b/api/lib/opentelemetry/adapter.rb
@@ -7,7 +7,20 @@
 require 'opentelemetry'
 
 module OpenTelemetry
-  # The basic interface for Adapter objects
+  # The basic interface for Adapter objects.
+  #
+  # Adapter class is intended to be subclassed, e.g.,
+  #
+  #   module Adapters
+  #     module Faraday
+  #       class Adapter < OpenTelemetry::Adapter
+  #       end
+  #     end
+  #   end
+  #
+  # then,
+  #
+  #   Adapters::Faraday.install(name: ..., version: ...)
   class Adapter
     class << self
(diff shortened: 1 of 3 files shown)


api/lib/opentelemetry/adapter.rb
```diff
@@ -0,0 +1,46 @@
...
+      def install(config = {})
+        @config = config
+        @http_formatter = config[:http_formatter] || default_http_formatter
+        @tracer = config[:tracer] || default_tracer
+        new.install
+      end
+
+      private
+
+      def default_http_formatter
```
mwear: I'm not sure how I feel about this being on the Adapter base class. While many adapters will need to propagate context, not all will. Ideally, formatters would be available on the tracer, and could be accessed when needed through the tracer reference. At one point, they were there, but named tracers caused them to move to the TracerFactory. 

In principle, I like the idea of instrumentation being bound to a tracer, and it having access to everything it needs through it. However, given the current architecture, it might make sense for instrumentation to have direct access to the TracerFactory and just lookup the tracer there. The same considerations apply to meters, when we get around to them.

If the current named tracers / named meters architecture holds, I don't think this Adapter, as its currently implemented is what we need. However, this could be useful if we change directions with named tracers / meters, which does not appear to be out of the question at this time.

mwear: It would be ideal if we could bind instrumentation to a tracer and meter at install time. If we could do that, I could see this adapter class being the right place to hold that information and the adapter class would have a reference to the config, the tracer, and meter. This would be ideal as it would keep named tracers and named meters out instrumentation.

The thing that I don't like is having to bind to the propagator here, and we should discuss if there are ways around that. One potential way out would be to have an accessor on the Tracer that just delegates to the TracerFactory.

Overall, I would prefer this approach to having to directly use the named tracers / meters in instrumentation. I think this will be a good discussion for tomorrow's SIG meeting. So, to revise my stance, this could be on the right track pending the outcome of our discussion.

mwear: Continuing to thing out loud, we are currently in the midst of a context propagation overhaul, which is going to lift context propagation to the top level (and off of the TracerFactory), so that actually solves one of the problems. When that work is done, we'll have a Propagation object that will have a reference to an HTTPInjector / Extractor, BinaryInjector / Extractor, and likely a RackInjector / Extractor.

So, I think this is probably going to work for us and its shape will probably look something like:

```ruby
  class Adapter
    class << self
      attr_reader :tracer, :meter, :propagation, :config

      def install(tracer:, meter:, propagation:, config:)
        ...
      end
    end
  end
```

(which is pretty close to what we have)...

Ultimately, I could see the `Adapter` class taking the role of the `Instrumentation` class that I describe in #19. It could override the `inherited` hook to populate an `AdapterRegistry` that could be used to auto-install instrumentation. I could see the `Adapter` class gain a `library_present?` method (for whether or not the underlying gem is present in the app) and an `enabled?` method, for whether or not the instrumentation should be installed.

I think I have a better picture of how these different pieces fit together. We should still discuss whether that picture is accurate or not. If it is, I think we have a reasonable path forward.

duonoid: I've made changes, and am hoping "`propagation_format" is ok, or did you prefer "`propagation`"? :heavy_check_mark:

fbogsany: > have an accessor on the Tracer that just delegates to the TracerFactory

I think this is generally useful. Most of the time we will have a Tracer handy, and it'd be convenient to access all the things through it instead of the TracerFactory.

duonoid:
Per recent discussions surrounding #148, #149: we decided to close this PR as "premature" for the time-being.  We expect a superclass interface to emerge after more adapter implementations are produced.

PR #157, opened 2019-11-20
API: Add HTTP to Status convenience mapper

duonoid:
# Overview
Adds a `Trace::Status.http_to_status(code)` method to facilitate mapping from http status code to `Status`.

## Background
* https://github.com/open-telemetry/opentelemetry-specification/issues/306
* [Status Canonical Code](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#statuscanonicalcode)
* [HTTP to Status Mapping Guidance](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#status)
* Extracted from #149 

## Related
* #149

### DIFF
--- a/api/lib/opentelemetry/trace/util/http_to_status.rb
+++ b/api/lib/opentelemetry/trace/util/http_to_status.rb
@@ -1,0 +1,50 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Trace
+    module Util
+      # Convenience methods, not necessarily required by the API specification.
+      module HttpToStatus
+        # Implemented according to
+        # https://github.com/open-telemetry/opentelemetry-specification/issues/306
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#status
+        #
+        # @param code Numeric HTTP status
+        # @return Status
+        def http_to_status(code) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
+          case code.to_i
+          when 100..299
+            new(const_get(:OK))
+          when 300..399
+            new(const_get(:OK))
+            # TODO: or DEADLINE_EXCEEDED in case of a loop
+          when 401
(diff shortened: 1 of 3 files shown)


api/lib/opentelemetry/trace/util/http_to_status.rb
```diff
@@ -0,0 +1,50 @@
...
+        #
+        # @param code Numeric HTTP status
+        # @return Status
+        def http_to_status(code) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
+          case code.to_i
+          when 100..299
+            new(const_get(:OK))
+          when 300..399
+            new(const_get(:OK))
+            # TODO: or DEADLINE_EXCEEDED in case of a loop
```
fbogsany: How would we figure that out from `code` alone? I'd leave that to the caller to figure out and simply collapse this and the previous case.

api/lib/opentelemetry/trace/util/http_to_status.rb
```diff
@@ -0,0 +1,50 @@
...
+          when 401
+            new(const_get(:UNAUTHENTICATED))
+          when 403
+            new(const_get(:PERMISSION_DENIED))
+          when 404
+            new(const_get(:NOT_FOUND))
+          when 429
+            new(const_get(:RESOURCE_EXHAUSTED))
+          when 400..499
```
fbogsany: Are Ruby cases evaluated sequentially? Just wondering about the behaviour for overlapping `when` conditions.

api/lib/opentelemetry/trace/util/http_to_status.rb
duonoid: :heavy_check_mark:

api/lib/opentelemetry/trace/util/http_to_status.rb
duonoid: My intention was to have more well-specified codes get handled early, then have "safety nets" catch the stranglers.  Once the first match is made, the case statement should exit.

fbogsany: Checked the Ruby docs - confirms that first successful `when` clause wins.

PR #164, opened 2020-01-07
Instrumentation Auto Installer

mwear:
This PR introduces an instrumentation discovery and auto-installer system. It's a revised version my thoughts from #19. It is comprised of an `OpenTelemetry::Instrumentation::Adapter` base class that all adapter are expected to subclass. Subclasses will be automatically registered with the `OpenTelemetry.instrumentation_registry`. The API has all of the machinery to discover and install instrumentation, but does not do any installing itself. That is left to the SDK.

## SDK config

I have a draft PR for SDK configuration and global init (#165) . It's relevant to this PR and it has three main goals.

* Make it easy to setup an SDK by providing sensible defaults and easy way to override them
* Establish a sequence of initialization, which is roughly
  * setup logging
  * setup propagation
  * setup the tracer_factory and meter_factory
  * install instrumentation (as the last step)
* Make it easy for users to configure and install instrumentation

There will be two ways to install instrumentation. 
  * Install adapters individually
  * Install all adapters that are present on the system by way of the discovery system
 
This is how each of those would look. **Note**: this is part of the SDK config draft PR, but an important piece of the instrumentation puzzle:

#### Install adapters individually

```ruby
  c.use 'OpenTelemetry::Adapters::Sinatra'
  c.use 'OpenTelemetry::Adapters::Faraday', {some_config: some_value} #config can optionally be passed
end
```

#### Install all adapters with default config

```ruby
  c.use_all
end
```

#### Install all adapters with config overrides

Any adapter configs will be passed onto the adapter during installation. Providing them is optional.

```ruby
  c.use_all {
    'OpenTelemetry::Adapters::Faraday' => {some_config: some_value}
    ....
  }
end
```

## OpenTelemetry::Instrumentation::Adapter

The Adapter class holds all metadata and configuration for an instrumentation adapter. All instrumentation adapter packages should include a subclass of `Instrumentation::Adapter` that will register it with `OpenTelemetry.instrumentation_registry` and make it available for discovery and installation by an SDK.

A typical subclass of Adapter will provide the an install block, a present block, and possibly a compatible block.

### Example Adapter Subclass

```ruby
module OpenTelemetry
  module Adapters
    module Sinatra
      class Adapter < OpenTelemetry::Instrumentation::Adapter
        install do |config|
          # install instrumentation, either by library hook or applying
          # a monkey patch
        end

        # determine if the target library is present
        present do
          Gem.loaded_specs.key?('sinatra')
        end

        # if the target library is present, is it compatible?
        compatible do
          Gem.loaded_specs['sinatra'].version > MIN_VERSION
        end
      end
    end
  end
end
```

### Additional Details
* The adapter name and version will inferred from the namespace of the class. In this example, they'd be 'OpenTelemetry::Adapters::Sinatra' and OpenTelemetry::Adapters::Sinatra::VERSION, but can be explicitly set using the `adapter_name` and `adapter_version` methods if necessary.
* The adapter class establishes a convention for disabling an adapter by environment variable and local configuration. An adapter disabled by environment variable will take precedence over local config. The convention for environment variable name is the library name, upcased with '::' replaced by underscores, and '_ENABLED' appended. For example: `OPENTELEMETRY_ADAPTERS_SINATRA_ENABLED = false`.
* Instrumentation libraries can use the adapter subclass to easily gain a reference to its named tracer. For example: `OpenTelemetry::Adapters::Sintatra.instance.tracer`

## Instrumentation::Registry

All adapter subclasses become automatically registered with the `OpenTelemetry::instrumentation_registry` via the inherited hook. At present, the registry contains the following functionality:

* `lookup(adapter_name)` - Return an adapter subclass instance by  name
* `install(adapter_names, adapter_config_map = {})` install specific adapters by name, with optional config per adapter
* `install_all(adapter_config_map = {})` - Install all instrumentation registered (that is present and compatible) with optional config per adapter.


api/lib/opentelemetry/instrumentation.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
```
fbogsany: Do we need to update copyright notices for 2020?

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The Adapter class holds all metadata and configuration for an
+    # instrumentation adapter. All instrumentation adapter packages should
+    # include a subclass of +Instrumentation::Adapter+ that will register
+    # it with +OpenTelemetry.instrumentation_regsitry+ and make it available for
```
fbogsany: s/instrumentation_regsitry/instrumentation_registry/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    # A typical subclass of Adapter will provide the adapter_name,
+    # adapter_version, an install block, a present block, and possibly
+    # a compatible block. Below is an example:
+    #
+    # module OpenTelemetry
+    #   module Adapters
+    #     module Sinatra
+    #       class Adapter < OpenTelemetry::Instrumentation::Adapter
+    #         library_name 'OpenTelemetry::Adapters::Sinatra'
+    #         library_version OpenTelemetry::Adapters::Sinatra::VERSION
```
fbogsany: Can we rely on conventions to avoid having to specify the `library_name` and `library_version`? I see you have defaults for both, but in this case they would be equivalent to:
```ruby
library_name 'OpenTelemetry::Adapters::Sinatra::Adapter'
library_version '0.0.0'
```

It feels like we could make more assumptions here, and 💥 in informative ways if the assumptions are invalid.

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    #     end
+    #   end
+    # end
+    #
+    # All subclasses of OpenTelemetry::Instrumentation::Adapter are automatically
+    # registered with OpenTelemetry.instrumentation_registry which is used by
+    # SDKs for for instrumentation discovery and installation.
+    #
+    # Instrumentation libraries can use the adapter subclass to easily gain
+    # a referene to its named tracer. For example:
```
fbogsany: s/referene/reference/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    #
+    # OpenTelemetry::Adapters::Sintatra.instance.tracer
+    #
+    # The adapter class establishes a convention for disabling an adapter
+    # by environment variable and local configuration. An adapter disabled
+    # by environment variable will take precedent over local config. The
```
fbogsany: s/precedent/precedence/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    # convention for enviroment variable name is the library name, upcased with
```
fbogsany: s/enviroment/environment/

api/lib/opentelemetry/instrumentation/adapter.rb
fbogsany: Also, I think you renamed these `adapter_name` and `adapter_version`

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    #         library_name 'OpenTelemetry::Adapters::Sinatra'
+    #         library_version OpenTelemetry::Adapters::Sinatra::VERSION
+    #
+    #         install do |config|
+    #           # install instrumentation, either by library hook or applying
+    #           # a monkey patch
+    #         end
+    #
+    #         # determine if the target library is present
+    #         present do |config|
```
fbogsany: The `config` param is not supported by the implementation.

api/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -0,0 +1,84 @@
...
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The instrumentation Registry contains information about instrumentation
+    # adapters available and facilitates discovery, installation and
+    # configuration. This functonality is primarily useful for SDK
+    # impelementors.
```
fbogsany: s/impelementors/implementors/

api/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -0,0 +1,84 @@
...
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The instrumentation Registry contains information about instrumentation
+    # adapters available and facilitates discovery, installation and
+    # configuration. This functonality is primarily useful for SDK
```
fbogsany: s/functonality/functionality/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+    # end
+    #
+    # All subclasses of OpenTelemetry::Instrumentation::Adapter are automatically
+    # registered with OpenTelemetry.instrumentation_registry which is used by
+    # SDKs for for instrumentation discovery and installation.
+    #
+    # Instrumentation libraries can use the adapter subclass to easily gain
+    # a reference to its named tracer. For example:
+    #
+    # OpenTelemetry::Adapters::Sintatra.instance.tracer
```
duonoid: s/Sintatra/Sinatra/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,219 @@
...
+        @name = name
+        @version = version
+        @install_blk = install_blk
+        @present_blk = present_blk
+        @compatible_blk = compatible_blk
+        @config = {}
+        @installed = false
+      end
+
+      # Install adapter with the given config. The present? and compatbile?
```
duonoid: s/compatbile/compatible/

api/lib/opentelemetry/instrumentation/adapter.rb
mwear: I added the ability to infer name and version. The name is inferred as the namespace if the class is namespaced, otherwise it is the class name. On the off chance the class name is nil (this would happen for an anonymous class) it will be 'unknown'.

Version is similar. It defaults to the `VERSION` constant under the namespace if it exists, the class if it doesn't, and '0.0.0` if neither exist.

I'm trying to avoid :boom:ing if I can since the defaults are lazily evaluated when the instance is created.

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+                             end
+        end
+
+        def infer_version
+          return unless (inferred_name = infer_name)
+
+          mod = inferred_name.split('::').map(&:to_sym).inject(Object) do |object, const|
+            object.const_get(const)
+          end
+          mod.const_get(:VERSION)
```
fbogsany: See? That inferencing is not complicated at all! 🙈 :trollface:

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+            @adapter_name = adapter_name
+          else
+            @adapter_name ||= infer_name || 'unknown'
+          end
+        end
+
+        # Optionally set the version of this adapter. If not explicitly set,
+        # the version will default to the version constant under namespace of
+        # the class, or the version constant under the class name if it does not
+        # have a namespace. If a version constant cannot be found, it defaults
```
fbogsany: More accurate to refer to `VERSION constant` instead of `version constant` in this comment?

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+
+        @config = config unless config.nil?
+        instance_exec(@config, &@install_blk)
+        @tracer ||= OpenTelemetry.tracer_factory.tracer(name, version)
+        @installed = true
+      end
+
+      # Whether or not this adapter is installable in the current process. Will
+      # be true when the adapter defines an install block, it's not disabled
+      # by enviroment or config, and the target library present and compatible.
```
fbogsany: s/enviroment/environment/

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+    #         # if the target library is present, is it compatible?
+    #         compatible do
+    #           Gem.loaded_specs['sinatra'].version > MIN_VERSION
+    #         end
+    #       end
+    #     end
+    #   end
+    # end
+    #
+    # The adapter name and version will inferred from the namespace of the
```
elskwid:
```suggestion
    # The adapter name and version will be inferred from the namespace of the
```

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+    # class. In this example, they'd be 'OpenTelemetry::Adapters::Sinatra' and
+    # OpenTelemetry::Adapters::Sinatra::VERSION, but can be explicitly set using
+    # the +adapter_name+ and +adapter_version+ methods if necessary.
+    #
+    # All subclasses of OpenTelemetry::Instrumentation::Adapter are automatically
+    # registered with OpenTelemetry.instrumentation_registry which is used by
+    # SDKs for for instrumentation discovery and installation.
```
elskwid:
```suggestion
    # SDKs for instrumentation discovery and installation.
```

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+        def adapter_name(adapter_name = nil)
+          if adapter_name
+            @adapter_name = adapter_name
+          else
+            @adapter_name ||= infer_name || 'unknown'
+          end
+        end
+
+        # Optionally set the version of this adapter. If not explicitly set,
+        # the version will default to the version constant under namespace of
```
elskwid:
```suggestion
        # the version will default to the version constant under the namespace of
```

api/lib/opentelemetry/instrumentation/adapter.rb
elskwid: Agreed, I think `VERSION constant` expresses what you're doing.

api/lib/opentelemetry/instrumentation/adapter.rb
elskwid: 😅

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+        @name = name
+        @version = version
+        @install_blk = install_blk
+        @present_blk = present_blk
+        @compatible_blk = compatible_blk
+        @config = {}
+        @installed = false
+      end
+
+      # Install adapter with the given config. The present? and compatible?
```
elskwid:
```suggestion
      # Install adapter with the given config. The present? and compatible? methods
```

api/lib/opentelemetry/instrumentation/adapter.rb
```diff
@@ -0,0 +1,244 @@
...
+        return false unless installable?(config)
+
+        @config = config unless config.nil?
+        instance_exec(@config, &@install_blk)
+        @tracer ||= OpenTelemetry.tracer_factory.tracer(name, version)
+        @installed = true
+      end
+
+      # Whether or not this adapter is installable in the current process. Will
+      # be true when the adapter defines an install block, it's not disabled
```
elskwid:
```suggestion
      # be true when the adapter defines an install block, is not disabled
```

api/lib/opentelemetry/instrumentation/adapter.rb
elskwid:
```suggestion
      # by environment or config, and the target library present and compatible.
```

api/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -0,0 +1,87 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The instrumentation Registry contains information about instrumentation
+    # adapters available and facilitates discovery, installation and
+    # configuration. This functionality is primarily useful for SDK
+    # implementors.
+    class Registry
```
elskwid: Not a blocker at all but would `#install` and `#install_all` fit better at a higher level? It feels like the `Registry` might be the wrong place to have the responsibility for installing adapters rather than just being the registry.

From the API perspective:

```ruby
OpenTelemetry.instrumentation_registry.install(['some_adapter'])
OpenTelemetry.instrumentation_registry.install_all
```

versus

```ruby
OpenTelemetry.install_adapters(['some_adapter'])
OpenTelemetry.install_all_adapters
```

Ideally, I would like something like below but this may be complicated with the instance of the registry and whatnot.

```ruby
OpenTelemetry.instrumentation.install(['some_adapter'])
OpenTelemetry.instrumentation.install
 ```

File: api/lib/opentelemetry/instrumentation/registry.rb [hunk 21]
```diff
@@ -0,0 +1,87 @@
...
+      # Install the specified adapters with optionally specified configuration.
+      #
+      # @param [Array<String>] adapter_names An array of adapter names to
+      #   install
+      # @param [optional Hash<String, Hash>] adapter_config_map A map of
+      #   adapter_name to config. This argument is optional and config can be
+      #   passed for as many or as few adapters as desired.
+      def install(adapter_names, adapter_config_map = {})
+        @lock.synchronize do
+          adapter_names.each do |adapter_name|
```
elskwid: You could wrap `adapter_names` in `Array()` to allow this method to accept a single name or an array.

```ruby
def install(adapter_names, adapter_config_map = {})
  @lock.synchronize do
    Array(adapter_names).each do |adapter_name|

  # ...
end
```

api/test/opentelemetry/instrumentation/adapter_test.rb
```diff
@@ -0,0 +1,274 @@
...
+      if const == names[-1]
+        object.const_set(:VERSION, version) if version
+        object.const_set(const, Class.new(OpenTelemetry::Instrumentation::Adapter))
+      else
+        object.const_set(const, Module.new)
+      end
+    end
+  end
+
+  def with_env(new_env)
```
elskwid: 😍

api/lib/opentelemetry/instrumentation/registry.rb
mwear: These are ultimately not user facing. They will get called by the SDK, so I went with what would be most useful for the SDK. With my draft pr for SDK configuration users will use this as:

```ruby
  c.use 'OpenTelemetry::Adapters::Sinatra'
  c.use 'OpenTelemetry::Adapters::Faraday', {some_config: some_value} #config can optionally be passed
end
```

or 

```ruby
  c.use_all
optionally be passed
end
```

Given that, I think it makes sense to have them as written. Let me know if you feel strongly about this.

api/lib/opentelemetry/instrumentation/registry.rb
mwear: Similar to my comment above. This is an API method that will be used by the SDK and the SDK will always pass in an array. It's a simple change to make if you think there's a use for it though.

api/lib/opentelemetry/instrumentation/registry.rb
elskwid: @mwear Thank you for the explanation. If this is internal then I think we're fine.

api/lib/opentelemetry/instrumentation/registry.rb
elskwid: @mwear Gotcha. I'm good with this as is - especially given that it's internal. Thank you.

mwear:
Thanks for the reviews everyone. I'm going to go ahead and merge this!

PR #166, opened 2020-01-14
Add Rack instrumentation

duonoid:
# Overview
Adds `rack` instrumentation adapter implementation, ported from existing `dd-trace-rb` functionality.

`rack` instrumentation is required by `rails` instrumentation, which does not exist yet.

## Details
* adds a working example, runnable via docker-compose, e.g.,
```
docker-compose run --rm ex-adapter-rack
bash-5.0$ ruby trace_demonstration.rb
```
* adds functionality to trace 'front-end' request (queue time) and retain middleware names
* `QueueTime`implementation is directly from dd-trace-rb
* config options include`:retain_middleware_names`, `:allowed_response_headers`, `:allowed_request_headers`

## Related
* #67 - port datadog instrumentation
* #148 - faraday instrumentation
* #149 - sinatra instrumentation


adapters/rack/Gemfile
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
```
fbogsany: Needs 2020 copyright year.

adapters/rack/lib/opentelemetry/adapters/rack/util/queue_time.rb
```diff
@@ -0,0 +1,47 @@
...
+      module Util
+        # QueueTime simply...
+        module QueueTime
+          REQUEST_START = 'HTTP_X_REQUEST_START'.freeze
+          QUEUE_START = 'HTTP_X_QUEUE_START'.freeze
+          MINIMUM_ACCEPTABLE_TIME_VALUE = 1_000_000_000
+
+          module_function
+
+          def get_request_start(env, now = Time.now.utc)
```
fbogsany: Given the 3 early returns and that `now` is not used until the end of the method, it might be better to default to `now = nil` and defer the `Time.now.utc` until `now` is needed.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+        # TODO: implement 'resource name'
+        class TracerMiddleware
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env)
+            # since middleware may only be initialized once (memoized),
+            # but we want to be sure to have a clean slate on each request:
+            dup._call env
```
mwear: Can we avoid this if you don't rely on global state, in this case, instance variables? I don't see a `dup` in the original DD instrumentation.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+              rack_request.url.chomp(rack_request.fullpath)
+            end
+          end
+
+          def rack_request
+            @rack_request ||= ::Rack::Request.new(env)
+          end
+
+          def record_and_reraise_error(error)
+            # TODO: implement span.set_error?
```
mwear: Currently there are not semantic conventions around errors. I think this is something that needs to be addressed at the specification level. I'm not exactly sure what we should do in the interim. My feeling is that these should probably be represented as events on the span.

adapters/rack/Gemfile
duonoid: All updated, thanks!  :heavy_check_mark:

adapters/rack/lib/opentelemetry/adapters/rack/util/queue_time.rb
duonoid: I've updated as suggested, Thanks!  :heavy_check_mark:

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+            Array(config[:allowed_request_headers])
+          end
+
+          # @return Hash
+          def allowed_response_headers(headers)
+            return {} if headers.nil?
+
+            {}.tap do |result|
+              Array(allowed_response_header_names).each do |header|
+                new_key_name = build_attribute_name('http.response.headers.', header)
```
mwear: Similarly, can we do any caching here?

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+            span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
+
+            allowed_response_headers(headers).each { |k, v| span.set_attribute(k, v) }
+          end
+
+          # @return Hash
+          def allowed_request_headers
+            {}.tap do |result|
+              Array(allowed_request_header_names).each do |header|
+                rack_header = "HTTP_#{header.to_s.upcase.gsub(/[-\s]/, '_')}"
```
mwear: Since the allowed headers are static, can we find some way to cache the header name and the attribute name so we don't have to create a new strings every time?

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+            end
+          end
+
+          def allowed_request_header_names
+            Array(config[:allowed_request_headers])
+          end
+
+          # @return Hash
+          def allowed_response_headers(headers)
+            return {} if headers.nil?
```
mwear: Is this return value going to be mutated? If not, can we make this hash a frozen constant?

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+                rack_header = "HTTP_#{header.to_s.upcase.gsub(/[-\s]/, '_')}"
+                if env.key?(rack_header)
+                  result[build_attribute_name('http.request.headers.', header)] = env[rack_header]
+                end
+              end
+            end
+          end
+
+          def allowed_request_header_names
+            Array(config[:allowed_request_headers])
```
mwear: Can we memoize this so we don't have to (possibly) create an array on each call?

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+                  if matching_header
+                    result[new_key_name] = headers[matching_header]
+                  end
+                end
+              end
+            end
+          end
+
+          def allowed_response_header_names
+            Array(config[:allowed_response_headers])
```
mwear: Could this be memoized? Another option might be to have a more sophisticated config (rather than just a hash). We could consider trying to abstract and borrow prior art from dd-trace here.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,248 @@
...
+              'http.method' => env['REQUEST_METHOD'],
+              'http.url' => full_http_request_url,
+              'http.host' => env['HOST'],
+              'http.scheme' => env['rack.url_scheme'],
+              'http.base_url' => base_url # TODO: check attribute naming/semantics
+            }.merge(allowed_request_headers)
+          end
+
+          def full_http_request_url
+            # NOTE: dd-trace-rb has implemented 'quantization'? (which lowers url cardinality)
```
duonoid: https://github.com/open-telemetry/opentelemetry-specification/pull/416

duonoid:
I've resolved the outstanding TODOs as well as made the performance-oriented modifications that @mwear suggested (Thanks!).  Probably ready for review at this point.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've refactored to avoid using instance variables.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I set `span.status`, kept a TODO note, kept the rescue clause intact (per SIG discussion).

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I added some caching :heavy_check_mark:

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: :heavy_check_mark:

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've added some caching :heavy_check_mark:

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've adding caching of the value in the class's singleton for the time-being.

adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb
```diff
@@ -0,0 +1,58 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Rack
+      # The Adapter class contains logic to detect and install the Rack
+      # instrumentation adapter
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
```
mwear: Mental note, this is the 3rd time I've seen this. I'm starting to think this should be an optional part of the DSL.

adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          retain_middleware_names if config[:retain_middleware_names]
+          configure_default_quantization
+
+          if (app = config[:application])
+            app.use Middlewares::TracerMiddleware
+          else
+            # monkey patch Rack::Builder:
+            require_relative 'patches/rack_builder'
+            ::Rack::Builder.prepend Rack::Patches::RackBuilder
```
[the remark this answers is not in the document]
duonoid: Sorry about that confusion... I have reverted the related commit to avoid auto-patching `Rack::Builder`.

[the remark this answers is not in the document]
mwear: Not a problem. Thanks for understanding!

Rakefile
```diff
@@ -63,6 +63,12 @@ GEM_INFO = {
       OpenTelemetry::Adapters::Faraday::VERSION
     }
   },
+  "opentelemetry-adapters-rack" => {
+    version_getter: ->() {
+      require './lib/opentelemetry/adapters/rack/version.rb'
+      OpenTelemetry::Adapters::Rack::VERSION
+    }
+  }
```
mwear:
```suggestion
  },
```

adapters/rack/test/opentelemetry/adapters/rack/adapter_test.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+require_relative '../../../../lib/opentelemetry/adapters/rack/adapter'
```
mwear: This is causing a circular require and a warning:

```
/Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17: warning: /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17: warning: loading in progress, circular require considered harmful - /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:5:in  `<main>'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:5:in  `select'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:17:in  `block in <main>'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:17:in  `require'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/test/opentelemetry/adapters/rack/adapter_test.rb:9:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/test/opentelemetry/adapters/rack/adapter_test.rb:9:in  `require_relative'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65:in  `require_relative'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17:in  `require_relative'
```

adapters/rack/test/opentelemetry/adapters/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,200 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+# require Adapter so .install method is found:
+require_relative '../../../../../lib/opentelemetry/adapters/rack'
```
mwear: I also see a circular require here:

```
/Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65: warning: /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65: warning: loading in progress, circular require considered harmful - /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:5:in  `<main>'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:5:in  `select'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:17:in  `block in <main>'
	from /Users/matthewwear/.gem/ruby/2.6.5/gems/rake-12.3.3/lib/rake/rake_test_loader.rb:17:in  `require'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/test/opentelemetry/adapters/rack/middlewares/tracer_middleware_test.rb:10:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/test/opentelemetry/adapters/rack/middlewares/tracer_middleware_test.rb:10:in  `require_relative'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack.rb:17:in  `require_relative'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65:in  `<top (required)>'
	from /Users/matthewwear/Repos/opentelemetry-ruby/adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb:65:in  `require_relative'
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,267 @@
...
+          end
+
+          # @return Hash
+          def allowed_request_headers(env)
+            {}.tap do |result|
+              self.class.allowed_rack_request_headers.each do |hash|
+                result[self.class.build_attribute_name('http.request.headers.', hash[:header])] = env[hash[:rack_header]] if env.key?(hash[:rack_header])
+              end
+            end
+          end
```
mwear: I think this optimization has gone awry. Can't we eagerly convert the allowed request headers into the rack normalized versions to use a keys in a map index the attribute names E.g `{'HTTP_FOO' => 'http.request.headers.foo', 'HTTP_BAR' => 'http.request.headers.bar'}`? You can then just iterate through the hash to looking for the key and you'll already have a pre-computed attribute name for it if it's found?

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,267 @@
...
+
+                next unless key
+
+                # cache string for next time:
+                hash[:search_for_keys] |= [key]
+
+                result[hash[:attribute_name]] = headers[key]
+              end
+            end
+          end
```
mwear: This optimization seems similarly problematic. Can't we just store an array of allowed_response headers and then a map of header name to attribute name. E.g. ['foo', 'bar']. After a match on 'foo' and a case insensitive match on 'BAR' the map would look like {'foo' => 'http.response.headers.foo', 'BAR' => 'http.response.headers.bar'}.

adapters/rack/lib/opentelemetry/adapters/rack/adapter.rb
```diff
@@ -0,0 +1,65 @@
...
+      # The Adapter class contains logic to detect and install the Rack
+      # instrumentation adapter
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
+
+          retain_middleware_names if config[:retain_middleware_names]
+          configure_default_quantization
+
+          config[:application]&.use Middlewares::TracerMiddleware
```
mwear: I think we should leave the `.use` call to the user. I don't see this in dd-trace and I think it could cause issues if the user calls `use` for this middleware and we also call it behind the scenes.

adapters/rack/test/test_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+
+require 'rack'
+
+require 'opentelemetry/sdk'
+
+require 'minitest/autorun'
+require 'webmock/minitest'
+
+# global opentelemetry-sdk setup:
+sdk = OpenTelemetry::SDK
```
mwear: We can use the `SDK.configure` and streamline this setup.

duonoid:
@mwear Thank you for that last round of feedback!  I believe I have addressed and resolved the points you mentioned.

adapters/rack/example/trace_demonstration2.rb
```diff
@@ -0,0 +1,25 @@
...
+require 'bundler/setup'
+
+Bundler.require
+
+# setup fake rack application:
+builder = Rack::Builder.new
+app = ->(_env) { [200, { 'Content-Type' => 'text/plain' }, ['All responses are OK']] }
+builder.run app
+
+# demonstrate integration using config[:application] instead of 'builder.use':
```
mwear: I'm not sure that this demonstration is applicable any longer. We require an explicit `use` like the original Datadog instrumentation does.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+          # @return Hash
+          def allowed_response_headers(headers)
+            return EMPTY_HASH if headers.nil?
+
+            {}.tap do |result|
+              self.class.allowed_response_headers.each do |key, value|
+                if headers.key?(key)
+                  result[value] = headers[key]
+                else
+                  case_insensitive_entry = headers.detect { |h| h.first.upcase == key }
```
mwear: Detect on a hash is going to allocate an array per entry. This could be improved by using `headers.each` and exiting the loop early on a match.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+        # by way of its middleware system
+        #
+        # Notable implementation differences from dd-trace-rb:
+        # * missing: 'span.resource', which is set to span.name
+        # * missing: config[:distributed_tracing]
+        # * missing: span.set_error() -- spec level change
+        class TracerMiddleware # rubocop:disable Metrics/ClassLength
+          class << self
+            def allowed_rack_request_headers
+              @allowed_rack_request_headers ||= {}.tap do |result|
```
mwear: This might be cleaner if we: lose the intermediary `allowed_request_header_names` method and use `each_with_object` instead of `{}.tap`. E.g.

```

@allowed_rack_request_headers ||= \
  Array(config[:allowed_request_headers]).each_with_object({}) do |header, memo|
    ...
  end
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+                end
+              end
+            end
+
+            def allowed_request_header_names
+              @allowed_request_header_names ||= Array(config[:allowed_request_headers])
+            end
+
+            def allowed_response_headers
+              @allowed_response_headers ||= {}.tap do |result|
```
mwear: Similarly, I think we could lose the intermediary `allowed_response_header_names` method and use `each_with_object` instead of `{}.tap`. E.g.

```

@allowed_response_headers ||= \
  Array(config[:allowed_response_headers]).each_with_object({}) do |header, memo|
    ...
  end
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+            {}.tap do |result|
+              self.class.allowed_rack_request_headers.each do |key, value|
+                result[value] = env[key] if env.key?(key)
+              end
+            end
+          end
+
+          # @return Hash
+          def allowed_response_headers(headers)
+            return EMPTY_HASH if headers.nil?
```
mwear: We could probably return early if allowed_response_headers are empty also.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+            # set_attribute('http.route', ...
+            # e.g., "/users/:userID?
+            span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
+
+            allowed_response_headers(headers).each { |k, v| span.set_attribute(k, v) }
+          end
+
+          # @return Hash
+          def allowed_request_headers(env)
+            {}.tap do |result|
```
mwear: We could return `EMPTY_HASH` if allowed_rack_request_headers is empty.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+          EMPTY_HASH = {}.freeze
+
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+            original_env = env.dup
+
+            parent_context = OpenTelemetry.tracer_factory.http_text_format.extract(env)
```
mwear: Now that context prop has merged, this method is going to look something like this:

```ruby
def call(env)
  extracted_context = OpenTelemetry.propagation.extract(env)

  # restore extracted context in this process

    frontend_context = create_frontend_span(env) if queue_time_present?

    request_span = tracer.start_span('request span', parent_context: frontend_context || extracted_context)

    tracer.with_span(request_span) do |span|
      begin
        @app.call(env)
      rescue StandardError => e
        span.set_error(e)
        raise e
      ensure
        add_attributes(request_span, ...)
        request_span.finish
        finish_frontend_span(frontend_context) if frontend_context
      end
    end
  end
end

# return 
def create_fontend_span(env)
  span = Tracer.start_span(...)
  extracted_context.with_value(current_span_key, span)
end

def finish_frontend_span(frontend_context)
  span = frontend_context[current_span_key]
  span.finish
end

def current_span_key
  OpenTelemetry::Trace::Propagation.current_span_key,
end
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,262 @@
...
+            # NOTE: get_request_start may return nil
+            request_start_time = OpenTelemetry::Adapters::Rack::Util::QueueTime.get_request_start(env)
+            frontend_span = tracer.start_span('http_server.queue', # NOTE: span kind of 'proxy' is not defined
+                                              with_parent_context: parent_context,
+                                              # NOTE: initialize with as many attributes as possible:
+                                              attributes: {
+                                                'component' => 'http',
+                                                'service' => config[:web_service_name],
+                                                # NOTE: dd-trace-rb differences:
+                                                # 'trace.span_type' => (http) 'proxy'
```
mwear: I think `kind: :server` will work for both the frontend and request span.

adapters/rack/test/opentelemetry/adapters/rack/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,186 @@
...
+
+          it 'records "service" attribute' do
+            _(frontend_span.attributes['service']).must_equal 'my_service_name'
+          end
+        end
+      end
+    end
+  end
+
+  describe 'config[:quantization]' do
```
mwear: Do we still have quantization?

adapters/rack/example/trace_demonstration2.rb
duonoid: I have updated the example to integrate explicitly (`use`).

Note that the `config[:application]` behavior should be similar to the `MiddlewareNamePatcher` https://github.com/DataDog/dd-trace-rb/blob/9301de22e173dcbd6a4870b6ed1195e71018086c/lib/ddtrace/contrib/rack/patcher.rb#L90

adapters/rack/test/opentelemetry/adapters/rack/middlewares/tracer_middleware_test.rb
duonoid: We don't have a full Quantization implementation (reverted in 78acca7cd6d3274f7c0a352107c8413ad528e62f and a69f94ee834ccd6ca977b2e0abbed99af53d6bf0), but the callable `config[:quantization]` still exists, which is passed the `request_uri_or_path_info`.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I tried to get as close to these suggestions as possible, but there were a couple of hurdles.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,257 @@
...
+              #
+              # REQUEST_URI is only available depending on what web server is running though.
+              # So when its not available, we want the original, unmutated PATH_INFO, which
+              # is just the relative path without query strings.
+              request_span_name = create_request_span_name(env['REQUEST_URI'] || original_env['PATH_INFO'])
+
+              rack_request = ::Rack::Request.new(env)
+              # Sets as many attributes as are available before control
+              # is handed off to next middleware.
+              request_span = tracer.start_span(request_span_name,
```
mwear: This makes the frontend_span and request_span siblings, but in dd-trace they are a parent and child. One thing to keep in mind is that `Tracer#trace` and `Tracer#start_span` automatically make the span the active span. For OpenTelemetry `Tracer#start_span` assumes you are explicitly managing context. `Tracer#in_span` and `Tracer#with_span` will activate a span.

Please see my comment again: https://github.com/open-telemetry/opentelemetry-ruby/pull/166#discussion_r376588828

The key points are:
```ruby
#create frontend_span returns a context with the frontend span as the current span
frontend_context = create_frontend_span(env)

#the request span is explicitly parented using the frontend_context 
request_span = tracer.start_span('request span', parent_context: frontend_context || extracted_context)
```

You don't have to use that exact code, but you do need to manage context the same way. Also, it'd be nice to see a test that the frontend span parents the request span in this situation.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,257 @@
...
+              #
+              # Because of this, we prefer to use REQUEST_URI, if available, which is the
+              # relative path + query string, and doesn't mutate.
+              #
+              # REQUEST_URI is only available depending on what web server is running though.
+              # So when its not available, we want the original, unmutated PATH_INFO, which
+              # is just the relative path without query strings.
+              request_span_name = create_request_span_name(env['REQUEST_URI'] || original_env['PATH_INFO'])
+
+              rack_request = ::Rack::Request.new(env)
```
mwear: We can avoid allocating the Rack::Request as it's just a wrapper around the `env`. It looks like we have wrapper methods on top of that and some backwards compatibility special handling to top it off. If it's not too much trouble to work on the env itself, we could yak shave an allocation.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,257 @@
...
+            end
+          end
+
+          private
+
+          # return Context
+          def create_frontend_span(env)
+            # NOTE: get_request_start may return nil
+            request_start_time = OpenTelemetry::Adapters::Rack::Util::QueueTime.get_request_start(env)
+            record_frontend_span = config[:record_frontend_span] && !request_start_time.nil?
```
mwear: We can probably just return early instead of assigning a local here.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,257 @@
...
+          # * http.url
+          def request_span_attributes(env:, full_http_request_url:, full_path:, base_url:)
+            {
+              'component' => 'http',
+              'http.method' => env['REQUEST_METHOD'],
+              'http.url' => full_http_request_url,
+              'http.host' => env['HOST'],
+              'http.scheme' => env['rack.url_scheme'],
+              'http.target' => full_path,
+              'http.base_url' => base_url # NOTE: 'http.base_url' isn't officially defined
```
mwear: I think we can just go with `http.scheme`, `http.host` and `http.target` as per the semantic conventions.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
mwear: If this had anything to do with one of the hurdles. let me know and we can look at it.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
mwear: I should be completely clear that, if we are using some of the convenience methods and it would be hard to live without them, we can leave this as is. Only yak shave if it's easy.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
mwear: After further thought I suspect this was one of the hurdles. We have a couple of options. The first would be ideal, but would require a change to `in_span` to make it work.

```ruby
def call(env)
  extracted_context = OpenTelemetry.propagation.extract(env)
  
  frontend_context = create_frontend_span(env, extracted_context)
  
  tracer.in_span(..., with_parent_context: frontend_context || extracted_context) do |request_span|
    app.call
  end
  
ensure
  finish_frontend_span(frontend_context)
end

def create_frontend_span(env, extracted_context)
  span = tracer.start_span(..., with_parent_context: extracted_context)
  extracted_context.set_value(current_span_key, span)
end

def finish_frontend_span(context)
  context[current_span_key]&.finish
end
```

For this to work properly, `in_span` would have to have the record and reraise behavior built in. This is relevant to this issue: https://github.com/open-telemetry/opentelemetry-ruby/issues/181.

Without `in_span` having the record and reraise behavior we to drop down to explicitly managed context. Seeing this written down makes me think we might need a few helpers to make explicit management easier:

```ruby
def call(env)
  extracted_context = OpenTelemetry.propagation.extract(env)
  
  frontend_context = create_frontend_span(env, extracted_context)

  request_context = create_request_span(frontend_context || extracted_context)

  OpenTelemetry::Context.with_current(request_context) do
    begin
      @app.call
    rescue => e
      #record and reraise
    ensure
      finish_span(request_context)
      finish_span(frontend_context)
    end
  end
end

def create_frontend_span(env, extracted_context)
  span = tracer.start_span(..., with_parent_context: extracted_context)
  extracted_context.set_value(current_span_key, span)
end

def create_request_span(context)
  span = tracer.start_span(..., with_parent_context: context)
  context.set_value(current_span_key, span)
end

def finish_span(context)
  context[current_span_key]&.finish
end
```

duonoid: @mwear ok, besides the error reporting, I think I got it shape as described now.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: After removing the extra attributes (`request_span_attributes`), I was able to inline the leftover `Rack::Request#fullpath` functionality without too much difficulty.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've removed the extra `http.url` and `http.base_url` attributes.

duonoid:
Hey @mwear , thanks for that last round of suggestions!  I think I've resolved the points so far.

Notable outstanding is the error handling (i.e., "record and re-raise").

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+            request_span = request_context[current_span_key]
+
+            # restore extracted context in this process:
+            OpenTelemetry::Context.with_current(request_context) do
+              begin
+                @app.call(env).tap do |status, headers, response|
+                  set_attributes_after_request(request_span, status, headers, response)
+                end
+              rescue StandardError => e
+                record_and_reraise_error(e, request_span: request_span)
```
mwear: Now that `in_span` records and reraises errors, you could consider rewriting call as follows:

```ruby
def call(env)
            original_env = env.dup
            extracted_context = OpenTelemetry.propagation.extract(env)
            frontend_context = create_frontend_span(env, extracted_context)

            OpenTelemetry::Context.with_current(frontend_context || extracted_context) do
              request_span_name = create_request_span_name(env['REQUEST_URI'] || original_env['PATH_INFO'])
              tracer.in_span(request_span_name,
                             attributes: request_span_attributes(env: env),
                             kind: :server) do |request_span|
                @app.call(env).tap do |status, headers, response|
                  set_attributes_after_request(request_span, status, headers, response)
                end
              end
            end
          ensure
            finish_span(frontend_context)
          end
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+              ensure
+                finish_span(frontend_context)
+                finish_span(request_context)
+              end
+            end
+          end
+
+          private
+
+          def create_request_span(env, context)
```
mwear: If you use the alternative call above, this method won't exactly be necessary.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+            # REQUEST_URI is only available depending on what web server is running though.
+            # So when its not available, we want the original, unmutated PATH_INFO, which
+            # is just the relative path without query strings.
+            request_span_name = create_request_span_name(env['REQUEST_URI'] || original_env['PATH_INFO'])
+
+            # Sets as many attributes as are available before control
+            # is handed off to next middleware.
+            span = tracer.start_span(request_span_name,
+                                     with_parent_context: context,
+                                     # NOTE: try to set as many attributes via 'attributes' argument
```
mwear: If we're keeping this method, we can probably loose the NOTE.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+                                     #       instead of via span.set_attribute
+                                     attributes: request_span_attributes(env: env),
+                                     kind: :server)
+
+            context.set_value(current_span_key, span)
+          end
+
+          # return Context with the frontend span as the current span
+          def create_frontend_span(env, extracted_context)
+            # NOTE: get_request_start may return nil
```
mwear: This is probably a useful comment, but we might be able to loose the following notes.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+            OpenTelemetry::Trace::Propagation::ContextKeys.current_span_key
+          end
+
+          def tracer
+            OpenTelemetry::Adapters::Rack::Adapter.instance.tracer
+          end
+
+          ### request_span
+
+          # Per https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#http-server-semantic-conventions
```
mwear: This was super useful during development and review, although, I don't that this comment should hang around in the final version.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+            else
+              request_uri_or_path_info
+            end
+          end
+
+          # catch exceptions that may be raised in the middleware chain
+          # Note: if a middleware catches an Exception without re raising,
+          # the Exception cannot be recorded here.
+          def record_and_reraise_error(error, request_span:)
+            request_span.record_error(error)
```
mwear: `Span#record_error` sets the status for us now. If you use the alternate call (that uses `in_span`) we won't need this method. If we keep it, it doesn't need to set the status.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+require_relative '../util/queue_time'
+
+module OpenTelemetry
+  module Adapters
+    module Rack
+      module Middlewares
+        # TracerMiddleware propagates context and instruments Rack requests
+        # by way of its middleware system
+        #
+        # Notable implementation differences from dd-trace-rb:
```
mwear: This is a useful comment for review purposes, but should probably not be in the final version.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,253 @@
...
+            return EMPTY_HASH if self.class.allowed_rack_request_headers.empty?
+
+            {}.tap do |result|
+              self.class.allowed_rack_request_headers.each do |key, value|
+                result[value] = env[key] if env.key?(key)
+              end
+            end
+          end
+
+          # @return Hash
```
mwear: I'm not sure we need to annotate this return. The rest of the methods do not have `@return` anyhow.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've rewritten, as suggested :+1:

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,193 @@
...
+            OpenTelemetry::Adapters::Rack::Adapter.instance.tracer
+          end
+
+          ### request_span
+
+          def request_span_attributes(env:)
+            {
+              'component' => 'http',
+              'http.method' => env['REQUEST_METHOD'],
+              'http.host' => env['HOST'],
```
mwear: This should be:

```suggestion
              'http.host' => env['HTTP_HOST'] || 'unknown',
```

`HOST` is normalized to `HTTP_HOST` by Rack. The 'unknown' is debatable, but I have that this can be, and regularly is nil, which is not a valid attribute value. The example app will expose this:

```
I, [2020-02-24T17:59:31.953568 #89019]  INFO -- : Adapter: OpenTelemetry::Adapters::Rack was successfully installed
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="/",
 kind=:server,
 status=
  #<OpenTelemetry::Trace::Status:0x00007f9562141b40
   @canonical_code=0,
   @description="">,
 parent_span_id="0000000000000000",
 child_count=0,
 total_recorded_attributes=7,
 total_recorded_events=0,
 total_recorded_links=0,
 start_timestamp=2020-02-24 17:59:31 -0800,
 end_timestamp=2020-02-24 17:59:31 -0800,
 attributes=
  {"component"=>"http",
   "http.method"=>"GET",
   "http.host"=>nil,
   "http.scheme"=>"http",
   "http.target"=>"/",
   "http.status_code"=>200,
   "http.status_text"=>"OK"},
 links=nil,
 events=nil,
 library_resource=
  #<OpenTelemetry::SDK::Resources::Resource:0x00007f95648c3a10
   @labels={"name"=>"OpenTelemetry::Adapters::Rack", "version"=>"0.0.0"}>,
 span_id="7b9ccc4a17b7355c",
 trace_id="9e6ad9993469c5443ef979de4a565bba",
 trace_flags=#<OpenTelemetry::Trace::TraceFlags:0x00007f9564888af0 @flags=1>>
#<Rack::MockResponse:0x00007f95621a83b8>
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,193 @@
...
+
+          def current_span_key
+            OpenTelemetry::Trace::Propagation::ContextKeys.current_span_key
+          end
+
+          def tracer
+            OpenTelemetry::Adapters::Rack::Adapter.instance.tracer
+          end
+
+          ### request_span
```
mwear: We can lose this comment.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,193 @@
...
+          def create_frontend_span(env, extracted_context)
+            request_start_time = OpenTelemetry::Adapters::Rack::Util::QueueTime.get_request_start(env)
+
+            return unless config[:record_frontend_span] && !request_start_time.nil?
+
+            span = tracer.start_span('http_server.queue',
+                                     with_parent_context: extracted_context,
+                                     attributes: {
+                                       'component' => 'http',
+                                       'service' => config[:web_service_name],
```
mwear: We should lose this attribute, or at least have a default for it, as it will be `nil` if one is not provided, and `nil` is not a valid attribute value. I think getting rid of it is the best option for now as it's not an official semantic convention and I think this should come from the application resource when we have a specification for it.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,193 @@
...
+                                       'start_time' => request_start_time
```
mwear: One more issue. This needs to be a number, not a timestamp.

```suggestion
                                       'start_time' => request_start_time.to_f
```

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
duonoid: I've removed the 'service' attribute for the time-being.  :+1:

mwear:
Last call for comments. It'd be nice to merge this tomorrow after the SIG meeting.

adapters/rack/lib/opentelemetry/adapters/rack/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,190 @@
...
+                             attributes: request_span_attributes(env: env),
+                             kind: :server) do |request_span|
+                @app.call(env).tap do |status, headers, response|
+                  set_attributes_after_request(request_span, status, headers, response)
+                end
+              end
+            end
+          ensure
+            finish_span(frontend_context)
+          end
```
fbogsany: This seems pretty convoluted and incorrect. If I understand correctly, with the "frontend" span enabled, we'll end up with 2 spans nested like:
```yaml
name: http_server.queue
start_timestamp: T
end_timestamp: T+D
kind: server
attributes:
  - start_time: request_start_time
children:
  - name: $REQUEST_URI
    start_timestamp: T-ish
    end_timestamp: T+D-ish
    kind: server
    ...
```
There are several problems here:
1. The duration of the 2 spans is about the same, so the queue time is only represented in an attribute of the outer span. It is more useful to create the `http_server.queue` span with `start_timestamp: request_start_time`, which will more accurately represent the LB request duration as the span's duration.
1. Both spans are marked as `kind: server`, which is logically incorrect and would be reported as so by some vendors' trace quality reports. This is supposed to represent time between a proxy/LB receiving a request and the service handling it, there has been chatter about representing this with something like `kind: proxy` (https://github.com/open-telemetry/opentelemetry-specification/issues/51 and https://github.com/open-telemetry/opentelemetry-specification/pull/337).
1. For analysis backends deriving SLIs from trace data, it is possible they may count requests twice if both spans are `kind: server`.
1. Calling the frontend span `http_server.queue` is inaccurate - it actually represents (almost) the entire request processing time from the perspective of the LB/proxy.

Alternatives:
1. Rename the frontend span to `http_server.proxy` or similar. Change the `kind` of the request span to `internal` if the frontend span is present.
1. Use a single span with `start_timestamp: request_start_time, kind: :server` and add a timed event for the service start time (i.e. "now").
1. Create both spans as children of the `extracted_context`, with the frontend span having `start_timestamp: request_start_time, end_timestamp: now` and the request span having `start_timestamp: now`. This allows the frontend span to represent queuing time only.
1. Create 2 spans for the frontend: a `kind: :server` and a `kind: :client` with the same start time and duration but a different resource. Specifying a different resource is not supported by the API presently.

mwear: My vote would be for option 1 or 2. The original datadog instrumentation did have `span.kind` = `proxy`. `internal` would work for now and we can change it to `proxy` when and if it gets added.

As a side issue, I think we are going to have problems with nested server spans. Currently Sinatra will set its span kind as `server`. Using Rack with Sinatra will result in nesting of server spans. We will need to come up with a way to resolve this issue as a user could use Sinatra without Rack, and that would be what you want.

The same is true for client spans with Faraday + net http.

duonoid: I've updated, using Option 1.

mwear: I'm happy with this. @fbogsany WDYT?

duonoid:
I've resolved the merge conflicts.

mwear:
The conflicts are back.

duonoid:
conflicts: resolved. :smile:

fbogsany:
I approved. I'll still extract my comment to an issue, since I think the spec will have to address this at some point, and we may need to revisit. This is 👍 for now tho.

PR #169, opened 2020-01-22
Update release schedule

mwear:
Our release schedule was out of date. I updated it to reflect past releases and with estimates for future dates. The dates are just guesses; we should try to make them reasonable guesses, but they aren't binding. We can and should modify the proposed roadmap through comments (on this PR) and at our SIG meeting.

Fixes #169.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,17 +115,16 @@ for a complete example.
 
 ## Release Schedule
 
-OpenTelemetry Ruby is under active development. Our goal is to release an
-_alpha_ version of the library at the end of September 2019. This release isn't
-guaranteed to conform to a specific version of the specification, and future
-releases will not attempt to maintain backwards compatibility with the alpha
-release.
+OpenTelemetry Ruby is under active development. Below is the release schedule
+for the Ruby library. The first version of the release isn't guaranteed to
+conform  to a specific version of the specification, and future releases will
+not attempt to maintain backward compatibility with the alpha release.
 
 | Component                   | Version       | Target Date       | Release Date      |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,22 +115,24 @@ for a complete example.
...
+| Component                       | Version       | Target Date       | Release Date      |
+| ------------------------------- | ------------- | ----------------- | ----------------- |
+| Tracing API                     | Alpha v0.2.0  | September 30 2019 | November 13 2019  |
+| Tracing SDK                     | Alpha v0.2.0  | September 30 2019 | November 13 2019  |
+| Trace Context Propagation       | Alpha v0.2.0  | September 30 2019 | November 13 2019  |
+| Jaeger Trace Exporter           | Alpha v0.2.0  | September 30 2019 | November 13 2019  |
+| Metrics API                     | Alpha v0.3.0  | February 24 2020  | Unknown           |
+| Metrics SDK                     | Alpha v0.3.0  | February 24 2020  | Unknown           |
+| Prometheus Metrics Exporter     | Alpha v0.3.0  | February 24 2020  | Unknown           |
+| Correlation Context Propagation | Alpha v0.3.0  | February 24 2020  | Unknown           |
```
fbogsany: Feels like this might be done sooner. For the rest, I think Feb 24 is OK as a guess.

mwear: As discussed at the SIG meeting, I think we safe going with Feb 24th for all of these. As the date approaches we can evaluate where we are, and re-prioritize things if it makes sense.

PR #165, opened 2020-01-07
Draft: Sdk Config

mwear:
This PR is currently a draft of SDK configuration and global init. It has three main goals.

* Make it easy to setup an SDK by providing sensible defaults and easy way to override them
* Establish a defined sequence of initialization. This allows us to make certain gaurantees about which systems and globals are available at each state. The sequence is roughly:
  * setup logging
  * setup propagation
  * setup the tracer_factory and meter_factory
  * install instrumentation (as the last step)
* Make it easy for users to configure and install instrumentation

It is essentially a builder that exposes and API to setup the SDK, which will in turn setup the proper components on the OpenTelemetry module.

#### Mimimal Setup

```ruby
OpenTelemetry::SDK.configure

```

This will setup the SDK with:
  - default propagation: W3C Trace Context and W3C Correlation Context
  - basic export: `SimpleSpanProcessor` with `ConsoleExporter`
  - STDOUT logging
  - No instrumentation installed

#### Customized setup

These are the options current available and subject to change as the draft progresses

```ruby
OpenTelemetry::SDK.configure do |c|
  c.tracer_factory = ...
  c.meter_factory = ...
  c.correlation_context_manager = ...
  c.propagation = ...
  c.logger = Logger.new('/dev/null')
  c.add_span_processor(...)
  c.use 'OpenTelemetry::Adapters::Sinatra', {enabled: true}
  c.use 'OpenTelemetry::Adapters::Faraday', {ignore_filters: [/domain\.com/]}
  c.use 'SomeCustom::Adapter::Lib'
end
```

### Installing Instrumentation

This is relevant to #164.

There will be two ways to install instrumentation. 
  * Install adapters individually
  * Install all adapters that are present on the system by way of the discovery system
 
This is how each of those would look. **Note**: this is part of the SDK config draft PR, but an important piece of the instrumentation puzzle:

#### Install adapters individually

```ruby
OpenTelemetry::SDK.configure do |c|
  c.use 'OpenTelemetry::Adapters::Sinatra'
  c.use 'OpenTelemetry::Adapters::Faraday', {some_config: some_value} #config can optionally be passed
end
```

#### Install all adapters with default config

```ruby
OpenTelemetry::SDK.configure do |c|
  c.use_all
end
```

#### Install all adapters with config overrides

Any adapter configs will be passed onto the adapter during installation. Providing them is optional.

```ruby
OpenTelemetry::SDK.configure do |c|
  c.use_all {
[...]

PR #171, opened 2020-01-22
Preliminary sdk config

mwear:
This PR is an implementation of SDK config for currently implemented features. It's based off of #165, but with speculative, future feature configuration removed. With this implementation you can configure logging, tracer factory, span processors and install instrumentation.

Here are some examples:

### Minimal setup

If a block is not specified it will setup defaults
* STDOUT logging
* SDK::Trace::TracerFactory
* SimpleSpanProcessor with ConsoleSpanExporter
* No instrumentation

```ruby
OpenTelemetry::SDK.configure
```

### Customizing the setup

The following can be customized:
* Logger
* Tracer Factory
* Span Processors
* Instrumentation: installed individually, or all, but not both

Customize the setup and specify instrumentation individually

```ruby
  c.tracer_factory = ...
  c.logger = Logger.new('/dev/null')
  c.add_span_processor(...)
  c.use 'OpenTelemetry::Adapters::Sinatra', {enabled: true}
  c.use 'OpenTelemetry::Adapters::Faraday', {ignore_filters: [/domain\.com/]}
  c.use 'SomeCustom::Adapter::Lib'
end
```

Customize the setup and install all instrumentation

```ruby
  c.tracer_factory = ...
  c.logger = Logger.new('/dev/null')
  c.add_span_processor(...)
  c.use_all 'OpenTelemetry::Adapters::Something' => {config_opt: true}
end
```

### DIFF
--- a/sdk/lib/opentelemetry/sdk.rb
+++ b/sdk/lib/opentelemetry/sdk.rb
@@ -22,6 +22,36 @@ module OpenTelemetry
     # @yieldparam [Configurator] configurator Yields a configurator to the
     #   provided block
     #
+    # Example usage:
+    #   Without a block defaults are installed without any instrumentation
+    #
+    #     OpenTelemetry::SDK.configure
+    #
+    #   Install instrumentation individually with optional config
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.use 'OpenTelemetry::Adapters::Faraday', tracer_middleware: SomeMiddleware
+    #     end
+    #
+    #   Install all instrumentation with optional config
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.use_all 'OpenTelemetry::Adapters::Faraday' => { tracer_middleware: SomeMiddleware }
+    #     end
+    #
+    #   Add a span processor
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.add_span_processor SpanProcessor.new(SomeExporter.new)
+    #     end
+    #
+    #   Configure everything
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.tracer_factor = SomeTracerFactory.new
+    #       c.add_span_processor SpanProcessor.new(SomeExporter.new)
+    #       c.use_all
+    #     end
     def configure
       configurator = Configurator.new
       yield configurator if block_given?
--- a/sdk/lib/opentelemetry/sdk.rb
+++ b/sdk/lib/opentelemetry/sdk.rb
@@ -15,9 +15,22 @@ require 'opentelemetry'
 module OpenTelemetry
   # SDK provides the reference implementation of the OpenTelemetry API.
   module SDK
+    extend self
+
+    # Configures SDK and instrumentation
+    #
+    # @yieldparam [Configurator] configurator Yields a configurator to the
+    #   provided block
+    #
+    def configure
+      configurator = Configurator.new
+      yield configurator if block_given?
+      configurator.configure
+    end
   end
 end
 
+require 'opentelemetry/sdk/configurator'
 require 'opentelemetry/sdk/internal'
 require 'opentelemetry/sdk/resources'
 require 'opentelemetry/sdk/trace'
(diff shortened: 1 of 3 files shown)


sdk/lib/opentelemetry/sdk.rb
```diff
@@ -15,9 +15,52 @@
...
+    #   Add a span processor
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.add_span_processor SpanProcessor.new(SomeExporter.new)
+    #     end
+    #
+    #   Configure everything
+    #
+    #     OpenTelemetry::SDK.configure do |c|
+    #       c.tracer_factor = SomeTracerFactory.new
```
fbogsany: s/tracer_factor/tracer_factory/

sdk/lib/opentelemetry/sdk.rb
```diff
@@ -15,9 +15,52 @@
...
+    #     end
+    def configure
+      configurator = Configurator.new
+      yield configurator if block_given?
+      configurator.configure
+    end
   end
 end
 
+require 'opentelemetry/sdk/configurator'
```
fbogsany: Naive question: should this `require` come before the `Configurator.new` or does it not matter (does the constant resolve happen at the point of call or at the point of compilation?).

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      def initialize
+        @adapter_names = []
+        @adapter_config_map = {}
+        @span_processors = []
+        @use_mode = USE_MODE_UNSPECIFIED
+      end
+
+      def tracer_factory
+        @tracer_factory ||= Trace::TracerFactory.new
```
fbogsany: Is it necessary to support replacing the tracer factory? I can see you'd want to do that in order to configure the tracer factory, but the documentation earlier suggests replacing the implementation.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -0,0 +1,114 @@
...
+        @adapter_config_map[adapter_name] = config if config
+      end
+
+      # Install all registered instrumentation. Configuration for specific
+      # adapters can be provided with the optional +adapter_config_map+
+      # parameter. Only +use+ or +use_all+, but not both when installing
+      # instrumentation. A call to +use+ after +use_all+ will result in an
+      # exception.
+      #
+      # @param [Hash<String,Hash>] adapter_config_map A map with string keys
```
fbogsany: optional

sdk/lib/opentelemetry/sdk.rb
mwear: The constant resolution will happen at the point of call. I can move the require if that'd make folks more comfortable.

sdk/lib/opentelemetry/sdk/configurator.rb
mwear: I suppose this doesn't make a whole lot of sense. It seems like a user would be using a different SDK, or a hybrid SDK of sorts. In either case, we don't have to support that for the default SDK, and I don't think we should, unless a use case becomes apparent.

sdk/lib/opentelemetry/sdk.rb
fbogsany: No, this is fine - just wanted to know 😄

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -0,0 +1,111 @@
...
+      end
+
+      def configure_span_processors
+        processors = @span_processors.empty? ? [default_span_processor] : @span_processors
+        processors.each { |p| @tracer_factory.add_span_processor(p) }
+      end
+
+      def default_span_processor
+        Trace::Export::SimpleSpanProcessor.new(
+          Trace::Export::ConsoleSpanExporter.new
```
elskwid: This is nice!

mwear:
Relevant to: https://github.com/open-telemetry/opentelemetry-specification/issues/390

PR #172, opened 2020-01-22
Add `rest-client` instrumentation adapter

grantbdev:
# Overview

Add `rest-client` instrumentation adapter similar to the existing adapter for Faraday.

The `rest-client` gem uses `restclient` internally for file/folder naming, so I used that style for the adapter code. The gem's README also titleizes the name as simply "REST Client", so that is used in some spots but if that is too confusing I could try to clarify or consolidate on a single name.

## Related
* Effort toward resolving #67
* Largely based on #148 and the subsequent changes to code added there


adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
```diff
@@ -0,0 +1,74 @@
...
+          end
+
+          private
+
+          def trace_request
+            span = tracer.start_span(
+              url,
+              attributes: {
+                'component' => 'http',
+                'http.method' => method.to_sym,
```
mwear: A symbol is not a valid attribute value. It should be a string.

adapters/restclient/test/test_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+
+require 'rest-client'
+
+require 'opentelemetry/sdk'
+
+require 'minitest/autorun'
+require 'webmock/minitest'
+
+# global opentelemetry-sdk setup:
+sdk = OpenTelemetry::SDK
```
mwear: You can use OpenTelemetry::SDK.configure here as well. See #177 for an example.

adapters/restclient/lib/opentelemetry/adapters/restclient/adapter.rb
```diff
@@ -0,0 +1,33 @@
...
+    module RestClient
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
+          config[:request_patch] ||= Patches::Request
+          patch_request
+        end
+
+        present do
+          Gem.loaded_specs.include?('rest-client')
```
mwear: We should actually be checking `defined?(::RestClient)`. The reason - a gemspec can be loaded, but the gem can go unrequired. When that happens we reference an undefined constant. The instrumentation installation handles this gracefully, but dumps a nasty stack trace in the logs. We should check for constants in the target library instead. Sorry for setting a bad example earlier on. I've made the same changes to Faraday and Sinatra in #177.

adapters/restclient/lib/opentelemetry/adapters/restclient/adapter.rb
```diff
@@ -0,0 +1,33 @@
...
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module RestClient
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
```
mwear: I'm taking a mental note of how. many times I see this in adapters. It might make sense to make this an optional part of the DSL at some point.

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
```diff
@@ -0,0 +1,74 @@
...
+  module Adapters
+    module RestClient
+      module Patches
+        # Module to prepend to RestClient::Request for instrumentation
+        module Request
+          def execute(&block)
+            return super(&block) if disable_span_reporting?
+
+            trace_request do |span|
+              propagate_context(span)
```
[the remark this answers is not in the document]
grantbdev: @mwear It can be moved and the `propagate_context` wrapper method can be replaced. I was only following the style of the Faraday adapter. Would you like me to do your suggested changes?

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
grantbdev: @mwear The Faraday adapter should be updated as well then. Should I handle that in this PR?

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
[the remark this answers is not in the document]
mwear: I don't think the extra indirection buys us anything, so I am inclined to make that change. If you agree, and wouldn't mind making the same change to Faraday, that'd be great. I'm open to discussion if you prefer the way things are current structured though.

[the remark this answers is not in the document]
grantbdev: @mwear I agree on removing the indirection.

The only thing that gives me pause about moving the propagation line is that's also where it's done in `dd-trace-rb`. However, I think it would be logically the same to place in `trace_request` before the `yield`.

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
mwear: Good catch (https://rubydoc.info/gems/faraday/Faraday/Env#method-instance_method). Go ahead and include it in this PR.

grantbdev:
@mwear Thank you for the feedback! I've pushed up my changes.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -14,10 +14,10 @@ def call(env)
 
             tracer.in_span(env.url.to_s,
                           attributes: { 'component' => 'http',
-                                        'http.method' => env.method,
+                                        'http.method' => env.method.to_s.upcase,
```
mwear: Could we have a map from symbol to normalized string to avoid object allocations on each span?

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
```diff
@@ -0,0 +1,69 @@
...
+  module Adapters
+    module RestClient
+      module Patches
+        # Module to prepend to RestClient::Request for instrumentation
+        module Request
+          def execute(&block)
+            return super(&block) if disable_span_reporting?
+
+            trace_request do |span|
+              propagator.inject(span.context, processed_headers)
```
mwear: Is there any problem moving this to line 41? It would make things easier to reason through if so.

grantbdev:
@mwear Feedback addressed.

adapters/faraday/example/faraday_no_op.rb
```diff
@@ -0,0 +1,25 @@
...
+require_relative '../lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware'
+# Demonstrate disabling span reporting
+class NoOp < OpenTelemetry::Adapters::Faraday::Middlewares::TracerMiddleware
+  def disable_span_reporting?(env)
+    env.url.to_s =~ /example.com/
+  end
+end
+
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Adapters::Faraday', tracer_middleware: NoOp
```
grantbdev: @mwear @elskwid Based on our discussion, I've updated the example files to use the proper installation and to have the no-op examples in separate files.

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
```diff
@@ -0,0 +1,69 @@
...
+          # span reporting rules.
+          def disable_span_reporting?
+            false
+          end
+
+          private
+
+          def trace_request # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+            span = tracer.start_span(
+              url,
```
mwear: This will lead to high-cardinality span names. See: https://github.com/open-telemetry/opentelemetry-ruby/pull/175#discussion_r373245633.

grantbdev: @mwear We could use [`"HTTP #{METHOD_NAME}"`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#name). If that sounds good I'll update both PRs.

Only other thing I can think of that would add specificity without introducing a lot more would be to include URL host/domain.

grantbdev:
@mwear Updated this PR with lower-cardinality span names and a few other tweaks I've found from working on more adapters.

Do you think we should add setting of `span.status` to each adapter?
https://github.com/mutations/opentelemetry-ruby/commit/35204f07cb543ec3e533365f6cb148fb0f4ca866#diff-d5fef5fa085cedfa6cea4c783a6bd8c2R47

adapters/restclient/lib/opentelemetry/adapters/restclient/adapter.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module OpenTelemetry
+  module Adapters
+    module RestClient
+      # The Adapter class contains logic to detect and install the RestClient
+      # instrumentation adapter
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
+          config[:request_patch] ||= Patches::Request
```
mwear: Is there any reason to make the request patch configurable at this point?

adapters/restclient/lib/opentelemetry/adapters/restclient/patches/request.rb
```diff
@@ -0,0 +1,62 @@
...
+              super(&block)
+            end
+          end
+
+          private
+
+          def trace_request # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+            http_method = method.upcase
+            span = tracer.start_span(
+              "HTTP #{http_method}",
```
mwear: I'm not sure that I am crazy about this span name either. Ultimately we are going to want the http quantization that is currently in the Rack PR to help us name things in accordance with OpenTelemetry guidelines. We should discuss what we should do in the interim.

mwear: I think using just `{verb}` or `{verb} {host}` would be good interim solutions. The latter might be best if it's not too much work, since the goal is to replace it soon anyways.

adapters/restclient/lib/opentelemetry/adapters/restclient/adapter.rb
grantbdev: https://github.com/open-telemetry/opentelemetry-ruby/pull/176#discussion_r374187096

Not that I know of, but I wanted to verify with you before removing.

grantbdev:
@mwear Feedback addressed.

grantbdev:
PR updated to use `"HTTP #{method_name}"` span names as discussed in the SIG meeting.

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -11,35 +11,40 @@ module Middlewares
...
+            tracer.in_span(
+              "HTTP #{http_method}",
+              attributes: {
+                'component' => 'http',
+                'http.method' => http_method,
+                'http.url' => env.url.to_s
+              },
+              kind: :client
+            ) do |span|
+              propagator.inject(span.context, env.request_headers)
```
mwear: Since the context prop PR was merged this can be changed to:

```suggestion
             OpenTelemetry.propagation.inject(env.request_headers)
```

adapters/faraday/lib/opentelemetry/adapters/faraday/middlewares/tracer_middleware.rb
```diff
@@ -11,35 +11,40 @@ module Middlewares
...
           attr_reader :app
 
-          # Outbound requests should only need to inject the current span.
-          def propagate_context(span, env)
-            propagator.inject(span.context, env.request_headers)
-          end
-
           def propagator
             OpenTelemetry.tracer_factory.http_text_format
           end
```
mwear: This can probably be removed, unless you'd prefer to delegate to `OpenTelemetry.propagation`.

grantbdev:
@mwear PR updated and I made sure to use the block that provides the span context for `restclient` like you suggested for `excon`.

I'm not sure that we have anything explicitly testing propagation, let me know if you have ideas on how to accomplish that.

mwear:
@grantbdev  Thanks for bringing up propagation tests. It is something we should test and in general, we should port over applicable tests when converting Datadog instrumentation. Here is an example from dd-trace-rb: https://github.com/DataDog/dd-trace-rb/blob/master/spec/ddtrace/contrib/rest_client/request_patch_spec.rb#L196-L209

By default OpenTelemetry.propagation.http_injectors is empty, making inject effectively a no-op. You could setup a custom injector for your test, or you could just use the trace context injector. Currently, there is a debate as to whether or not trace context should be a default injector. You can use the Datadog test as template and setup an injector like so:

```ruby
before do
  # these are currently empty, but this will future proof the test
  @orig_injectors = OpenTelemetry.propagation.http_injectors
  OpenTelemetry.propagation.http_injectors = [OpenTelemetry::Trace::Propagation.http_trace_context_injector]
end

after do
  OpenTelemetry.propagation.http_injectors = @orig_injectors
end
```

grantbdev:
@mwear Tests added, let me know what you think. I wasn't sure if we should use the `TraceParent` class directly to generate the header value in the test and I didn't know how to get the context for that approach either.

mwear:
👀 ing good @grantbdev. Thanks for addressing all my feedback. I am slightly weirded out that the assertions check for `headers['Traceparent']` when the injector uses `traceparent` as the key. Out of curiosity, do you know if that is the libraries doing that normalization?

Anyhow, I am happy with where this is. Would you like me to leave this open for you to take a look @fbogsany, or should I go ahead and merge?

grantbdev:
@mwear I think the capitalization is coming from the libraries. So far, all of them except `ethon` (which I haven't opened a PR for yet) have capitalized the `traceparent` header.

mwear:
This has been open for a bit and hasn't received any new comments. I'd like to merge it, but will need a rebase first.

grantbdev:
@mwear Rebased

mwear:
Thanks @grantbdev!

PR #175, opened 2020-01-23
Add `excon` instrumentation adapter

grantbdev:
# Overview

Add `excon` instrumentation adapter similar to the existing adapter for Faraday.

## Related
* Effort toward resolving #67


grantbdev:
@mwear I've updated this PR to address all the relevant feedback from #172.

adapters/excon/lib/opentelemetry/adapters/excon/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,100 @@
...
+  module Adapters
+    module Excon
+      module Middlewares
+        # Excon middleware for instrumentation
+        class TracerMiddleware < ::Excon::Middleware::Base
+          def request_call(datum) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+            begin
+              unless disable_span_reporting?(datum) || datum.key?(:otel_span)
+                tracer.start_span(
+                  datum[:path],
```
mwear: This could lead to high cardinality span names. See: https://github.com/open-telemetry/opentelemetry-specification/pull/416/files. We probably need a quantization approach in the long run. We can discuss options in the short term. Note, this issue exists in Rack, Faraday, and RestClient as well. Are there any opinions on how we should handle this now (as in this PR).

adapters/excon/lib/opentelemetry/adapters/excon/adapter.rb
```diff
@@ -0,0 +1,36 @@
...
+
+module OpenTelemetry
+  module Adapters
+    module Excon
+      # The Adapter class contains logic to detect and install the Excon
+      # instrumentation adapter
+      class Adapter < OpenTelemetry::Instrumentation::Adapter
+        install do |config|
+          require_dependencies
+          config[:tracer_middleware] ||= Middlewares::TracerMiddleware
```
mwear: As with `Rest::Client`, I'm not sure that a configuration option for the middware makes sense.

grantbdev:
@mwear PR feedback addressed.

These spec statements seem conflicting:
> Implementations MUST set status if the HTTP communication failed or an HTTP error status code is returned (e.g. above 3xx).

> Don't set a status message if the reason can be inferred from `http.status_code` and `http.status_text` already.

mwear:
The [api-tracing spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#status) has this to say:

> Status interface represents the status of a finished Span. It's composed of a canonical code in conjunction with an optional descriptive message.

My interpretation is that a canonical status code should be set, but the optional descriptive message shouldn't be if it can be inferred from the http status code (or message).

grantbdev:
PR updated to use `"HTTP #{method_name}"` span names as discussed in the SIG meeting.

adapters/excon/lib/opentelemetry/adapters/excon/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,118 @@
...
+                    'component' => 'http',
+                    'http.host' => datum[:host],
+                    'http.method' => http_method,
+                    'http.scheme' => datum[:scheme],
+                    'http.target' => datum[:path]
+                  },
+                  kind: :client
+                ).tap do |span|
+                  datum[:otel_span] = span
+                  propagator.inject(span.context, datum[:headers])
```
mwear: Since the context prop PR merged this should be:

```suggestion
                  tracer.with_span(span) do
                    OpenTelemetry.propagation.inject(datum[:headers])
                  end
```

The reason for `with_span` is to make this span the current span in the current context. `in_span` will do that automatically, but `start_span` doesn't. `in_span` is not a good fit for this situation because it will finish the span after executing the block.

adapters/excon/lib/opentelemetry/adapters/excon/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,118 @@
...
+                datum.delete(:otel_span)
+              end
+            end
+          rescue StandardError => e
+            OpenTelemetry.logger.debug(e.message)
+          end
+
+          def propagator
+            OpenTelemetry.tracer_factory.http_text_format
+          end
```
mwear: We probably don't need this with the previous suggestion.

grantbdev:
@mwear PR updated. Thanks!

mwear:
Like https://github.com/open-telemetry/opentelemetry-ruby/pull/172#issuecomment-583563367, we should test propagation here as well. This is the corresponding datadog test: https://github.com/DataDog/dd-trace-rb/blob/master/spec/ddtrace/contrib/excon/instrumentation_spec.rb#L166-L188.

mwear:
👀 ing good here too @grantbdev. Let me know if you want to take a look here as well @fbogsany.

mwear:
Same as #172, this could a use a rebase and I'll merge it.

mwear:
I used to have a button where I could merge in master, but it's no longer here. I guess we'll need another rebase.

grantbdev:
@mwear PR rebased.

PR #176, opened 2020-01-24
Add `redis` instrumentation adapter

grantbdev:
# Overview

Add `redis` instrumentation adapter. This is the first database client adapter, so I made it conform to those [conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-database.md) as best I could understand. I was particularly uncertain on what the span names should be, but based on the [spec guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/api-tracing.md#span) I decided to go with the command name for single command requests and "pipeline" for a pipeline request. Let me know if there is anything that should be changed or added.

A lot of the formatting code is copied from `dd-trace-rb`. It was unclear to me whether in OpenTelemetry there are length limits on span attributes, so I kept the truncation behavior from `dd-trace-rb` let me know if the truncation limit should be changed or removed entirely.

## Related
* Effort toward resolving #67


adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
```diff
@@ -0,0 +1,78 @@
...
+            end
+
+            response
+          end
+
+          # Override implementation (subclass) to determine per-connection
+          # span reporting rules.
+          def disable_span_reporting?(*_args)
+            false
+          end
```
mwear: I don't see how we have per-connection span reporting. Either this module is prepended to the redis client, or the no-op version is. We either get all spans, or no spans. We already have the ability to do this by installing the instrumentation (with c.use), or not installing the instrumentation. Correct me if I'm mistaken on this point.

adapters/redis/lib/opentelemetry/adapters/redis/utils.rb
```diff
@@ -0,0 +1,82 @@
...
+  module Adapters
+    module Redis
+      # Utility functions
+      module Utils
+        STRING_PLACEHOLDER = ''.encode(::Encoding::UTF_8).freeze
+        PLACEHOLDER = '?'
+        VALUE_MAX_LEN = 50
+        CMD_MAX_LEN = 500
+
+        def self.format_command(command_args)
```
mwear: We have generally been using extend self for modules used as a bag of methods.

adapters/redis/test/opentelemetry/adapters/redis/adapter_test.rb
```diff
@@ -0,0 +1,89 @@
...
+      _(span.attributes['component']).must_equal 'redis'
+      _(span.attributes['db.type']).must_equal 'redis'
+      _(span.attributes['db.instance']).must_equal '0'
+      _(span.attributes['db.statement']).must_equal "SET v1 0\nSET v2 0"
+      _(span.attributes['db.url']).must_equal 'redis://example.com:8321'
+      _(span.attributes['net.peer.name']).must_equal 'example.com'
+      _(span.attributes['net.peer.port']).must_equal 8321
+    end
+  end
+
```
mwear: These tests look fairly slim. I would expect to see the logic for more operations other than set to be tested. I also noticed that the the `Utils` class is a port of `Datadog::Contrib::Redis::Quantize`. There is a test for that in dd-trace, but not one here. 

Can you ensure that we have testing for OTel that is equivalent to what is in dd-trace?

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
grantbdev: @mwear "per-connection" may be misleading, but the arguments provided to this method would allow someone to create their own patch that would could disable spans for certain methods, keys, values, etc. That may not be a use case that needs to be supported and could be simplified/removed, but I kept it for parity with the functionality of the HTTP client adapters so far.

grantbdev:
@mwear PR updated. I ported the tests of the ported utils. I would like more input on what more of the adapter should be tested.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
mwear: I would err on the side of YAGNI at this stage in the project unless compelling use cases become apparent. My preference would be to remove it for now. That would also apply to the http client instrumentation

grantbdev: @mwear Should the ability to provide a custom patch/middleware for adapters still exist? It was only used in tests/examples for overriding this method, but I'm not sure if there is another use case or if it is desired for dependency injection.

grantbdev:
PR updated to add test assertions on the span names so it's easier for reviewers to see what it looks like.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
```diff
@@ -0,0 +1,68 @@
...
+
+            {
+              'component' => 'redis',
+              'db.type' => 'redis',
+              'db.instance' => options[:db].to_s,
+              'db.url' => "redis://#{host}:#{port}",
+              'net.peer.name' => host,
+              'net.peer.port' => port
+            }
+          end
```
mwear: It looks like `db` is changeable at runtime, but is everything else going to be static data once the client is initialized? If so, can we memoize and freeze this data instead of making it each time? The `db.name` can be merged in with `db.statement` in call.

adapters/redis/test/opentelemetry/adapters/redis/adapter_test.rb
```diff
@@ -0,0 +1,67 @@
...
+        redis.incr('v1')
+        redis.get('v1')
+      end
+
+      _(exporter.finished_spans.size).must_equal 1
+      _(span.name).must_equal 'pipeline'
+      _(span.attributes['component']).must_equal 'redis'
+      _(span.attributes['db.type']).must_equal 'redis'
+      _(span.attributes['db.instance']).must_equal '0'
+      _(span.attributes['db.statement']).must_equal "SET v1 0\nINCR v1\nGET v1"
```
mwear: We will eventually have a array valued attributes. This might be a good use for them, but for now, this works.

adapters/redis/test/opentelemetry/adapters/redis/adapter_test.rb
```diff
@@ -0,0 +1,67 @@
...
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+require_relative '../../../../lib/opentelemetry/adapters/redis'
+require_relative '../../../../lib/opentelemetry/adapters/redis/patches/client'
+
+describe OpenTelemetry::Adapters::Redis::Adapter do
```
mwear: Our tests are looking a little slim compared to those in dd-trace: https://github.com/DataDog/dd-trace-rb/blob/master/spec/ddtrace/contrib/redis/redis_spec.rb. Make sure we have an equivalent test for everything that applies. I realize we might not have ported everything, but there is a test for spans generated from the auth command, for example, that I think we should preserve.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
grantbdev: > is everything else going to be static data once the client is initialized?

@mwear I don't think so. https://github.com/redis/redis-rb#sentinel-support and https://github.com/redis/redis-rb#cluster-support seem to indicate to me that the port and even host can be different between requests. Using the example code I was able to see different hosts/ports for the spans using Sentinel. Unfortunately it does not appear that `fakeredis` which we are using for the tests supports Sentinel or clusters.

> The db.name can be merged in with db.statement in call.

Can you explain this more?

grantbdev:
@mwear Feedback addressed.

Something to consider: [ddtrace's #trace automatically catches errors and sets the span status.](https://github.com/DataDog/dd-trace-rb/blob/930d8210ab2bc6f54b35503aa41be9bb674c1c93/lib/ddtrace/tracer.rb#L278) I've started in-lining that logic, but it's something that may be good to add to `#in_span`.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
mwear: I was just saying that if the config was static aside from `db.name` we could have a frozen hash and merge('db.name' => ..., 'db.statement') with it. It's a moot point given that config will change at runtime. Thanks for your research!

mwear:
We might want to change `in_span` automatically rescue, record the error, and reraise. I'll open an issue about this and we can discuss it at the SIG.

mwear:
I'm not totally sure about the inlining, but everything else looks good to me. I created an with some questions / proposals around error handling: https://github.com/open-telemetry/opentelemetry-ruby/issues/181. I think we should discuss this at the SIG and determine what to do about the inlining there. Right now, to replicate what is happening in dd-trace would have to inline exception handling for every Tracer#in_span call, which is definitely not ideal.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
```diff
@@ -0,0 +1,86 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Redis
+      module Patches
+        # Module to prepend to Redis::Client for instrumentation
+        module Client
+          def call(*args, &block)
```
fbogsany: The gem has a couple of other `call_*` methods: `call_pipelined` (note the difference from `call_pipeline`) and `call_loop`. Have you considered patching `process` instead? The upside is that it is a bottleneck method that is called on all `call*` paths. The downside is that errors are sometimes raised outside its scope, so they're harder to intercept.

grantbdev: @fbogsany I did not consider those because `dd-trace-rb` does not patch those methods.

fbogsany: Relevant bits of Redis client: https://github.com/redis/redis-rb/blob/41395e9a1923fe72fad8403e60cb98627d515a85/lib/redis/client.rb#L124-L215

grantbdev:
@mwear This PR (and my other adapter PRs) are rebased and the exception handling has been simplified.

adapters/redis/lib/opentelemetry/adapters/redis/patches/client.rb
mwear: I made an issue to track improvements: #197.

mwear:
Last call for comments on this. It'd be nice to merge this tomorrow after the SIG meeting.

PR #180, opened 2020-02-06
Allow non-string Resource label values.

### LINKED ISSUE #158: Resource should support the same attribute types as span attributes
I think this is just a documentation change in [Resource](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/sdk/lib/opentelemetry/sdk/resources/resource.rb) to match [Span#set_attribute](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/api/lib/opentelemetry/trace/span.rb).

Addresses https://github.com/open-telemetry/opentelemetry-specification/pull/356

fhwang:
Fixes #158.


fhwang:
I think I'm set up with the CLA -- commenting here in hopes this will fire the Github check ...

fhwang:
test comment

fhwang:
@mwear any thoughts on how I can fix the CLA issue? I thought I went through the required steps but I probably missed something.

elskwid:
@fhwang I believe you need to post a comment that says `I signed it`

fhwang:
I signed it

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -21,9 +21,10 @@ class << self
           # @return [Resource]
           def create(labels = {})
             frozen_labels = labels.each_with_object({}) do |(k, v), memo|
-              raise ArgumentError, 'label keys and values must be strings' unless k.is_a?(String) && v.is_a?(String)
+              raise ArgumentError, 'label keys must be strings' unless k.is_a?(String)
+              raise ArgumentError, 'label values must be strings, integers, floats, or booleans' unless Internal.valid_value?(v)
```
fbogsany: This requires a documentation change as well (the `@param` in the comment above).

fhwang:
I signed it

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -15,15 +15,16 @@ class << self
 
           # Returns a newly created {Resource} with the specified labels
           #
-          # @param [Hash<String, String>] labels Hash of key-value pairs to be used
+          # @param [Hash{String=>String, Numeric, Boolean>} labels Hash of key-value pairs to be used
```
mwear: I think we have been taking the easy way out by using `Hash<String, Object>`, but this is more accurate and an established convention according to the [yardoc site](https://yardoc.org/types#see_results). @fbogsany, should we go forward with this and change our other docs accordingly. Follow up question, if so, should we do that as part of this PR? Followup to the followup, eventually we'll be supporting array value attributes, which will level up the documentation complications, should we worry about this, or am I getting too far ahead.

fhwang: If we want to update the other docs in the same way, I'm happy to do that next.

mwear: That sounds good @fhwang. Do you want to do that as part of this PR or as a followup?

Also, taking a second look at this doc change, it does appear as though we have a lingering `>` and we should probably add spaces around the `=>`. Can we change this to `[Hash{String => String, Numeric, Boolean}]`?

fhwang: I suppose I'll do that work as a followup PR, in the interest of keeping PRs small when possible.

I've made the requested changes.

PR #182, opened 2020-02-12
Improve YARD types for attributes.

fhwang:


api/lib/opentelemetry/trace/link.rb
```diff
@@ -22,14 +22,15 @@ class Link
...
+      # @return [Hash{String => String, Numeric, Boolean}]
       attr_reader :attributes
 
       # Returns a new immutable {Link}.
       #
       # @param [SpanContext] span_context The context of the linked {Span}.
-      # @param [optional Hash<String, Object>] attributes A hash of attributes for
-      #   this link. Attributes will be frozen during Link initialization.
+      # @param [optional Hash{String => String, Numeric, Boolean}] attributes A
+      #   hash of attributes for this event. Attributes will be frozen during
```
mwear: This should be `link`, not event.

fhwang: Sorry about that -- got a little sloppy with cut-and-paste. Should be fixed now.

Issue #181, opened 2020-02-12
Spans and Errors


mwear:
### Background

As we convert instrumentation, ensuring that we preserve error recording from the original instrumentation is becoming a problem. One of the big bottlenecks to this problem is that there is not clear guidance on how to record errors in OpenTelemetry. Several proposals have surfaced recently, notably: 

* https://github.com/open-telemetry/opentelemetry-specification/pull/427
* https://github.com/open-telemetry/opentelemetry-specification/pull/432

I wrote some Ruby code to see how handling an error would look for each of the proposals. It's in this [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/427#issuecomment-584424963), but I'll cut and paste it for reference:

Example for #427 

```ruby
begin
  buggy_method     # raises an exception when called
rescue => e
  span.add_event(name: 'error',
                 attributes: {
                   'error.type' => e.class,
                   'error.message' => e.message,
                   'error.stack' => e.backtrace
                 })
end
```

Example for #432

```ruby
begin
  buggy_method     # raises an exception when called
rescue => e
  span.add_event(name: 'error',
                 attributes: {
                   'log.type' => e.class,
                   'log.severity' => 'error',
                   'log.message' => e.message,
                   'frame.file' => e.backtrace_locations[0].absolute_path,
                   'frame.line' => e.backtrace_locations[0].lineno,
                   'frame.func' => e.backtrace_locations[0].label,
                   'frame.stack' => stack_to_json(e)
                 })
end

def stack_to_json(e)
  JSON.dump(
    e.backtrace_locations.map do |loc|
      { 'frame.file' => loc.absolute_path, 'frame.line' => loc.lineno, 'frame.func' => loc.label }
    end
  )
end
```

### Should we add a Span#add_error or Span#record_error convenience method?

In writing that example, one thing became clear to me - we're going to want a convenience method to simplify recording errors.

The Go SIG has reached the same conclusion:

* https://github.com/open-telemetry/opentelemetry-go/issues/248
* https://github.com/open-telemetry/opentelemetry-go/pull/473

I'm wondering if we should preemptively add a convenience method. It could even no-op for the time being, or perhaps speculatively record an event with semantic conventions we expect to change. Possible names could be:

* Span#add_error
* Span#record_error

### Should in_span record and reraise exceptions?

[...]

PR #184, opened 2020-02-14
Preliminary error recording

mwear:
Based on #181 and our discussions at the SIG meeting, we decided to go ahead with preliminary error recording. This PR adds a `Span#record_error` that records an error as an event with the semantic conventions laid out in https://github.com/open-telemetry/opentelemetry-specification/pull/427. The stack trace is stored as a newline delimited string until we have support for array-value attributes. This also updates `Tracer#in_span` to record and reraise exceptions.

We expect that the implementation of `Span#record_error` will change as it becomes officially spec'd, but this will give us something that we can use now. It will be helpful for a couple of open PRs: #166 and #176.


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -43,10 +43,17 @@ def active_span_context(context = nil)
...
+      # On exit, the Span that was active before calling this method will be reactivated. If an
+      # exception occurs during the execution of the provided block, it will be recorded on the
+      # span and reraised.
       def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, sampling_hint: nil, with_parent: nil, with_parent_context: nil)
         span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, sampling_hint: sampling_hint, with_parent: with_parent, with_parent_context: with_parent_context)
         with_span(span) { |s| yield s }
+      rescue Exception => e # rubocop:disable Lint/RescueException
+        span.record_error(e)
+        span.status = Status.new(Status::UNKNOWN_ERROR,
+                                 description: "Unhandled exception of type: #{e.class}")
```
elskwid: 👍

mwear:
Merging this as it should be useful for getting #166 and #176 finished.

PR #185, opened 2020-02-15
Add support for array-valued attributes.

### LINKED ISSUE #160: Add support for array-valued attributes
See https://github.com/open-telemetry/opentelemetry-specification/pull/368

fhwang:
Resolves #160.


sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -16,14 +16,39 @@ def boolean?(value)
...
+                         end
+          if !first_element_type
+            first_element_type = element_type
+          elsif element_type != first_element_type
+            return false
+          end
+        end
+        true
+      end
+
```
mwear: It might be a improvement if we remove this, and then write things as I have in the following comment.

sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -16,14 +16,39 @@ def boolean?(value)
...
       end
 
+      def valid_array_value?(value)
+        value.is_a?(Array) && value.all? { |elt| valid_primitive_value?(elt) } && homogenous_element_types?(value)
+      end
+
+      def valid_value?(value)
+        valid_primitive_value?(value) || valid_array_value?(value)
+      end
+
```
mwear: And then write this as:

```ruby
      def valid_value?(value)
        value.instance_of?(String) || value == false || value == true || value.is_a?(Numeric) || valid_array_value?(value)
      end

      def valid_array_value?(value)
        return false unless value.is_a?(Array)

        case value.first
        when String
          value.all? { |v| v.instance_of?(String) }
        when TrueClass, FalseClass
          value.all? { |v| boolean?(v) }
        when Numeric
          value.all? { |v| v.is_a?(Numeric) }
        else
          false
        end
      end
```

WDYT?

api/lib/opentelemetry/trace/span.rb
```diff
@@ -77,10 +77,10 @@ def set_attribute(key, value)
       #
       # @param [optional String] name Optional name of the event. This is
       #   required if a block is not given.
-      # @param [optional Hash{String => String, Numeric, Boolean}] attributes
-      #   One or more key:value pairs, where the keys must be strings and the
-      #   values may be string, boolean or numeric type. This argument should
-      #   only be used when passing in a name.
+      # @param [optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}]
+      #   attributes One or more key:value pairs, where the keys must be
+      #   strings and the values may be string, boolean or numeric type. This
```
fbogsany: Comment should be updated to reflect that `values may be (array of) string, boolean or numeric type.`

sdk/lib/opentelemetry/sdk/internal.rb
fbogsany: Probably also need to handle the empty array case.

mwear: Good call.

fhwang: I could be convinced otherwise on this, but I think there's some value to having standalone predicates named `valid_primitive_value?` and `homogenous_element_types?`. `valid_primitive_value?` gets used twice on my original code, where the suggested code buries the second check amid the `case` statement in `valid_array_value?`. And it seems more readable to highlight `homogenous_element_types?` as a specific, distinct condition of validity for array values.

fhwang: @mwear any thoughts on my comment from Sunday?

mwear: @fhwang I strongly prefer the code that I suggested both for its simplicity and its performance. I've provided a benchmark for each implementation for both string and numeric arrays. The best situation for your implementation is the string case, where it is 2.58x slower than my suggestion. The worse case for your implementation is the numeric case, where it is 8.21x slower than my suggestion. The reason there is such a difference in performance between the two implementations is because my suggestion does a a few upfront checks, but a single comparison for each array element. Your implementation does as many as 10 comparisons per array element depending on the type.

### Benchmark Results

The benchmark results are below. My implementation is "valid array value 1", your is "valid array value 2":

```
valid array value 1 string
                        97.090k i/100ms
valid array value 2 string
                        44.723k i/100ms
valid array value 1 string
                          1.608M (± 2.3%) i/s -      8.058M
valid array value 2 string
                        624.411k (± 3.5%) i/s -      3.131M

Comparison:
valid array value 1 string:  1608102.0 i/s
valid array value 2 string:   624411.4 i/s - 2.58x slower

valid array value 1 numeric
                        91.511k i/100ms
valid array value 2 numeric
                        15.725k i/100ms
valid array value 1 numeric
                          1.369M (± 1.4%) i/s -      6.863M
valid array value 2 numeric
                        166.807k (± 1.8%) i/s -    849.150k

Comparison:
valid array value 1 numeric:  1368667.9 i/s
valid array value 2 numeric:   166807.2 i/s - 8.21x slower
```

The Benchmark:

```ruby
require 'benchmark/ips'
require 'securerandom'

string_array = Array.new(10) { SecureRandom.hex(8) }
numeric_array = Array.new(10) { rand(10) }

# my suggestion
def valid_array_value1?(value)
  return false unless value.is_a?(Array)
  return true if value.empty?

  case value.first
  when String
    value.all? { |v| v.instance_of?(String) }
  when TrueClass, FalseClass
    value.all? { |v| boolean?(v) }
  when Numeric
    value.all? { |v| v.is_a?(Numeric) }
  else
    false
  end
end

# original code
def homogenous_element_types?(value)
  first_element_type = nil
  value.each do |element|
    element_type = case element
                   when String then String
                   when TrueClass, FalseClass then :boolean
                   when Numeric then Numeric
                   end
    if !first_element_type
      first_element_type = element_type
    elsif element_type != first_element_type
      return false
    end
  end
  true
end

def valid_primitive_value?(value)
  value.instance_of?(String) || value == false || value == true || value.is_a?(Numeric)
end

def valid_array_value2?(value)
  value.is_a?(Array) && value.all? { |elt| valid_primitive_value?(elt) } && homogenous_element_types?(value)
end

Benchmark.ips do |x|
  x.report('valid array value 1 string') do
    valid_array_value1?(string_array)
  end
  x.report('valid array value 2 string') do
    valid_array_value2?(string_array)
  end
  x.compare!
end

Benchmark.ips do |x|
  x.report('valid array value 1 numeric') do
    valid_array_value1?(numeric_array)
  end
  x.report('valid array value 2 numeric') do
    valid_array_value2?(numeric_array)
  end
  x.compare!
end
```

fhwang: @mwear Thanks for digging into this. It definitely seems like your suggested change is faster. I actually think it wasn't just calling `valid_primitive_value?` on all elements, because when I restricted that call to just the first it was still noticeably slower. At any rate I've implemented the change you suggested.

sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -20,10 +20,30 @@ def valid_key?(key)
         key.instance_of?(String)
       end
 
-      def valid_value?(value)
+      def valid_primitive_value?(value)
```
mwear: I don't think it's accurate to refer to any values in ruby as being primitive since all are objects.

fhwang: I introduced the term "primitive" because it seemed like an easy way to add some readability by distinguishing between composite types (arrays) and everything else. Do you think there's a better term for use? I'm just drawing "primitive" from some quick Googling (cf https://en.wikipedia.org/wiki/List_of_data_structures#Data_types) but it's possible there's a better term.

Absent that, I suppose this could just be inlined into `valid_value?` but that seems like a step back in readability.

mwear: I totally understand the intent and my main issue is that it means something else in most other programming languages. Otherwise, it's a really good word for the purpose. I would be fine with inlining as it saves us from having to come up with a better name (the hardest problem in CS). Some suggestions for alternative names, if we want to go that route, would be `...basic_value` or `...simple_value`. I'm not crazy about them, but they'd work.

fbogsany:
I don't feel the refactor of the Jaeger exporter is warranted by the small change required here. It makes this a lot harder to review. If we feel a refactor is necessary (I don't think it is), I'd rather see that in a separate PR.

fhwang:
The extraction of the SpanEncoder allows for two small improvements: Removing the need for `rubocop:disable Metrics/ClassLength` in d7190baa3c328d383daeff9a7d40d7cc82e30f99 and making it easier to add unit tests in 98343a7e6f69d683575c0e78a5ddff478c9eb05a . Do we feel that adding test coverage is not a priority at this time?

I'm happy to do the extraction in a standalone PR first if that would make this easier to review.

fbogsany:
> I'm happy to do the extraction in a standalone PR first if that would make this easier to review.

I'd really appreciate that. I have a short attention span 😄

fhwang:
@mwear @fbogsany Jaeger support for array attributes is now ready for review in this PR.

Issue #170, opened 2020-01-22
BatchSpanProcessor exporter not interrupted test is flaky


mwear:
This test fails intermittently (both in local runs and on CI). We should fix it.

```
  1) Failure:
OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor::export timeout::normally#test_0001_exporter is not interrupted [/home/circleci/project/sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb:227]:
Expected: :not_interrupted
  Actual: :called
```

fbogsany:
Sigh. This has come up for a [related test](https://github.com/open-telemetry/opentelemetry-ruby/pull/153) before and [was anticipated in the original PR](https://github.com/open-telemetry/opentelemetry-ruby/pull/133#discussion_r340676385).

fbogsany:
💩 fix is to adjust the `exporter_sleeps_for_millis`. Better fixes welcome, though.

fhwang:
Looks like this issue was fixed by deleting the test in 031eb4d1271455775de613cbdc6a10b6b647aa01?

PR #187, opened 2020-02-18
Add `Net:HTTP` instrumentation adapter

grantbdev:
# Overview

Add `Net::HTTP` instrumentation adapter similar to the other HTTP client adapters.

## Related
* Effort toward resolving #67

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -69,6 +69,12 @@ GEM_INFO = {
       OpenTelemetry::Adapters::Faraday::VERSION
     }
   },
+  "opentelemetry-adapters-net_http" => {
+    version_getter: ->() {
+      require './lib/opentelemetry/adapters/net/http/version.rb'
+      OpenTelemetry::Adapters::Net::HTTP::VERSION
+    }
+  },
   "opentelemetry-adapters-redis" => {
     version_getter: ->() {
       require './lib/opentelemetry/adapters/redis/version.rb'
--- a/adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
+++ b/adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
@@ -1,0 +1,62 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Net
+      module HTTP
+        module Patches
+          # Module to prepend to Net::HTTP for instrumentation
+          module Instrumentation
+            HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
+            USE_SSL_TO_SCHEME = { false => 'http', true => 'https' }.freeze
+
+            def request(req, body = nil, &block)
+              # Do not trace recursive call for starting the connection
+              return super(req, body, &block) unless started?
+
+              tracer.in_span(
+                HTTP_METHODS_TO_SPAN_NAMES[req.method],
+                attributes: {
+                  'component' => 'http',
+                  'http.method' => req.method,
+                  'http.scheme' => USE_SSL_TO_SCHEME[use_ssl?],
+                  'http.target' => req.path,
+                  'peer.hostname' => @address,
+                  'peer.port' => @port
+                },
+                kind: :client
+              ) do |span|
+                OpenTelemetry.propagation.inject(req)
+
+                super(req, body, &block).tap do |response|
+                  annotate_span_with_response!(span, response)
+                end
+              end
+            end
+
+            private
+
+            def annotate_span_with_response!(span, response)
+              return unless response&.code
+
+              status_code = response.code.to_i
+
+              span.set_attribute('http.status_code', status_code)
+              span.status = OpenTelemetry::Trace::Status.http_to_status(
+                status_code
+              )
+            end
+
+            def tracer
+              Net::HTTP::Adapter.instance.tracer
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,6 +69,14 @@ commands:
             bundle exec appraisal install
             bundle exec appraisal rake test
       - run:
+          name: Bundle + CI (Adapters - Net::HTTP)
+          command: |
+            cd adapters/net_http
+            gem uninstall -aIx bundler
+            gem install --no-document bundler -v '~> 2.0.2'
+            bundle install --jobs=3 --retry=3
+            bundle exec rake test
+      - run:
           name: Bundle + CI (Adapters - Redis)
           command: |
             cd adapters/redis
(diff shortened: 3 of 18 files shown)


adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
```diff
@@ -0,0 +1,58 @@
...
+    module Net
+      module HTTP
+        module Patches
+          # Module to prepend to Net::HTTP for instrumentation
+          module Instrumentation
+            USE_SSL_TO_SCHEME = { false => 'http', true => 'https' }.freeze
+
+            def request(req, body = nil, &block) # rubocop:disable Metrics/MethodLength
+              tracer.in_span(
+                "HTTP #{req.method}",
```
mwear: Given this naming convention, we could cache the span name in a map with method as the key, and  `HTTP {method}` as the value.

grantbdev: @mwear Done. Is there a way you would like to provide for a fallback in case the lookup returns `nil`?

mwear: You could make the cache adaptive: `Hash.new do |h, k| { h[k] = "HTTP #{k}" }`.

grantbdev:
@mwear The reason you're not getting a span with that particular error is because the error occurs before the part of `Net::HTTP` that is instrumented.
The `SocketError` is encountered in the `start` but the instrumentation is hooked into the inner `request` logic:
https://github.com/ruby/ruby/blob/6ecafe0c9865c4187bbf4037fb4e4ac4115e3847/lib/net/http.rb#L481-L485

The same behavior should exist in `dd-trace-rb` since the same methods are being instrumented. If we need to catch this error, we would have to override a higher method. The test suite shows an example of an error that is caught and added to the span because it is in the scope of the instrumentation.

grantbdev:
@mwear It appears that the reason why there are two `Net::HTTP` spans is that the `request` method has a recursive case that is encountered by the way Faraday is using it:
https://github.com/ruby/ruby/blob/6ecafe0c9865c4187bbf4037fb4e4ac4115e3847/lib/net/http.rb#L1468-L1474

We could skip instrumentation on the outer layer if you'd like by guarding creating a span when `started?` is false.

mwear:
Thanks for research @grantbdev, it's much appreciated! 

I think we can continue to patch the request method, with the understanding that some exceptions will  happen before it's reached. We can consider adding more instrumentation around establishing a connection later, if we want to. For now, sticking to what dd-trace does seems like the pragmatic thing to do.

While we're here, I think it makes sense to address the multiple spans resulting from the recursive call to request when a connection is not been started. It seems easy enough to gate span creation on `started?`. Let me know if that's a reasonable ask.

grantbdev:
@mwear Span creation gated. I tried adding a test case for this scenario as well, but I think the patching that WebMock does to `Net::HTTP` prevents the double-span problem from being reachable in the first place since it doesn't replicate the recursion: https://github.com/bblimke/webmock/blob/bf418ec25424ed63899d9196546e05c5ba82ee70/lib/webmock/http_lib_adapters/net_http.rb#L100-L112

Rakefile
```diff
@@ -69,6 +69,12 @@ GEM_INFO = {
       OpenTelemetry::Adapters::Faraday::VERSION
     }
   },
+  "opentelemetry-adapters-net-http" => {
```
mwear: This should be:

```suggestion
  "opentelemetry-adapters-net_http" => {
```

adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
```diff
@@ -0,0 +1,62 @@
...
+  module Adapters
+    module Net
+      module HTTP
+        module Patches
+          # Module to prepend to Net::HTTP for instrumentation
+          module Instrumentation
+            HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
+            USE_SSL_TO_SCHEME = { false => 'http', true => 'https' }.freeze
+
+            def request(req, body = nil, &block) # rubocop:disable Metrics/MethodLength
```
mwear: Rubocop complains that we are unnecessarily disabling `Metrics/MethodLength` when I run tests. Can you confirm this is the case for you, and remove the `rubocop:disable` if so?

grantbdev:
@mwear Feedback addressed

mwear:
Last call for comments. It'd be nice to merge this tomorrow after the SIG meeting.

adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
```diff
@@ -0,0 +1,62 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Net
+      module HTTP
+        module Patches
+          # Module to prepend to Net::HTTP for instrumentation
+          module Instrumentation
+            HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
```
fbogsany: I'm a little concerned about the lazy initialization here. I'd prefer to eagerly init this and freeze the hash. I'm sort of on the fence, so just flagging the concern.

adapters/net_http/lib/opentelemetry/adapters/net/http/patches/instrumentation.rb
```diff
@@ -0,0 +1,62 @@
...
+
+            def annotate_span_with_response!(span, response)
+              return unless response&.code
+
+              status_code = response.code.to_i
+
+              span.set_attribute('http.status_code', status_code)
+              span.status = OpenTelemetry::Trace::Status.http_to_status(
+                status_code
+              )
```
fbogsany: Do we have any semantic conventions around identifying errors based on response codes? Or are we only flagging spans as errors if an exception is raised?

mwear: There are not semantic conventions for errors at all yet. Our `Span#record_error` method uses a set of currently proposed conventions, with the understanding that those could change. It's currently only used for unhandled exceptions. This does bring up a good point though and that is, do we want to synthesize errors based on response codes. If yes, we should probably do this across the board in our HTTP instrumentation adapters. I'll create an issue to discuss and track this.

mwear: https://github.com/open-telemetry/opentelemetry-ruby/issues/198

mwear:
@grantbdev: This failed on JRuby due to a flaky test (#170). I'd like to see this green before merging. Can you rerun the workflow? If that's not a possible a `git commit --amend` ` git push -f` will trigger a new build.

mwear:
Welp... it has merge conflicts now too.

grantbdev:
@mwear PR rebased and build is green.

PR #188, opened 2020-02-19
Add `ethon` instrumentation adapter

grantbdev:
# Overview

Add `ethon` instrumentation adapter based on the adapter for `rest-client` and `ddtrace` code for `ethon`.

While the purpose of this adapter is to instrument HTTP requests in particular, it is my understanding that the protocol could be something else without being able to detect and there are also times where the HTTP method/verb is unknowable.

`ddtrace` also had a parent span for `Ethon::Multi`, but that is currently not provided here because I do not think the those spans could be compliant with the specification.

Related to #67

### DIFF
--- a/adapters/ethon/lib/opentelemetry/adapters/ethon/patches/easy.rb
+++ b/adapters/ethon/lib/opentelemetry/adapters/ethon/patches/easy.rb
@@ -1,0 +1,101 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Ethon
+      module Patches
+        # Ethon::Easy patch for instrumentation
+        module Easy
+          ACTION_NAMES_TO_HTTP_METHODS = Hash.new do |h, k|
+            # #to_s is required because user input could be symbol or string
+            h[k] = k.to_s.upcase
+          end
+          HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
+
+          def http_request(url, action_name, options = {})
+            @otel_method = ACTION_NAMES_TO_HTTP_METHODS[action_name]
+            super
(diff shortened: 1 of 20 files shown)


grantbdev:
@mwear In simple usage, the HTTP method is not set in the code. Using `N/A` was the fallback used in `dd-trace-rb`. Technically, we are already inferring that it is being used for an HTTP request which may not always be the case since it supports other protocols as well. We could go all-in with the HTTP assumption and say that if the method is specified, then it is a `GET` request.

PR updated to add the instrumentation to `Ethon::Multi` to make the easy spans work in the multi situation. I did not port over the parent multi spans though, because I don't think they would make sense or be compliant with the HTTP data semantic conventions.

adapters/ethon/lib/opentelemetry/adapters/ethon/patches/easy.rb
```diff
@@ -0,0 +1,98 @@
...
+  module Adapters
+    module Ethon
+      module Patches
+        # Ethon::Easy patch for instrumentation
+        module Easy
+          HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
+
+          def http_request(url, action_name, options = {})
+            # #to_s is required because user input could be symbol or string
+            @otel_method = action_name.to_s.upcase
```
mwear: Could we avoid upcasing here if we change the cache like so: 

```
HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k.to_s.upcase}" }
```

grantbdev:
@mwear Feedback addressed. I couldn't adopt your suggestion directly since then the HTTP methods wouldn't be normalized for the span attribute, but there is now a caching hash for the method names that feeds into the caching hash for the span names.

mwear:
@grantbdev Any chance you can rebase so we can merge this?

grantbdev:
@mwear PR rebased

mwear:
Thanks @grantbdev!

PR #190, opened 2020-02-19
Named tracer/meter updates

mwear:
The original TracerFactory and MeterFactory names were wildly unpopular. The spec now calls them TracerProvider and MeterProvider. See: https://github.com/open-telemetry/opentelemetry-specification/pull/449. The PR updates our codebase to use the new names.


elskwid:
Nice! 🎉

PR #192, opened 2020-02-25
Fix ::Rack::Utils namespace resolution

duonoid:
# Overview

Fixes `Rack::Utils` resolution is conflicted when using rack instrumentation, since it includes Rack::Utils as well.

Problem is mentioned here: https://github.com/open-telemetry/opentelemetry-ruby/pull/166#pullrequestreview-363831738


mwear:
@duonoid This is out of date with master and it won't let me merge it in. Can you rebase?

PR #200, opened 2020-02-27
Add `concurrent-ruby` instrumentation adapter

grantbdev:
# Overview
Add `concurrent-ruby` instrumentation adapter based on the related `ddtrace` code.

As in `dd-trace-rb`, this instrumentation does not generate spans. Instead, it ensures that context propagation works across new threads when using `concurrent-ruby`.

Related to #67


adapters/concurrent-ruby/example/concurrent-ruby.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'rubygems'
+require 'bundler/setup'
+
+Bundler.require
+
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Adapters::ConcurrentRuby'
+end
+
+tracer = OpenTelemetry::Adapters::ConcurrentRuby::Adapter.instance.tracer
```
mwear: This might be present in the other PRs, but really, you should be getting a named tracer from the OpenTelemetry.tracer_provider for tracing application code. Something like:

```
OpenTelemetry.tracer_provider.tracer('default')
```

What you have works, but we should set a good example in our examples 🙂 .

mwear:
@grantbdev Any chance you can rebase so we can merge this?

grantbdev:
@mwear Feedback addressed on the example and PR rebased

mwear:
This has a conflict now @grantbdev.

grantbdev:
@mwear Conflicts fixed, rebased on latest.

mwear:
Thanks!

Issue #201, opened 2020-02-29
Span vs Context


mwear:
## Span vs Context

The context propagation changes from OTEP 66 have lead to a few wrinkles in our APIs that need be addressed. `Tracer#start_span` returns a span. Often times this is what you want, but sometimes you really want that span in a `Context` object. On the other hand, sometimes you have a `Context` object and you want the current span from it.

### What should Tracer#start_span return?

We could return a context instead a span, but I'm not sure that would improve things. It's worth considering though. Go has take the approach of returning both a context and span from `start_span`. Multiple return values would be possible, but not idiomatic in Ruby, and sadly would cost an additional array allocation. I suppose we can still consider that as an option.

#### I have a span, but I want it in a context

An option is: 

```
Tracer#in_span(span) { ctx = Context.current }
```

Is this good enough or should we add more convienence methods around this?

#### I have a context, but I want a span

Currently you have to do this:

```
Context.current[OpenTelemetry::Trace::Propagation::ContextKeys.current_span_key]
```

### Proposal ContextUtils

While the above options work I think we can improve this situation. The OpenTelemetry::Context object is meant to be a general purpose context that could possibly be extracted out from OpenTelemetry and standalone as its own package. As such, adding a convenience methods to the Context directly should be avoided, but we can introduce a `OpenTelemetry::ContextUtils` class. Here's how I would see that working.

```
# read the current span from a given context 
ContextUtils.span_from(context)

# read the span context from a given context 
ContextUtils.span_context_from(context)

# return a new context with a given span and parent context
ContextUtils.set_span(span, parent: Context.current)

# return a new context with a given span context and parent context
ContextUtils.set_span(span, parent: Context.current)
```

While something like https://github.com/open-telemetry/oteps/issues/68 could be a better long term solution, this will improve some difficulties that we're running into with the current API.

fbogsany:
I think this sort of thing should be the responsibility of objects in the Trace API rather than sticking them in a separate utils module. Having seen example uses in #202, the code doesn't seem any cleaner to me.

mwear:
Related: https://github.com/open-telemetry/opentelemetry-specification/issues/455

Oberon00:
If you are interested in this still, please take a look at https://github.com/open-telemetry/opentelemetry-specification/pull/875

fbogsany:
I think this issue can be closed @mwear.

PR #202, opened 2020-02-29
ContextUtils

mwear:
See #201 for discussion and rationale. I made that as a separate issue as I think there could be a larger discussion. 

This PR contains an implementation of the ContextUtils proposal to add convenience methods to operate on a context and also applies in the codebase where applicable.

### DIFF
--- a/api/lib/opentelemetry/context_utils.rb
+++ b/api/lib/opentelemetry/context_utils.rb
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # The ContextUtils module contains convenience methods for inserting and
+  # data stored in a Context object.
+  module ContextUtils
+    extend self
+
+    # Returns the Span from given context, if one exists, nil otherwise.
+    #
+    # @param [Context] context The context to read the current span from
+    # @return [Span, nil]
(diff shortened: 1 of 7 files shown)


mwear:
I think #108 could benefit from some of the convenience methods introduced here.

api/lib/opentelemetry/context_utils.rb
```diff
@@ -0,0 +1,61 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # The ContextUtils module contains convenience methods for inserting and
```
fbogsany: Missing a word in the comment here? `inserting and ...`?

fbogsany:
So far, this doesn't feel like much of an improvement to me, probably because the methods are stuck in the `ContextUtils` class. There is probably a more convenient place for them.

Higher level concern: it feels like we've exposed a lot more of the internals of the API than I recall, especially the ability to set and get the current span and extracted span context. This is more of a "spidey sense" thing than anything concrete, but it does feel like the API is becoming a lot more complex (both in implementation and interface) due to various indirections.

mwear:
Good points. The convenience methods for working with a span context are only useful if you're writing a propagator. I think we need some help around reading a span from a context, or returning a new context with an updated span, but those probably have a better place. We can discuss where that might be.

I'll close this and we can discuss in the related issue.

PR #206, opened 2020-03-05
Extract Jaeger::Exporter::SpanEncoder class.

fhwang:
This PR is a standalone refactor to reduce the size of #185.


exporters/jaeger/lib/opentelemetry/exporters/jaeger/exporter.rb
```diff
@@ -79,43 +80,9 @@ def encoded_batches(span_data)
...
-          links&.map do |link|
-            Thrift::SpanRef.new(
-              'refType' => Thrift::SpanRefType::CHILD_OF,
-              'traceIdLow' => int64(link.context.trace_id[16, 16]),
-              'traceIdHigh' => int64(link.context.trace_id[0, 16]),
-              'spanId' => int64(link.context.span_id)
-            )
-          end
+        def encoded_span(span_data)
+          @span_encoder ||= SpanEncoder.new
```
fbogsany: I don't think there's any value in lazily initializing the encoder. I'd just initialize it in `initialize`. At that point, you might also want to inline this method at its only callsite.

fhwang: I suppose there's the miniscule optimization of not initializing the encoder if no spans are ever sent, but that's pretty minimal. I'll change it now.

Issue #151, opened 2019-11-13
Sidekiq integration


danielmbarlow:
We are using OpenCensus to do this and have a [PR open on that project](https://github.com/census-instrumentation/opencensus-ruby/pull/102).

Are there plans to implement a Sidekiq integration on OpenTelemetry?

elskwid:
Hi @danielmbarlow!

Yes, we absolutely have plans to implement a Sidekiq integration (we're calling them adapters for now 😄). These are part of #67 for now. We've started with some simple adapters: Faraday in #148 and Sinatra in #149. The goal is to come up with a blueprint for how to write adapters for this library. We'd love your feedback, ideas, and contributions.

Are you interested in working on the Sidekiq adapter given that you've worked on the same for Opencensus?

danielmbarlow:
> We've started with some simple adapters
I'll watch those PRs to see the kind of approach you're taking, thanks.

> Are you interested in working on the Sidekiq adapter 
Quite likely. I'll have to get approval from my project manager first, though. Will be in touch.

danielmbarlow:
I haven't been given approval to work on this so far. We put a lot of time into developing the OpenCensus sidekiq integration, and I don't think I'll get a green light any time soon to work on it's replacement.

fbogsany:
I'll take a look at the OpenCensus Sidekiq integration. We will probably start work on this next week.

fbogsany:
@robertlaurin can you please comment on this issue so I can assign it to you?

robertlaurin:
> 
> 
> @robertlaurin can you please comment on this issue so I can assign it to you?

👋

mwear:
I'm closing this because we have sidekiq instrumentation from #212. If we need any improvements or changes, we can track those separately.

PR #207, opened 2020-03-11
SDK::Trace::Span defers attr validation until finish is called.

fhwang:

Issue #10, opened 2019-06-04
SDK: supported versions of Ruby


fbogsany:
What versions/implementations of Ruby should we support?
* MRI 2.5.0 and newer? In general, N-1?
* JRuby?
* Graal/Truffle?

indrekj:
In MRI case probably same as in https://www.ruby-lang.org/en/downloads/branches/ `normal maintenance`, maybe `security maintenance` as well. Don't really see a point supporting versions that are not supported by the ruby-core team.

fbogsany:
+1 for `normal maintenance` for MRI

luong-komorebi:
+1 for security maintenance for MRI. For JRuby and Grall, let's consider them later since they already have the capability of running java code from Ruby 
https://github.com/oracle/truffleruby/blob/master/doc/user/jruby-migration.md#using-java-from-ruby
https://github.com/jruby/jruby/wiki/CallingJavaFromJRuby

fbogsany:
Shopify has a need to support both JRuby and TruffleRuby. The former is already targeted by CI. I’ll add TruffleRuby as well.

fbogsany:
We've enabled CI for JRuby, TruffleRuby, and MRI 2.4 - 2.6. This is currently run for Shopify PRs. Running for everyone is gated on #77. Closing this, though, since we've covered all the bases for now.

duaneking:
With respect,  this is a bad policy that actively harms the ruby and open telemetry ecosystems, and makes it massively harder to do the due diligence required for updates and maintenance of legacy projects that you are explicating saying here you want to support upgrades to.

PR #212, opened 2020-03-25
Instrument Sidekiq gem

robertlaurin:
Work in progress pull request for a [Sidekiq gem](https://github.com/mperham/sidekiq) adapter.

Original issue here https://github.com/open-telemetry/opentelemetry-ruby/issues/151#issuecomment-600379132


adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/server/tracer_middleware.rb
```diff
@@ -0,0 +1,40 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Sidekiq
+      module Middlewares
+        module Server
+          class TracerMiddleware
+            def call(_worker, msg, _queue)
+              parent_context = OpenTelemetry.propagation.extract(msg, extractors: OpenTelemetry.propagation.job_extractors)
```
robertlaurin: Hey @mwear I was wondering if you could provide a bit of guidance on how to the existing code base to support job extractors and injectors here.  I just kind of shoehorned it in to get a better understanding of the codepath, doesn't seem right though.

I'm still getting familiar with the code base so sorry for the rough state of this PR.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -0,0 +1,37 @@
...
+  module Adapters
+    module Sidekiq
+      module Middlewares
+        module Client
+          class TracerMiddleware
+            def call(worker_class, job, _queue, _redis_pool)
+              tracer.in_span(
+                worker_class,
+                attributes: {
+                  jid: job['jid'],
```
fbogsany: Might be nicer to call this `job_id`.

Also, attribute keys need to be strings. We don't support symbols. We don't validate this at the moment, but #207 may add some of that back.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -0,0 +1,37 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Adapters
+    module Sidekiq
+      module Middlewares
+        module Client
+          class TracerMiddleware
+            def call(worker_class, job, _queue, _redis_pool)
```
fbogsany: We should include the queue name as an attribute. https://github.com/open-telemetry/opentelemetry-specification/pull/418 calls this `messaging.destination`.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -0,0 +1,37 @@
...
+              tracer.in_span(
+                worker_class,
```
fbogsany: Is `worker_class` a string? If not, we should convert it here.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -0,0 +1,37 @@
...
+                attributes: {
+                  jid: job['jid'],
+                  created_at: job['created_at'],
```
fbogsany: Technically, this shouldn't be required. Trace analysis should be able to infer it from the parent span (the producer).

fbogsany: Uh, well, that's embarrassing -- this *is* the producer. Do we have the `created_at` at this point? Does it matter? Is the span start time a reasonable substitute?

api/lib/opentelemetry/trace/propagation/job_trace_context_extractor.rb
```diff
@@ -0,0 +1,56 @@
...
+        # Returns a new JobTraceContextExtractor that extracts context using the
+        # specified header keys
+        #
+        # @param [String] traceparent_key The traceparent header key used in the carrier
+        # @param [String] tracestate_key The tracestate header key used in the carrier
+        # @return [JobTraceContextExtractor]
+        def initialize(traceparent_key: 'traceparent',
+                       tracestate_key: 'tracestate')
+          @traceparent_key = traceparent_key
+          @tracestate_key = tracestate_key
```
fbogsany: I wonder if we really need `tracestate` here? It is typically intended as a representation of vendor-specific trace information as traces traverse systems that may be owned by different companies and instrumented in different ways. That doesn't apply here.

Arguing against myself: the jobs may make HTTP requests to external services that should form part of the same trace, and if we had: A -http-> B -sidekiq-> B -http-> A, then the `tracestate` needs to be propagated back to A.

Thoughts @mwear ?

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -129,6 +133,19 @@ def default_http_extractors
           OpenTelemetry::CorrelationContext::Propagation.rack_http_extractor
         ]
       end
+
+      def default_job_injectors
+        [
+          OpenTelemetry::Trace::Propagation.job_trace_context_injector,
+
```
fbogsany: nit: empty line.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
robertlaurin: Good catch, when `sidekiq/testing` is used it comes in as a string which is why I hadn't explicitly `.to_s` the worker_class.  However in a non test environment it is sending in the class constant.

The `job` hash contains job class in string form so I'll pull from there instead.

robertlaurin: When double checking this I noticed that the job hash contains different information when using the [ActiveJob setup](https://github.com/mperham/sidekiq/wiki/Active-Job#active-job-setup)

With ActiveJob setup
```rb
{
  "retry"=>0,
  "queue"=>"default",
  "class"=>"ActiveJob::QueueAdapters::SidekiqAdapter::JobWrapper",
  "wrapped"=>SimpleJob,
  "args"=> [
    {
      "job_class"=>"SimpleJob",
      "job_id"=>"8a34d851-ab9d-484b-ada4-320e8e591911",
      "provider_job_id"=>nil,
      "queue_name"=>"default",
      "priority"=>nil,
      "arguments"=>[],
      "executions"=>0,
      "exception_executions"=>{},
      "locale"=>"en",
      "timezone"=>"UTC",
      "enqueued_at"=>"2020-03-27T16:06:24Z"
    }
  ],
  "jid"=>"fcc6967ada593724d55e15a9",
  "created_at"=>1585325184.135893
}
```

Without ActiveJob setup
```rb
{
  "retry"=>0,
  "queue"=>"default",
  "class"=>"SimpleJob",
  "args"=>[],
  "jid"=>"d9bf4951a65e81456a1c7990",
  "created_at"=>1585325499.9481258
}
```

My question regarding the above is, should it be a configuration option to declare whether or not you are using sidekiq with the active job queue adapter?  

Or should this adapter try and handle capturing the job class implicitly doing something like
```rb
job_class = job['wrapped'] || job['class']
```

Not sure what's more in the spirit of this gem.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
robertlaurin: >Is the span start time a reasonable substitute?

I think it is reasonable, not sure if it's worth splitting hairs here but the span start time would be after the actual `created_at` timestamp.  

Theres another time value that could provide some value, which is the `enqueued_at` time.  It's available in the server middleware, but also on client middleware when using the sidekiq activejob adapter.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/server/tracer_middleware.rb
mwear: :wave: @robertlaurin, sorry for not getting back to you sooner, but work done in #214 makes the answer to this question somewhat easier. Instrumentation adapters should use injectors and extractors registered on the top level propagation object, `OpenTelemetry.propagation`. I did some reorganization to that module in #214. Previously it had `http_injectors` and `http_extractors`. Despite their (previous) name, they will work for many non-http text scenarios. In order to make this as user friendly as possible, I added text injectors / extractors to the propagation object.

With these changes, you should not need special job injectors or extractors. In #214 the top level propagation object looks like this:

```
#inject methods take an optional setter callable
OpenTelemetry.propagation.http.inject(carrier, context)
OpenTelemetry.propagation.text.inject(carrier, context)

# extract methods take an optional getter callable
OpenTelemetry.propagation.http.extract(carrier, context)
OpenTelemetry.propagation.text.extract(carrier, context)
```

For Sidekiq you can use `OpenTelemetry.propagation.text`. We should probably consult @fbogsany to see what he thinks of the changes in #214 though.

api/lib/opentelemetry/trace/propagation/job_trace_context_extractor.rb
mwear: @fbogsany I concur with the argument you pose to yourself. If the job makes a request to another a system that depends on `tracestate` it will break the trace if it isn't sent.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
fbogsany: It's nice if we can auto-extract. My recollection is that this is a royal PITA. We have some weirdness in https://github.com/Shopify/shopify-tracing/blob/master/lib/shopify/tracing/adapters/sidekiq_adapter.rb and https://github.com/Shopify/shopify-tracing/blob/master/lib/shopify/tracing/adapters/resque_like_adapter.rb to try to deal with it, and I believe that broke at least one dependent service.

Maybe we can try the auto-extraction, and if we find breakage when people try to use this in production, we can add the config option.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
fbogsany: Hrm... Why not both? Let's add both `created_at` and `enqueued_at`.

robertlaurin:
I signed it

fbogsany:
You need to add an Appraisals file like https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/faraday/Appraisals

adapters/sidekiq/Appraisals
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+appraise 'sidekiq-6.0' do
+  gem 'sidekiq', '~> 6.0'
```
robertlaurin, line 4: I'm not familiar with the best practices around using Appraisal.  What releases should be included for appraisal?

CI is also [failing](https://circleci.com/gh/open-telemetry/opentelemetry-ruby/1717?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-checks-link) on the `6.0` release of sidekiq when testing against ruby2.4 which is to be expected.

>Ruby <2.5 is no longer supported.

https://github.com/mperham/sidekiq/blob/master/6.0-Upgrade.md#upgrade

mwear: I'm not sure that we have established any hard rules about what versions we should test using appraisal. At a minimum, I'd suggest that we test against the major versions in the supported range, and consider testing against the latest minor versions. In general, I think we just want reasonable assurance the adapter works for the versions we expect it to.

fbogsany: [Ruby 2.4 support has ended](https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/). In #10 we decided not to support versions not supported by Ruby Core. We should bump the minimum to 2.5 everywhere, which will fix this problem too. I’ll do that in another PR and then this can be rebased.

mwear: If we run into this again, which is probably likely, we can remove the generated gemfiles from source control, and generate them before each run. That would allow us to put conditionals in the Appraisals file based on version. For example:

```ruby
if RUBY_VERSION >= '2.5.0'
  appraise 'sidekiq-6.0' do
    gem 'sidekiq', '~> 6.0'
  end
end
```

We'd just have to add `bundle exec rake appraisal generate` to the circle ci script.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/adapter.rb
```diff
@@ -0,0 +1,60 @@
...
+        end
+
+        def add_server_middleware
+          ::Sidekiq.configure_server do |config|
+            config.server_middleware do |chain|
+              chain.add Middlewares::Server::TracerMiddleware
+            end
+          end
+
+          ::Sidekiq.configure_client do |config|
```
mwear, line 45: It looks like we add Client::TracerMiddleware twice. Once here and once on line 31. Is that intentional?

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -0,0 +1,38 @@
...
+  module Adapters
+    module Sidekiq
+      module Middlewares
+        module Client
+          class TracerMiddleware
+            def call(_worker_class, job, _queue, _redis_pool)
+              tracer.in_span(
+                job['wrapped']&.to_s || job['class'],
+                attributes: {
+                  'job_id' => job['jid'],
```
mwear, line 17: I'm not exactly sure what we should do about attributes that don't fall under a semantic convention. I think we should definitely record this, but I'm wondering if it should have a prefix of some kind. I'm also wondering if this is a common enough attribute that it deserves to have a convention defined. To be clear, I'm not exactly asking for anything here, just kind of thinking out loud on this.

fbogsany: Should this be `messaging.message_id`? From https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/adapter.rb
fbogsany: Adding it twice is intentional. See https://github.com/mperham/sidekiq/wiki/Middleware#client-middleware-registered-in-both-places. It is done incorrectly here, though. The `config.client_middleware` below should be in the `configure_server` block above.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/server/tracer_middleware.rb
```diff
@@ -0,0 +1,40 @@
...
+          class TracerMiddleware
+            def call(_worker, msg, _queue)
+              parent_context = OpenTelemetry.propagation.text.extract(msg)
+              tracer.in_span(
+                msg['wrapped']&.to_s || msg['class'],
+                attributes: {
+                  'job_id' => msg['jid'],
+                  'messaging.destination' => msg['queue'],
+                  'created_at' => msg['created_at'],
+                  'enqueued_at' => msg['enqueued_at'],
```
fbogsany, line 21: I feel like these are better recorded as events rather than tags (especially since attribute values can’t be time stamps). https://github.com/open-telemetry/opentelemetry-ruby/blob/cd4b44daa4453e78eb4e2c2ced80ee84f680eb65/api/lib/opentelemetry/trace/span.rb#L88

E.g.:
```ruby
span.add_event(name: 'enqueued_at', timestamp: msg['enqueued_at'])
```

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
robertlaurin: I think `messaging.message_id` fits here, going to use that.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/adapter.rb
robertlaurin: Thanks for catching that, it looks like `Sidekiq::Testing.inline!` was hiding the problem a bit here as it was still producing the spans in the way I would expect them to (for the most part).  

I've since fixed this (and the tests) as well as tested it against a simple rails app to confirm it's behaving the way it should be.

adapters/sidekiq/Appraisals
fbogsany: https://github.com/open-telemetry/opentelemetry-ruby/pull/226 removes Ruby 2.4 support.

adapters/sidekiq/lib/opentelemetry/adapters/sidekiq/middlewares/client/tracer_middleware.rb
mwear: Of the already specified attributes, it's the closest match, but I'm not sure that it's semantically the same. I believe this field is supposed to represent the unique id assigned to a message as part of the underlying messaging protocol. Because background jobs involve operations such as consume, and produce, and listen on queues they intersect with messaging concepts, but aren't a perfect fit. For sidekiq, the underlying messaging system is Redis, which again, falls weirdly in between a datastore and a message broker. I think we can go with `messaging.message_id` for now, but perhaps have a discussion about conventions for background jobs more generally. I'll try to find out what other SIGs are doing for background jobs. There are similar systems, such as Celery for Python, for example.

mwear:
I'm guessing this will go green with #226.

PR #214, opened 2020-03-29
Context prop updates

mwear:
### Context prop updates
This PR is an effort towards bridging the gap between OTEP-66 and the resulting spec work. 

### Propagators as composable injectors / extractors

OTEP-66 divided propagators into injectors and extractors, but the resulting spec work did not keep that division. Instead it specifies that there should be a mechanism to register a global propagator, and that there should be a composite propagator option to allow propagation of multiple formats.

Dividing propagators into injectors and extractors has some benefits. One, is that you can injector fewer or more types of context than you wish to extract. The format for injection and extraction can vary due to an intermediary, such as Rack. There are also situations where ordering of injection and extraction matters, and being able to provide a collections of injectors and extractors allows propagation to preserve that ordering. There are cases where you may want a format to be the last to inject, but the first to extract, for example.

In order to keep the benefits of separate injectors and extractors, but comply with the propagator interface that is specified, this PR introduces `Propagator` and `CompositePropagator` classes that compose separate injectors and extractors into an object that exposes inject and extract methods that delegate to the underlying injectors and extractors. cc: @dyladan 

### More formats supported by OpenTelemetry.propagation

Before this PR the top-level propagation object exposed inject and extract methods directly and it had accessors to set the global http injectors and extractors. For example,

```ruby
OpenTelemetry.http_injectors = [...]
OpenTelemetry.http_extractors = [....]
OpenTelemetry.propagation.inject(carrier, ctx)
OpenTelemetry.propagation.extract(carrier, ctx)
```

The resulting spec work did not keep this structure and instead recommended having a single global propagator. As a result of work in #212 and other SIG discussions it's become clear that we have a need for both text and http formats. To that end, the top level propagation object exposes accessors for http and text propagators.

```ruby
OpenTelemetry.propagation.http = CompositePropagator.new(injectors, extractors)
OpenTelemetry.propagation.text = CompositePropagator.new(injectors, extractors)

OpenTelemetry.propagation.http.inject(carrier, context)
OpenTelemetry.propagation.text.inject(carrier, context)
...
```

Before this PR the various injectors and extractors were named after, or had HTTP implications. They've been refactored to be more agnostic and generally use the `Text` prefix instead.

cc: @fbogsany


mwear:
A couple of changes is this PR were the namespace for TraceContext, as well as the names for injectors and extractors. While we we're here, I'm wondering if we like what we have, or if we should consider changing things further. 

This is what we currently have:

`OpenTelemetry::CorrelationContext::Propagation`
  - `TextInjector`
  - `TextExtractor`

`OpenTelemetry::Trace::Propagation::TraceContext`
  - `TextInjector`
  - `TextExtractor`

Would it be better if we put both under `OpenTelemetry::Context::Propagation` and possibly refer to the injectors / extractors as HttpTextInjector / Extractor. 

`OpenTelemetry::Context::Propagation::CorrelationContext`
  - `HttpTextInjector`
  - `HttpTextExtractor`

`OpenTelemetry::Context::Propagation::TraceContext`
  - `HttpTextInjector`
  - `HttpTextExtractor`

We could still keep two formats (http and text) on the top level propagation object. The `HttpText` prefix indicates the injectors and extractors are usable for both purposes. The main point of differentiation is the choice of key names / formats.

adapters/ethon/test/opentelemetry/adapters/ethon/adapter_test.rb
```diff
@@ -17,18 +17,20 @@
...
-    OpenTelemetry.propagation.http_injectors = [
-      OpenTelemetry::Trace::Propagation.http_trace_context_injector
-    ]
+    # this is currently a noop but this will future proof the test
+    @orig_propagator = OpenTelemetry.propagation.http
+    propagator = OpenTelemetry::Context::Propagation::Propagator.new(
+      OpenTelemetry::Trace::Propagation::TraceContext.text_injector,
+      OpenTelemetry::Trace::Propagation::TraceContext.text_extractor
+    )
+    OpenTelemetry.propagation.http = propagator
```
fbogsany: We seem to be using this pattern in a lot of tests. Can we extract it to test helpers?

fbogsany:
:-1: I'm not keen on the `HttpText` prefix.

I don't think the injectors and extractors should be reparented under the `OpenTelemetry::Context::Propagation` module. That violates the principle of layering trace and correlation context propagation on top of the base context propagation component.

adapters/ethon/test/opentelemetry/adapters/ethon/adapter_test.rb
mwear: The tests are in separate gems. There is a need to share utilities between the different instrumentation packages, but we don't have a good place them yet. I could stick this in the test_helper for the API gem, and require it here, but that doesn't seem like the best option. I'm wondering if we should punt on this until we have a common place to put it.

PR #221, opened 2020-04-04
more gemspec fixes

mwear:
Dashes, underscores, and combinations thereof are fun.


mwear:
Thanks @elskwid! Finallly, all of the adapters have been built and released 🎉.

PR #223, opened 2020-04-09
opentelemetry-adapters-all gem

mwear:
This gem is an all-in-one distribution of all of the officially supported instrumentation adapters. This will likely be the go-to method that users will use to add instrumentation dependencies to their application. With this package users will only need the following to get off the ground with opentelemetry.

Gemfile

```ruby
source `https://rubygems.org`

gem `opentelemetry-sdk`
gem `opentelemetry-adapters-all`
```
App
```ruby
  c.use_all
end

# ...
```

### DIFF
--- a/adapters/all/README.md
+++ b/adapters/all/README.md
@@ -1,6 +1,6 @@
 # opentelemetry-adapters-all
 
-The `opentelemetry-adapters-all` gem is an all-in-one distribution of community maintained instrumentation adapters. Instrumentation adapters are packaged as individual gems for flexibilty and maintainability. Instead of having to require each adapter individually, applications can depend on this all-in-one gem as a convenient alternative.
+The `opentelemetry-adapters-all` gem is an all-in-one distribution of community maintained instrumentation adapters. Instrumentation adapters are packaged as individual gems for flexibility and maintainability. Instead of having to require each adapter individually, applications can depend on this all-in-one gem as a convenient alternative.
 
 ## What is OpenTelemetry?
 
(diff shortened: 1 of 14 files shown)


adapters/all/README.md
```diff
@@ -0,0 +1,84 @@
+# opentelemetry-adapters-all
+
+The `opentelemetry-adapters-all` gem is an all-in-one distribution of community maintained instrumentation adapters. Instrumentation adapters are packaged as individual gems for flexibilty and maintainability. Instead of having to require each adapter individually, applications can depend on this all-in-one gem as a convenient alternative.
```
fbogsany, line 3: s/flexibilty/flexibility/

adapters/all/README.md
```diff
@@ -0,0 +1,84 @@
...
+Install the gem using:
+
+```
+gem install opentelemetry-adapters-all
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-adapters-all` in your `Gemfile`.
+
+
+The `opentelemetry-api` has functionality to discover the instrumentation adapters that an application depends on. It maintains an registry of discovered adapters that SDKs can use to automatically install the instrumentation for you. These instructions pertain to the offical `opentelemetry-sdk` implementation. Consult the documentation for your SDK if you are using an alternative implementation.
```
fbogsany, line 26: s/an registry/a registry/

adapters/all/opentelemetry-adapters-faraday.gemspec
```diff
@@ -0,0 +1,47 @@
...
+
+  spec.add_dependency 'opentelemetry-adapters-concurrent_ruby', version
+  spec.add_dependency 'opentelemetry-adapters-ethon', version
+  spec.add_dependency 'opentelemetry-adapters-excon', version
+  spec.add_dependency 'opentelemetry-adapters-faraday', version
+  spec.add_dependency 'opentelemetry-adapters-net_http', version
+  spec.add_dependency 'opentelemetry-adapters-rack', version
+  spec.add_dependency 'opentelemetry-adapters-redis', version
+  spec.add_dependency 'opentelemetry-adapters-restclient', version
+  spec.add_dependency 'opentelemetry-adapters-sinatra', version
```
fbogsany, line 38: This is going to require the versions of all adapters to be kept in sync. This is fine for now, but probably won’t be in the future. It might be useful to open an issue to remind us of this.

Likewise, it’d be helpful to document this for release management, so we remember to bump versions and release all the adapters as part of the general release process. If we *want* to coordinate versions of all the gems, we should automate the version bumps, but I don’t think that’s the right thing to do in the long term.

mwear: Right now I am keeping to strict versioning to allow for changes in the API. As things stabilize, I agree, we'll want to relax this policy. I'll update the release docs with the process for releasing this gem and get an issue started to note and track expected changes to it.

mwear:
@fbogsany I went ahead preemptively removed the version synchronization. It seems like it was probably overkill and not completely necessary. I did keep a note about publishing the gem, as it is dependent on its dependencies being published first.

PR #228, opened 2020-04-16
Add license for sidekiq

mwear:
Sidekiq needs a license so we can build the gem ;)


mwear:
Built and released 🔥: https://rubygems.org/gems/opentelemetry-adapters-sidekiq

Issue #96, opened 2019-09-16
Code examples: gRPC instrumentation example


fbogsany:

fbogsany:
@mwear @dazuma no-one is currently working on this. Do we need it for the Alpha 0.2 release, or should we punt it to Alpha 0.3?

mwear:
I don't think this should hold up our v0.2 alpha. I'm in favor of punting to Alpha 0.3.

fbogsany:
@pinglamb can you please comment on this issue so that I can assign it to you?

pinglamb:
@fbogsany 

I am working on this in https://github.com/open-telemetry/opentelemetry-ruby/pull/229

PR #229, opened 2020-04-22
Added GRPC example

pinglamb:
Hi, I have added a basic GRPC example (#96).

I took reference of the Java and Go example and used `GRPC::ServerInterceptor` for injecting the trace logic. Other parts are mostly copied from Ruby GRPC example (with some modifications), including the `.proto` file and server/client implementation.

I have 2 parts I am not sure about, which would like to seek some suggestions/guidance:

* How the parent context should be extracted? I saw in the `http` example it uses `OpenTelemetry.propagation.http.extract(env)`, should I add `grpc` util or just extract it from `call.metadata` in the Interceptor. Also, what metadata should I extract from there?
*  Should I assign the `grpc status` to `span status` when error happens?

I found that `protobuf` seems to have issue with Ruby 2.7 (https://github.com/protocolbuffers/protobuf/issues/7070). I will test further see if it is my configuration or `protobuf` issue.

Any feedback is welcome, it is my first time contribution. Thank you very much.


fbogsany:
Thanks for working on this!

The protobuf issue with Ruby 2.7 has been a blocker for us (Shopify) too. I’ll check internally to see if and how we’re working around it.

Assigning `grpc status` to `span status` is required: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#status

The semantic conventions for gRPC are documented here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/rpc.md#grpc

The specification doesn’t describe how propagation is supposed to work with gRPC, but the [Go implementation](https://github.com/open-telemetry/opentelemetry-go/blob/master/plugin/grpctrace/grpctrace.go) simply sets the keys in the metadata. I *think* we want to use the `text` propagators rather than `http` for that, since the latter (for Ruby) expects Rack formatted keys. @mwear can definitely correct me on that.

pinglamb:
Thanks @fbogsany for pointing out the spec.

For the span name, as the `grpc` lib will `underscore` the rpc method name (https://github.com/grpc/grpc/blob/v1.28.1/src/ruby/lib/grpc/generic/service.rb#L98), we have to `camelize` it back.

For the `peer` attributes, we only have a string containing both the ip address and port so I added a regex to parse it.

pinglamb:
I also assigned the `context` to be `OpenTelemetry.propagation.text.extract(call.metadata)`, not sure if I am doing it right.

pinglamb:
I have studied the Go implementation again and did some adjustments:

* `opentelemetry-go` made the `grpctrace` a plugin (https://github.com/open-telemetry/opentelemetry-go/tree/master/plugin/grpctrace) which contains the server/client interceptors for different scenarios. I copied the idea and the interceptors are grouped under a single module `GRPCTrace` under the example `lib` folder which can potentially be extracted out in the future
* Understand better about the `context` propagation, used the `OpenTelemetry.propagation.http.inject` and `OpenTelemetry.propagation.http.extract` for the propagation in the interceptors (referencing Go implementation again)

Two hacks are needed so far because the `grpc` interceptors does not have enough info exposed (it is still an experimental API) for building the span:

* For client interceptor, it can't access the `peer` info from the `call` so `instance_variable_get` needs to be used to access underlying data
* For server interceptor, the RPC method name is lost (which is available for client interceptor). Right now, I try to reconstruct it from the actual `method` object

I will continue to explore for `StreamServer` and `StreamClient`

fbogsany:
@pinglamb are you still working on this?

pingnamo:
I can continue to work on this, any comments so far? Thanks!

pinglamb:
I can continue to work on this, any comments so far? Thanks!

Asafb26:
@pinglamb have any progress with this one?

kaylareopelle:
Hi @pinglamb! 👋  We've moved code related to instrumentation into a different repo: [open-telemetry/opentelemetry-ruby-config](https://github.com/open-telemetry/opentelemetry-ruby-contrib)

We appreciate your PR and would be happy to take another look in that context. Thank you!

PR #260, opened 2020-05-12
Fix Sinatra example in docker-compose.yml

highb:
This commit fixes the Sinatra example so that it runs out of the box
like the standard `ex-http` example when called with
`docker-compose run ex-adapter-sinatra`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -66,6 +66,7 @@ services:
 
   ex-adapter-sinatra:
     <<: *base
+    command: ./start_server.sh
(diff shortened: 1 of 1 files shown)


highb:
I'm running it by my boss/legal now. :)

highb:
I signed it. (I think?)

highb:
I signed it

highb:
I've opened a support ticket to get this sorted.

highb:
@fbogsany Looks like the CLA bot is happy now.

PR #261, opened 2020-05-14
Rename excon component attribute from http to excon

dmathieu:
The rack middleware uses the same component value. So excon and rack
requests are currently impossible to differenciate.

The value `excon` seems like the most explicit one to me. I'm open to any alternative though.


mwear:
Technically, we are supposed to remove the component attribute from this (and all other) instrumentation. See: https://github.com/open-telemetry/opentelemetry-specification/pull/447. We should replace this with the InstrumentationLibrary (see: https://github.com/open-telemetry/oteps/pull/84) and make that available on SpanData for export.

I will try to write some issues up about this shortly.

dmathieu:
That definitely makes sense, and feels much better in the end.
I've opened #264 and #265. I suppose we can continue there.

Issue #236, opened 2020-05-11
Add documentation for Excon


mwear:
Add documentation for Excon instrumentation adapter. The implementation exists [here](https://github.com/open-telemetry/opentelemetry-ruby/tree/master/adapters/excon), but we should have a readme describing its behavior and options.

mwear:
Fixed by #262

PR #262, opened 2020-05-14
Add excon readme

dmathieu:
This adds documentation for the excon adapter, closing #236.

### DIFF
--- a/adapters/excon/README.md
+++ b/adapters/excon/README.md
@@ -1,0 +1,50 @@
+# Excon Adapter
+
+The excon adapter is a community-maintained instrumentation adapter for the [excon][excon-home] HTTP library.
+
+## How do I get started?
+
+Install the gem using:
+
+```
+gem install opentelemetry-adapters-excon
+```
+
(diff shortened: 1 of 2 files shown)


adapters/excon/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+```
+gem install opentelemetry-adapters-excon
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-adapters-excon` to your `Gemfile`.
+
+## Usage
+
+To install the instrumentation adapter, use `call` with the name of the instrumentation.
```
mwear, line 17: I think you mean to say `use` and not `call`. Also, it's worth pointing out that there are two ways that you can "install" instrumentation. One is using a selective install, where you explicitly list the instrumentation, and the other is using installing all available instrumentation. It would look as follows:

```ruby
  c.use_all
end
```

dmathieu:
Thank you. I've added a mention of `use_all`.
I didn't setup two subsections, as those are quite short and that seemed superfluous. I'm happy to do so anyway though.

dmathieu:
The test failure seems to be due to a flaky test.

Issue #230, opened 2020-04-23
resource autodetection


mwear:
It would be nice to have resource auto-detection, similar to what exists in opentelemetry-js. The logic was largely ported from OpenCensus Node: 

https://github.com/open-telemetry/opentelemetry-js/pull/899

There is little to no specification for it at this time, there is a related issue:
https://github.com/open-telemetry/opentelemetry-specification/issues/535

fbogsany:
@robertlaurin can you please comment here so I can assign this one to you? 😄

robertlaurin:
Hey 👋, I'm going to work on this :)

fbogsany:
I referenced the OpenTelemetry Collector's implementation of resource auto-discovery on today's SIG call. Turns out it's a little more Kubernetes-specific (and much more complicated) than I thought: https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/master/processor/k8sprocessor. I'm not sure how much we can usefully extract from it.

fbogsany:
Do we consider this fixed by #263 or do we want to leave this open to add resource detection for platforms other than GCP?

fbogsany:
From last week's call, we decided to close this with #263 and open new issues for platforms other than GCP.

PR #263, opened 2020-05-14
Add support for automated resource detection

robertlaurin:
This adds support for populating a resource with with Telemetry data
as well as Google Cloud Platform environment metata data.

For #230

Example usage
```rb
  c.resource = OpenTelemetry::SDK::Resources::AutoDetector.detect
end
```


sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
```diff
@@ -0,0 +1,52 @@
...
+        module GoogleCloudPlatform
+          extend self
+
+          def detect
+            gcp_env = Google::Cloud::Env.new
+            return Resource.create unless gcp_env.compute_engine?
+
+            resource_labels = {}
+            resource_labels[CLOUD_RESOURCE[:provider]] = 'gcp'
+            resource_labels[CLOUD_RESOURCE[:account_id]] = gcp_env.project_id || ''
```
robertlaurin, line 22: This deserves a second set of eyes I'm not totally sure this is the correct value being set here, I was following the work done on the JS library https://github.com/open-telemetry/opentelemetry-js/pull/899/files#diff-2ae8d53e5ba7424080459738d6bc210bR45

From the spec 
>The cloud account id used to identify different entities.

But this is using `project-id` from https://cloud.google.com/compute/docs/storing-retrieving-metadata#default

sdk/lib/opentelemetry/sdk/correlation_context.rb
```diff
@@ -9,7 +9,7 @@
 
 module OpenTelemetry
   module SDK
-    # Contains operational implementataions of the CorrelationContext::Manager
+    # Contains operational implementations of the CorrelationContext::Manager
```
mwear, line 12: nice 🙏

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
```diff
@@ -0,0 +1,31 @@
...
+        extend self
+
+        DETECTORS = [
+          OpenTelemetry::SDK::Resources::Detectors::GoogleCloudPlatform,
+          OpenTelemetry::SDK::Resources::Detectors::Telemetry
+        ].freeze
+
+        def detect
+          resources = DETECTORS.map(&:detect)
+          resources.reduce(OpenTelemetry::SDK::Resources::Resource.create) do |empty_resource, detected_resource|
```
mwear, line 24: Any chance we could rename `empty_resource`? It is likely non-empty at some point and is really the accumulated resource. I don't have a specific name in mind. Two names I often use, which are completely generic are `memo`, or `acc`.

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
```diff
@@ -0,0 +1,54 @@
...
+              resource_labels[CONTAINER_RESOURCE[:name]] = ENV['CONTAINER_NAME'] || ''
+            end
+
+            Resource.create(resource_labels)
+          end
+
+          private
+
+          def hostname
+            ENV['HOSTNAME'] || `hostname`&.strip
```
mwear, line 46: Any reason not to use `Socket.gethostname` instead: https://ruby-doc.org/stdlib-2.7.1/libdoc/socket/rdoc/Socket.html#method-c-gethostname?

sdk/opentelemetry-sdk.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.5.0'
 
+  spec.add_dependency 'google-cloud-env'
```
mwear, line 28: This is the first 3rd party runtime dependency we're adding. I'm not saying we shouldn't add it, just pointing that out.

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
```diff
@@ -0,0 +1,31 @@
...
+require 'opentelemetry/sdk/resources/detectors/telemetry'
+
+module OpenTelemetry
+  module SDK
+    module Resources
+      # AutoDetector contains detect class method for running all detectors
+      module AutoDetector
+        extend self
+
+        DETECTORS = [
```
mwear, line 17: Do we intend to add anything else here? The JS implementation has an AWS detector and environment variable detector (OTEL_RESOURCE_LABELS).

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
```diff
@@ -0,0 +1,31 @@
...
+          OpenTelemetry::SDK::Resources::Detectors::GoogleCloudPlatform,
+          OpenTelemetry::SDK::Resources::Detectors::Telemetry
+        ].freeze
+
+        def detect
```
mwear, line 22: These detectors are running in serial. Right now it's not a big deal, but if we add additional detectors, especially ones with network calls, this could end up blocking for longer than we'd like. Have you considered parallelizing this at all?

sdk/lib/opentelemetry/sdk/resources/detectors/telemetry.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module OpenTelemetry
+  module SDK
+    module Resources
+      module Detectors
+        # Telemetry contains detect class method for determining instrumentation resource labels
+        module Telemetry
+          extend self
+
+          def detect
```
mwear, line 15: We will always have a Telemetry::SDK resource and we are required by the spec to create one. I'm not sure that this should be a detector, as I think auto detection should be optional.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
```diff
@@ -38,7 +39,8 @@ def initialize
         def tracer(name = nil, version = nil)
           name ||= ''
           version ||= ''
-          @mutex.synchronize { @registry[Key.new(name, version)] ||= Tracer.new(name, version) }
+          resource = @resource.merge(OpenTelemetry::SDK::Resources::Resource.create('name' => name, 'version' => version))
```
mwear, line 42: The name and version here should technically be the `InstrumentationLibrary`. We should be adding this to the span separately: https://github.com/open-telemetry/oteps/pull/84. Recording this as a Resource was an existing mistake we shouldn't carry forward.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -27,7 +27,7 @@ def initialize
         @text_injectors = nil
         @span_processors = []
         @use_mode = USE_MODE_UNSPECIFIED
-        @tracer_provider = Trace::TracerProvider.new
+        @resource = OpenTelemetry::SDK::Resources::Resource.create
```
mwear, line 30: We should be able to default this to the Telemetry SDK Resource. We're required to have one by the spec and this would be a good place to initialize it. I talk about it more in a comment further down.

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
robertlaurin: No reason not to, I'll make use of that instead.

sdk/opentelemetry-sdk.gemspec
robertlaurin: I actually had a conversation with @fbogsany regarding this after I pushed this up, which is why I converted it back to draft.  He advised me that this functionality should be extracted out of the SDK and made into its own gem so not to add dependencies for something that will be optional.

My plan here will be to create a `detectors/` folder at the root of the repository containing the gem and it will include the auto detection functionality as well as all the detectors.  

What do you think @mwear ?

sdk/lib/opentelemetry/sdk/resources/detectors/telemetry.rb
robertlaurin: >We will always have a Telemetry::SDK resource and we are required by the spec to create one.

I missed that when going over the specs, I will remove this as a detector as you're right it does not belong in auto detection.   Could you link me to the spec that outline this (I haven't found it yet), I think adding this behaviour to the library fits this PR.

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
robertlaurin: For the env detector it wasn't completely clear to me what the behaviour should be there.  I saw this https://github.com/open-telemetry/opentelemetry-specification/issues/572 but I wasn't sure if any standardized convention was set yet.  With a little direction I'll gladly add it.

Regarding the AWS detector, I don't have access to an AWS instance readily available so I didn't want to put out something that might not work in a real world environment.  For example with the GCP detector I also validated it against a production environment.

sdk/opentelemetry-sdk.gemspec
mwear: I think that makes sense and generally mirrors what we're doing in JS where resources are their own npm package.

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
mwear: I think people are going to want one for AWS one way or another. We could approach this in one of two ways. We could make an issue for it and let someone pick it up if they want it. Optionally, we could start the process by creating one using the sample output here: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
mwear: Incidentally, this work is going on in #264.

robertlaurin: I think I've found myself to be a little confused here and regarding your question here https://github.com/open-telemetry/opentelemetry-ruby/pull/264#pullrequestreview-413928944 

I believe my change to reflect what is in the specification:

>New Tracer instances are always created through a TracerProvider (see API). The name and version arguments supplied to the TracerProvider must be used to create a Resource instance which is stored on the created Tracer.

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#tracer-creation

So to try and answer your question on #264 I believe we need to keep the `library_resource` on the tracer so it is able to provide that information to the spans `getLibraryResource` method.

mwear: I think that portion of the specification might be outdated.


The [SDK Resource Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/resource/sdk.md) states:

> When used with distributed tracing, a resource can be associated with the TracerProvider when it is created. That association cannot be changed later. When associated with a TracerProvider, all Spans produced by any Tracer from the provider MUST be associated with this Resource.

That seems to indicate that there is a single resource that is referenced through the TracerProvider, that will be attached to all spans it generates. At a bare minimum, this Resource will be the [telemetry library resource](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/resource/semantic_conventions#telemetry-sdk). If there are additional resources detected for a cloud, container, or based on environment variables they will all be merged into a single resource. The way a resource will make it to the export pipeline is basically a reference that gets attached to: TracerProvider -> Tracer -> Span -> SpanData.

The name and version of an individual tracer belongs to the instrumentation library which will flow from Tracer -> Span -> SpanData. This is specified [here](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#obtaining-a-tracer).

My interpretation is that there is a single Resource per TracerProvider (with labels composed from many underlying resources) that is made available on spans and also an InstrumentationLibrary reference on each span.

The protos also seem to support this interpretation. 
See: 
* https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/trace/v1/trace.proto#L28-L45
* https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/common/v1/common.proto#L57-L62
* https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/resource/v1/resource.proto#L26-L34

Note is a `ResourceSpans` message that has repeated `InstrumentationLibrarySpans` that has repeated `Spans`.

I don't see where the `library_resource` fits in.

As you've pointed out, there is a discrepancy between the wording in the [SDK tracer creation spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#tracer-creation) and this interpretation. To my knowledge, this was the way things were before the instrumentation library was introduced and that part of the spec might be outdated. I'm piecing this together from conversations that have happened at the spec sig, other implementations I've seen, and the protos. I realize this is hardly a definitive answer. I'll try to write up a spec issue to get some clarification.

mwear: Also relevant: https://github.com/open-telemetry/oteps/blob/master/text/0083-component.md

mwear: Also: https://github.com/open-telemetry/opentelemetry-specification/issues/614

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
fbogsany: I think we should get the general framework in place with the GCP detector in this PR, and open issues for the other detectors.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
fbogsany: Each `Tracer` should have an `InstrumentationLibrary` with the `name` and `version`. The `TracerProvider` should have a `Resource`. The `Resource` is attached to every [ETSR](https://github.com/open-telemetry/opentelemetry-proto/blob/28e2774262a651c4c2e701c3a05820c3e43ead71/opentelemetry/proto/collector/trace/v1/trace_service.proto#L38-L45) in OTLP (technically to every `ResourceSpans`, but in practice the `repeated` in ETSR will be always 1 for ETSRs produced by clients - larger batches may be produced by the collector). Things get more complicated for other protocols, and we’ll need to figure out the mapping. We’re not alone here - the Jaeger exporter from OTel Go isn’t quite in the right shape yet.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -9,21 +9,17 @@ module SDK
     module Trace
       # {Tracer} is the SDK implementation of {OpenTelemetry::Trace::Tracer}.
       class Tracer < OpenTelemetry::Trace::Tracer
-        attr_reader :name
-        attr_reader :version
+        attr_reader :resource
```
fbogsany, line 12: This should be `instrumentation_library` I think.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -9,21 +9,17 @@ module SDK
...
 
         # @api private
         #
         # Returns a new {Tracer} instance.
         #
-        # @param [String] name Instrumentation package name
-        # @param [String] version Instrumentation package version
+        # @param [Resource] resource Containing name and version arguments supplied to the TracerProvider
```
fbogsany, line 18: `s/Resource/InstrumentationLibrary/`

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
ibawt: it will be more portable that way for sure, and shelling out is the worst.

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
```diff
@@ -0,0 +1,54 @@
...
+        module GoogleCloudPlatform
+          extend self
+
+          def detect
+            gcp_env = Google::Cloud::Env.new
+            resource_labels = {}
+
+            if gcp_env.compute_engine?
+              resource_labels[CLOUD_RESOURCE[:provider]] = 'gcp'
+              resource_labels[CLOUD_RESOURCE[:account_id]] = gcp_env.project_id || ''
```
ibawt, line 23: do we want blank strings if it's not there?  we'll end up increasing the payload by sizeof(key)

if you're wondering why they wouldn't be there, https://github.com/GoogleCloudPlatform/k8s-metadata-proxy is a thing.

ibawt: I have a feeling workload concealment does something similar as well.

fbogsany: Yeah, I prefer to not have the attribute rather than have an empty attribute.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
mwear: @fbogsany that matches my understanding and after raising in issue on the sec SIG it seems to be the current state of things. Through this process, I discovered: https://github.com/open-telemetry/opentelemetry-proto/pull/150, which, if accepted, might bring this full circle. We would end up having per tracer Resources that contain the instrumentation library and process level information.

sdk/opentelemetry-sdk.gemspec
robertlaurin: The work to split it out into a separate gem has been pushed up now.

sdk/lib/opentelemetry/sdk/resources/auto_detector.rb
robertlaurin: I expected this to need to be improved as more detectors are added, but my intention was to provide a starting point to be iterated on.  Hadn't really considered parallelizing it yet.

sdk/lib/opentelemetry/sdk/configurator.rb
robertlaurin: Changes regarding this comment here https://github.com/open-telemetry/opentelemetry-ruby/pull/263/commits/3cbcb13493fc7522dadb94baa7a0bf4806f929ac

Not sure if this follows the spec correctly as I wasn't able to find the specific parts around this requirement.

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
robertlaurin: Updated to remove keys with blank values.
https://github.com/open-telemetry/opentelemetry-ruby/pull/263/files#diff-f799130e405e5a80aa43ba547be52cdbR40

https://github.com/open-telemetry/opentelemetry-ruby/pull/263/files#diff-6e5e6a6b9a0554f85023aa90b68f94abR62-R76

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -27,13 +27,17 @@ def initialize
         @text_injectors = nil
         @span_processors = []
         @use_mode = USE_MODE_UNSPECIFIED
-        @tracer_provider = Trace::TracerProvider.new
+        @resource = Resources::Resource.telemetry_sdk
       end
 
       def logger
         @logger ||= Logger.new(STDOUT)
       end
```
robertlaurin, line 36: How strict are we with rubocop on this project?  Asking for a friend.
```
lib/opentelemetry/sdk/configurator.rb:11:5: C: Metrics/ClassLength: Class has too many lines. [102/100]
```

mwear: we are heavy users of `rubocop:disable`

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
robertlaurin: Piecing it together, I believe I've adjusted things to follow what was discussed above.

- the configurator is initialized a resource containing the [telemetry sdk](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/configurator.rb#L30) labels.
- the configurator optionally accepts a resource which is [merged with the default resource](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/configurator.rb#L37-L39)
- the tracer provider is initialized with the [resource from the configurator](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/configurator.rb#L96-L98)
- the tracer provider can initialize a tracer [(accepting name and version as optional params)](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L38-L42)
- the tracer has an instrumentation_library containing [name and version](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L25)
- when a tracer creates a new span it passes along the resource from the [tracer provider](https://github.com/robertlaurin/opentelemetry-ruby/blob/ec04cac756da0620b293594c162787ed75604e3f/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L57-L58)

robertlaurin: >I don't see where the library_resource fits in.
https://github.com/robertlaurin/opentelemetry-ruby/blob/09aa570537bada624810e8725050ca221e39a19b/sdk/lib/opentelemetry/sdk/trace/span.rb#L239

Should this be renamed or removed entirely, we still need to associate spans with the resource correct?

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -27,13 +27,17 @@ def initialize
...
 
+      def resource=(new_resource)
+        @resource = @resource.merge(new_resource)
```
fbogsany, line 38: Do we want to `merge` here or just replace? In either case, we should document the behavior.

sdk/lib/opentelemetry/sdk/resources/constants.rb
```diff
@@ -0,0 +1,124 @@
...
+          # The version string of the service API or implementation.
+          version: 'service.version'
+        }.freeze
+
+        # Attributes describing the telemetry library.
+        TELEMETRY_SDK_RESOURCE = {
+          # The name of the telemetry library.
+          name: 'telemetry.sdk.name',
+
+          # The language of telemetry library and of the code instrumented with it.
```
fbogsany, line 31: `... of the telemetry ...`

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: We definitely want `merge`, since we always want to have the default SDK resource labels. So this is just a request to document the method.

resource-detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          if gcp_env.compute_engine?
+            resource_labels[resource_constants::CLOUD_RESOURCE[:provider]] = 'gcp'
+            resource_labels[resource_constants::CLOUD_RESOURCE[:account_id]] = gcp_env.project_id
+            resource_labels[resource_constants::CLOUD_RESOURCE[:region]] = gcp_env.instance_attribute('cluster-location')
+            resource_labels[resource_constants::CLOUD_RESOURCE[:zone]] = gcp_env.instance_zone
+
+            resource_labels[resource_constants::HOST_RESOURCE[:hostname]] = hostname
+            resource_labels[resource_constants::HOST_RESOURCE[:id]] = gcp_env.lookup_metadata('instance', 'id')
+            resource_labels[resource_constants::HOST_RESOURCE[:name]] = gcp_env.lookup_metadata('instance', 'hostname')
```
fbogsany, line 29: At some point, it might be worth adding a helper method (or methods) to `OpenTelemetry::SDK::Resources`, like:
```ruby
OpenTelemetry::SDK::Resources.host_labels(
  hostname: hostname,
  id: gcp_env.lookup_metadata('instance', 'id'),
  name: gcp_env.lookup_metadata('instance', 'hostname'),
)
```

resource-detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
```diff
@@ -0,0 +1,27 @@
...
+      module AutoDetector
+        extend self
+
+        DETECTORS = [
+          OpenTelemetry::Resource::Detectors::GoogleCloudPlatform
+        ].freeze
+
+        def detect
+          resources = DETECTORS.map(&:detect)
+          resources.reduce(OpenTelemetry::SDK::Resources::Resource.create) do |accumulated, detected|
```
fbogsany, line 20: I think this can all be abbreviated to:
```ruby
DETECTORS.map(&:detect).reduce(:merge)
```

sdk/lib/opentelemetry/sdk/configurator.rb
robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/pull/263/commits/ec14e075b254d72f2eca2284a2548c79de45b59d#diff-938e8162d01853e91a23d9ab6d45636aR37-R41

resource-detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
robertlaurin: Updated.

resource-detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
robertlaurin: I agree, that would be a good helper.  I don't think it fits any of the use cases that are part of this PR however (I wouldn't use it in the google detector as an example).  Is it something you would like as part of this PR?

fbogsany:
Related: https://github.com/open-telemetry/oteps/pull/111

fbogsany:
@robertlaurin can you please resolve any out-of-date comments above and rebase? This LGTM and I’ve approved it, but it wasn’t clear to me if you still had unresolved questions. Selfishly, I need this merged for my work on the OTLP exporter 😄 .

robertlaurin:
@fbogsany I've resolved every I believe to be resolved 😄 

There's only two points left that I'd like to leave to some one else to resolve.

I don't think this one is a problem but I raised it just to get a second set of eyes on it.
https://github.com/open-telemetry/opentelemetry-ruby/pull/263#discussion_r425281233

The long message thread here, once again I think this one is good but I'd like to leave it to either you or @mwear to sign off on it.
https://github.com/open-telemetry/opentelemetry-ruby/pull/263#discussion_r430006630
https://github.com/open-telemetry/opentelemetry-ruby/pull/263#discussion_r430013414

sdk/lib/opentelemetry/sdk/resources/detectors/google_cloud_platform.rb
fbogsany: `project-id` seems wrong, but the metadata doesn't supply the account ID. I would suggest leaving this as-is and opening an issue in the spec repo asking for clarification.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
fbogsany: I think the library resource should be removed. I'm going to PR the change I think is needed separately. I'll re-read the PR, but I don't think you need to change anything in this PR.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -16,14 +16,12 @@ class Tracer < OpenTelemetry::Trace::Tracer
         #
         # Returns a new {Tracer} instance.
         #
-        # @param [String] name Instrumentation package name
-        # @param [String] version Instrumentation package version
+        # @param [InstrumentationLibrary] resource Containing name and version arguments supplied to the TracerProvider
```
fbogsany, line 19: This change is incorrect. We create the `InstrumentationLibrary` inside the method from the `name` and `version` params.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -56,7 +54,8 @@ def internal_create_span(result, name, kind, trace_id, span_id, parent_span_id,
             attributes = attributes&.merge(result.attributes) || result.attributes
             active_trace_config = OpenTelemetry.tracer_provider.active_trace_config
             active_span_processor = OpenTelemetry.tracer_provider.active_span_processor
-            Span.new(context, name, kind, parent_span_id, active_trace_config, active_span_processor, attributes, links, start_timestamp || Time.now, @resource, @instrumentation_library)
+            resource = OpenTelemetry.tracer_provider.resource
+            Span.new(context, name, kind, parent_span_id, active_trace_config, active_span_processor, attributes, links, start_timestamp || Time.now, resource, @instrumentation_library)
```
fbogsany, line 58: I *think* this is incorrect. Or at least unnecessary. We should remove the `resource` parameter -- for now, can you just pass `nil`?

resource-detectors/lib/opentelemetry/resource/detectors.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
```
mwear, line 3: New files should have 2020. There are a number of files with 2019. We can also considering removing the year (as a separate issue). The JS SIG recently did this: https://github.com/open-telemetry/opentelemetry-js/pull/1150.

resource-detectors/test/opentelemetry/detectors/google_cloud_platform_test.rb
```diff
@@ -0,0 +1,79 @@
...
+
+      it 'returns a resource with gcp attributes' do
+        _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
+        _(detected_resource_labels).must_equal(expected_resource_labels)
+      end
+
+      describe 'and a nil resource value is detected' do
+        let(:project_id) { nil }
+
+        it 'returns a with that label' do
```
mwear, line 65: Is there a missing word here, resource perhaps?

resource-detectors/test/opentelemetry/detectors/google_cloud_platform_test.rb
```diff
@@ -0,0 +1,79 @@
...
+          _(detected_resource_labels.key?('cloud.account.id')).must_equal(false)
+        end
+      end
+
+      describe 'and an empty string resource value is detected' do
+        let(:project_id) { '' }
+
+        it 'returns a with that label' do
```
mwear, line 73: Missing word?

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -56,7 +54,8 @@ def internal_create_span(result, name, kind, trace_id, span_id, parent_span_id,
             attributes = attributes&.merge(result.attributes) || result.attributes
             active_trace_config = OpenTelemetry.tracer_provider.active_trace_config
             active_span_processor = OpenTelemetry.tracer_provider.active_span_processor
-            Span.new(context, name, kind, parent_span_id, active_trace_config, active_span_processor, attributes, links, start_timestamp || Time.now, @resource, @instrumentation_library)
+            resource = OpenTelemetry.tracer_provider.resource
```
mwear, line 57: Is there any reason not to assign this as an instance variable in the constructor?

sdk/lib/opentelemetry/sdk/trace/tracer.rb
mwear: What the issue here exactly? A span should have a reference to a resource.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
mwear: The comment is incorrect, but the code looks fine.

fbogsany: Yep, sorry - I should have been more specific: just the comment is incorrect.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
fbogsany: A span needs a reference to an `InstrumentationLibrary`. The `Resource` is global (a singleton). I don’t see the need to attach the same Resource reference to every span when it is always available from `OpenTelemetry.tracer_provider.resource`.

mwear: Unfortunately, that's a current design flaw of the Ruby library. Technically, multiple TracerProviders should be able to exist in a single process. They should be able to have different resources, but share the same export pipeline. So, the resource is not exactly a global singleton. This was discussed somewhat in https://github.com/open-telemetry/opentelemetry-specification/issues/508 and I am supposed to be writing some clarifying language for the spec about this topic.

For the Ruby SDK. I think it's fine to have a single TracerProvider configured by default, as long as we allow additional ones to be instantiated. We'll need to make sure that references go through TracerProvider instances, rather than the classes. There are few places we were need to clean up static references in this regard.

mwear: I'll try to get some more clarification about the multiple TracerProvider situation at the spec SIG tomorrow. I've heard that it's a use case, but given that you can only have one global TracerProvider at the API level, I'm not really sure how this works in practice, or if it's necessarily a use case that needs to be preserved.

mwear: This is the relevant portion of the spec: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider

> Normally, the TracerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default TracerProvider.

> Notwithstanding any global TracerProvider, some applications may want to or have to use multiple TracerProvider instances, e.g. to have different configuration (like SpanProcessors) for each (and consequently for the Tracers obtained from them), or because its easier with dependency injection frameworks. Thus, implementations of TracerProvider SHOULD allow creating an arbitrary number of TracerProvider instances.

Noting that the spec lists this as a SHOULD, not a MUST.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
robertlaurin: Nope no reason not to.  Updated.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
robertlaurin: If I'm following correctly, for the `sdk/lib/opentelemetry/sdk/trace/tracer.rb` code could we not "just" pass in a reference to the `tracer_provider` that created it?  The `tracer_provider` is accessed in multiple places within the `tracer.rb` class already, why not pass it in as part of the initializer for the tracer?

That would solve the issue of supporting multiple `tracer_provider`s for this particular code path, no?

It would also remove some of the implicit coupling that the tracer has on the tracer_provider.
https://github.com/robertlaurin/opentelemetry-ruby/blob/cf02a49c38e230f805dcfd9bbb39b8b9a8caabfb/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L27

mwear: You're following correctly @robertlaurin. I think the core issue is that because of the multiple TracerProvider use case, each span needs a reference to a resource. Ultimately, this is because TracerProviders can be initialized with different resources, but share the same export pipeline. As you point out, we can support this use case pretty easily passing a TracerProvider to a Tracer during initialization and maintaining a reference to it.

mwear: I'm not sure how much of this needs to be addressed in this PR. It would be fine to create issues for additional work that isn't directly related.

robertlaurin: I agree that I don't want to grow this PR into additional work that isn't directly related. 
 I think this change here https://github.com/open-telemetry/opentelemetry-ruby/pull/263/commits/a43ce3a580aa10b3f34e5cda80915e42412ceffc leaves the code in a good places to be improved on.

fbogsany:
LGTM. Needs a rebase, but otherwise 👍

PR #264, opened 2020-05-15
Setup instrumentation library

dmathieu:
Following @mwear's comment in #261, this is a proposal to implement the instrumentation library pattern, as I understand it from open-telemetry/oteps/pull/84.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/tracer.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/tracer.rb
@@ -24,6 +24,7 @@ module OpenTelemetry
           @name = name
           @version = version
           @resource = Resources::Resource.create('name' => name, 'version' => version)
+          @instrumentation_library = InstrumentationLibrary.new(name, version)
         end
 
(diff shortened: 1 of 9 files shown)


sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -11,6 +11,7 @@ module Trace
       class Tracer < OpenTelemetry::Trace::Tracer
         attr_reader :name
         attr_reader :version
+        attr_accessor :instrumentation_library
```
dmathieu, line 14: We could setup a setter here which would only take name and version. We already have those values from the initializer though, so we could reuse them from there too (the name would then be the full class name then, which isn't awesome IMHO).

Another possibility would be to have a `name` setter in the adapter?

dmathieu:
Thank you for the feedback @mwear. I've updated my PR to reflect your comment.

mwear:
There could be changes on the horizon if this is accepted: https://github.com/open-telemetry/opentelemetry-proto/pull/150. I'm merging this PR as it represents the current state of the specification.

PR #273, opened 2020-06-11
[do not merge] Add opentelemetry-datadog-exporter

ericmustin:
#### Summary
This PR aims to provide an exporter to convert opentelemetry spans into Datadog specific format and to export them to a datadog-agent. 

Additionally, there is a Propagator for injecting and extracting datadog specific headers into distributed traces.

Discussed briefly with @mwear @fbogsany and @martinisoft in the last SIG meeting and wanted to put this in to mark progress of where the work stands.

- At a high level, the opentelemetry-exporter-datadog is similar to the existing exporters, such as jaeger. The key difference is that, unlike the Jaeger implementation, the Datadog exporter can not rely on the existing SimpleSpanProcessor or BatchSpanProcessor. This is because Datadog's trace-agent (_a daemon which runs as a background process, and receives/aggregates/and transmits traces to an api-intake endpoint_), must receive complete Traces (ie: a trace which for which all span's that have been started have been finished). In order to accommodate this constraint, this PR also contains a specialised DatadogSpanProcessor, which is quite similar to the BatchSpanProcessor but only exports complete traces. 

- Additionally, Datadog contains vendor specific distributed tracing headers which must be extracted and injected, and so this PR also contains a Propagator with an injector and extractor that can be added via the Composite Propagator.

- Otherwise, the logic is very similar to existing exporters. 
  - A Span Processor hooks into specific span lifecycle events and manages a completed span/traces queue. 
  - A worker thread periodically exports complete traces from that queue in datadog format. 
  - A writer+transport flushes those complete traces to the trace-agent, and will leverage the existing writer/transport implementation from dd-trace-rb.

#### Notes / TODOs
- [ ] this will move to a separate `opentelemetry-ruby-contrib` repo, and shouldn't be merged into this repo.
- [x] TODO: ~tests for `span_encoder`,  `datadog_span_processor` , datadog `exporter` and datadog `propagator`~ Added tests to the rakefile and circeci config (to be removed when this lands in `-contrib` repo). 
- [x] TODO: There's still some tests needed regarding forced_flushing in the `datadog_span_processor` but was hoping to have the main implementation reviewed first.
- [x] TODO: ~Distributed Tracing Propagation of trace context via http headers~
- [ ] TODO: confirm sampling logic. In some other languages (i believe python), spans that are not sampled are not passed to the span_processor's `on_finish` hook. However it appears even spans with a sampling rate of 0 are passed to on_finish in ruby? Is this the case?
- [x] TODO: ~`span.context.tracestate` needs to be exposed to the Struct created by `to_span_data`, in order to handle the datadog concept of an "x-datadog-origin" header... can make a separate PR for this~ PR Merged https://github.com/open-telemetry/opentelemetry-ruby/pull/276
- [x] TODO: ~the span_encoder's translation to a Datadog "Resource" (basically, the route or endpoint) is incomplete.~ This has been added.
- [x] TODO: ~UDS transport is not yet implemented~ This has been added
- [x] TODO: ~error details recorded as event's are not yet taken into account~ This has been added
- [x] TODO: ~README and docs~ This has been added. As we get closer to having this in a place where it can be merged I'd be happy to add an example to the `example` directory
- [x] TODO: ~general linting and code cleanliness~
- [x] TODO:  ~Sinatra instrumentation that I need to make a PR for~ Merged PR here: https://github.com/open-telemetry/opentelemetry-ruby/pull/275

#### Usage

- 

```ruby
  c.use 'OpenTelemetry::Adapters::Sinatra'
  c.add_span_processor(
    OpenTelemetry::Exporters::Datadog::DatadogSpanProcessor.new(
      exporter: OpenTelemetry::Exporters::Datadog::Exporter.new(
          service_name: 'my-service', agent_url: "http://localhost:8126"
        )
      )
    )  
end
```

- Here's an example OpenTelemetry Trace in Datadog's UI 😄  
dd trace


ericmustin:
@mwear @fbogsany 👋  Hey y'all, hope all is well. I've updated the PR notes here to reflect where this PR is currently, and tried to make everything much clearer for the reviewer. I think a review would be beneficial at this point, there are still a few TODOs, but mostly I could really use some feedback on whether the approach of the DatadogSpanProcessor is agreeable, since this is the largest change vs existing exporter implementations. Please ping if any Qs, also available via the Gitter if anything pops up, cheers!

exporters/datadog/lib/opentelemetry/exporters/datadog/datadog_span_processor.rb
```diff
@@ -0,0 +1,199 @@
...
+
+        # datadog trace-agent endpoint requires a complete trace to be sent
+        # threadsafe may block on lock
+        def on_start(span)
+          context = span.context
+          trace_id = context.trace_id
+
+          lock do
+            if all_spans_count(traces_spans_count) >= max_queue_size
+              OpenTelemetry.logger.warn('Max spans for all traces, spans will be dropped')
```
marcotc, line 61: One alternative way to handle a full queue is to drop a random completed trace (dropping a full trace, not a single span). We currently do that in the [`ddtrace` gem](https://github.com/DataDog/dd-trace-rb/blob/c6fbf2410a60495f1b2d8912bf7ea7dc63422141/lib/ddtrace/buffer.rb#L34-L36).

It allows for a more fair usage of the queue when under stress load, and will create proportional representation of code paths being instrumented at stress time.

exporters/datadog/lib/opentelemetry/exporters/datadog/exporter/span_encoder.rb
```diff
@@ -0,0 +1,199 @@
...
+            datadog_spans = []
+
+            default_tags = get_default_tags(tags) || {}
+
+            otel_spans.each do |span|
+              trace_id, span_id, parent_id = get_trace_ids(span)
+              span_type = get_span_type(span)
+
+              datadog_span = ::Datadog::Span.new(nil, span.name,
+                                                 service: service,
```
marcotc, line 52: It looks like it will only be possible to have one `service` name for the whole application being instrumented. This might limit the usage, given Datadog's backend can separate different services into distinct buckets for aggregation.
Is this limitation acceptable here?

exporters/datadog/opentelemetry-exporters-datadog.gemspec
```diff
@@ -0,0 +1,41 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'ddtrace'
```
marcotc, line 29: You would probably have a minimum version of `ddtrace` supported here.

I suggest taking at look at the modules and classes you reference in this PR and find when they were introduced with a suitable implementation in `ddtrace`.

ericmustin: So the main thing is the `Datadog::Writer` module, which it has just occurred to me may get deprecated/removed in future? Would it make more sense to just move to using the `AsyncWriter` now and making `0.35.0` a min version?

exporters/datadog/lib/opentelemetry/exporters/datadog/datadog_span_processor.rb
ericmustin: that's a good idea, i've updated the processor to drop a random unfinished trace in this case

exporters/datadog/lib/opentelemetry/exporters/datadog/exporter/span_encoder.rb
ericmustin: I think this is by design, or at least, this is how python's exporter approaches service naming. i am going to check to see whether this was an oversight or not, as I suppose you could technically leverage the `span_type` field to format the different instrumented library service names as`<service_name>-<span_type_shortname>` so it would look like `my_service-redis` `my_service-net` etc etc

fbogsany:
> it appears even spans with a sampling rate of 0 are passed to `on_finish` in ruby? Is this the case?

OpenTelemetry makes a distinction between “sampled” and “recording”. It is possible for a sampler to decide to record spans that are not sampled. In that case, we create a SDK span with the sampled bit not set. Only if the span is not sampled and not recording do we create an API span, which will not be passed to `on_start` or `on_finish`.

https://github.com/open-telemetry/opentelemetry-ruby/blob/0af11280508c8f754bab54a9d2cb482f76983282/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L59-L61

ericmustin:
@fbogsany Thanks for the feedback. Re:

> Only if the span is not sampled and not recording do we create an API span, which will not be passed to on_start or on_finish.

In practice I believe what this means is that I will need to be able to access the sampling rate of each span, instead of just simply the `sampled` bit `true/false`, since with something like the `ProbabilitySampler` there would be potentially be spans both unrecorded and unsampled. This will be necessary to effectively "upscale" metrics related to hits/errors/latency etc for spans that are passed to the processor/exporter. I believe there are two ways I could retrieve it, either in the processor and then passed to the exporter with every span, or by making a PR to core to add an additional piece of information to the `SpanData` struct. I'd would be happy to do the latter if you would be open to accepting a PR for that? I don't want to unnecessarily bloat the SpanData struct but I do think this could be a common use case for exporters.

ericmustin:
@fbogsany thinking about this a bit more, the `span.tracer_config.sampler.probability` only applies to the `ProbabilitySampler` and isn't a consistent property across all samplers.  Additionally the rate is applied to every span if i understand it correctly? So this probably doesn't belong on the SpanData struct and would instead just be something to check for in the processor (once, since it's set at the tracer level), and then pass into the `exporter.export` method directly?

fbogsany:
The default behaviour of the probability sampler is to respect the parent span’s sampled flag and to apply probability sampling to root spans and remote parents: https://github.com/open-telemetry/opentelemetry-ruby/blob/0099668e9ad7eedf32bb496e135e8220f1e49c61/sdk/lib/opentelemetry/sdk/trace/samplers.rb#L76-L78

It *is* possible to apply sampling to every span, but that’s an inefficient configuration. The probability sampling is based on the trace ID, so the sampling decision will be the same for all spans with the same trace ID.

The probability value itself is not exposed, although you could parse it out of the sampler’s `.description` (the format is described by the OTel specification).

Sampling is still an area of active discussion for the spec, so the current structures could change. It’d be worth discussing requirements or ideas in the spec repo.

ericmustin:
@fbogsany I see, thanks for the explanation...Looking through this with fresh eyes, it seems like in practice the case of "not sampled _AND_ not recorded" does occur with the `ProbabilitySampler`, [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/0099668e9ad7eedf32bb496e135e8220f1e49c61/sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb#L35), So i'll need to grab the sampling rate and pass to the exporter. in the datadog context, the rate has to applied to every span, even though in the Otel context the rate is only being used 1 time per trace at the root span/remote span level.  I think it will be fine to just simply grab the configured rate one time `probability ||= get_rate_from_description(span)` and apply it naively to every single span exported. it would mean some inaccuracy on a very small scale but across a production system would show accurate hit counts.

ericmustin:
> So i'll need to grab the sampling rate and pass to the exporter.

Upon closer inspection I don't believe this is being made available to the Processor, or is accessibe through the spans passed to `on_start` and `on_finish`. I think in order to account for probability sampling rates, it will require a span's `tracer_config.sampler` to be added to the `SpanData` struct

ericmustin:
closing this since it's fallen out of date and we've moved dev over to a datadog hosted repo. for posterity for those interested, a beta exporter can be found here https://rubygems.org/gems/opentelemetry-exporters-datadog

PR #276, opened 2020-06-12
add tracestate to span data

ericmustin:
#### summary

This PR adds a span's `context.tracestate` to the SpanData Struct which gets created by `to_span_data` and is used in conjunction with trace processors and exporters.  

#### description

The reasoning behind this is that there may be some vendor specific arbitrary metadata which is generally passed via distributed tracing headers and, as far as I understand, would live on the `context.tracestate` when extracted. 

Examples of this could be: 
 - baggage items from open tracing, which can be [extracted/injected via http headers](https://opentracing.io/docs/overview/inject-extract/). although this isn't particularly popular, [it's implemented](https://github.com/lightstep/lightstep-tracer-objc/blob/e7e0b6fd68d05f032970ff0cc548c27b877854ba/Pod/Classes/LSTracer.m#L151) in some different vendors. 
- For vendor specific metadata, which is how [opentelemetry-python is using tracestate](https://github.com/open-telemetry/opentelemetry-python/blob/3fa15aa2ab86a8808406f0bc07c18732bf1bc7b0/ext/opentelemetry-ext-datadog/src/opentelemetry/ext/datadog/propagator.py#L77) 

In opentelemetry-ruby, however, for exporters which encode these traces in a vendor specific format, those details in the tracestate are not available since they are not passed into the SpanData struct. This PR adds them to the SpanData struct and adds a very simple test to validate it.

#### notes

Let me know if I've missed anything here, tried to go through everything, but still getting familiar with the repo, so it's possible i've missed a test or usage somewhere.

### DIFF
--- a/sdk/test/integration/global_tracer_configurations_test.rb
+++ b/sdk/test/integration/global_tracer_configurations_test.rb
@@ -38,6 +38,12 @@ describe OpenTelemetry::SDK, 'global_tracer_configurations' do
       it 'are all SpanData' do
         _(finished_spans.collect(&:class).uniq).must_equal([sdk::Trace::SpanData])
       end
+
+      it 'are all SpanData which have trace_flags and tracestate span context' do
+        finish_span_keys = finished_spans.collect(&:members).flatten.uniq
+        _(finish_span_keys).must_include(:tracestate)
+        _(finish_span_keys).must_include(:trace_flags)
+      end
     end
   end
 
--- a/sdk/test/integration/global_tracer_configurations_test.rb
+++ b/sdk/test/integration/global_tracer_configurations_test.rb
@@ -16,10 +16,11 @@ describe OpenTelemetry::SDK, 'global_tracer_configurations' do
     end
   end
   let(:tracer) { provider.tracer(__FILE__, sdk::VERSION) }
+  let(:parent_context) { nil }
   let(:finished_spans) { exporter.finished_spans }
 
   before do
-    tracer.in_span('root') do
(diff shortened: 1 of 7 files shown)


sdk/test/integration/global_tracer_configurations_test.rb
```diff
@@ -38,6 +38,12 @@
       it 'are all SpanData' do
         _(finished_spans.collect(&:class).uniq).must_equal([sdk::Trace::SpanData])
       end
+
+      it 'are all SpanData which have trace_flags and tracestate span context' do
+        finish_span_keys = finished_spans.collect(&:members).flatten.uniq
+        _(finish_span_keys).must_include(:tracestate)
+        _(finish_span_keys).must_include(:trace_flags)
+      end
```
fbogsany, line 46: I think this test is overkill, since the previous one ensures all the finished spans are `SpanData`.

fbogsany:
I'd love to see a test for non-nil `tracestate` -- ideally as end-to-end as possible, so we ensure that if a span is created from context that has non-nil `tracestate`, then that propagates all the way to an exporter.

sdk/test/integration/global_tracer_configurations_test.rb
ericmustin: Fair enough, I can remove this and work on adding a more robust end-to-end tracestate test that starts a span with a non-nil tracestate context, finishes+processes it, and verifies it is rec'd by the exporter with matching tracestate

Issue #231, opened 2020-04-30
Add Binary Thrift over HTTP as a transport option for `OpenTelemetry::Exporters::Jaeger::Exporter`


konfortes:
Hey guys, 
I'm having some trouble with setting the `OpenTelemetry::Exporters::Jaeger::Exporter` with `opentelemetry-collector`.
This is how I configure it:

```
collector_host = 'otel-collector.opentelemetry.svc.cluster.local'
collector_port = 14268

OpenTelemetry::SDK.configure do |c|
  c.add_span_processor(
    OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
      OpenTelemetry::Exporters::Jaeger::Exporter.new(
        service_name: 'ruby-app', host: collector_host, port: collector_port
      )
    )
  )
end
```

it seems like the collector host succesfully resolved because I **DON'T** 
get this error anymore: `/usr/local/bundle/gems/opentelemetry-exporters-jaeger-0.4.0/lib/opentelemetry/exporters/jaeger/transport.rb:14:in** 'connect': getaddrinfo: Name or service not known (SocketError)`

But I can't see any logs related to this ruby-app in the collector nor see any traces in Jaeger.
Am I missing something?

konfortes:
the opentelemetry-collector expects traces in `/api/traces` so I tried the Jaeger collector directly, but the problem persists. keep getting `Unable to connect to Jaeger Agent`

fbogsany:
I won't get to this this week, so punting it to Alpha 0.6. In reality, it should be done well before then, and can be released as a 0.5.1 version of just the Jaeger gem.

PR #277, opened 2020-06-13
OTLP exporter

fbogsany:
This adds a gem containing an exporter for OTLP over HTTP (trace only).

## To do:
- [x] Generate protos 
- [x] Distinguish failure modes and retry with exponential back off
- [x] Translate Status
- [x] Avoid tracing the export request (needs SDK support)
- [x] ~GZIP~ (Requires upstream support)
- [x] Unit tests
- [x] Manual (and perhaps automated?) integration test against the collector
- [x] Array-valued attributes
- [x] ~Handling redirects~ (Punting to a subsequent PR)
- [x] Rubocop warnings (somehow not generated by CI 🤔 )
- [x] user specified headers
- [x] group by `span.resource`
- [x] environment variable config
- [ ] rename the gem to `opentelemetry-exporter-otlp-http` (gRPC support will be added later as `opentelemetry-exporter-otlp-grpc`)

## Blockers
- [x] We’re going to translate to the current `master` protobuf definitions, including https://github.com/open-telemetry/opentelemetry-proto/pull/157. The collector does not yet support this version, so our end-to-end tests will be blocked until the collector receiver is updated.


exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+          case value
+          when String
+            kv.value.string_value = value
+          when Integer
+            kv.value.int_value = value
+          when Float
+            kv.value.double_value = value
+          when true, false
+            kv.value.bool_value = value
+          # TODO: when Array
```
fbogsany, line 245: Array support has been implemented in the proto and collector now, so this needs to be done.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+              FAILURE
+            end
+          rescue Net::OpenTimeout, Net::ReadTimeout
+            retry if backoff?(retry_count: retry_count += 1)
+            return FAILURE
+          end
+        end
+
+        def handle_redirect(location)
+          # TODO: figure out destination and reinitialize @http and @path
```
fbogsany, line 118: Need to handle redirects -- this code is c&p from Shopify's implementation, and needs to be adjusted to work in this implementation.

fbogsany:
There are two open issues:
1. Array-valued attributes
1. Handling redirects

Otherwise this is ready for review.

fbogsany:
@ericmustin I'd appreciate your 👀 on this, since you have some experience implementing exporters. 🙏

exporters/otlp/test/opentelemetry/exporters/otlp/exporter_test.rb
```diff
@@ -0,0 +1,218 @@
...
+
+  describe '#export' do
+    let(:exporter) { OpenTelemetry::Exporters::OTLP::Exporter.new(host: '127.0.0.1', port: 55681) }
+
+    before do
+      OpenTelemetry.tracer_provider = OpenTelemetry::SDK::Trace::TracerProvider.new(OpenTelemetry::SDK::Resources::Resource.telemetry_sdk)
+    end
+
+    it 'integrates with collector' do
+      skip unless ENV['TRACING_INTEGRATION_TEST']
```
fbogsany, line 27: I've run this locally with:
```sh
TRACING_INTEGRATION_TEST=1 bundle exec rake test
```
and the collector:
```sh
./otelcol --config logging.yml
```
where `logging.yml` is:
```yaml
receivers:
  otlp:
    protocols:
      http:
exporters:
  logging:
extensions:
  zpages:
service:
  extensions: [zpages]
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [logging]
```

It produces the output:
```
2020-07-29T22:05:48.751-0400	INFO	loggingexporter/logging_exporter.go:102	TraceExporter	{"#spans": 1}
```
And metrics:
```
Franciss-MacBook-Pro:otlp bogsanyi$ curl http://localhost:8888/metrics
# HELP otelcol_exporter_send_failed_spans Number of spans in failed attempts to send to destination.
# TYPE otelcol_exporter_send_failed_spans counter
otelcol_exporter_send_failed_spans{exporter="logging",service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be"} 0
# HELP otelcol_exporter_sent_spans Number of spans successfully sent to destination.
# TYPE otelcol_exporter_sent_spans counter
otelcol_exporter_sent_spans{exporter="logging",service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be"} 1
# HELP otelcol_process_runtime_heap_alloc_bytes Bytes of allocated heap objects (see 'go doc runtime.MemStats.HeapAlloc')
# TYPE otelcol_process_runtime_heap_alloc_bytes gauge
otelcol_process_runtime_heap_alloc_bytes{service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be"} 4.49036e+06
# HELP otelcol_process_runtime_total_alloc_bytes Cumulative bytes allocated for heap objects (see 'go doc runtime.MemStats.TotalAlloc')
# TYPE otelcol_process_runtime_total_alloc_bytes gauge
otelcol_process_runtime_total_alloc_bytes{service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be"} 9.477112e+06
# HELP otelcol_process_runtime_total_sys_memory_bytes Total bytes of memory obtained from the OS (see 'go doc runtime.MemStats.Sys')
# TYPE otelcol_process_runtime_total_sys_memory_bytes gauge
otelcol_process_runtime_total_sys_memory_bytes{service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be"} 7.2286456e+07
# HELP otelcol_receiver_accepted_spans Number of spans successfully pushed into the pipeline.
# TYPE otelcol_receiver_accepted_spans counter
otelcol_receiver_accepted_spans{receiver="otlp",service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be",transport="grpc"} 1
# HELP otelcol_receiver_refused_spans Number of spans that could not be pushed into the pipeline.
# TYPE otelcol_receiver_refused_spans counter
otelcol_receiver_refused_spans{receiver="otlp",service_instance_id="d5ec4873-08ce-4142-b8ea-c282714a95be",transport="grpc"} 0
```

ericmustin:
@fbogsany yup happy to look, should be able to block off some time in the coming days...one thing that jumps out is it isn't immediately clear to me how `Avoid tracing the export request (needs SDK support)` is getting accomplished? Is there a separate PR here or something implicit in the `Net::HTTP` instrumentation?

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+        def shutdown
+          @shutdown = true
+          @http.finish if @http.started?
+        end
+
+        private
+
+        def send_bytes(bytes)
+          retry_count = 0
+          untraced do
```
fbogsany, line 77: `Net::HTTP` requests in this block should be untraced.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+            request = Net::HTTP::Post.new(@path)
+            request.body = bytes
+            request.add_field('Content-Type', 'application/x-protobuf')
+            # TODO: enable gzip when https://github.com/open-telemetry/opentelemetry-collector/issues/1344 is fixed.
+            # request.add_field('Content-Encoding', 'gzip')
+  
+            @http.start unless @http.started?
+            response = @http.request(request)
```
fbogsany, line 85: This request *should* be untraced.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+
+        def handle_redirect(location)
+          # TODO: figure out destination and reinitialize @http and @path
+          location = response['location']
+          location = URI(location).relative? ? (@uri + location).to_s : location
+          initialize_http(location) # Assume this is our new destination from now on.
+        end
+
+        def untraced
+          @tracer.with_span(OpenTelemetry::Trace::Span.new) { yield }
```
fbogsany, line 125: This implements "untracing". It is only effective if the active sampler respects the parent's sampling decision. We create a parent span that is not sampled and not recorded (and, incidentally, not finished). Child spans within this block should also be not sampled and not recorded.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
ibawt: you should put a limit on the redirects is about all I can recommend.

exporters/otlp/lib/opentelemetry/exporters/otlp/version.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Copyright 2019 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Exporters
+    module OTLP
+      ## Current OpenTelemetry OTLP exporter version
```
ibawt, line 10: idk if you want to indicate that this is 0.5.0 version of the proto as well, or is that implied by "exporter"

fbogsany: It's really just the gem version. In general, we're currently bumping the versions of all the gems in this repo in lockstep, as we release new Alpha versions of the SDK and API gems.

I'm not sure how to convey the version of the proto that we support. Perhaps just in the README.md?

ibawt: hrm...it may be a non issue at the moment.  if 0.5.0 is guarunteeing that subsequent versions of the proto can still be parsed  and no one actually used otlp maybe it's ok.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+            kv.value.string_value = value
+          when Integer
+            kv.value.int_value = value
+          when Float
+            kv.value.double_value = value
+          when true, false
+            kv.value.bool_value = value
+          # TODO: when Array
+          end
+          kv
```
ericmustin, line 247: mostly just due to me not being very familiar with the `KeyValue` proto, but what would be the behavior for when a kv pair's value does not get matched by the case statement? does `kv.value` default to nil? and is that something that should be getting encoded and sent to the collector or should it get dropped pre-emptively, or is there a default that makes more sense?

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
ericmustin: this is very nifty!

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+        # Default timeouts in seconds.
+        KEEP_ALIVE_TIMEOUT = 30
+        OPEN_TIMEOUT = 5
+        READ_TIMEOUT = 5
+        RETRY_COUNT = 5
+        PATH = '/v1/trace'
+        private_constant(:KEEP_ALIVE_TIMEOUT, :OPEN_TIMEOUT, :READ_TIMEOUT, :RETRY_COUNT, :PATH)
+
+        def initialize(host:,
+                       port:,
```
ericmustin, line 33: from a vendor perspective, we have found that there's a benefit to having host (and port, to a lesser extent) be something that can be configured via a standardised environment variable.

I am unfortunately _not_ the best person to explain the use cases here but one example where, i believe, it has value would be in k8s, if you have a collector running ...(_gestures vaguely_) in each node...per...cluster?

In that situation, the application container may not know the host ip a collector would be listening on ahead of time and it would need to be set as an env var using the downward api

- https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#the-downward-api

- https://docs.datadoghq.com/agent/kubernetes/apm/?tab=daemonset#setup
```
Configure your application pods to pull the host IP in order to communicate with the Datadog Agent: Use the downward API to pull the host IP; the application container needs the DD_AGENT_HOST environment variable that points to status.hostIP.

    apiVersion: apps/v1
    kind: Deployment
     # ...
        spec:
          containers:
          - name: "<CONTAINER_NAME>"
            image: "<CONTAINER_IMAGE>"/"<TAG>"
            env:
              - name: DD_AGENT_HOST
                valueFrom:
                  fieldRef:
                    fieldPath: status.hostIP
```

Just one example but I'm sure there's many others where env var's make life easier for users, or would be required to use in a production environment. 

Obviously, fetching an env var and passing in it's value when initialising the otlp exporter is something a user could roll themselves quite easily. But as things scale having something standard and that get's picked up by default is a tremendous help for reducing onboarding pain.

looking at https://github.com/open-telemetry/opentelemetry-collector/tree/master/examples/demo i see something called `OTEL_AGENT_ENDPOINT` in the docker-compose....is this a standard?

Anyway, if there's something in the spec around this i think it's worth adding support for it. Also, the value of `OTEL_AGENT_ENDPOINT` in that example implies the spec may end up supporting a `url` value that would need to have host/port extracted from it.

otoh, if there's nothing in the spec it may just be that this implementation right now is constrained by that, but if there's a reasonable or already somewhat adopted env var, adding that support in early i think will help users

exporters/otlp/README.md
```diff
@@ -0,0 +1,83 @@
+# opentelemetry-exporters-otlp
+
+The `opentelemetry-exporters-otlp` gem provides an OTLP exporter for OpenTelemetry for Ruby. Using `opentelemetry-exporters-otlp`, an application can configure OpenTelemetry to export collected tracing data to [the OpenTelemetry Collector][opentelemetry-collector-home].
```
robertlaurin, line 3: Might be nice to make `OTLP` in `...provides an OTLP exporter...` a hyperlink to the repo for OTLP so anyone exploring this repo can find more information easily.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
fbogsany: https://github.com/open-telemetry/opentelemetry-specification/pull/699 is an attempt to specify all the things for OTLP, with a particular focus on environment variables for configuration. When that lands, this should be a lot clearer.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+            request = Net::HTTP::Post.new(@path)
+            request.body = bytes
+            request.add_field('Content-Type', 'application/x-protobuf')
+            # TODO: enable gzip when https://github.com/open-telemetry/opentelemetry-collector/issues/1344 is fixed.
+            # request.add_field('Content-Encoding', 'gzip')
+  
+            @http.start unless @http.started?
+            response = @http.request(request)
+
+            case response
```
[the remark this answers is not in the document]
fbogsany: Yep - I mentioned this in the SIG meeting: we ideally want some structured mechanism to report errors. Logging is a little too unstructured. Metrics would be nice, but some people might want logs or structured error reporting. I'm not sure what to do there - it seems like this would be useful more generally, across languages, but I worry we then have a very complicated API with callbacks for all sorts of things.

exporters/otlp/lib/opentelemetry/exporters/otlp/version.rb
fbogsany: The proto has been declared "stable" for tracing, so in theory, changes should be backwards compatible from now on. I think we should document the supported version of the proto in the README and ensure we bump the major version when the supported version of the proto changes in an incompatible way (most likely for metrics at this point), and document the change in the release notes.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
mwear: We discussed this briefly at the last SIG meeting and I just made an issue, but we should be able to implement an untraced method using context: https://github.com/open-telemetry/opentelemetry-ruby/issues/313. This doesn't necessarily need to happen in this PR though.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+          true
+        end
+
+        def encode(span_data)
+          Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.encode(
+            Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.new(
+              resource_spans: [
+                Opentelemetry::Proto::Trace::V1::ResourceSpans.new(
+                  resource: Opentelemetry::Proto::Resource::V1::Resource.new(
+                    attributes: OpenTelemetry.tracer_provider.resource.label_enumerator.map { |key, value| as_otlp_key_value(key, value) },
```
mwear, line 159: The spec makes concessions for having multiple tracer providers in a single process. It's possible for multiple tracer providers to share the same exporter, and in that case, resources can differ by span. We may have some other issues with spec compliance existing around this poorly understood usecase, but we could support it here by also grouping by `span.resource`.

Here's the relevant spec details:

https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#tracerprovider

> Normally, the TracerProvider is expected to be accessed from a central place. Thus, the API SHOULD provide a way to set/register and access a global default TracerProvider.

> Notwithstanding any global TracerProvider, some applications may want to or have to use multiple TracerProvider instances, e.g. to have different configuration (like SpanProcessors) for each (and consequently for the Tracers obtained from them), or because its easier with dependency injection frameworks. Thus, implementations of TracerProvider SHOULD allow creating an arbitrary number of TracerProvider instances.

Nothing that this is spec'd as *SHOULD* and not *MUST*.

This issue also has some relevant discussion: https://github.com/open-telemetry/opentelemetry-specification/issues/508.

Here's an example from JS: https://github.com/open-telemetry/opentelemetry-js/blob/master/packages/opentelemetry-exporter-collector/src/transform.ts#L263-L282

I think the multiple TracerProvider use case is better solved with Resource Scopes (https://github.com/open-telemetry/oteps/issues/78), but since it's currently part of the spec (as a SHOULD), I'm inclined to go with it for now. I'm willing to be talked out of this position though.

fbogsany:
I've rebased this and made some minor fixes for API changes. I haven't addressed feedback yet, or the open issues.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
fbogsany: I’m going to remove this from the current PR and figure it out in a subsequent PR. AFAIK the collector won’t redirect, though IDK whether Envoy in front of a collector pool might. In any case, this PR is large enough and late enough already.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
fbogsany: 👍 I’ll stick with the current method for this PR. We can work out the context-based mechanism in #313.

exporters/otlp/lib/opentelemetry/exporters/otlp/version.rb
fbogsany: I updated the README to clarify the supported version of the protocol.

robertlaurin:
> Rubocop warnings (somehow not generated by CI 🤔 )

https://github.com/robertlaurin/opentelemetry-ruby/blob/2913c9e13a12e7d73e864c85a5806ddfc38f3ff9/.circleci/config.yml#L61-L69

`bundle exec rake` will run all the rake tasks (including rubocop) and I think OTLP is missing from CI all together right now.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -0,0 +1,277 @@
...
+        OPEN_TIMEOUT = 5
+        READ_TIMEOUT = 5
+        WRITE_TIMEOUT = 5
+        RETRY_COUNT = 5
+        private_constant(:KEEP_ALIVE_TIMEOUT, :OPEN_TIMEOUT, :READ_TIMEOUT, :WRITE_TIMEOUT, :RETRY_COUNT)
+
+        def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_SPAN_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: 'localhost:55681/v1/trace'), # rubocop:disable Metrics/AbcSize
+                       insecure: config_opt('OTEL_EXPORTER_OTLP_SPAN_INSECURE', 'OTEL_EXPORTER_OTLP_INSECURE', default: false),
+                       certificate_file: config_opt('OTEL_EXPORTER_OTLP_SPAN_CERTIFICATE', 'OTEL_EXPORTER_OTLP_CERTIFICATE'),
+                       headers: config_opt('OTEL_EXPORTER_OTLP_SPAN_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS'), # TODO: what format is expected here?
```
fbogsany, line 35: @mwear this allows passing in headers to be added to each request. The format of the environment variable doesn't appear to be specified. What should we support here? I assume we also need to support explicitly passing a `Hash`?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -0,0 +1,277 @@
...
+              FAILURE
+            end
+          rescue Net::OpenTimeout, Net::ReadTimeout
+            retry if backoff?(retry_count: retry_count += 1)
+            return FAILURE
+          end
+        end
+
+        def handle_redirect(location)
+          # TODO: figure out destination and reinitialize @http and @path
```
fbogsany, line 139: Punting this to a later PR.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
ericmustin: I looked into this a bit. JS appears to accept only [an Object](https://github.com/open-telemetry/opentelemetry-js/blob/ff3ee0f6d1a3625e2dbc191501358181e99667fd/packages/opentelemetry-exporter-collector/src/platform/node/CollectorExporterNodeBase.ts#L43) (Hash, for all intents and purposes), and [does validation](https://github.com/open-telemetry/opentelemetry-js/blob/ff3ee0f6d1a3625e2dbc191501358181e99667fd/packages/opentelemetry-exporter-collector/src/util.ts#L25-L38), throwing out any malformed key/value pairs and coercing the values into strings

They don't have an env var option which, i would assume will have to be a string of comma separated `key:value` pairs, ie: `header_one:value_one,header_two:value_two`.

Optimistically i'd say we could support array of k/v pairs, hash, and csv separated string of k/v pairs. realistically it's probably hard to document and cumbersome to validate and we should maybe stick to some subset of those. CSV string and hash feels pretty flexible to me imo, but fine with any choices here.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -0,0 +1,275 @@
...
+
+        def encode(span_data) # rubocop:disable Metrics/MethodLength
+          Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.encode(
+            Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.new(
+              resource_spans: span_data
+                .group_by(&:resource)
+                .map do |resource, span_datas|
+                  Opentelemetry::Proto::Trace::V1::ResourceSpans.new(
+                    resource: Opentelemetry::Proto::Resource::V1::Resource.new(
+                      attributes: resource.label_enumerator.map { |key, value| as_otlp_key_value(key, value) }
```
ericmustin, line 177: (same comment applies to instrumentation_library mappings)

Can span's have `nil` values for `resource` and `instrumentation_library`? In that case will invoking `resource.label_enumerator.map` and `il.name` / `il.version` cause an exception? And I guess generally what's the expectation for how to handle span's without a resource or instrumentation library?

exporter/otlp/opentelemetry-exporter-otlp.gemspec
```diff
@@ -0,0 +1,40 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'google-protobuf', '>= 3.4.1.1', '< 4'
```
ericmustin, line 28: There appears to be some jruby wonkiness with this dependancy, it would seem? Have you had a chance to look into the test failures on jruby?

ericmustin (commented):
I had a few small questions about making sure we safely encode spans that may not have resource and instrumentation_library, and left suggestions about the custom header support, but i'm fine with any approach here, and if we can't reach consensus i'm also fine pushing this feature to a later PR/release. There also look to be one or two ENV Var tests that are still marked as TODOs, if you want to punt on those to a later PR i think that's completely fine as well.

I think this is very close imo, the only real blocker to me is that jruby appears to be having some dependancy issues with the added gems.

exporter/otlp/opentelemetry-exporter-otlp.gemspec
fbogsany: I took a brief look. It seems as though we need to install the Java protobuf package using maven. I’m tempted to not support jruby with this gem for now, but I’m not sure how to do that yet.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: I’ll go with CSV string and hash. I’ll try to get that wrapped up on Tuesday.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: For `resource`, you’d have to explicitly install the tracer provider with a `nil` resource. Both the configurator and the provider have non nil defaults. The tracer constructor always instantiates an InstrumentationLibrary, so that should be non nil. The Span constructor is considered a private API, so a Span should only be created by the Tracer. Technically, SpanData should only be created by Span. I think it is reasonable to assume that the instrumentation library can’t be nil. I’m less sure about the resource. I’ll re-read the spec to see if it always requires a resource - certainly, using the SDK as designed, the resource should be non nil.

exporter/otlp/opentelemetry-exporter-otlp.gemspec
genebean: FWIW, every app I have been using this gem with is running in JRuby

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
ericmustin: Did some light testing of this and as usual you are correct, I think this is fine, apologies for the extra work here. I had missed the [default `@resource` being set to `telemetry_sdk`](https://github.com/open-telemetry/opentelemetry-ruby/blob/181165c9761e013241fb9d6eb5f2eddd8473c7d1/sdk/lib/opentelemetry/sdk/configurator.rb#L30) that was added in v0.5.0 (which is great, btw!). I think we're probably fine here and if people are using the tracer in unholy ways/messing with private api's, they are on their own.

exporter/otlp/opentelemetry-exporter-otlp.gemspec
ericmustin: So I looked into this a bit, as Datadog supports jRuby and has a GRPC integration (which relies on this gem). And as It turns out we [don't support GRPC instrumentation on jruby](https://github.com/DataDog/dd-trace-rb/pull/1126#issue-457957841) since the google-protobuf gem itself does not support jruby (and grpc relies on this gem). It's impacting a lot of ruby library folks, it would seem. This is the issue in question, for context: https://github.com/protocolbuffers/protobuf/issues/1594

It's a long issue, and I can't say I have fully grok'd things, but tl;dr it sounds like there's not enough free hands right now within jruby/google to add support at the moment.

There might be some viable workarounds on the thread but it's not clear to me, someone on there mentioned a pure ruby implementation: https://github.com/ruby-protobuf/protobuf , but it doesn't seem very actively maintained at the moment. The reasonable path forward here to me would be not include jruby support for now, mostly because I am not clever enough to know how to work around this.

Fwiw there are definitely some large enterprises out there who are both JRuby users and users of tracing products.

fbogsany: To be clear, I won’t remove jruby support from OTel Ruby in general, just from the OTLP exporter. The Jaeger exporter should be a viable alternative for jruby shops until we can get protobuf working.

genebean: Based on that, I guess it's a really good thing https://github.com/open-telemetry/opentelemetry-ruby/issues/231 is already in the upcoming milestone. I think you are right that this should provide a viable alternative and, for me, may actually simplify things since my local dev environments are using local Jaeger instances for validation of trace data.

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
```diff
@@ -0,0 +1,252 @@
...
+        end
+
+        private
+
+        def send_bytes(bytes)
+          retry_count = 0
+          untraced do
+            request = Net::HTTP::Post.new(@path)
+            request.body = bytes
+            request.add_field('Content-Type', 'application/x-protobuf')
```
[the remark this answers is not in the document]
fbogsany: Done

exporters/otlp/lib/opentelemetry/exporters/otlp/exporter.rb
fbogsany: Done

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: Done

fbogsany:
I think I've done everything I said I would, except for:
> rename the gem to `opentelemetry-exporter-otlp-http`

I honestly lack the energy to do this. I'd rather merge what we have here and take some time afterwards to talk through how to handle the [`OTEL_EXPORTER_OTLP_PROTOCOL` config option](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/protocol/exporter.md#configuration-options). I'm less convinced now that the gem rename is the right approach (but I also don't know what the "right thing" is).

ericmustin (approved):
> I'd rather merge what we have here and take some time afterwards to talk through how to handle the OTEL_EXPORTER_OTLP_PROTOCOL config option. I'm less convinced now that the gem rename is the right approach (but I also don't know what the "right thing" is).

Concur, let's not let perfect be the enemy of good as this is very robust.

PR #280, opened 2020-06-22
Spec compliance for span and trace IDs

### LINKED ISSUE #232: Spec compliance for in-memory representation of span and trace IDs
The [specification for `SpanContext`](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#spancontext) requires that:

> `TraceId` A valid trace identifier is a 16-byte array with at least one non-zero byte.
>
> `SpanId` A valid span identifier is an 8-byte array with at least one non-zero byte.

Our representation is 16 and 32 character hex strings:
https://github.com/open-telemetry/opentelemetry-ruby/blob/79b047a38d43782bf0733c7338a5c00aec92c278/api/lib/opentelemetry/trace.rb#L11-L42

We should conform to the spec. This actually makes a few things better, including fewer allocations overall and making translation to things that *don't* use hex strings much easier to understand and prove correct.

fbogsany:
Fixes #232

### DIFF
--- a/adapters/ethon/test/opentelemetry/adapters/ethon/adapter_test.rb
+++ b/adapters/ethon/test/opentelemetry/adapters/ethon/adapter_test.rb
@@ -99,7 +99,7 @@ describe OpenTelemetry::Adapters::Ethon::Adapter do
             _(easy.instance_eval { @otel_span }).must_be_nil
             _(
               easy.instance_eval { @otel_original_headers['traceparent'] }
-            ).must_equal "00-#{span.trace_id}-#{span.span_id}-01"
+            ).must_equal "00-#{span.trace_id.unpack1('H*')}-#{span.span_id.unpack1('H*')}-01"
           end
         end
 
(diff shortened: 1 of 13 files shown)


adapters/ethon/test/opentelemetry/adapters/ethon/adapter_test.rb
```diff
@@ -99,7 +99,7 @@ def stub_response(options)
             _(easy.instance_eval { @otel_span }).must_be_nil
             _(
               easy.instance_eval { @otel_original_headers['traceparent'] }
-            ).must_equal "00-#{span.trace_id}-#{span.span_id}-01"
+            ).must_equal "00-#{span.trace_id.unpack1('H*')}-#{span.span_id.unpack1('H*')}-01"
```
fbogsany, line 102: This is a very common pattern in the adapter tests. It’s worth thinking about extracting this to a helper in the API or SDK. It *can* be accomplished with something like:
```ruby
OpenTelemetry::Trace::Propagation::TraceContext::TraceParent.from_context(span).to_s
```
but that’s pretty wordy, may not work exactly like this, and doesn’t really reveal the intent terribly well. Something like the following would be better:
```ruby
OpenTelemetry::Trace.to_traceparent(trace_id: span.trace_id, span_id: span.span_id, sampled: true)
```
I guess that’s even longer, but I think it is clearer.

mwear: +1 on wanting a helper. TBH, I like the first option better, because the second implicitly couples the Trace module to W3C trace context. Both are wordy, and I could see sticking it somewhere else altogether. 

A related issue is that we don't have a good way to share a) test helpers, b) shared logic between adapters. We should think about how we want to solve that moving forward.

fbogsany:
👍 I created #284 to track this.

PR #282, opened 2020-06-24
add sampler to SpanData

ericmustin:
### Summary

This PR adds the span's `trace_config.sampler` to the SpanData struct. The motivation behind this is that an exporter may be required to know the type of sampling that has been applied to span's to be exported, as some sampler's will choose to neither sample nor record certain spans, and those spans will not be passed to a span processor (and, by extension, the exporter). By making the sampler available to the SpanData struct, this information can be read/parsed from the `sampler.description`.

Specifically, the ProbabilitySampler [does not record span's it does not sample](https://github.com/open-telemetry/opentelemetry-ruby/blob/a76c4285729caee021fb5557b21f696f62ba4c6d/sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb#L35). So knowing the `probability` rate which of spans it did choose to sample, [contained in the description](https://github.com/open-telemetry/opentelemetry-ruby/blob/a76c4285729caee021fb5557b21f696f62ba4c6d/sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb#L23), can help vendor specific exporter's convey information so that they can then more accurately "upscale" metrics related to the exported span's, such as "hits" or "errors".

This PR relates to a conversation in a related PR, found here: https://github.com/open-telemetry/opentelemetry-ruby/pull/273#issuecomment-647854791

### Notes

- If there is some alternative way to access the sampler description in a span processor or exporter I would be open to using that approach, this was the only way I could think of. I would prefer to avoid bloating the SpanData struct with unnecessary info, but do believe that information about the sampler's logic which created the span could be useful at export time.

### DIFF
--- a/exporters/jaeger/.rubocop.yml
+++ b/exporters/jaeger/.rubocop.yml
@@ -10,7 +10,7 @@ Metrics/AbcSize:
 Metrics/LineLength:
   Enabled: false
 Metrics/MethodLength:
-  Max: 21
+  Max: 22
 Metrics/ParameterLists:
   Enabled: false
 Naming/FileName:
--- a/exporters/jaeger/test/opentelemetry/exporters/jaeger/exporter/span_encoder_test.rb
+++ b/exporters/jaeger/test/opentelemetry/exporters/jaeger/exporter/span_encoder_test.rb
@@ -60,7 +60,7 @@ describe OpenTelemetry::Exporters::Jaeger::Exporter::SpanEncoder do
     )
   end
 
-  def create_span_data(attributes: nil, events: nil, links: nil, trace_id: OpenTelemetry::Trace.generate_trace_id, trace_flags: OpenTelemetry::Trace::TraceFlags::DEFAULT, tracestate: nil)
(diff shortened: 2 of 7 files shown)


fbogsany:
The spec provides no guidance on the content of SpanData (in fact, it only requires than an exportable representation of a span exist in the SDK). The `trace_config` and related information are not readable from the Java SDK’s representation, for example. I don’t think we should add them here without a discussion at the spec level — if you need this here, then you’ll need it in SDKs for other languages too.

That said, I’m not sure you *do* need it. The SDK spec requires 3 built-in Samplers, but you’re free to provide your own and to recommend use of your own Sampler implementation to your customers. In particular, you could ensure that your Sampler implementation never returns `NOT_RECORD`, in which case your Span Processor should always receive all spans and your Span Exporter will only receive sampled spans. The downside of that is that instrumentation will never have the option of avoiding expensive computation. I don’t know that we’re leveraging that in any auto-instrumentation yet.

ericmustin:
@mwear @fbogsany k, fair enough. i'll close this for now, as frankly i'm just looking for the lowest friction path forward here. in this case it sounds like including a `<VendorSpecific>ProbabilitySampler`,  (likely just inheriting from the `ProbabilitySampler` ) accomplishes that.

PR #283, opened 2020-06-24
prepare adapter bugfix release

mwear:
This PR bumps the versions of the sinatra adapter and the adapters-all gem. The bugfix release will include the changes introduced in #275.

cc: @mat-rumian


mat-rumian:
Thanks a lot @mwear for notification :) Already checked, works fine 👍

Issue #300, opened 2020-07-20
uninitialized constant OpenTelemetry::Instrumentation::Base


PuntoDiGoccia:
I'm trying auto-inspection on `rails`

`Gemfile`:
```
gem 'opentelemetry-api'
gem 'opentelemetry-sdk'
gem 'opentelemetry-exporters-jaeger'
gem 'opentelemetry-instrumentation-all'
```

`application.rb`:
```
require 'opentelemetry/api`
require 'opentelemetry/sdk'
require 'opentelemetry/exporters/jaeger'
require 'opentelemetry/instrumentation/all`
```

`bin/rails server` fails with the message in subject.
Looking at the stack, RubyGem gets confused: 
in 
`opentelemetry-api-0.5.0/lib/opentelemetry.rb`: `require 'opentelemetry/instrumentation'` (line 14)
loads 
`opentelemetry-instrumentation-sinatra-0.5.0/lib/opentelemetry/instrumentation.rb` 
instead of 
`opentelemetry-api-0.5.0/lib/opentelemetry/instrumentation.rb`

Possibly easiest solution is to change `require` -> `require_relative`

fbogsany:
Fix in #301. Thanks for the report! I reproduced the failure in a small test app & verified the fix. Note that your first `require` should be `require 'opentelemetry'` rather than `require 'opentelemetry/api'`.

mwear:
Fix released in https://rubygems.org/gems/opentelemetry-api/versions/0.5.1

Issue #298, opened 2020-07-18
ProbabilitySampler.sample_trace_id? should account for spec compliant trace_id


ericmustin:
I believe that https://github.com/open-telemetry/opentelemetry-ruby/pull/280 introduced an edge case bug in the `sample_trace_id?` private method on `ProbabilitySampler`
https://github.com/open-telemetry/opentelemetry-ruby/blob/a76c4285729caee021fb5557b21f696f62ba4c6d/sdk/lib/opentelemetry/sdk/trace/samplers/probability_sampler.rb#L58

Where, for a `ProbabilitySampler` with sampling rate < 1.0, the `sample_trace_id?` method will always return `true`. I believe this is because the trace_id is not in the expected format for the comparison to work as intended, and the LOC needs to be updated something like:

`@probability == 1.0 || trace_id.unpack1('H*')[16, 16] < @id_upper_bound `

I may be mistaken but some testing locally seems to support this, just flagging this now (sorry to notice this only just after v0.5 release 😢 !)

fbogsany:
Thanks very much for reporting this @ericmustin! I'll fix this ASAP (tonight or tomorrow). We'll need a 0.5.1 release of the SDK for this, unfortunately - sorry @mwear! 😞

mwear:
Fix released in: https://rubygems.org/gems/opentelemetry-sdk/versions/0.5.1

PR #302, opened 2020-07-20
prepare for 0.5.1 release

mwear:
This PR bumps the API and SDK versions and adds changelog entires in preparation for a patch release that will address #298 and #300. After this merges I can cut a new gem versions for opentelemetry-api and opentelemetry-sdk.


mwear:
I realized we also need a patch release of the SDK. I'll merge #299, and update this to bump versions for both the API and SDK.

fbogsany:
Thanks @mwear !

PR #316, opened 2020-08-12
Fix serialized SpanData's span_id and trace_id

Asafb26:
in version 0.5.1 of opentelemetry-ruby we changed the way we generate the `trace_id` and `span_id`
now the correspondent generators return a raw random binary string which not unpacked.
I'm not sure about the effect of this change for the entire system, but its effect at least for some of the SpanExporters which fail to serialize it.
this PR should fix the SpanData serialization by using `unpack1('H*')` for those ids, but we may do it in a higher level (maybe in `SpanContext` initializer?)

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -238,8 +238,8 @@ module OpenTelemetry
             @events,
             @library_resource,
             @instrumentation_library,
-            context.span_id,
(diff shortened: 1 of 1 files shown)


fbogsany:
This should actually be fixed at a lower level, in the span exporters. The in-memory format is specified to be 8 bytes for the span ID and 16 bytes for the trace ID. This format should be preserved all the way through SpanData to the exporters, which can then translate it appropriately.

PR #320, opened 2020-08-17
Add rake task for resource detectors

robertlaurin:
- The original PR I forgot to add in the rake tasks 🤦 
- The rake tasks folder naming assumptions break with the name `resource-detectors` https://github.com/open-telemetry/opentelemetry-ruby/blob/master/Rakefile#L209 so I've renamed it to `resource_detectors`


Rakefile
```diff
@@ -57,6 +57,12 @@ GEM_INFO = {
       OpenTelemetry::Exporters::Jaeger::VERSION
     }
   },
+  "opentelemetry-resource_detectors" => {
+    version_getter: ->() {
+      require './lib/opentelemetry/resource-detectors/version.rb'
```
fbogsany, line 62: Should this be `resource_detectors/version.rb` or `resource/detectors/version.rb`? I'm a little confused with all the renaming below.

robertlaurin: Definitely the latter, fixed.

PR #323, opened 2020-08-19
Implement GetCorrelations

### LINKED ISSUE #315: Retrieve all available correlations for a context
Hi,
I have a case which needed to retrieve a hash with all available correlations assigned for the current context,
I see that OpenTelemetry API and the SDK not supported to retrieve this but only a specific key.
Is there a reason behind this decision? can I archive it?

fbogsany:
Fixes #315

### DIFF
--- a/api/lib/opentelemetry/correlation_context/manager.rb
+++ b/api/lib/opentelemetry/correlation_context/manager.rb
@@ -9,7 +9,8 @@ module OpenTelemetry
     # No op implementation of CorrelationContext::Manager
     class Manager
       NOOP_BUILDER = Builder.new
-      private_constant :NOOP_BUILDER
+      EMPTY_VALUES = {}.freeze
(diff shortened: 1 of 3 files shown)


api/lib/opentelemetry/correlation_context/manager.rb
```diff
@@ -24,6 +25,10 @@ def value(key, context: Context.current)
         nil
       end
 
+      def values(context: Context.current)
```
fbogsany, line 28: I called this `values` to line up with `set_value`, `value`, `remove_value`, but `correlations` may be more appropriate.

WDYT @mwear?

mwear: I think `values` is the perfect choice. The correlation context manager is accessible through the `OpenTelemetry` module as `OpenTelemetry.correlations`. So, this would end up being `OpenTelemetry.correlations.values`, which is much better than `OpenTelemetry.correlations.correlations`.

PR #324, opened 2020-08-19
Fix gemspec name for resource detector

robertlaurin:
Missing conventional ruby file matching gem name in the `/lib` folder.

### DIFF
--- a/resource_detectors/lib/opentelemetry-resource_detectors.rb
+++ b/resource_detectors/lib/opentelemetry-resource_detectors.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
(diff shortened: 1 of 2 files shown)


resource_detectors/opentelemetry-resource-detectors.gemspec
```diff
@@ -9,7 +9,7 @@ $LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require 'opentelemetry/resource/detectors/version'
 
 Gem::Specification.new do |spec|
-  spec.name        = 'opentelemetry-resource-detectors'
+  spec.name        = 'opentelemetry-resource_detectors'
```
fbogsany, line 12: Does the name of this file need to change too?

robertlaurin: It does! Updated.

PR #333, opened 2020-08-24
Fix exception semantic conventions

fbogsany:
Implement https://github.com/open-telemetry/opentelemetry-specification/pull/765 and conform to https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md


sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -127,11 +127,11 @@ def add_event(name: nil, attributes: nil, timestamp: nil)
         #
         # @return [void]
         def record_error(error)
```
mwear, line 129: Should we update the method name to the the spec'd `record_exception` as well: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#record-exception

PR #338, opened 2020-08-26
Rename CorrelationContext to Baggage

fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/pull/857


api/lib/opentelemetry/baggage/propagation/context_keys.rb
```diff
@@ -18,7 +18,7 @@ module ContextKeys
         # Returns the context key that correlations are indexed by
         #
         # @return [Context::Key]
-        def correlation_context_key
+        def baggage_key
           CORRELATION_CONTEXT_KEY
```
robertlaurin, line 22: Would it be fitting to rename this as well?

sdk/lib/opentelemetry/sdk/baggage/manager.rb
```diff
@@ -6,10 +6,10 @@
 
 module OpenTelemetry
   module SDK
-    module CorrelationContext
+    module Baggage
       # Manages correlation context
       class Manager
-        CORRELATION_CONTEXT_KEY = OpenTelemetry::CorrelationContext::Propagation::ContextKeys.correlation_context_key
+        CORRELATION_CONTEXT_KEY = OpenTelemetry::Baggage::Propagation::ContextKeys.baggage_key
```
robertlaurin, line 12: Similar to the above, is there rational I'm missing for keeping the naming correlation context here?

api/lib/opentelemetry/baggage/propagation/text_map_extractor.rb
```diff
@@ -7,41 +7,41 @@
...
         # Returns a new TextMapExtractor that extracts context using the specified
         # header key
         #
-        # @param [String] correlation_context_key The correlation context header
+        # @param [String] baggage_key The baggage header
         #   key used in the carrier
         # @return [TextMapExtractor]
-        def initialize(correlation_context_key: 'otcorrelations')
-          @correlation_context_key = correlation_context_key
+        def initialize(baggage_key: 'Baggage')
```
mwear, line 22: It's worth mentioning that this change is still under discussion: https://github.com/open-telemetry/opentelemetry-specification/issues/879. It will probably go through, although, I believe we should go with all lowercase `baggage`, as the canonical key. For http it doesn't matter, but for case-sensitive scenarios I believe we want to standardize on using all lower case.

api/lib/opentelemetry/baggage/propagation/text_map_injector.rb
```diff
@@ -7,45 +7,45 @@
...
         # Returns a new TextMapInjector that injects context using the specified
         # header key
         #
-        # @param [String] correlation_context_header_key The correlation context header
+        # @param [String] baggage_header_key The baggage header
         #   key used in the carrier
         # @return [TextMapInjector]
-        def initialize(correlation_context_key: 'otcorrelations')
-          @correlation_context_key = correlation_context_key
+        def initialize(baggage_key: 'Baggage')
```
mwear, line 22: same comment regarding the header name.

api/lib/opentelemetry/baggage/propagation/text_map_extractor.rb
fbogsany: I speculated that the change would go through. If it's rejected, I can change back. I canonicalized the key.

PR #341, opened 2020-08-26
feat!: add timeout for force_flush and shutdown

fbogsany:
[The spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/sdk.md#span-processor) allows language implementers to support an optional timeout parameter for `ForceFlush` and `Shutdown`, for Span Exporters and Span Processors. E.g.

> Language library authors can decide if they want to make the flush timeout configurable.

We've implicitly expected users to leverage `Timeout.timeout` for this purpose, however @robertlaurin and I were implementing this in Shopify's OTLP exporter, and it requires dealing with a ton of corner cases (since the `Timeout::Error` can be raised anywhere). If we instead add the optional `timeout` parameter and pass it down the chain (updating appropriately to account for passing time), we have much more control over where exactly we handle timeouts.

In many respects, this is more important for `force_flush`, where we want to ensure that internal state remains consistent if a timeout occurs, but it will probably not be uncommon for exporters to call `force_flush` as part of `shutdown`.

This PR adds an optional `timeout:` parameter to both methods and propagates that through various call chains.

## TODO:
- [x] Add tests for timeouts

### DIFF
--- a/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
+++ b/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
@@ -30,12 +30,14 @@ module OpenTelemetry
         # @param [Enumerable<OpenTelemetry::SDK::Trace::SpanData>] span_data the
         #   list of recorded {OpenTelemetry::SDK::Trace::SpanData} structs to be
         #   exported.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] the result of the export.
-        def export(span_data)
+        def export(span_data, timeout: nil)
           return FAILURE if @shutdown
 
+          start_time = Time.now
           encoded_batches(span_data) do |batch|
-            return FAILURE if @shutdown
+            return FAILURE if @shutdown || OpenTelemetry::SDK::Internal.maybe_timeout(timeout, start_time)&.zero?
 
             @client.emitBatch(batch)
           end
@@ -44,7 +46,9 @@ module OpenTelemetry
         # Called when {OpenTelemetry::SDK::Trace::Tracer#shutdown} is called, if
         # this exporter is registered to a {OpenTelemetry::SDK::Trace::Tracer}
         # object.
-        def shutdown
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        def shutdown(timeout: nil)
           @shutdown = true
         end
 
--- a/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
+++ b/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
@@ -37,7 +37,7 @@ module OpenTelemetry
 
           start_time = Time.now
           encoded_batches(span_data) do |batch|
-            return FAILURE if @shutdown || OpenTelemetry::SDK::Internal.maybe_timeout(timeout, start_time)&.zero?
+            return FAILURE if @shutdown || OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)&.zero?
 
             @client.emitBatch(batch)
           end
--- a/exporter/jaeger/lib/opentelemetry/exporter/jaeger/collector_exporter.rb
+++ b/exporter/jaeger/lib/opentelemetry/exporter/jaeger/collector_exporter.rb
@@ -36,8 +36,9 @@ module OpenTelemetry
         # @param [Enumerable<OpenTelemetry::SDK::Trace::SpanData>] span_data the
         #   list of recorded {OpenTelemetry::SDK::Trace::SpanData} structs to be
         #   exported.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] the result of the export.
-        def export(span_data)
+        def export(span_data, timeout: nil)
           return FAILURE if @shutdown
 
           encoded_batches(span_data).each do |batch|
@@ -56,7 +57,9 @@ module OpenTelemetry
         # Called when {OpenTelemetry::SDK::Trace::Tracer#shutdown} is called, if
         # this exporter is registered to a {OpenTelemetry::SDK::Trace::Tracer}
         # object.
-        def shutdown
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        def shutdown(timeout: nil)
           @shutdown = true
         end
 
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -25,10 +25,8 @@ module OpenTelemetry
 
         # Default timeouts in seconds.
         KEEP_ALIVE_TIMEOUT = 30
-        OPEN_TIMEOUT = 5
-        READ_TIMEOUT = 5
         RETRY_COUNT = 5
-        private_constant(:KEEP_ALIVE_TIMEOUT, :OPEN_TIMEOUT, :READ_TIMEOUT, :RETRY_COUNT)
+        private_constant(:KEEP_ALIVE_TIMEOUT, :RETRY_COUNT)
 
         def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_SPAN_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: 'localhost:55681/v1/trace'), # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
                        insecure: config_opt('OTEL_EXPORTER_OTLP_SPAN_INSECURE', 'OTEL_EXPORTER_OTLP_INSECURE', default: false),
@@ -45,8 +43,6 @@ module OpenTelemetry
           @http.use_ssl = insecure.to_s.downcase == 'false'
           @http.ca_file = certificate_file unless certificate_file.nil?
           @http.keep_alive_timeout = KEEP_ALIVE_TIMEOUT
-          @http.open_timeout = OPEN_TIMEOUT
-          @http.read_timeout = READ_TIMEOUT
 
           @path = uri.path
           @headers = case headers
@@ -64,17 +60,20 @@ module OpenTelemetry
         # @param [Enumerable<OpenTelemetry::SDK::Trace::SpanData>] span_data the
         #   list of recorded {OpenTelemetry::SDK::Trace::SpanData} structs to be
         #   exported.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] the result of the export.
-        def export(span_data)
+        def export(span_data, timeout: nil)
           return FAILURE if @shutdown
 
-          send_bytes(encode(span_data))
+          send_bytes(encode(span_data), timeout: timeout)
         end
 
         # Called when {OpenTelemetry::SDK::Trace::Tracer#shutdown} is called, if
         # this exporter is registered to a {OpenTelemetry::SDK::Trace::Tracer}
         # object.
-        def shutdown
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        def shutdown(timeout: nil)
           @shutdown = true
           @http.finish if @http.started?
         end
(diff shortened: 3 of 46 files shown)


ericmustin (commented):
👍 Lgtm, can approve when the Todo to update the processor tests is added.

fbogsany:
> Same here. I meant to ask about this at the SIG meeting. Will tests be part of this PR?

Yep. We've built out the same mechanism in the OTLP exporter from our internal instrumentation gem. We're going to write the tests there and then port them here when we're happy with them.

fbogsany:
I've updated this - I had to battle with a ton of merge conflicts 😞 . I'll take another look through it and see what tests we're missing.

I've fixed some problems @robertlaurin and I identified in our internal implementation of this, particularly adding a separate export mutex around the `@exporter.export` call in `BatchSpanProcessor` - the contract states that there will never be concurrent calls to `export`, but that was possible if someone called `force_flush` or `shutdown` from another thread.

This also modifies the signature of `Exporter.export` to take an optional `timeout:` parameter. This is not in the spec, but I believe it is necessary to implement `force_flush` with a timeout, without resorting to `Timeout.timeout` (which is what we're trying to avoid here).

exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
```diff
@@ -30,12 +30,14 @@ def initialize(host: ENV.fetch('OTEL_EXPORTER_JAEGER_AGENT_HOST', 'localhost'),
...
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] the result of the export.
-        def export(span_data)
+        def export(span_data, timeout: nil)
           return FAILURE if @shutdown
 
+          start_time = Time.now
           encoded_batches(span_data) do |batch|
-            return FAILURE if @shutdown
+            return FAILURE if @shutdown || OpenTelemetry::SDK::Internal.maybe_timeout(timeout, start_time)&.zero?
```
fbogsany, line 40: Not super awesome that we're reaching into an `Internal` SDK module here. Would we prefer duplication in each exporter gem?

exporter/jaeger/lib/opentelemetry/exporter/jaeger/collector_exporter.rb
```diff
@@ -37,8 +37,9 @@ def initialize(endpoint: ENV.fetch('OTEL_EXPORTER_JAEGER_ENDPOINT', 'http://loca
         # @param [Enumerable<OpenTelemetry::SDK::Trace::SpanData>] span_data the
         #   list of recorded {OpenTelemetry::SDK::Trace::SpanData} structs to be
         #   exported.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] the result of the export.
-        def export(span_data)
+        def export(span_data, timeout: nil)
```
fbogsany, line 42: Note that we ignore the `timeout:` parameter here because we don't have a clean way to reach into the transport to adjust lower-level timeouts. I think that's a good "future work" item to figure out, but I don't think it should be in scope for this PR.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -107,8 +106,10 @@ def invalid_url?(url)
           true
         end
 
-        def send_bytes(bytes) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
+        def send_bytes(bytes, timeout:) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
           retry_count = 0
+          timeout ||= @timeout
```
fbogsany, line 111: It is possible for `timeout:` to be `nil` here (e.g. if `export` is called from `SimpleSpanProcessor`), in which case we use the configured timeout, which defaults to 10 seconds.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -117,6 +118,11 @@ def send_bytes(bytes) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCompl
             # TODO: enable gzip when https://github.com/open-telemetry/opentelemetry-collector/issues/1344 is fixed.
             # request.add_field('Content-Encoding', 'gzip')
 
+            remaining_timeout = OpenTelemetry::SDK::Internal.maybe_timeout(timeout, start_time)
+            return FAILURE if remaining_timeout.zero?
```
fbogsany, line 122: The spec only allows `SUCCESS` and `FAILURE` - there's no differentiation between `TIMEOUT` and `FAILURE`, unlike `force_flush` and `shutdown`.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -117,6 +118,11 @@ def send_bytes(bytes) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCompl
...
+
+            @http.open_timeout = remaining_timeout
+            @http.read_timeout = remaining_timeout
```
fbogsany, line 125: We would like to set `@http.write_timeout` here too, but that isn't available in Ruby 2.5. Shopify's internal implementation has a version check to allow setting that for Ruby versions > 2.5. Do we want to do the same thing here?

sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb
```diff
@@ -23,36 +23,34 @@ def initialize(span_exporters)
...
-              merge_result_code(result_code, span_exporter.export(spans))
+          def export(spans, timeout: nil)
+            start_time = Time.now
+            results = @span_exporters.map do |span_exporter|
+              span_exporter.export(spans, timeout: Internal.maybe_timeout(timeout, start_time))
             rescue => e # rubocop:disable Style/RescueStandardError
               OpenTelemetry.logger.warn("exception raised by export - #{e}")
               FAILURE
             end
+            results.uniq.max || SUCCESS
```
fbogsany, line 36: The `|| SUCCESS` applies if `@span_exporters` is empty. I'd be happy to check that and return early instead.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -254,6 +254,7 @@ def to_span_data
       let(:exporter_sleeps_for_millis) { exporter_timeout_millis + 700 }
 
       it 'is interrupted by a timeout' do
+        skip # TODO: fix the timeout testing
```
fbogsany, line 257: The current test explicitly assume `Timeout.timeout` in the implementation. I'll fix this to be more useful.

exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
robertlaurin: Is this something that could live in the soon to exist "utility gem"?  However simple the method may be, we probably want to keep it consistent across exporters right?  If not then I would just duplicate it into each gem.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: I think it makes sense to add it if we're diligent enough to drop the version check when 2.5 hits EoL.

exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
fbogsany: I think that’s a good place for it, yes.

robertlaurin: Should we pull this into `common`?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -25,10 +25,9 @@ class Exporter # rubocop:disable Metrics/ClassLength
 
         # Default timeouts in seconds.
         KEEP_ALIVE_TIMEOUT = 30
-        OPEN_TIMEOUT = 5
-        READ_TIMEOUT = 5
         RETRY_COUNT = 5
-        private_constant(:KEEP_ALIVE_TIMEOUT, :OPEN_TIMEOUT, :READ_TIMEOUT, :RETRY_COUNT)
+        WRITE_TIMEOUT_SUPPORTED = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
```
robertlaurin, line 29: Added ruby version check for write_timeout support.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: You've since added `TIMEOUT`, I think I should update this now right?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: Added write_timeout.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -152,6 +160,11 @@ def send_bytes(bytes) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticCompl
             retry if backoff?(retry_count: retry_count += 1)
             return FAILURE
           end
+        ensure
+          # Reset timeouts to defaults for the next call.
```
robertlaurin, line 164: I think we want to port this over from our own internal implementation.

robertlaurin (approved):
Approving my own changes 😄 

If either of you want to take a look I think this is good to go now @mwear @ericmustin

exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb
robertlaurin: Does it make sense to have the SDK depend on `common`?

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -115,6 +116,31 @@
       _(result).must_equal(SUCCESS)
     end
 
+    it 'returns FAILURE on timeout' do
```
fbogsany, line 119:
```suggestion
    it 'returns TIMEOUT on timeout' do
```

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -115,6 +116,31 @@
...
+      stub_request(:post, 'https://localhost:55681/v1/trace').to_return(status: 200)
+      span_data = create_span_data
+      result = exporter.export([span_data], timeout: 0)
+      _(result).must_equal(TIMEOUT)
+    end
+
+    it 'returns FAILURE on timeout after retrying' do
```
fbogsany, line 126:
```suggestion
    it 'returns TIMEOUT on timeout after retrying' do
```

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
robertlaurin: I think this test case can just be completely removed.

sdk/lib/opentelemetry/sdk/trace/export/multi_span_exporter.rb
```diff
@@ -23,36 +23,34 @@ def initialize(span_exporters)
...
-          def merge_result_code(result_code, new_result_code)
-            if result_code == SUCCESS && new_result_code == SUCCESS
-              # If both errors are success then return success.
-              SUCCESS
-            else
-              # At this point at least one of the code is FAILURE, so return FAILURE.
-              FAILURE
+              processor.shutdown(timeout: remaining_timeout)
             end
+            results.uniq.max
```
fbogsany, line 53:
```suggestion
            results.uniq.max || SUCCESS
```

fbogsany (commented):
I can't approve my own PR 😄 . This LGTM, though. I think there are just some Rubocop offences to clean up.

mwear (approved):
LGTM

PR #343, opened 2020-08-26
Fix docs for array valued attributes

### LINKED ISSUE #342: Add support for arrays of primitives as attribute values
This looks like it is just a documentation issue. We need to update https://github.com/open-telemetry/opentelemetry-specification/blob/master/spec-compliance-matrix.md as well - based on the incorrect documentation 😄 , I marked this as not supported.

fbogsany:
Fixes #342


api/lib/opentelemetry.rb
```diff
@@ -9,7 +9,6 @@
 require 'opentelemetry/error'
 require 'opentelemetry/context'
 require 'opentelemetry/baggage'
-require 'opentelemetry/internal'
```
fbogsany, line 12: I deleted the API Internal module because nothing uses it.

Issue #336, opened 2020-08-25
Update resource autodetection to match spec


fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/pull/811. I think the main thing we're not handling at the moment is the `OTEL_RESOURCE_ATTRIBUTES` env var.

PR #347, opened 2020-08-27
Environment variable resource detection

robertlaurin:
Closes: https://github.com/open-telemetry/opentelemetry-ruby/issues/336


resource_detectors/.yardopts
```diff
@@ -0,0 +1,8 @@
+--no-private
+--title=OpenTelemetry Resource Detectors
+--markup=markdown
+--main=README.md
+./lib/opentelemetry/resource/detectors/**/*.rb
+./lib/opentelemetry/resource/detectors.rb
+-
+README.md
```
robertlaurin, line 8: Following the great comment left here :) https://github.com/open-telemetry/opentelemetry-ruby/pull/326#issuecomment-678486553

robertlaurin:
>The SDK MUST extract information from the `OTEL_RESOURCE_ATTRIBUTES` environment
variable and merge this, as the secondary resource, with any resource
information provided by the user, i.e. **the user provided resource information
has higher priority.**

Re-reading [this](https://github.com/open-telemetry/opentelemetry-specification/pull/811/files#diff-b01b7033ffecc15c531ea83fa6248619R99) I'm realizing I haven't quite brought this into spec compliance yet so I'm going to move this into draft.

mwear:
I interpret the spec you referenced to mean that the environment variable detector should be part of the SDK, rather than the detectors package. Does that match your understanding?

robertlaurin:
> I interpret the spec you referenced to mean that the environment variable detector should be part of the SDK, rather than the detectors package. Does that match your understanding?

Yeah there are two parts here, this sounds like it should be part of the SDK *and* there needs to be priority assigned to user provided resource info which I've not made any consideration for.  

What qualifies as user provided resource info exactly?

fbogsany:
> What qualifies as user provided resource info exactly?

Explicitly configured resources, so:
```ruby
  c.resource = OpenTelemetry::SDK::Resources::Resource.create('answer' => 42)
end
```

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -36,11 +36,12 @@ def logger
...
       # Accepts a resource object that is merged with the default telemetry sdk
       # resource. The use of this method is optional, and is provided as means
-      # to add additional resource information.
+      # to include additional resource information.
+      # If a resource key collision occurs the passed in resource takes priority.
       #
       # @param [Resource] new_resource The resource to be merged
       def resource=(new_resource)
-        @resource = @resource.merge(new_resource)
+        @resource = new_resource.merge(@resource)
```
robertlaurin, line 44: Had the flip it so the supplied resource takes precedence.

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -31,11 +31,22 @@ def create(labels = {})
...
           def telemetry_sdk
-            create(
+            resource_labels = {
               Constants::TELEMETRY_SDK_RESOURCE[:name] => 'opentelemetry',
               Constants::TELEMETRY_SDK_RESOURCE[:language] => 'ruby',
               Constants::TELEMETRY_SDK_RESOURCE[:version] => OpenTelemetry::SDK::VERSION
-            )
+            }
+
+            resource_pairs = ENV['OTEL_RESOURCE_ATTRIBUTES']
```
robertlaurin, line 40: I think it makes sense to have this baked into the default resource that gets created during configuration.

sdk/test/test_helper.rb
```diff
@@ -12,3 +12,13 @@
 require 'minitest/autorun'
 
 OpenTelemetry.logger = Logger.new('/dev/null')
+
+def with_env(new_env)
```
robertlaurin, line 16: I moved this into the test_helper as it is used in a few places now.

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: 🤔 is that true? Maybe I need to understand the sequence of initialization...

fbogsany: What does the call sequence look like for users? This?
```ruby
  c.resource = OpenTelemetry::Resource::Detectors::AutoDetector.detect
  c.resource = OpenTelemetry::SDK::Resources::Resource.create('foo' => 'bar')
end
```

I am always confused by the difference between `Resource#merge` and `Hash#merge` 😞

robertlaurin: Let me add a test for it so we can see if it's doing the right thing.

PR #349, opened 2020-08-28
Safely navigate span variable during error cases

sribalakumar:
On facing issue https://github.com/open-telemetry/opentelemetry-ruby/issues/352, the noticed error at runtime is

`NoMethodError` while the actual error should be `SystemStackError` :

```
/otel-ruby/bundler_gems/ruby/2.6.3/bundler/gems/opentelemetry-ruby-4c8d39f1e592/lib/opentelemetry/trace/tracer.rb:64:in `ensure in in_span': undefined method `finish' for nil:NilClass (NoMethodError) from
/otel-ruby/bundler_gems/ruby/2.6.3/bundler/gems/opentelemetry-ruby-4c8d39f1e592/lib/opentelemetry/trace/tracer.rb:64:in `in_span'
......
```

`span` variable is not defined when the error happens in the middle of `start_span` call.

So introduce `&` safety navigation operator for `span` variable access in the error and ensure block.


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -55,13 +55,13 @@ def active_span_context(context = nil)
...
         with_span(span) { |s, c| yield s, c }
-      rescue Exception => e # rubocop:disable Lint/RescueException
+      rescue => e # rubocop:disable Style/RescueStandardError
         span.record_exception(e)
         span.status = Status.new(Status::UNKNOWN_ERROR,
                                  description: "Unhandled exception of type: #{e.class}")
         raise e
       ensure
-        span.finish
+        span.finish if defined?(span)
```
fbogsany, line 64: Under what circumstances will `span` **not** be defined here?

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -55,13 +55,13 @@ def active_span_context(context = nil)
       def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, with_parent: nil, with_parent_context: nil)
         span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, with_parent: with_parent, with_parent_context: with_parent_context)
         with_span(span) { |s, c| yield s, c }
-      rescue Exception => e # rubocop:disable Lint/RescueException
```
robertlaurin, line 58: Going over our [spec for handling errors](https://github.com/open-telemetry/opentelemetry-specification/blob/609a63784b2da7d59efe2a39822a23b09ae5b637/specification/error-handling.md#basic-error-handling-principles) it's not entirely clear to me why we went with `rescuing Exception`, which also will capture signal interrupts.

Do you have thoughts here? @mwear @fbogsany 

I think the example provided by @sribalakumar is a separate issue/consideration in itself, as we have no control over how other gems may patch the gems we instrument (alias_method vs prepend).  
>I was able to get the SystemStackError scenario on including prometheus-exporter which patches both Mysql2 and Redis using alias_method along side with OpenTelemetry instrumentation for Mysql2 and Redis which is using Module#prepend

api/lib/opentelemetry/trace/tracer.rb
sribalakumar: For `SystemStackError (stack level too deep)` and other cases when there are signal interrupts in the middle of `start_span` method call `span` is not defined.

api/lib/opentelemetry/trace/tracer.rb
sribalakumar: Yes the looping because of **alias_method vs prepend** is a separate issue in itself and it is commonly seen across all major APM providers.

- https://github.com/elastic/apm-agent-ruby/issues/379
- https://docs.newrelic.com/docs/agents/ruby-agent/troubleshooting/systemstackerror-stack-level-too-deep
- https://github.com/scoutapp/scout_apm_ruby/issues/309 (data dog integration)
- https://github.com/DataDog/dd-trace-rb/issues/862

We can track it as a separate issue as more users are bound to face this problem. The best we can do is [give a reference link that is similar to Newrelic](https://blog.newrelic.com/engineering/ruby-agent-module-prepend-alias-method-chains/) for the users to identify which external Gem could have caused the looping `SystemStackError (stack level too deep)` error.

sribalakumar: I have opened an issue https://github.com/open-telemetry/opentelemetry-ruby/issues/352 and have left some references on how other APM providers are handling this case.

fbogsany: @robertlaurin the spec for handling errors relates to errors raised within the API and SDK (and related components like exporters). The spec for *reporting* errors is here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/exceptions.md

It doesn't actually say anything about the types of errors to be reported, just how to report them. I think we opted for the widest scope here because we want to report *anything* raised by code wrapped in `in_span`. In general, I think that's the right approach (e.g. if the operation was interrupted, I think that should be reported as an error). In this specific case, it is problematic.

fbogsany:
I'd really like to defer to @mwear for this.

api/lib/opentelemetry/trace/tracer.rb
mwear: This code actually came from dd-trace when we ported over the Redis instrumentation. We wanted to preserve the behavior of the analogous `Tracer#trace` method see: https://github.com/DataDog/dd-trace-rb/blob/master/lib/ddtrace/tracer.rb#L261-L302. The goal here, is the same as expressed in the comment in the dd-trace source (link above, comment pasted below).

```
        # Here we really want to catch *any* exception, not only StandardError,
        # as we really have no clue of what is in the block,
        # and it is user code which should be executed no matter what.
        # It's not a problem since we re-raise it afterwards so for example a
        # SignalException::Interrupt would still bubble up.
```

I also think we are doing the right thing here. Perhaps @ericmustin can weigh in as to whether you've seen reports from users where this behavior causes any issues?

mwlang: All in all, rescuing `Exception` here strikes me as the right choice and is ok since we are re-raising the exception.  As the comment Matt referenced says, "it is the user code which should be executed no matter what."  

Rescuing `Exception` here likely isn't masking/exposing an issue introduced by the Otel framework and we definitely want to know about all the exceptions that do come through this block. 

We're capturing all the possible errors against the current span (if one was successfully started), which is good, and re-raising.  Which means those special exceptions we'd normally shy away from rescuing can still propagate to do their magic -- also good.  

However, the possibility that OTel itself is the fault and may fail to start the current span  leads to necessarily having the `if defined?(span)` conditional in the following block, which _is_ a correct change to make here.

sribalakumar: @mwlang 

 >However, the possibility that OTel itself is the fault and may fail to start the current span leads to necessarily having the if defined?(span) conditional in the following block, which is a correct change to make here.

We will have to check `if defined?(span)` in the `rescue` block as well to handle exceptions which happens in the middle of the current span construction.

Something like:

```ruby
def in_span(...)
  ...
  ...
rescue Exception => e # rubocop:disable Lint/RescueException
  if defined?(span)
    span.record_exception(e)
    span.status = Status.new(Status::UNKNOWN_ERROR,
                                 description: "Unhandled exception of type: #{e.class}")
  end
  raise e
ensure
  span.finish if defined?(span)
end
```

This should also be fine and should solve the problem. Let me know if we can modify the PR accordingly.

sribalakumar: Perhaps if we want to go with rescuing all `Exception` then we can modify like

```ruby
def in_span(...)
  span = nil
  # Span construction below
  ...
  ...
rescue Exception => e # rubocop:disable Lint/RescueException
  span&.record_exception(e)
  span&.status = Status.new(Status::UNKNOWN_ERROR,
                                description: "Unhandled exception of type: #{e.class}")
  raise e
ensure
  span&.finish
end
```

robertlaurin: I think this is the correct approach, it would be great to see a test alongside this change to prevent any future regressions around this behaviour.

fbogsany (approved):
Thanks @sribalakumar !

api/lib/opentelemetry/trace/tracer.rb
sribalakumar: @robertlaurin Thanks, I have modified the PR as we discussed. Will add the test cases in a day or two.

sribalakumar: I am resolving this conversation and updating the PR description and name accordingly.

The original cause which is independent will be recorded in issue https://github.com/open-telemetry/opentelemetry-ruby/issues/352

Issue #244, opened 2020-05-12
dalli instrumentation adapter


mwear:
Convert Datadog instrumentation for dalli to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md	)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/dalli)

sribalakumar:
@fbogsany @mwear 

We have done the [Dalli porting](https://github.com/RashmiRam/opentelemetry-ruby/tree/v0.4.0-ruby-2.4-compatible/adapters/dalli ) and Iam interested in contributing this enhancement as PR.

Let me know if the work is already started by someone else ?

fbogsany:
@sribalakumar can you please coordinate with @robertlaurin? He has implemented this as well & we planned to contribute it this week. I think it would be useful to get both implementations up as PRs that we can review together, and cherry-pick the best of both.

robertlaurin:
I've put up a draft PR porting the internal shopify implementation here https://github.com/open-telemetry/opentelemetry-ruby/pull/355

sribalakumar:
@fbogsany @robertlaurin Thanks, I'll be able to send the alternate PR (we are patching at `Server.request` level, as in Datadog) over this weekend.

sribalakumar:
@robertlaurin @fbogsany I had a chance to go through https://github.com/open-telemetry/opentelemetry-ruby/pull/355 and its pretty much the [same definitions](https://github.com/RashmiRam/opentelemetry-ruby/tree/v0.4.0-ruby-2.4-compatible/adapters/dalli) that we have also back ported from Datadog. 

So to avoid duplicate efforts I have started with https://github.com/open-telemetry/opentelemetry-ruby/pull/363

Issue #339, opened 2020-08-26
Optionally trim attribute keys & values?


fbogsany:
Related to https://github.com/open-telemetry/opentelemetry-specification/issues/504

Java provides optional attribute value truncation: https://github.com/open-telemetry/opentelemetry-java/blob/a021a59b9433691c06ab2265b4c574eab825b0ed/sdk/tracing/src/main/java/io/opentelemetry/sdk/trace/SpanBuilderSdk.java#L195-L197
```java
    if (traceConfig.shouldTruncateStringAttributeValues()) {
      value = StringUtils.truncateToSize(value, traceConfig.getMaxLengthOfAttributeValues());
    }
```

We have some ad-hoc truncation and UTF8 validation in the Redis instrumentation, for example: https://github.com/open-telemetry/opentelemetry-ruby/blob/1a680bbd489eee18cdf78b24a99e3cc0333c41e5/instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb#L60-L80

It is likely worth extracting this into a common utility module, and implementing attribute truncation in a single place with a configurable limit (most likely in the SDK `Span` class).

plantfansam:
I believe this issue [has been addressed](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/span.rb#L350-L356). Should we close or is there more to do here?

fbogsany:
I think #1136 is relevant here.

PR #355, opened 2020-09-01
Add Dalli instrumentation gem

robertlaurin:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/244

### DIFF
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
@@ -14,27 +14,27 @@ module OpenTelemetry
       CMD_MAX_LEN = 500
 
       OPNAME_MAPPING = {
-        'get'           => 'get',
-        'cas'           => 'get',
-        'set'           => 'set',
-        'add'           => 'add',
-        'replace'       => 'replace',
-        'delete'        => 'delete',
-        'incr'          => 'incr',
-        'decr'          => 'decr',
-        'flush'         => 'flush',
-        'write_noop'    => 'noop',
-        'version'       => 'version',
+        'get' => 'get',
+        'cas' => 'get',
+        'set' => 'set',
+        'add' => 'add',
+        'replace' => 'replace',
+        'delete' => 'delete',
+        'incr' => 'incr',
+        'decr' => 'decr',
+        'flush' => 'flush',
+        'write_noop' => 'noop',
+        'version' => 'version',
         'send_multiget' => 'getkq',
-        'append'        => 'append',
-        'prepend'       => 'prepend',
-        'stats'         => 'stat',
-        'reset_stats'   => 'stat',
-        'multi_set'     => 'setq',
-        'multi_add'     => 'addq',
+        'append' => 'append',
+        'prepend' => 'prepend',
+        'stats' => 'stat',
+        'reset_stats' => 'stat',
+        'multi_set' => 'setq',
+        'multi_add' => 'addq',
         'multi_replace' => 'replaceq',
-        'multi_delete'  => 'deleteq',
-        'touch'         => 'touch',
+        'multi_delete' => 'deleteq',
+        'touch' => 'touch'
         # 'sasl_authentication' => 'auth_negotiation',
         # 'sasl_authentication' => 'auth_request',
       }.freeze
@@ -49,7 +49,7 @@ module OpenTelemetry
         command = [operation, *args].join(' ').strip
         command = utf8_encode(command, binary: true, placeholder: placeholder)
         truncate(command, CMD_MAX_LEN)
-      rescue => e
+      rescue StandardError => e
         OpenTelemetry.logger.debug("Error sanitizing Dalli operation: #{e}")
         placeholder
       end
@@ -78,4 +78,3 @@ module OpenTelemetry
     end
   end
 end
-
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
@@ -1,0 +1,81 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # Utility functions
+    module Utils
+      extend self
+
+      STRING_PLACEHOLDER = ''.encode(::Encoding::UTF_8).freeze
+      CMD_MAX_LEN = 500
+
+      OPNAME_MAPPING = {
+        'get'           => 'get',
+        'cas'           => 'get',
+        'set'           => 'set',
+        'add'           => 'add',
+        'replace'       => 'replace',
+        'delete'        => 'delete',
+        'incr'          => 'incr',
+        'decr'          => 'decr',
+        'flush'         => 'flush',
+        'write_noop'    => 'noop',
+        'version'       => 'version',
+        'send_multiget' => 'getkq',
+        'append'        => 'append',
+        'prepend'       => 'prepend',
+        'stats'         => 'stat',
+        'reset_stats'   => 'stat',
+        'multi_set'     => 'setq',
+        'multi_add'     => 'addq',
+        'multi_replace' => 'replaceq',
+        'multi_delete'  => 'deleteq',
+        'touch'         => 'touch',
+        # 'sasl_authentication' => 'auth_negotiation',
+        # 'sasl_authentication' => 'auth_request',
+      }.freeze
+
+      def opname(operation, multi)
+        lookup_name = multi ? "multi_#{operation}" : operation.to_s
+        OPNAME_MAPPING[lookup_name] || operation.to_s
+      end
+
+      def format_command(operation, args)
+        placeholder = "#{operation} BLOB (OMITTED)"
+        command = [operation, *args].join(' ').strip
+        command = utf8_encode(command, binary: true, placeholder: placeholder)
+        truncate(command, CMD_MAX_LEN)
+      rescue => e
+        OpenTelemetry.logger.debug("Error sanitizing Dalli operation: #{e}")
+        placeholder
+      end
+
+      def truncate(string, size)
+        string.size > size ? "#{string[0...size - 3]}..." : string
+      end
+
+      def utf8_encode(str, binary: false, placeholder: STRING_PLACEHOLDER)
+        str = str.to_s
+
+        if binary
+          # This option is useful for "gracefully" displaying binary data that
+          # often contains text such as marshalled objects
+          str.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
+        elsif str.encoding == ::Encoding::UTF_8
+          str
+        else
+          str.encode(::Encoding::UTF_8)
+        end
+      rescue StandardError => e
+        OpenTelemetry.logger.debug("Error encoding string in UTF-8: #{e}")
+
+        placeholder
+      end
+    end
+  end
+end
+
--- a/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
+++ b/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
@@ -14,7 +14,7 @@ describe OpenTelemetry::Instrumentation::Dalli::Instrumentation do
   let(:exporter) { EXPORTER }
   let(:span) { exporter.finished_spans.first }
   let(:host) { ENV.fetch('TEST_MEMCACHED_HOST') { '127.0.0.1' } }
-  let(:port) { (ENV.fetch('TEST_MEMCACHED_PORT') { 11211 }).to_i }
+  let(:port) { (ENV.fetch('TEST_MEMCACHED_PORT') { 11_211 }).to_i }
   let(:dalli) { ::Dalli::Client.new("#{host}:#{port}", {}) }
 
   before do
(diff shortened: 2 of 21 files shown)


instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # Utility functions
+    module Utils
```
robertlaurin, line 10: The utils here are shameless ported from some existing instrumentation, there is some room for feedback/improvement here.

robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/issues/339 this is also feels very relevant here

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/client.rb
```diff
@@ -0,0 +1,62 @@
...
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Dalli
+      module Patches
+        # Module to prepend to Dalli::Client for instrumentation
+        module Client
```
ericmustin, line 12: Haven't used Dalli in production before, could you explain a bit the benefit of instrumenting the higher level `Dalli::Client.perform` and `Dalli::Client.get_multi_yielder`, as opposed to `Dalli::Server.request`?

Basically, by choosing to wrap `Dalli::Client.get_multi_yielder` we introduce a bit of a tradeoff it seems, where we're losing some visibility into the underlying `host/port` info of the server, since `get_multi_yielder` calls `server.request` directly without going thru perform, if i'm reading this correctly? https://github.com/petergoldstein/dalli/blob/ca8b286de230546c3bfe2789e9b6bc5a2c619570/lib/dalli/client.rb#L429

So in those cases we'd lose the  `net.peer.name` and `net.peer.port` attributes, (since capturing it for a query that hits multiple servers introduces a bit of a race condition), and if there are errors we may not be able to attribute it to a specific server. But, as a tradeoff, we get better visibility into the timing of the entire operation, which is what the user is performing and likely cares more about. Am I stating that correctly? And I guess, more importantly is this intentional and something you feel is a greater value add? I defer to your opinion here but think it's worth being explicit about.

My other thought here is maybe we wrap both? IE, wrap the multi yielder, but also generate child spans for each `server.request`. That way if say, there's a connection issue or latency on a specific server, it's clear which one, while with the `get_multi` parent span we also _get_ (_hehe_) the visibility into what the broader user query was? Not sure if that's worth the implementation effort but just throwing it out there

fbogsany: Thanks for the suggestion. When writing our instrumentation, we saw `Client.perform` as the "Chokepoint method for instrumentation", and I guess we didn't dig deeper. Looking at `Server.request`, I see that's also marked as "Chokepoint method for instrumentation", and it is likely better to instrument at that level, closer to the wire. I think we should instrument *only* at that level, rather than introducing additional spans at the higher level.

instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
```diff
@@ -0,0 +1,58 @@
...
+      _(span.attributes['db.statement']).must_equal 'set foo bar 0 0'
+      _(span.attributes['net.peer.name']).must_equal '127.0.0.1'
+      _(span.attributes['net.peer.port']).must_equal 11211
+    end
+
+    it 'after dalli#get_multi' do
+      dalli.get_multi('foo', 'bar')
+
+      _(exporter.finished_spans.size).must_equal 1
+      _(span.name).must_equal 'SEND_MULTIGET'
```
fbogsany, line 51: 🤔 I'd prefer to see actual `memcache` commands here. IIRC, this should be a `GET`. For cases where Dalli maps the command names to something else, it'd be good to map them back to the real command names.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
ericmustin: Seems fine to me, i don't have strong opinions on the default placeholder but that's the only thing maybe we may want to change. not a big deal at all, ship it imo

ericmustin (commented):
Generally this looks good to me, the changes over to `::Server` I think simplify things. All that's left here is probably a test for an error case, and perhaps some tests for the behavior around multi_get/sets. the attributes look in line with the spec, if there's any other optional attributes on there that are reasonable to add, go for it, but if not no worries https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md

Ping me when this is out of draft mode and happy to approve after giving another once over for the above suggestions

Also, great work on this!

instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
```diff
@@ -0,0 +1,93 @@
...
+      _(span.attributes['db.statement']).must_equal 'getkq foo bar'
+      _(span.attributes['net.peer.name']).must_equal host
+      _(span.attributes['net.peer.port']).must_equal port
+    end
+
+    it 'after error' do
+      dalli.set('foo', 'bar')
+      exporter.reset
+
+      dalli.instance_variable_get(:@ring).servers.first.stub(:write, ->(_bytes) { raise Dalli::NetworkError }) do
```
robertlaurin, line 76: I'm calling `.set` so the client sets up a server in the ring instance variable, then grabbing it for a stub. I want to try and simulate an actual error that could crop up.  There wasn't a better way obvious to me on how to do this.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,81 @@
...
+        'prepend'       => 'prepend',
+        'stats'         => 'stat',
+        'reset_stats'   => 'stat',
+        'multi_set'     => 'setq',
+        'multi_add'     => 'addq',
+        'multi_replace' => 'replaceq',
+        'multi_delete'  => 'deleteq',
+        'touch'         => 'touch',
+        # 'sasl_authentication' => 'auth_negotiation',
+        # 'sasl_authentication' => 'auth_request',
```
robertlaurin, line 39: There are two underlying write requests for the `sasl_authentication` so I'm not sure if it should map to one or the other here.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,81 @@
...
+      }.freeze
+
+      def opname(operation, multi)
+        lookup_name = multi ? "multi_#{operation}" : operation.to_s
```
robertlaurin, line 43: I don't like [repeating the multi check](https://github.com/petergoldstein/dalli/blob/master/lib/dalli/server.rb#L289) for getting the op code, but I didn't think of a better way of doing it.  Open to suggestions for improvement here.

instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
ericmustin: lgtm, i also have no ideas here off the top of my head

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
ericmustin: maybe we just add a heuristic in the `opname` method to make this choice, and return like `'auth'` from opname mapping? my thought is it's better to be accurate than necessarily have the tightest code
```
if (temp_opname = OPNAME_MAPPING[lookup_name] || operation.to_s) && temp_opname == 'auth'
  // additional heuristic to check if its negotiation or request
else
  return temp_opname
end
```

ericmustin (approved):
Left some small feedbacks on remaining open questions but generally comfortable having this merged so happy to approve now, assuming we can clear up those Jruby test failures if they are not transient (tough to tell if they're transient or not, we can discuss at SIG mtg if it's recurring)

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
fbogsany: It's a pretty trivial boolean check - I think this is fine.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,81 @@
...
+        OPNAME_MAPPING[lookup_name] || operation.to_s
+      end
+
+      def format_command(operation, args)
+        placeholder = "#{operation} BLOB (OMITTED)"
+        command = [operation, *args].join(' ').strip
+        command = Utils.utf8_encode(command, binary: true, placeholder: placeholder)
```
fbogsany, line 50: Do you need to prepend with `Utils.` here? We're in the same module.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,81 @@
...
+      rescue => e
+        OpenTelemetry.logger.debug("Error sanitizing Dalli operation: #{e}")
+        placeholder
+      end
+
+      def truncate(string, size)
+        string.size > size ? "#{string[0...size - 3]}..." : string
+      end
+
+      def utf8_encode(str, options = {})
```
fbogsany, line 61: Why the hash options here instead of keyword args with defaults?

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
```diff
@@ -0,0 +1,81 @@
...
+        placeholder = "#{operation} BLOB (OMITTED)"
+        command = [operation, *args].join(' ').strip
+        command = Utils.utf8_encode(command, binary: true, placeholder: placeholder)
+        Utils.truncate(command, CMD_MAX_LEN)
+      rescue => e
+        OpenTelemetry.logger.debug("Error sanitizing Dalli operation: #{e}")
+        placeholder
+      end
+
+      def truncate(string, size)
```
fbogsany, line 57: I really wish we could do this truncation while formatting. I looked for options to do that, but didn't find anything. My concern is that we could have a multi-megabyte binary blob that we're going to go to the trouble of encoding, only to throw away or truncate the result.

instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
fbogsany: Is there an equivalent of `Dalli::Server.any.stub(...)` to avoid reaching in to the instance variable?

Otherwise this seems fine to me. You could go to the trouble of pulling in Semian, but I'm not sure it's worth it yet.

fbogsany: Hmm, doesn't look like Semian would help here - it doesn't include support for Dalli AFAICT.

robertlaurin: >Is there an equivalent of Dalli::Server.any.stub(...) to avoid reaching in to the instance variable?

I've been looking, and it seems minitest doesn't have this baked in and I didn't really want to pull in another test dependency for just this one test.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
robertlaurin: Nope, removed.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
robertlaurin: I just copied it over from DD 🙈 , fixed.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
fbogsany: I'm kinda OK with handling this case later. As it is, the default in `opname` will catch this and turn it into `"sasl_authentication"` in both cases.

fbogsany (approved):
LGTM after the merge conflicts are fixed. Thanks @robertlaurin!

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb
robertlaurin: Yeah trying to dig into the `sasl_authentication` method here https://github.com/petergoldstein/dalli/blob/master/lib/dalli/server.rb#L701-L730 and I'm not sure how I could extract more meaningful info out of it right now.  This puts us in a better place to iterate on it going forward when/if needed.

ericmustin: works for me 👍

fbogsany (approved):

fbogsany:
There's a conflict in the Rakefile and this gem should be added to the `releases.yml` file. Otherwise LGTM, thanks!

PR #357, opened 2020-09-02
fixes the various rubocop offenses across the board

mwlang:
Resolves all the rubocop offenses reported when running `rake each:rubocop` in the project root folder.


instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -9,13 +9,15 @@ module Instrumentation
     module Sidekiq
       module Middlewares
         module Client
+          # TracerMiddleware propagates context and instruments Sidekiq client
+          # by way of its middlware system
```
fbogsany, line 13:
```suggestion
          # by way of its middleware system
```

fbogsany:
Thanks @mwlang !

PR #363, opened 2020-09-07
Fork safety for batch processor

### LINKED ISSUE #310: Fork safety in BatchSpanProcessor
`BatchSpanProcessor` starts a thread to handle the export loop. This needs to be restarted in a child process after `fork`. Also, we need to ensure that any spans that may have been buffered in the parent process are *not* also sent from the child process after `fork`. The way we've accomplished this in Shopify's instrumentation gem is:

```ruby
        def reset_on_fork(restart_thread: true)
          pid = Process.pid
          return if @pid == pid
          @pid = pid
          @records.clear
          @thread = Thread.new { work } if restart_thread
        end
```

where `reset_on_fork` is called from `initialize` and `on_finish`, and `reset_on_fork(restart_thread: false)` is called from `work` and `force_flush`.

sribalakumar:
Fixes #310


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -90,6 +93,7 @@ def force_flush
               batch = snapshot.shift(@batch_size).map!(&:to_span_data)
               result_code = @exporter.export(batch)
               report_result(result_code, batch)
+              reset_on_fork(restart_thread: false) if @keep_running
```
fbogsany, line 96: This should happen inside the `snapshot = lock { ... }` block, before `shift`ing the spans. We want to avoid accidentally sending buffered spans from the child.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -121,9 +125,19 @@ def work
               end
 
               export_batch(batch)
+              reset_on_fork(restart_thread: false)
```
fbogsany, line 128: This should move up inside the `batch = lock do ... end` block.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
sribalakumar: Thanks, have corrected it.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
sribalakumar: Have made the corrections.

fbogsany (approved):
Thanks @sribalakumar!

PR #367, opened 2020-09-09
fix!: Rename Resource labels to attributes

### LINKED ISSUE #359: Should resource labels be renamed to resource attributes?
Some enterprising developer on opentelemetry-js seems to think so https://github.com/open-telemetry/opentelemetry-js/pull/1419 , and it sounds like this is now spec. Should ruby also migrate `resource.labels` to be `resource.attributes` ?

fbogsany:
Fixes #359


mwear (approved):
LGTM

PR #384, opened 2020-09-15
fix: Don't generate a span ID unnecessarily

fbogsany:
We generate a span ID in the SDK's `Tracer#start_span` that we never use. I think it used to be required for sampling, but that requirement was removed. This removes the extraneous work.


mwear (approved):
LGTM

PR #396, opened 2020-09-17
docs: Remove release schedule

fbogsany:
The release schedule is horribly out of date. Some other repos have chosen to post a feature status instead, but most have nothing at all. I've opted here for no information rather than bad information.


mwear (approved):
This has always been confusing and out of date. Good to see it go.

PR #398, opened 2020-09-18
refactor: context only parenting

mwear:
This PR is a draft showing what changes would be required to restrict explicit span parenting to context only (rather than allowing a span, or context). This is an effort towards validating that: https://github.com/open-telemetry/opentelemetry-specification/issues/510 is viable in Ruby.

There were two options to provide an explicit parent to `start_span` and friends. One was named `with_parent` and expected an argument to be a span. The other was named `with_parent_context` where the argument was a context. In this PR we removed `with_parent_context` and updated `with_parent` to accept a context rather than a span.

I also added a `Tracer#context_with_span` convenience method which is the equivalent of `ctx = tracer.with_span(span) { |_, c| c }`, which was something I found myself doing in tests when I removed explicit span based parenting.

The original `with_parent` option was only used in tests. Instrumentation used `with_parent_context`, further validating context based parenting want we want.


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -85,14 +85,9 @@ def start_root_span(name, attributes: nil, links: nil, start_timestamp: nil, kin
       #
       # Parent context can be either passed explicitly, or inferred from currently activated span.
       #
-      # @param [optional Span] with_parent Explicitly managed parent Span, overrides
-      #   +with_parent_context+.
-      # @param [optional Context] with_parent_context Explicitly managed. Overridden by
-      #   +with_parent+.
```
fbogsany, line 91: We need an updated version of this doc.

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -85,14 +85,9 @@ def start_root_span(name, attributes: nil, links: nil, start_timestamp: nil, kin
...
-      #
       # @return [Span]
-      def start_span(name, with_parent: nil, with_parent_context: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil)
-        span_context = with_parent&.context || active_span_context(with_parent_context)
+      def start_span(name, with_parent: nil, attributes: nil, links: nil, start_timestamp: nil, kind: nil)
+        span_context = active_span_context(with_parent)
```
fbogsany, line 90: Can we make `active_span_context` private or `@api private`? Or is it required by the spec? (We don’t use it anywhere except in the API and SDK `Tracer` classes.)

api/benchmarks/tracer_bench.rb
```diff
@@ -31,13 +31,8 @@
...
   end
 
-  x.report 'start span with parent' do
-    span = tracer.start_span('test_span', with_parent: parent_span)
-    span.finish
-  end
-
   x.report 'start span with parent context' do
-    span = tracer.start_span('test_span', with_parent_context: parent_span.context)
+    span = tracer.start_span('test_span', with_parent: parent_span.context)
```
fbogsany, line 35: This is not correct (it wasn’t correct before either). `with_parent:` wants a `Context`, but `parent_span.context` is a `SpanContext`.

api/lib/opentelemetry/trace/tracer.rb
mwear: I made it private. I think we can remove this method completely by simplifying the current span situation. I'll look into opening a separate PR that does that tomorrow.

fbogsany (approved):
I realize this is just a prototype, but it looks like a good improvement to me 👍

fbogsany:
https://github.com/open-telemetry/opentelemetry-specification/pull/875 also requires `on_start` for span processors to accept a `parent_context` parameter. This requires a little refactoring because `Span#initialize` only has a `SpanContext` and not the parent `Context`.

mwear:
This is ready for review and can merge when folks are comfortable with it. The corresponding spec changes merged earlier today: https://github.com/open-telemetry/opentelemetry-specification/pull/875.

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -21,6 +21,15 @@ def current_span(context = nil)
         context.value(CURRENT_SPAN_KEY) || Span::INVALID
       end
 
+      # Returns a context containing the span, derived from the optional parent
+      # context, or the current context if one was not provided.
+      #
+      # @param [optional Context] context The context to use as the parent for
+      #  the returned context
+      def context_with_span(span, context = Context.current)
```
fbogsany, line 29: Thoughts on making the second parameter named rather than positional? E.g. `parent_context: Context.current`

robertlaurin (approved):
This makes sense to me.

fbogsany (approved):

Issue #246, opened 2020-05-12
GraphQL instrumentation adapter


mwear:
Convert Datadog instrumentation for GraphQL to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/graphql)

robertlaurin:
Added in https://github.com/open-telemetry/opentelemetry-ruby/pull/527

PR #410, opened 2020-09-20
feat: Add GraphQL instrumentation gem

ericmustin:
This PR addressing #246 , by adding a gem that provides instrumentation for the [Ruby GraphQL gem](https://github.com/rmosolgo/graphql-ruby).

The implementation is largely in line with the existing hooks that the Gem provides in it's tracing directory https://github.com/rmosolgo/graphql-ruby/tree/master/lib/graphql/tracing , and this is largely a port of the [Datadog instrumentation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/graphql) simply because that's the one I'm most familiar with. 

I think it's robust in terms of capturing query details and timing specifics, but it does require the user to provide an array of the schema they want to instrument, which imo is not ideal. I haven't looked too deeply into whether other implementations have benefits for some magic around "just tracing all the right schema ootb" but if there is, i'd be happy to incorporate that.

The tests are a little weak right now around non-happy path cases, the example repo itself is very modest (i am not an active graphql user 😓 ) , and I still need to look around the OTEL ecosystem to see what other language implementations / Otel-Spec want in terms of attribute specs/naming details/etc.

Anyway, any and all feedback is appreciated and I think i'm at the point where this can start to be reviewed

TODO:

- [ ] Document required `schema` config options more clearly
- [ ] Add Tests for fail cases, various `schema` cases
- [ ] True up the Attribute names with what other language implementations have + what's required by spec
- [ ] Improve example, ideally provide a more robust example, or something against a containerised +seeded db, vs returning static info.
- [ ] Determine if there's a cleaner or more user-friendly way to specify which Schemas to instrument (something that wouldn't require `use_all` to receive graphql specific config).
- [ ] Ideally, make a PR against upstream [graph-ruby](https://github.com/rmosolgo/graphql-ruby) to include our tracing class, which would bring us to parity with what other OSS and 3rd Party Vendors have done (maybe we want to wait till we're GA?).


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
...
+          require_relative 'patches/opentelemetry_graphql_tracing'
+        end
+
+        def patch_graphql(config = {})
+          config[:schemas].each { |s| patch_schema!(s) }
+        end
+
+        def target_version
+          Integration.version
+        end
```
fbogsany, line 34: What is this for? I haven't found a reference to it (yet).

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
...
+            schema.define do
+              use(
+                OpenTelemetry::Instrumentation::GraphQL::Patches::OpenTelemetryGraphQLTracing
+              )
+            end
+          end
+        # Protect our instrumentation from bad user input in schema config option
+        rescue StandardError => e
+          OpenTelemetry.logger.error("Unable to patch schema #{schema}: #{e.message}")
+          true
```
fbogsany, line 51: I don't think this is necessary?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+        class OpenTelemetryGraphQLTracing < ::GraphQL::Tracing::PlatformTracing
+          self.platform_keys = {
+            'lex' => 'lex.graphql',
+            'parse' => 'parse.graphql',
+            'validate' => 'validate.graphql',
+            'analyze_query' => 'analyze.graphql',
+            'analyze_multiplex' => 'analyze.graphql',
+            'execute_multiplex' => 'execute.graphql',
+            'execute_query' => 'execute.graphql',
+            'execute_query_lazy' => 'execute.graphql'
```
fbogsany, line 21: Shopify's instrumentation also has 'lazy_loader' here.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -0,0 +1,56 @@
...
+
+        private
+
+        def require_dependencies
+          require_relative 'patches/opentelemetry_graphql_tracing'
+        end
+
+        def patch_graphql(config = {})
+          config[:schemas].each { |s| patch_schema!(s) }
+        end
```
fbogsany, line 30: Interestingly, in Shopify's instrumentation, we just:
```ruby
GraphQL::Schema.prepend(Shopify::Tracing::Adapters::GraphQLAdapter::SchemaPatch)
```
where
```ruby
        module SchemaPatch
          def initialize
            super
            @tracers << GraphQLTracer
          end
        end
```
It looks like we're assuming the Schema's `@tracer` is expected to implement `#trace`. I wonder if our implementation predates GraphQL's platform tracing 🤔 .

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+              kind: :client
+            ) do |span|
+              if key == 'execute_multiplex'
+                operations = data[:multiplex].queries.map(&:selected_operation_name).join(', ')
+                span.set_attribute('operation', operations)
+              end
+
+              if key == 'execute_query'
+                span.set_attribute(:selected_operation_name, data[:query].selected_operation_name)
+                span.set_attribute(:selected_operation_type, data[:query].selected_operation.operation_type)
```
fbogsany, line 37:
```suggestion
                span.set_attribute(:selected_operation_type, data[:query].selected_operation&.operation_type)
```
From our code, it looks like there's a chance that `selected_operation` is `nil`.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+                span.set_attribute(:query_string, data[:query].query_string)
+              end
```
fbogsany, line 39: Have you considered computing the `attributes` hash before the block and passing it into the `in_span` call directly? That's going to be more efficient than setting the attributes individually inside the block.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+            'parse' => 'parse.graphql',
+            'validate' => 'validate.graphql',
+            'analyze_query' => 'analyze.graphql',
+            'analyze_multiplex' => 'analyze.graphql',
+            'execute_multiplex' => 'execute.graphql',
+            'execute_query' => 'execute.graphql',
+            'execute_query_lazy' => 'execute.graphql'
+          }
+
+          def platform_trace(platform_key, key, data)
```
fbogsany, line 24: We additionally trace `execute_field` and `execute_field_lazy`, but rather than creating a span for each field, we append the fields to a `field_contexts` array hanging off the current `Span`, and monkey-patch `finish` to add the array as an attribute on the span. It's ugly, and I wonder if we can do it another way here. In general, it's useful to users to know which fields are evaluated as part of the query.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
fbogsany: Looks like we're being unnecessarily sleazy here. I believe we can just:
```ruby
GraphQL::Schema.tracer(OpenTelemetry::Instrumentation::GraphQL::Patches::OpenTelemetryGraphQLTracing.new)
```
The [`tracers` method](https://github.com/rmosolgo/graphql-ruby/blob/88a2ebf89e14c7b09b6e9c259b8b4c67ff62d214/lib/graphql/schema.rb#L1578-L1580) collects tracers from up the inheritance hierarchy. I guess the question is: do we want auto-instrumentation to instrument *all* schemas or just specific ones listed by the user?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
ericmustin: 👍 leftover from the port, will remove

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
ericmustin: i was concerned about users passing in a string, or a class which doesn't have the `define` method, etc. but if it's swallowed by broader error handling i can remove

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
ericmustin: My initial thought here is that we should instrument all the schemas auto-magically, since by and large user's will just be configuring things via the tracer's `c.use_all`. If the goal is "Time to Value",  and simply passing in the otel tracing class to the `.tracer` method permits that, it feels like a big time to value win.  I do have concerns that this approach could be noisy and generate a lot of spans OOTB, which might impact users with storage costs or performance overhead. I'm not sure how much of a real concern that is, but it _seems_ like a thing someone would want to tweak. It might be worth having `schemas` as an optional config setting. If undefined we patch everything, via `GraphQL::Schema.tracer(OpenTelemetry::Instrumentation::GraphQL::Patches::OpenTelemetryGraphQLTracing.new)`. If defined we patch only the schemas specified in the config's `schemas` array.

I'd like to poke around other lang's and vendors here and see how they approach things or if they have surfaced drawbacks to this approach, but generally what do you think about the above?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
ericmustin: Makes sense, agreed, will update

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
ericmustin: ah, that's interesting. It sounds like valuable info imo, so i think we should try to support it. Let me see if i can find a more canonical way to do so.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
fbogsany: Sorry, I meant specifically returning `true`. The method isn't expected to return anything.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
fbogsany: > It might be worth having `schemas` as an optional config setting. If undefined we patch everything, via `GraphQL::Schema.tracer(OpenTelemetry::Instrumentation::GraphQL::Patches::OpenTelemetryGraphQLTracing.new)`. If defined we patch only the schemas specified in the config's `schemas` array.
> 
> I'd like to poke around other lang's and vendors here and see how they approach things or if they have surfaced drawbacks to this approach, but generally what do you think about the above?

Very much 👍

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
fbogsany: Shopify's tracing gem doesn't support Events, so I forgot about those. I think that might be a better way to handle `execute_field` and `execute_field_lazy`.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+              yield
+            end
+          end
+
+          def tracer
+            GraphQL::Instrumentation.instance.tracer
+          end
+
+          def platform_field_key(type, field)
+            "#{type.graphql_name}.#{field.graphql_name}"
```
robertlaurin, line 49: From our SIG call today, this is the key that turned an 11 span trace into a 2500 span trace.  However I really do think the information that comes from this field is valuable.  

I'm uncertain as to whether or not this should be on by default, I know for us (Shopify) we would want to configure this to be off.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          def platform_authorized_key(type)
+            "#{type.graphql_name}.authorized"
```
robertlaurin, line 53: Another noisy key, I did not see a lot of value from this when I had it enabled but I am also not an expert on GraphQL.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/patches/opentelemetry_graphql_tracing.rb
```diff
@@ -0,0 +1,63 @@
...
+          end
+
+          def platform_resolve_type_key(type)
+            "#{type.graphql_name}.resolve_type"
```
robertlaurin, line 57: Some comment as the authorized_key, noisy, uncertain of the value.

robertlaurin:
Hey @ericmustin, I recently reworked the graphql instrumentation for our internal tracing gem so this stuff is still fairly fresh.  With your blessing in me doing so I could take over and push this over the finish line.

ericmustin:
@robertlaurin yes absolutely please feel free to, i am bogged down with a few things for the rest of the quarter unfort, happy to review though

robertlaurin:
Hey @ericmustin, I guess I need your formal permission if I want to continue off of your PR.  Did we want to do that or should I just spin up a new PR.  Whatever you prefer here, I have no preference.

```
robertlaurin@robertlaurins-MacBook-Pro ~/s/f/e/opentelemetry-ruby> git push -f
remote: Permission to ericmustin/opentelemetry-ruby.git denied to robertlaurin.
fatal: unable to access 'https://github.com/ericmustin/opentelemetry-ruby.git/': The requested URL returned error: 403
```

ericmustin:
@robertlaurin ergh ok, sorry about that, let me see if i can add you to that repo...one secc

ericmustin:
@robertlaurin added you as a collaborator, does that help?

robertlaurin:
> @robertlaurin added you as a collaborator, does that help?

It did, but it lead to a new error.  I'm not at all familiar with this one.  
```
robertlaurin@robertlaurins-MacBook-Pro ~/s/f/e/opentelemetry-ruby> git push -f
Enumerating objects: 3029, done.
Counting objects: 100% (3027/3027), done.
Delta compression using up to 12 threads
Compressing objects: 100% (1170/1170), done.
Writing objects: 100% (2703/2703), 334.79 KiB | 25.75 MiB/s, done.
Total 2703 (delta 1617), reused 2312 (delta 1298), pack-reused 0
remote: Resolving deltas: 100% (1617/1617), completed with 253 local objects.
To https://github.com/ericmustin/opentelemetry-ruby.git
 ! [remote rejected] add_graphql -> add_graphql (refusing to allow an OAuth App to create or update workflow `.github/workflows/ci.yml` without `workflow` scope)
error: failed to push some refs to 'https://github.com/ericmustin/opentelemetry-ruby.git'
```

robertlaurin:
@ericmustin I'm not sure it makes too much sense fighting with this, maybe I just spin it off into a new PR.  I swear I'm not trying to steal credit 😄

ericmustin:
@robertlaurin no worries. yes please feel free to run with it.

robertlaurin:
@ericmustin Thanks!

PR #411, opened 2020-09-21
fix! implement new Span.Status spec

fbogsany:
This implements the new Span.Status spec: https://github.com/open-telemetry/opentelemetry-specification/pull/966

We need the mapping to OTLP to be specified in order to complete this, however that is strictly additive and this implements the proposal for OK & Unset -> OK and Error -> UnknownError.

### DIFF
--- a/api/lib/opentelemetry/trace/status.rb
+++ b/api/lib/opentelemetry/trace/status.rb
@@ -26,7 +26,7 @@ module OpenTelemetry
 
       # Initialize a Status.
       #
-      # @param [Integer] canonical_code One of the standard gRPC codes: https://github.com/grpc/grpc/blob/master/doc/statuscodes.md
+      # @param [Integer] canonical_code One of the canonical status codes below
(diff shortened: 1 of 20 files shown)


robertlaurin (approved):
This looks good to me

mwear (approved):
LGTM. We can and probably should set the `error: true` span attribute for the Jaeger exporter: https://github.com/open-telemetry/opentelemetry-ruby/blob/master/exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb#L121-L124. This can be done in a follow up.

fbogsany:
Updated to encode an error status in Jaeger, and to default to `UNSET`. We have treated `nil` as `UNSET`, and we have that expectation in a bunch of places (particularly tests), so the exporters are resilient to `span_data.status.nil?`. The spec is clear that `UNSET` is the default, though.

robertlaurin (approved):

PR #414, opened 2020-09-21
fix! Return results from force_flush and shutdown

### LINKED ISSUE #405: Consider returning success/fail/timeout for SpanProcessor shutdown and force_flush
See https://github.com/open-telemetry/opentelemetry-specification/pull/878

This is a *SHOULD* rather than *MUST*, but I think it makes sense.

fbogsany:
Fixes #405 

This adds return codes for `SpanProcessor#force_flush` and `SpanProcessor#shutdown`. They don't necessarily make sense on their own at the moment, but once #341 lands, this will allow callers to check for success/failure/timeout.

Since most of the processors live in the `Export` module, I overloaded the result codes we've already defined there, and added `TIMEOUT`.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export.rb
@@ -7,16 +7,21 @@
 module OpenTelemetry
   module SDK
     module Trace
-      # The Export module contains the built-in exporters for the OpenTelemetry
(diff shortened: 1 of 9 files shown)


ericmustin (approved):
lgtm

robertlaurin (approved):

Issue #408, opened 2020-09-18
Add URL quantization callback support to Sinatra instrumentation


fbogsany:
Our Rack instrumentation allows configuration of a URL quantization function https://github.com/open-telemetry/opentelemetry-ruby/blob/1c37b00dc82d490dace1f92739977aa4674d7039/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L122-L137

We should add the same functionality to the Sinatra instrumentation. @genebean reported this requirement in Gitter:
> I am using the auto instrumentation of the Sinatra but need to account for url paramters so that `/api/v2/status/queue/info/1600289926679` is instead traced as `/api/v2/status/queue/info/{job_id}` - any tips on how to do this?

genebean:
Thanks! Please make sure that `instrumentation/rack/README.md` and `instrumentation/sinatra/README.md` are updated to include how to put that to use... I don't see anything about it at this time.

mwear:
We can do this, but I think we should also use `env['sinatra.route']` as the span name. It is already a quantized low-cardinality string. For example, for this route:

```
get '/accounts/:id' do
  ...
end
```

`sinatra.route` for a request with an actual id always be `"GET /accounts/:id" `.

genebean:
@mwear would that be as simple as editing https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb and replacing `env['PATH_INFO']` with `env['sinatra.route']`? If so, I will put up a PR

PR #415, opened 2020-09-21
fix: default to sinatra.route for span name

genebean:
This solves most of what is being asked for in #408

### DIFF
--- a/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
+++ b/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
@@ -15,8 +15,10 @@ module OpenTelemetry
           end
 
           def call(env)
+            span_name = env['sinatra.route'] || env['PATH_INFO']
+
             tracer.in_span(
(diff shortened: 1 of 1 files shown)


mwear (approved):
I think we can drop the `http.route` attribute here: https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb#L46. Otherwise looks good to me.

robertlaurin (approved):

fbogsany:
> I think we can drop the `http.route` attribute here: https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb#L46. Otherwise looks good to me.

`http.route` is not _required_, but it _is_ part of the [semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions). I think we should probably retain it, in particular if we add support for overriding the quantization.

fbogsany (approved):

PR #417, opened 2020-09-22
feat: Add service_name setter to configurator

robertlaurin:
Adds a convenience method so setter the service.name resource attribute.
Closes #379

PR #426, opened 2020-10-04
feat: Add service_version setter to configurator

genebean:
PR #417 added a setter to the configurator for service.name - this follows suite by adding the same for service.version to simplify the setting of the other attribute requested by many tracing backends.

Prior to this change, setting the pair of these would have looked like this:

```ruby
c.service_name = 'Otel Demo App'
c.resource = OpenTelemetry::SDK::Resources::Resource.create(
  OpenTelemetry::SDK::Resources::Constants::SERVICE_RESOURCE[:version] => '0.6.0'
)
```

After this change, the same is accomplished with this:

```ruby
c.service_name = 'Otel Demo App'
c.service_version = '0.6.0'
```

### DIFF
--- a/exporter/jaeger/README.md
+++ b/exporter/jaeger/README.md
@@ -40,9 +40,8 @@ OpenTelemetry::SDK.configure do |c|
       # OpenTelemetry::Exporter::Jaeger::CollectorExporter.new(endpoint: 'http://192.168.0.1:14268')
     )
   )
-  c.resource = OpenTelemetry::SDK::Resources::Resource.create(
-    OpenTelemetry::SDK::Resources::Constants::SERVICE_RESOURCE[:name] => 'jaeger-example'
-  )
+  c.service_name = 'jaeger-example'
+  c.service_version = '0.6.0'
 end
 
 # To start a trace you need to get a Tracer from the TracerProvider
--- a/sdk/lib/opentelemetry/sdk/configurator.rb
+++ b/sdk/lib/opentelemetry/sdk/configurator.rb
@@ -54,6 +54,16 @@ module OpenTelemetry
         ).merge(@resource)
       end
 
+      # Accepts a string that is merged in as the service.version resource attribute.
+      # The most recent assigned value will be used in the event of repeated
+      # calls to this setter.
+      # @param [String] service_version The value to be used as the service version
+      def service_version=(service_version)
+        @resource = OpenTelemetry::SDK::Resources::Resource.create(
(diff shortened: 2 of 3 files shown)


fbogsany (approved):

robertlaurin (approved):

mwear:
I'm late to the party and ultimately fine with this change. I wanted to get folks pulse on a possible alternative. Rather than add one off configuration for various resource attributes, would it make sense to make setting them a little easier? Just so we don't end up with n-different ways to set resources, I was thinking we could repurpose the `c.resource` method to accept either a Resource, or a Hash of resrouce attributes. This way you could do something like:

```
sdk= OpenTelemetry::SDK.configure do |c|
  c.resource = { 'service.name' => 'my service', 'service.version' => '0.0.1', 'some.attr` => 'some value' }
 ...
end
```

I'm not sure if this a better solution, but I just wanted to throw it out there.

robertlaurin:
> I'm not sure if this a better solution, but I just wanted to throw it out there.

I agree with your sentiment. But if it ends up just being these two, then I think it is fine to leave this the way it is.  If we go for a third we should rework it to be a bit more flexible for adding resources.

fbogsany:
The current solution helps with maintaining semantic conventions without excessive verbosity. Repurposing the `c.resource=` method is nice, but doesn't help with either semantic conventions or verbosity.

mwear:
WFM. I think service name and service version are important enough to have as first class configure options. I suppose we can always consider repurposing `c.resource = ` as an additional niceity  in the future.

genebean:
It already does accept a hash. Note the custom attribute in this block:

```ruby
  c.resource = OpenTelemetry::Resource::Detectors::AutoDetector.detect
  c.resource = OpenTelemetry::SDK::Resources::Resource.create({
    OpenTelemetry::SDK::Resources::Constants::SERVICE_RESOURCE[:name] => service_name,
    OpenTelemetry::SDK::Resources::Constants::SERVICE_RESOURCE[:version] => version,
    'abs_component' => ENV['ABS_COMPONENT'],
  })
```

mwear:
I was recommending making the `OpenTelemetry::SDK::Resources::Resource.create` part optional, so that a hash could be passed in alone. It seems that looking up the attribute key constants is also a bit cumbersome. It was just an idea I wanted to float, but am happy with the first class config for service name and service version as-is.

genebean:
> I was recommending making the `OpenTelemetry::SDK::Resources::Resource.create` part optional

Ahh, didn't quite catch that that was the intent. I would be very much in favor of that.

PR #432, opened 2020-10-08
docs: beta

fbogsany:
Clarify beta status of tracing functionality. Shopify is starting to roll this out in production, and we think it is production-ready.


mwear (approved):
I definitely agree with the beta label.

PR #434, opened 2020-10-08
fix: set span name to sinatra.route

genebean:
The previous attempt at this failed due to sinatra.route not having been set yet. This moves the logic for using sinatra.route to later in the process and only updates the span name if sinatra.route exists.

### DIFF
--- a/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
+++ b/instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
@@ -15,10 +15,8 @@ module OpenTelemetry
           end
 
           def call(env)
-            span_name = env['sinatra.route'] || env['PATH_INFO']
-
             tracer.in_span(
-              span_name,
+              env['PATH_INFO'],
               attributes: { 'http.method' => env['REQUEST_METHOD'],
                             'http.url' => env['PATH_INFO'] },
               kind: :server,
@@ -46,6 +44,7 @@ module OpenTelemetry
             span.set_attribute('http.status_code', status)
             span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
+            span.name = env['sinatra.route'] if env['sinatra.route']
             span.status = OpenTelemetry::Trace::Status.http_to_status(status)
           end
         end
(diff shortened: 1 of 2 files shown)


instrumentation/sinatra/lib/opentelemetry/instrumentation/sinatra/middlewares/tracer_middleware.rb
```diff
@@ -46,6 +44,7 @@ def trace_response(span, env, resp)
             span.set_attribute('http.status_code', status)
             span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
             span.set_attribute('http.route', env['sinatra.route'].split.last) if env['sinatra.route']
+            span.name(env['sinatra.route'].split.last) if env['sinatra.route']
```
fbogsany, line 47:
```suggestion
            span.name = env['sinatra.route'].split.last if env['sinatra.route']
```

genebean: `name` looks to me to be a method just like `set_attribute` so I'd think the syntax is the same... am I missing something?

fbogsany: It's a "setter" (aka `attr_reader`), not a normal method: https://github.com/open-telemetry/opentelemetry-ruby/blob/bb1ad6be6f56e39b9c26e69febbc87ef19500c34/api/lib/opentelemetry/trace/span.rb#L114

fbogsany: Status is a similar thing: https://github.com/open-telemetry/opentelemetry-ruby/blob/bb1ad6be6f56e39b9c26e69febbc87ef19500c34/api/lib/opentelemetry/trace/span.rb#L103

genebean: Ahhh, thank you for educating me (seriously)

fbogsany (approved):

mwear (approved):

genebean:
I am looking at the test failure now. Totally forgot to run them locally :(

mwear:
> I am looking at the test failure now. Totally forgot to run them locally :(

Great, I was just going to mention the approval might have been a little premature.

genebean:
@fbogsany @mwear I updated the failing test and added a new one - I'd appreciate if you would take a second look to make sure they are in line with what you'd expect to see.

mwear:
LGTM

mwear:
Thanks @genebean as per the conversation on gitter, I'll push this out as a patch release.

PR #439, opened 2020-10-16
fix!: Move context/span methods to Trace module

fbogsany:
This implements the alternative to #425 outlined in https://github.com/open-telemetry/opentelemetry-ruby/pull/425#issuecomment-710096680


mwear (approved):
This looks good to me. I agree it's a bit more in the spirit of the spec.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -87,15 +87,11 @@ def create_frontend_span(env, extracted_context)
...
+            OpenTelemetry::Trace.context_with_span(span, parent_context: extracted_context)
           end
 
           def finish_span(context)
-            context[current_span_key]&.finish if context
-          end
-
-          def current_span_key
-            OpenTelemetry::Trace::Propagation::ContextKeys.current_span_key
+            OpenTelemetry::Trace.current_span(context)&.finish if context
```
fbogsany, line 94: I don't think we need the `&.` here. `current_span` will always return non-`nil`. If it returns `Span::INVALID`, we can safely call `.finish` on it - that's a no-op on the API `Span`.

PR #446, opened 2020-10-24
feat: Delayed Job instrumentation (ported from Datadog)

### LINKED ISSUE #251: DelayedJob instrumentation adapter
Convert Datadog instrumentation for DelayedJob to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/delayed_job)

johnnyshields:
Fixes #251 

Two questions:

1. Datadog had a `shutdown!` method which was called after job execution. Is this not needed in OpenTelemetry? I've commented it out in this PR.

2. ~~`span name`~~ --> resolved as per semantic conventions

Ready for review


johnnyshields:
@dazuma @ericmustin @fbogsany all tests green and ready for review

instrumentation/delayed_job/test/test_helper.rb
```diff
@@ -0,0 +1,49 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'delayed_job'
+require 'delayed_job_active_record'
+
+require 'opentelemetry/sdk'
+
+require 'minitest/autorun'
+require 'rspec/mocks/minitest_integration'
```
fbogsany, line 13: We've opted for Minitest rather than RSpec in this project. This is the first introduction of a RSpec feature. Is it necessary for the tests or can we mock without it?

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+              end
+            end
+
+            def instrument_invoke(job, &block) # rubocop:disable Metrics/AbcSize
+              return block.call(job) unless enabled?
+
+              tracer.in_span('delayed_job.invoke', kind: :consumer) do |span|
+                add_attributes(span, job)
+                span.set_attribute('delayed_job.attempts', job.attempts) if job.attempts
+                span.set_attribute('delayed_job.locked_by', job.locked_by) if job.locked_by
```
fbogsany, line 32: It is more efficient to compute the attributes before span creation and pass them into the `tracer.in_span` call.

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+                add_events(span, job)
+                begin
+                  yield job
+                rescue StandardError => e
+                  span.set_attribute('error', true)
+                  span.set_attribute('error.kind', e.class.name)
+                  span.set_attribute('message', e.message&.[](0...120))
+                  raise e
+                end
```
fbogsany, line 41: This is already handled in `tracer.in_span`, so you don't need it here.

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,92 @@
...
+
+            protected
+
+            def add_attributes(span, job)
+              span.set_attribute('component', 'delayed_job')
+              span.set_attribute('delayed_job.id', job.id)
+              span.set_attribute('delayed_job.name', job_name(job))
+              span.set_attribute('delayed_job.queue', job.queue) if job.queue
+              span.set_attribute('delayed_job.priority', job.priority)
+            end
```
fbogsany, line 59: The semantic conventions most applicable to job queuing systems are those for [messaging systems](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md). Based on this, I think we want attributes like:
```ruby
{
  'messaging.system' => 'delayed_job',
  'messaging.destination' => job.queue || 'unknown',
  'messaging.destination_kind' => 'queue',
  'messaging.message_id' => job.id,
  'messaging.delayed_job.name' => job_name(job),
  'messaging.delayed_job.priority' => job.priority
}
```

instrumentation/delayed_job/test/test_helper.rb
johnnyshields: I tried to use minitest's mocks, it was almost impossible to use and required installing 3+ extra gems. For mocks specifically I strongly recommend we use RSpec's mocks which are [compatible with minitest](https://relishapp.com/rspec/rspec-mocks/docs/outside-rspec/integrate-with-minitest)

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
johnnyshields: Hmm... I agree that "messaging" is the best fit among current semantic conventions. I will go with this for now and raise a separate proposal to add semantic conventions for jobs specifically.

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
johnnyshields: Done

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
johnnyshields: Done

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
johnnyshields: Done.

johnnyshields:
@fbogsany comments addressed. this is ready for another review

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,97 @@
...
+
+            def instrument_invoke(job, &block)
+              return block.call(job) unless enabled?
+
+              attributes = build_attributes(job)
+              attributes['messaging.delayed_job.attempts'] = job.attempts if job.attempts
+              attributes['messaging.delayed_job.locked_by'] = job.locked_by if job.locked_by
+              attributes['messaging.operation'] = 'process'
+              attributes['messaging.message_id'] = job.id
+              tracer.in_span("delayed_job.#{job_queue(job)}.process", attributes: attributes, kind: :consumer) do |span|
```
fbogsany, line 36:
```suggestion
              tracer.in_span("#{job_queue(job)} process", attributes: attributes, kind: :consumer) do |span|
```

Based on [Span name semantic conventions for messaging](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#span-name).

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,97 @@
...
+      module Middlewares
+        # Delayed Job plugin that instruments invoke_job and other hooks
+        class TracerMiddleware < Delayed::Plugin
+          class << self
+            def instrument_enqueue(job, &block)
+              return block.call(job) unless enabled?
+
+              attributes = build_attributes(job)
+              attributes['messaging.operation'] = 'send'
+              tracer.in_span("delayed_job.#{job_queue(job)}.send", attributes: attributes, kind: :producer) do |span|
```
fbogsany, line 21:
```suggestion
              tracer.in_span("#{job_queue(job)} send", attributes: attributes, kind: :producer) do |span|
```
Based on [Span name semantic conventions for messaging](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#span-name).

ericmustin:
@johnnyshields apols on the delay here, my fault. Will try to get a review on this this evening and followup on those ddog specific questions. First pass is this looks very strong

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/instrumentation.rb
```diff
@@ -0,0 +1,39 @@
...
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        MINIMUM_VERSION = Gem::Version.new('4.1')
+
+        install do |_config|
+          require_dependencies
+          register_tracer_middleware
+        end
+
+        present do
+          !defined?(::Delayed).nil? && gem_version >= MINIMUM_VERSION
```
robertlaurin, line 20: Curious why you're doing the explicit `nil` check here, we're doing `defined?(::Library)` everywhere else.  Are we missing an edge case?

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/instrumentation.rb
```diff
@@ -0,0 +1,39 @@
...
+        end
+
+        private
+
+        def gem_version
+          Gem.loaded_specs['delayed_job']&.version
```
robertlaurin, line 26: Is the safe operator needed?  If this returns `nil` it will raise when it tries to compare to the minimum version. If you remove it, then the error points to the source of the problem.

instrumentation/delayed_job/test/opentelemetry/instrumentation/delayed_job_test.rb
```diff
@@ -0,0 +1,75 @@
...
+      _(instrumentation.present?).must_equal true
+    end
+
+    it 'when delayed_job gem not installed' do
+      hide_const('Delayed')
+      _(instrumentation.present?).must_equal false
+    end
+
+    it 'when older gem version installed' do
+      allow_any_instance_of(Bundler::StubSpecification).to receive(:version).and_return(Gem::Version.new('4.0.3'))
```
robertlaurin, line 31: Let's use mini test here to keep things consistent with the rest of the repo.
```rb
Gem.stub(:loaded_specs, { 'delayed_job' => Gem::Specification.new { |s| s.version = '4.0.3' } }) do
  _(instrumentation.present?).must_equal false
end
```

instrumentation/delayed_job/test/test_helper.rb
robertlaurin: Out of curiosity which tests were you having trouble mocking?  I pretty exclusively used Rspec prior to working on this repo, but haven't found anything that doesn't have some sort of equivalent in minitest.

robertlaurin (commented):
This is looking great so far.  I noticed some commented out code, are you planning on continuing that or is that for another time?  If it's the latter I think it would be better to remove it.

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/instrumentation.rb
johnnyshields: I want this method to return exactly true or false (rather than nil).
* `defined?(::Delayed)` can return nil, so can't use it
* Rubocop doesn't like `!!defined?(::Delayed)`
* Therefore only remaining option is `!defined?(::Delayed).nil?`

johnnyshields:
@robertlaurin thanks for the review!

> commented out code

Datadog had a `shutdown!` method which was called after job execution, and this commented code is related to that. Is there a near-term plan to add this `shutdown!` method to Otel instrumenters? If not, I'll just delete it.

johnnyshields:
@robertlaurin I've removed the commented code for now. Ok to merge?

ericmustin:
@johnnyshields I think this is good it has just been, for various reasons, a bad week for me. I'd like to followup on the `shutdown!` method here as well. I plan to review this over the weekend, appreciate the patience here.

ericmustin:
So, generally everything looks great here, per usual nice work @johnnyshields . As background I don't have a lot of direct familiarity with using DelayedJob in production, so it's possible I've missed something here and defer to your opinion on this (i noticed you in the open Issues of delayed_job back in 2016 so i trust you're familiar with it 😅 )

Looking at the instrumentation this is ported from:

https://github.com/DataDog/dd-trace-rb/blob/2c4c5400fbf8c5f989815721468e810a7913a649/lib/ddtrace/contrib/delayed_job/plugin.rb#L48-L52

```
        def self.flush(worker, &block)
          yield worker

          tracer.shutdown! if tracer && tracer.enabled
        end
```

The `flush` method gets invoked around the `:execute` lifecycle callback

https://github.com/DataDog/dd-trace-rb/blob/2c4c5400fbf8c5f989815721468e810a7913a649/lib/ddtrace/contrib/delayed_job/plugin.rb#L80
```
          lifecycle.around(:execute, &method(:flush))
```

and here's the `execute` callback

https://github.com/collectiveidea/delayed_job/blob/b0adf2c3d976564b2ce7ce8b0cb424f4b1ee6ee4/lib/delayed/worker.rb#L171-L195

```
      self.class.lifecycle.run_callbacks(:execute, self) do
        loop do
          self.class.lifecycle.run_callbacks(:loop, self) do
            @realtime = Benchmark.realtime do
              @result = work_off
            end
          end

          count = @result[0] + @result[1]

          if count.zero?
            if self.class.exit_on_complete
              say 'No more jobs available. Exiting'
              break
            elsif !stop?
              sleep(self.class.sleep_delay)
              reload!
            end
          else
            say format("#{count} jobs processed at %.4f j/s, %d failed", count / @realtime, @result.last)
          end

          break if stop?
        end
      end
```

I think Datadog's instrumentation does this because the `Worker` is running as...a separate process typically? Is that correct? If so, ddog's tracer runs a background thread that handles flushing the span/trace buffer. I believe the issue is that if `:execute` completes, the process could exit and the background thread containing the span's/tracers in a buffer would never be flushed to..a datadog agent, in this case. So wrapping the `execute` calllback and ensuring `tracer.shutdown!` gets invoked ensures that any spans in memory get flushed.

So, for OpenTelemetry-Ruby, I think we have a similar paradigm, especially if you're using a `BatchSpanProcessor` to process/export traces, you'll need to call the `tracer_provider` `shutdown` method `https://github.com/open-telemetry/opentelemetry-ruby/blob/0099668e9ad7eedf32bb496e135e8220f1e49c61/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L52` To ensure that any freshly created span's from the worker get flushed to, in this case, the Otel-Collector, or jaeger collector, or what have you. @fbogsany @robertlaurin ,Am i correct here? (Sorry I have been jumping between a lot of different codebases recently and my mental models are starting to fail me). If so, we will want to add back in that commented out code to ensure we're flushing these spans here, otherwise in practice you'll end up with incomplete traces that are missing some spans from some traces.

Assuming the above is correct here, everything else in this implementation lgtm and i'd be keen to see this and `mongo` get merged in before our next release.

ericmustin (changes requested):
see note above. please feel free re-request review if i'm incorrect here (or when above feedback is implemented) and i'll approve

johnnyshields:
@ericmustin I've investigated this. _TLDR; I don't think any further change is needed here._

DelayedJob is multi-process, single-threaded per process. The process runs continuously; it does **not** respawn the process for each job.

Calling `shutdown` after each job doesn't work. Once you call shutdown, you can't reuse the tracer for spans for the next job. This breaks everything.

We could call `flush` after each job, however, I think this is bad practice. There is no problem to process multiple jobs and batch the spans for those jobs. I use a different telemetry library currently for DJ and I am **not** flushing after each job currently; it works fine.

The only time we would every actually need to `flush` is on process termination. A better way to do this would be for OpenTelemetry to trap `SIGTERM` and `SIGKILL` and flush when either of those are called.

ericmustin:
Thanks for the explanation here on delayed job, that definitely helps a lot. Fwiw, as I understand it, we wouldn't be calling `tracer.tracer_provider.shutdown` after _each_ job, but only after all jobs have been enqueued, 

ie this part: 
```
          if count.zero?
            if self.class.exit_on_complete
              say 'No more jobs available. Exiting'
              break
            elsif !stop?
              sleep(self.class.sleep_delay)
              reload!
            end
          else
            say format("#{count} jobs processed at %.4f j/s, %d failed", count / @realtime, @result.last)
          end

          break if stop?
```

This `execute` lifecycle only gets executed via `Delayed::Worker.new.start` (which is calling `work_off` under the hood). The more I work through it, the more concerns I have around attempting to force a flush+shutdown after `execute` though. It seems like cases exist where we'd shutdown the tracer too early, as `exit_on_complete` forces an exit after all jobs have been [_enqueued_ but not necessarily `invoked`](https://github.com/collectiveidea/delayed_job/blob/73bd1b50e719b336b70fcbb8dc4a37ec9b2f6f35/lib/delayed/command.rb#L77https://github.com/collectiveidea/delayed_job/blob/73bd1b50e719b336b70fcbb8dc4a37ec9b2f6f35/lib/delayed/command.rb#L77), And generally I am not familiar enough with the long tail of different use cases here to properly guard against them all to ensure we don't shutdown the tracer early, which IMO is considerably worse than not flushing a few stray spans. I do still think that possibility exists but, would not want perfect to be the enemy of good here (especially if attempting to do that ends up breaking tracing in more cases than it helps).

This is all a lot of sound and fury to say, lgtm, let's ship it.

ericmustin (approved):
LGTM, some discussion around forcing a flush on sigint/sigterm is probably worth understanding better but is not blocking here imo. 🚀

mwear (approved):
LGTM. Thanks @johnnyshields!

johnnyshields:
@dazuma please approve

instrumentation/delayed_job/lib/opentelemetry/instrumentation/delayed_job/instrumentation.rb
```diff
@@ -0,0 +1,39 @@
...
+        def gem_version
+          Gem.loaded_specs['delayed_job'].version
+        end
+
+        def require_dependencies
+          require_relative 'middlewares/tracer_middleware'
+        end
+
+        def register_tracer_middleware
+          ::Delayed::Worker.plugins << Middlewares::TracerMiddleware
```
fbogsany, line 34: I feel like we should be using the terminology of the thing we're instrumenting here, and call this `Plugins::TracerPlugin`.

fbogsany (changes requested):
I have a concern about terminology: not everything is "middleware". Otherwise, this LGTM.

johnnyshields:
@fbogsany change done to use Plugins namespace instead of Middleware

fbogsany (approved):

fbogsany:
@johnnyshields this needs a rebase too. To save yourself some time, please rebase #448 first. I'll merge that, then this can be rebased and I'll merge this.

robertlaurin (approved):

johnnyshields:
Rebased and ready to go

johnnyshields:
Rebased again

fbogsany:
Thanks @johnnyshields

johnnyshields:
Likewise, thank you!

PR #448, opened 2020-10-24
feat: Mongo Instrumentation (ported from Datadog)

johnnyshields:
Fixes https://github.com/open-telemetry/opentelemetry-ruby/issues/257

Ready for review


instrumentation/mongo/README.md
```diff
@@ -0,0 +1,53 @@
+# OpenTelemetry Mongo Instrumentation
+
+The OpenTelemetry Mongo Ruby gem is a community maintained instrumentation for [Mongo][mongo-home]. This is an HTTP client library that provides a common interface over many adaptors, such as Net::HTTP.
```
fbogsany, line 3:
```suggestion
The OpenTelemetry Mongo Ruby gem is a community maintained instrumentation for [Mongo][mongo-home].
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span.set_attribute('component', 'mongo-ruby')
+            span.set_attribute('db.type', 'mongodb')
+            span.set_attribute('db.instance', event.database_name)
+            span.set_attribute('mongo.command', event.command_name)
+            span.set_attribute('mongo.collection', collection) if collection
+            span.set_attribute('db.statement', serialized_command)
+            span.set_attribute('mongo.request_id', event.request_id)
+            span.set_attribute('mongo.op_id', event.operation_id)
+            span.set_attribute('peer.hostname', event.address.host)
+            span.set_attribute('peer.port', event.address.port)
```
fbogsany, line 40:
```suggestion
            span.set_attribute('net.peer.name', event.address.host)
            span.set_attribute('net.peer.port', event.address.port)
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span = tracer.start_span(name, kind: :client)
+            set_span(event, span)
+
+            serialized_command = CommandSerializer.new(event.command).serialize
+            collection = get_collection(event.command)
+            span.set_attribute('component', 'mongo-ruby')
+            span.set_attribute('db.type', 'mongodb')
+            span.set_attribute('db.instance', event.database_name)
+            span.set_attribute('mongo.command', event.command_name)
+            span.set_attribute('mongo.collection', collection) if collection
```
fbogsany, line 35:
```suggestion
            span.set_attribute('db.mongo.collection', collection) if collection
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            # https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/monitoring/publishable.rb#L38-L56
+            # TODO: decide naming convention
+            name = 'mongo.cmd'
+            # name = "#{event.database_name}.#{event.command_name}"
+            span = tracer.start_span(name, kind: :client)
+            set_span(event, span)
+
+            serialized_command = CommandSerializer.new(event.command).serialize
+            collection = get_collection(event.command)
+            span.set_attribute('component', 'mongo-ruby')
```
fbogsany, line 31:
```suggestion
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span.set_attribute('db.type', 'mongodb')
+            span.set_attribute('db.instance', event.database_name)
+            span.set_attribute('mongo.command', event.command_name)
```
fbogsany, line 34:
```suggestion
            span.set_attribute('db.operation', event.command_name)
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            # TODO: decide naming convention
+            name = 'mongo.cmd'
+            # name = "#{event.database_name}.#{event.command_name}"
+            span = tracer.start_span(name, kind: :client)
+            set_span(event, span)
+
+            serialized_command = CommandSerializer.new(event.command).serialize
+            collection = get_collection(event.command)
+            span.set_attribute('component', 'mongo-ruby')
+            span.set_attribute('db.type', 'mongodb')
```
fbogsany, line 32:
```suggestion
            span.set_attribute('db.system', 'mongodb')
```
This used to be `db.type` but appears to have changed to `db.system` at some point: https://github.com/open-telemetry/opentelemetry-specification/blob/323cd1d8811d3e3df08bec689f9dacd6cb782888/semantic_conventions/trace/database.yaml#L8

We will need to fix the Redis and MySQL instrumentation too.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span.set_attribute('db.instance', event.database_name)
```
fbogsany, line 33:
```suggestion
            span.set_attribute('db.name', event.database_name)
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span.set_attribute('mongo.command', event.command_name)
+            span.set_attribute('mongo.collection', collection) if collection
+            span.set_attribute('db.statement', serialized_command)
+            span.set_attribute('mongo.request_id', event.request_id)
+            span.set_attribute('mongo.op_id', event.operation_id)
```
fbogsany, line 38:
```suggestion
```
These are not part of the OTel semantic conventions for Mongo: https://github.com/open-telemetry/opentelemetry-specification/blob/323cd1d8811d3e3df08bec689f9dacd6cb782888/semantic_conventions/trace/database.yaml#L283-L295

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            span.set_attribute('peer.port', event.address.port)
+          end
+
+          def failed(event)
+            finish_event('failed', event) do |span|
+              if event.is_a?(::Mongo::Monitoring::Event::CommandFailed)
+                span.set_attribute('error', true)
+                span.set_attribute('error.kind', 'CommandFailed')
+                span.set_attribute('message', event.message)
+              end
```
fbogsany, line 49: The error should be reported in a similar fashion to https://github.com/open-telemetry/opentelemetry-ruby/blob/576bc41f45dc0286a5c4a3c01bdac5efd6a27028/sdk/lib/opentelemetry/sdk/trace/span.rb#L125-L132

In this case, I think it should be something like:
```suggestion
              if event.is_a?(::Mongo::Monitoring::Event::CommandFailed)
                add_event('exception',
                          attributes: {
                            'exception.type' => 'CommandFailed',
                            'exception.message' => event.message
                          })
              end
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            end
+          end
+
+          def succeeded(event)
+            finish_event('succeeded', event) do |span|
+              n = event.reply.fetch('n', nil)
+              span.set_attribute('mongo.n', n) unless n.nil?
+            end
+          end
```
fbogsany, line 58:
```suggestion
          def succeeded(event)
          end
```
This isn't part of the OpenTelemetry semantic conventions for Mongo.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+
+          def set_span(event, span)
+            Thread.current[THREAD_KEY] ||= {}
+            Thread.current[THREAD_KEY][event.request_id] = span
+          end
+
+          def clear_span(event)
+            return if Thread.current[THREAD_KEY].nil?
+
+            Thread.current[THREAD_KEY].delete(event.request_id)
```
fbogsany, line 92:
```suggestion
            Thread.current[THREAD_KEY]&.delete(event.request_id)
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+          def started(event) # rubocop:disable Metrics/AbcSize
+            # start a trace and store it in the current thread; using the `operation_id`
+            # is safe since it's a unique id used to link events together. Also only one
+            # thread is involved in this execution so thread-local storage should be safe. Reference:
+            # https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/monitoring.rb#L70
+            # https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/monitoring/publishable.rb#L38-L56
+            # TODO: decide naming convention
+            name = 'mongo.cmd'
+            # name = "#{event.database_name}.#{event.command_name}"
+            span = tracer.start_span(name, kind: :client)
```
fbogsany, line 26:
```suggestion
            collection = get_collection(event.command)
            span = tracer.start_span("#{collection || ''}.#{event.command_name}", kind: :client)
```
Based on example at https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md#mongodb

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,102 @@
...
+            set_span(event, span)
+
+            serialized_command = CommandSerializer.new(event.command).serialize
+            collection = get_collection(event.command)
```
fbogsany, line 30:
```suggestion
```
Moved this up to set the span name.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
johnnyshields: I think they should be... these will make it much easier to trace with respect to changes in the database, especially op_id which would allow the trace to be linked to the opLog

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
johnnyshields: Could we add this? This is the number of modified records. Again very useful to see in the trace and quickly explain latency.

johnnyshields:
@fbogsany thank you so much for the code review. Would it be possible to consider extending the OpenTelemetry schema to incorporate:

```
db.mongo.request_id
db.mongo.op_id
db.mongo.n  (or count)
```

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
fbogsany: Could you please raise this as an issue (or open a PR) in the spec repo? I suspect they should be `db.mongodb.request_id` and `db.mongodb.op_id`, based on existing semantic conventions.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
fbogsany: As above, could you please open an issue (or create a PR) for this in the spec repo? I think it'd require a prefix like `db.mongodb.` and should probably have a more meaningful name than `n`.

fbogsany:
> @fbogsany thank you so much for the code review. Would it be possible to consider extending the OpenTelemetry schema to incorporate:
> 
> ```
> db.mongo.request_id
> db.mongo.op_id
> db.mongo.n  (or count)
> ```

It is possible. Please open a spec issue (or PR) making the case for these.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
ericmustin: Ditto to @fbogsany comments here. I think it's a reasonable thing to add but i think it would gain broader support using a more descripitve naming convention

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,99 @@
...
+
+          def started(event) # rubocop:disable Metrics/AbcSize
+            # start a trace and store it in the current thread; using the `operation_id`
+            # is safe since it's a unique id used to link events together. Also only one
+            # thread is involved in this execution so thread-local storage should be safe. Reference:
+            # https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/monitoring.rb#L70
+            # https://github.com/mongodb/mongo-ruby-driver/blob/master/lib/mongo/monitoring/publishable.rb#L38-L56
+            # TODO: decide naming convention
+            collection = get_collection(event.command)
+            span = tracer.start_span("#{collection || ''}.#{event.command_name}", kind: :client)
```
ericmustin, line 25: Generally I think this fits the specification per https://github.com/open-telemetry/opentelemetry-specification/blob/acd6f771d6064b92738a96afd729c59ba8ab9c4d/specification/trace/semantic_conventions/database.md#mongodb

That being said it looks like: 

- python has a slightly different approach here, with `mongodb.<collection>.<operation>` https://github.com/open-telemetry/opentelemetry-python/blob/f3cdfa2cdb9cb1c442189e2ead3788f45d92352d/instrumentation/opentelemetry-instrumentation-pymongo/src/opentelemetry/instrumentation/pymongo/__init__.py#L66

- and js is even less descriptive but also prepends the name with `mongodb` https://github.com/open-telemetry/opentelemetry-js-contrib/blob/088a4d57afa7a36c8f86c9afaa82d18477c82e5c/plugins/node/opentelemetry-plugin-mongodb/src/mongodb.ts#L232 

whereas ruby is just `<collection>.<operation>`. It seems like it's not part of the spec but it couldn't hurt  to prepend with `mongodb`

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,99 @@
...
+
+require_relative 'command_serializer'
+
+module OpenTelemetry
+  module Instrumentation
+    module Mongo
+      module Middlewares
+        # Event handler class for Mongo Ruby driver
+        class Subscriber
+          THREAD_KEY = :__opentelemetry_mongo_spans__
```
ericmustin, line 15: Broadly speaking I know that while the Datadog Implementation relies on thread local storage for it's mongo instrumentation, I'm not sure what Otel wants to support or if we have more 1st class helpers to do this. I'd like to get some other opinions here.

instrumentation/mongo/test/test_helper.rb
```diff
@@ -0,0 +1,54 @@
...
+  def client
+    @client ||= Mongo::Client.new(["#{host}:#{port}"], database: database)
+  end
+
+  def database
+    'otel_test'
+  end
+
+  def host
+    ENV.fetch('TEST_MONGODB_HOST', '127.0.0.1')
```
ericmustin, line 48: I believe the datadog test approach relies on a separate mongodb container in the docker network. I think we'd need to add this env var pointing to the appropriate container in our docker-compose.yml https://github.com/open-telemetry/opentelemetry-ruby/blob/master/docker-compose.yml and add both the mongodb container and the relevant env vars to the ci.yml https://github.com/open-telemetry/opentelemetry-ruby/blob/master/.github/workflows/ci.yml

ericmustin (changes requested):
I think this broadly lgtm, left a few comments.

- there's a few outstanding items to do to make sure the test suite can run locally and in CI
- I'd like some more experienced eyes on the use of thread local storage here. it's possible there's a more canonical way to handle this within the otel api or sdk 
- a few minor nits on the attribute metadata naming and details

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
johnnyshields: I agree to prepend with mongodb to be consistent with python.

johnnyshields:
@ericmustin re: "use of thread local storage", I don't see any other way clean around this besides to use thread storage. The mongo-ruby-driver's instrumentation API gives us separate "start" and "end" callbacks, so we need to store the ID somewhere during the time between. They don't give us the equivalent of an "around (with yield)" callback. The only real alternative would be to hack-in our own monkey patch instrumentation which is ill-advised...

ericmustin:
@johnnyshields yup I think thread local is the way to go, I just mean that if we have some existing helpers within the API for managing this we should use those. I don't believe we do though, I think your implementation is good on this front.

instrumentation/mongo/test/test_helper.rb
johnnyshields: The CI passes fine here as-is. If you want to run this on a different CI, you will need to set `TEST_MONGODB_HOST` and `TEST_MONGODB_PORT` in that CI's runner script. I'm marking this as resolved as I don't see any code change is needed.

johnnyshields:
@ericmustin @fbogsany I believe I've addressed all comments. Can you please review again?

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
johnnyshields: Following discussion below, this should be acceptable for now, and if in the future OTEL ruby provides an API for thread storage we can refactor in a separate PR. Resolving discussion.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
```diff
@@ -0,0 +1,104 @@
...
+            span&.finish
+            clear_span(event)
+          end
+
+          def span_name(collection, command_name)
+            name = +'mongodb.'
+            name << "#{collection}." if collection
+            name << command_name
+            name.freeze
+          end
```
fbogsany, line 66:
```suggestion
          def span_name(collection, command_name)
            return command_name unless collection
            "#{collection}.#{command_name}"
          end
```
Despite what other languages have done, we should strive to follow the semantic conventions where possible. Deciding **not** to do that leads to further divergence rather than convergence.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
fbogsany: Ideally, we'd use `Context` for this, but that tends to assume consistent nesting of operations. I this case, it looks like we can't really avoid the thread-local.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
ericmustin: my bad @johnnyshields . Agreed, we should follow [The Spec](https://youtu.be/o31UcPeiUCc)

ericmustin (approved):
nice work here, i think everything has been addressed afaik

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/middlewares/subscriber.rb
johnnyshields: @fbogsany yep originally I actually tried using Context but then abandoned it.

johnnyshields:
Let's merge?

johnnyshields:
@mwear @dazuma please approve

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/instrumentation.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        def gem_version
+          Gem.loaded_specs['mongo']&.version
+        end
+
+        def require_dependencies
+          require_relative 'middlewares/subscriber'
+        end
+
+        def register_tracer_middleware
```
mwear, line 33: It might be more accurate to call this `register_subscriber` or something along those lines.

mwear (approved):
LGTM. One minor comment.

instrumentation/mongo/lib/opentelemetry/instrumentation/mongo/instrumentation.rb
fbogsany: It's also more accurate to call the `Middlewares` module something like `Subscribers`, or perhaps remove that module altogether and move `Subscriber` and friend to the `Mongo` module.

fbogsany (approved):

fbogsany:
LGTM after the CI failure is fixed.

johnnyshields:
@fbogsany specs fixed lets merge

fbogsany:
Thanks @johnnyshields. This needs a rebase before I can merge. I don't have permissions to rebase - can you please do that?

robertlaurin (approved):

johnnyshields:
Merged in master

johnnyshields:
Ready to merge

fbogsany:
Thanks @johnnyshields

PR #450, opened 2020-10-25
Correct "peer" attributes according to semantic conventions

johnnyshields:
Rename attributes according to https://opentracing.io/specification/conventions/

net.peer.name --> peer.hostname
net.peer.port --> peer.port


fbogsany:
@johnnyshields thanks for looking into this, however the semantic conventions this project follows are from OpenTelemetry, not OpenTracing. These are correct, according to https://github.com/open-telemetry/opentelemetry-specification/blob/323cd1d8811d3e3df08bec689f9dacd6cb782888/semantic_conventions/trace/general.yaml#L43-L50. I'm closing this PR. Please re-open it if you think this is incorrect.

PR #454, opened 2020-10-27
feat: instrument rails

robertlaurin:
This is the starting point for adding rails instrumentation, intentionally scoped down to tracing controller requests.

Todo: 
- [x] Add tests and support for exception apps
~- [ ] Add tests and support for inferring the destined controller action when it fails to hit the controller~


instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+              end
+            end
+          ensure
+            OpenTelemetry::Trace.current_span(extracted_context).finish if extracted_context
+          end
+
+          private
+
+          def request_destination_info(env)
+            ::Rails.application.routes.router.recognize(ActionDispatch::Request.new(env)) { |_route, params| return params }
```
robertlaurin, line 40: So I'm not sure if I should use this or not, we can actually infer the controller and action in this middleware **before** `@app.call` is made.  In the event that something goes wrong in a middleware we could use that information to normalize the operation name, but I don't know if I should consider this a cheap enough operation to do it on every request.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/railtie.rb
```diff
@@ -0,0 +1,29 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module Rails
+      # The Instrumentation class contains logic to detect and install the Rails
+      # instrumentation, while this Railtie is used to conventionally instrument
+      # the Rails application through its initialization hooks
+      class Railtie < ::Rails::Railtie
+        initializer 'opentelemetry.before_initialize' do |app|
+          app.middleware.insert_after(
+            ActionDispatch::RequestId,
```
robertlaurin, line 16: I don't really have any strong opinions on **where** we should put the tracing middleware by default.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/railtie.rb
```diff
@@ -0,0 +1,29 @@
...
+            OpenTelemetry::Instrumentation::Rails::Middlewares::TracerMiddleware
+          )
+        end
+
+        private
+
+        def config
+          Instrumentation.instance.config
+        end
```
fbogsany, line 25: What's this intended for? I don't see it used in this PR.

instrumentation/rails/test/test_helpers/configs/rails5.rb
```diff
@@ -0,0 +1,52 @@
...
+  require 'test_helpers/middlewares'
+
+  new_app = ::Rails5::Application.new
+  new_app.config.secret_key_base = 'secret_key_base'
+
+  # Ensure we don't see this Rails warning when testing
+  new_app.config.eager_load = false
+
+  # Prevent tests from creating log/*.log
+  new_app.config.logger = Logger.new('/dev/null')
```
fbogsany, line 26: You're already setting these above, right? Do you need them in both places?

instrumentation/rails/test/test_helpers/configs/rails6.rb
```diff
@@ -0,0 +1,59 @@
...
+  new_app.config.secret_key_base = 'secret_key_base'
+
+  # Ensure we don't see this Rails warning when testing
+  new_app.config.eager_load = false
+
+  # Prevent tests from creating log/*.log
+  new_app.config.logger = Logger.new('/dev/null')
+
+  # Required in Rails 6
+  new_app.config.hosts << 'example.org'
```
fbogsany, line 30: Same question here - do you need to set these in both places?

instrumentation/rails/test/test_helpers/configs/rails6.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  # Creates a lot of deprecation warnins on subsequent app initializations if not explicitly set.
```
fbogsany, line 32:
```suggestion
  # Creates a lot of deprecation warnings on subsequent app initializations if not explicitly set.
```

instrumentation/rails/test/test_helpers/configs/rails6.rb
```diff
@@ -0,0 +1,59 @@
...
+  new_app.config.action_view.finalize_compiled_template_methods = ActionView::Railtie::NULL_OPTION
```
fbogsany, line 33: Is this and the `config.hosts` above the only things different between Rails 5 and 6? If so, could you do the version check here and not have the versioned files?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/railtie.rb
robertlaurin: Cruft from the various iterations I played with.  I'll remove it until there's a use for it.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+            span.status = OpenTelemetry::Trace::Status.http_to_status(status)
+            span.set_attribute('http.status_code', status)
+            span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
+
+            controller = env['action_controller.instance']
+            return unless controller
+
+            span.name = "#{controller.class.name}.#{controller.action_name}"
+            span.set_attribute('rails.controller', controller.class.name)
+            span.set_attribute('rails.action', controller.action_name)
```
fbogsany, line 53: Although we find these very useful, they aren't part of the semantic conventions.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+
+          def set_attributes_after_request(span, env:, status:, headers:, response:)
+            span.status = OpenTelemetry::Trace::Status.http_to_status(status)
+            span.set_attribute('http.status_code', status)
+            span.set_attribute('http.status_text', ::Rack::Utils::HTTP_STATUS_CODES[status])
+
+            controller = env['action_controller.instance']
+            return unless controller
+
+            span.name = "#{controller.class.name}.#{controller.action_name}"
```
fbogsany, line 51: The spec [says](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions):
> Given an inbound request for a route (e.g. `"/users/:userID?"`) the `name` attribute of the span SHOULD be set to this route. If the route does not include the application root, it SHOULD be prepended to the span name.
>
> If the route cannot be determined, the `name` attribute MUST be set as defined in the general semantic conventions for HTTP.

Do we know the route here? If so, we should try to use that for the name instead. If not, then I think what you have here is right (and preferable).

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+            span.set_attribute('rails.controller', controller.class.name)
+            span.set_attribute('rails.action', controller.action_name)
+          end
+
+          def request_span_attributes(env:)
+            {
+              'http.method' => env['REQUEST_METHOD'],
+              'http.host' => env['HTTP_HOST'] || 'unknown',
```
fbogsany, line 59: If the host header is empty or absent, we should instead set either:
1. `http.server_name` and `net.host.port` OR
2. `net.host.name` and `net.host.port`

Note that `http.server_name` is supposed to be determined from configuration rather than headers - in our (Shopify's) case, I think we'd use the Shop's primary domain name.

I would probably try to use the second option here, since the first one is app-dependent.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+                    env: env,
+                    status: status,
+                    headers: headers,
+                    response: response
+                  )
+                end
+              end
+            end
+          ensure
+            OpenTelemetry::Trace.current_span(extracted_context).finish if extracted_context
```
fbogsany, line 34: I believe this is incorrect - I think the `current_span` of the `extracted_context` is the "propagation span" (which does nothing on `.finish`), not the span you created above with `tracer.in_span`.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+          end
+
+          private
+
+          def request_destination_info(env)
+            ::Rails.application.routes.router.recognize(ActionDispatch::Request.new(env)) { |_route, params| return params }
+          end
```
fbogsany, line 41: What's this for? I don't see it used in this PR.

robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/pull/454#discussion_r513594562

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env)
+            extracted_context = OpenTelemetry.propagation.http.extract(env)
+            OpenTelemetry::Context.with_current(extracted_context) do
+              tracer.in_span(env['PATH_INFO'], attributes: request_span_attributes(env: env), kind: :server) do |request_span|
+                @app.call(env).tap do |status, headers, response|
+                  set_attributes_after_request(
```
fbogsany, line 23: How does this work with exceptions? Even in the presence of exceptions, we should try to set some of the attributes (we just need to be more defensive than usual).

instrumentation/rails/test/opentelemetry/instrumentation/rails/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,106 @@
...
+    _(span.attributes['http.status_code']).must_equal 307
+  end
+
+  describe 'when the application has an exceptions controller configured' do
+    let(:rails_app) { build_app(use_exceptions_controller: true) }
+
+    it 'traces the request handled by the exceptions_app' do
+      get '/exception'
+
+      _(span.name).must_equal 'ExceptionsController.show'
```
fbogsany, line 91: Eek! Please, no! We should really try very hard for this to be the "original" controller.action pair.

instrumentation/rails/test/opentelemetry/instrumentation/rails/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,106 @@
...
+      _(span.kind).must_equal :server
+      _(span.status.ok?).must_equal true
```
fbogsany, line 93: This is not OK :trollface:

instrumentation/rails/test/opentelemetry/instrumentation/rails/middlewares/tracer_middleware_test.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      _(span.attributes['http.method']).must_equal 'GET'
+      _(span.attributes['http.host']).must_equal 'example.org'
+      _(span.attributes['http.scheme']).must_equal 'http'
+      _(span.attributes['http.target']).must_equal '/exception'
+      _(span.attributes['http.status_code']).must_equal 200
```
fbogsany, line 99: 🤔 I'd normally expect this to be an error code of some kind - a 500, in this case.

instrumentation/rails/test/test_helpers/middlewares/redirect_middleware.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+class RedirectMiddleware
+  def initialize(app, _options = {})
+    @app = app
+  end
+
+  def call(env)
+    return [307, {}, 'Temporary Redirect'] if env['PATH_INFO'] == '/redirection'
```
fbogsany, line 13: For the 🥜 gallery, this represents something we've seen in a production app at Shopify 😄. Really anything would be fine for this test - it just represents middleware that returns early, without calling the `@app`.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/railtie.rb
fbogsany: Our Shopify tracing middleware stuck it here because we wanted to record the request ID as an attribute. We're not doing that here (but maybe we should?).

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+            span.set_attribute('rails.action', controller.action_name)
+          end
+
+          def request_span_attributes(env:)
+            {
+              'http.method' => env['REQUEST_METHOD'],
+              'http.host' => env['HTTP_HOST'] || 'unknown',
+              'http.scheme' => env['rack.url_scheme'],
+              'http.target' => fullpath(env)
+            }
```
fbogsany, line 62: IMO we should also include the user agent. It's optional in the spec, but often quite useful.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
fbogsany: I feel like that should be a fallback rather than the default, but I don't really know the cost of the operation.

Does the `_route` contain any useful info, or just the `params`?

robertlaurin: The route object contains a lot of info, I spent some time digging in it and the most useful thing IMO was the `requirements` method which == params.

#### For our example controller
```rb
Route = #<ActionDispatch::Journey::Route:0x00007fa1f0719eb0>
Params = {:controller=>"example", :action=>"ok"}
```

#### If you hit the built in "welcome to rails" page
```rb
Route = []
# We get the namespace as well
Params = {:controller=>"rails/welcome", :action=>"index"}
```

The route object details can be found here https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/journey/route.rb

Note that params is just [`route.requirements`](https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/journey/route.rb#L91-L103).

instrumentation/rails/test/test_helpers/configs/rails6.rb
robertlaurin: The above config was not needed as the `.new` call doesn't carry over any of that config.  I've removed that duplication. In both places.

instrumentation/rails/test/test_helpers/configs/rails6.rb
robertlaurin: Yeah, I've moved some stuff around based on this comment.  I think it's an improvement 🤞

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
robertlaurin: Should these be removed?  I wouldn't expect the specs around conventions to cover every library and framework so for something like this are we able to take some liberties?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
fbogsany: From the code you pointed at:
```
      #   get 'photo/:id', :controller => 'photos', :action => 'show',
```
I think the spec would like us to use `photo/:id` as the span name, but I'm not entirely sure how to get that from the Route.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
fbogsany: IDK - in general, I've tried to hew pretty close to the semantic conventions in "official" auto-instrumentation. I'll leave it up to other folks to chime in on this point. Technically, we could extend this in our wrapper gem to add them 🤷 . That said, I wonder if the span name has enough already (e.g. these could be added back in using an attributeprocessor in the collector).

instrumentation/rails/test/opentelemetry/instrumentation/rails/middlewares/tracer_middleware_test.rb
robertlaurin: Updated the example to be less confusing :)

mwear (commented):
This is off to a good start. I have a couple of comments about how things should behave if Rack instrumentation in the mix and a few questions about the approach in general.

I am curious if you considered using ActiveSupport::Notifications for this instrumentation. While ActionController can be instrumented via a middleware, and I am more accustomed to seeing Rails components (in general) instrumented using ActiveSupport::Notifications. This can be accomplished by subscribing to the `process_action.action_controller` event. The event payload would look like this: https://edgeguides.rubyonrails.org/active_support_instrumentation.html#process-action-action-controller.

Glancing at the dd-trace instrumentation, I notice it monkey patches `process_action` in ActionController::Metal. The dd-trace approach and ActiveSupport approach are both measuring the same thing. The middleware approach measures something slightly different, and exactly what is measured will depend on the placement of the middleware.

I'm not tied to either approach, but I thought it might be worth a discussion.

Also, as I was looking at dd-trace, I noticed that it has some logic to detect if the controller that is rendering an action is a custom exception controller. Often users will set up a catchall exceptions controller, and if an exception occurs, it will handle the request. In the shuffle, spans can get attributed the exception controller, when you really want them to be attributed to the controller where the exception happened. We don't necessarily have to address that in this initial PR, but it's an enhancement we should keep in mind.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
mwear: This is a good question. I think the main goal with the semantic conventions is so that all instrumentation records a set of common attributes for spans representing common scenarios. I'm not completely sure what the guidance is on collecting more than is required. I can see arguments in both directions.

fbogsany:
Thanks for the feedback @mwear. @robertlaurin and I discussed many of the same things on a call today. Summary is:
1. Let's make this gem dependent on the Rack instrumentation gem, have a single (Rack) middleware, and create a single `server` span for the request.
1. We'll stick that span in the context (with an accessor in the Rack instrumentation), allowing things later in the call chain to add or overwrite information.
1. We planned to use a similar `ActionController::Metal` 🐒 patch to Datadog's, but if we can get away with ASN, that'd be preferable.
1. Use the presence of `"action_dispatch.exception"` in the request headers, where possible, to infer that we're in an exceptions controller.

In general, I have concerns about relying on ASN for instrumentation. My information may be out-of-date, but previously, if multiple subscribers were notified of the same event, and one of them raised, the remainder would not receive the notification. We have some complex and messy code in Shopify's tracing gem to work around this issue - it can lead to either failure to end a span or an attempt to end a span that doesn't exist.

I would greatly welcome evidence that I'm wrong about this in all the versions of Rails we intend to support. If I'm right, though, then I'd favour the `ActionController::Metal` 🐒 patch.

robertlaurin:
Something I'd like to raise for discussion is span naming for rails controller actions.

The spec says we *SHOULD* follow the low cardinality version of the path (`/users/:userID`)

Coming from a rails dev background I have a strong bias towards some variation of `ControllerName#action_name` as the operation name for the span. 

I have done a bit of work to see how viable each approach would be, and I'm in the opinion that using the Controller/action name is the least brittle/simpler approach in code.  

I think @fbogsany and I landed on the same page for this, but I wanted to engage more people to see what the sentiments were around this.

cc: @mwear @ericmustin

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
```diff
@@ -0,0 +1,79 @@
...
+      module Middlewares
+        # TracerMiddleware propagates context and instruments Rails requests
+        # by way of its middleware system
+        class TracerMiddleware
+          def initialize(app)
+            @app = app
+          end
+
+          def call(env)
+            extracted_context = OpenTelemetry.propagation.http.extract(env)
```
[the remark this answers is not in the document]
robertlaurin: I'm going to move the PR back to draft and rework it to depend on rack instrumentation as mentioned in the nice summary Francis posted.  I'll put this back up as ready for review when once the churn slows down on the PR and is more ready for finer review.

mwear:
@robertlaurin I am fine with either. I'm also accustomed to seeing spans named `Controller#action_name`, I think this makes sense, especially if you need to do any gymnastics to get the quantized path name.

@fbogsany (and Robert) The plan in https://github.com/open-telemetry/opentelemetry-ruby/pull/454#issuecomment-719105800 looks good to me!

instrumentation/rails/lib/opentelemetry/instrumentation/rails/middlewares/tracer_middleware.rb
robertlaurin: Considering we are leaning towards make the op name a combination of the two, I think we could probably get away with omitting it.

ericmustin:
@robertlaurin I still need to give this a good review, apols on the delay, just chiming in here to say:

> Coming from a rails dev background I have a strong bias towards some variation of ControllerName#action_name as the operation name for the span.

I agree with this very strongly.

robertlaurin:
@fbogsany @mwear @ericmustin Wanted to get some opinions on how Rails should depend on the rack instrumentation.

**Option 1)** I can make Rack an implicit dependency that is required as part of the gem, which comes with the benefit of it being an easier configuration process for the users of this gem.  They require the rails instrumentation, configure it and don't have to worry about rack at all.

This approach raised a couple things:
  - Should spans generated by the middleware should be considered instrumented by rack or rails (we tag the instrumentation library)?
  - The rack tracing middleware assumes it has it's own `instance.config` and `instance.tracer`, should this be configurable as a middleware parameter?  The install flow for rack will not get ran in this setup, or is not obvious how it would in this setup.

**Option 2)** Rack can be an explicit dependency of Rails instrumentation.  This would mean if you wanted your controller requests traced you would include the rack instrumentation gem and configure it to run and the rails instrumentation would look for it's presence and then add the middleware if it was present.

- Spans would be marked as instrumented by Rack
- All the existing rack configuration and its documentation would be available to users of the gem
- Risk running into configuration order problems if rails install flow runs before rack is required (they need to know about each other to add the middleware).

Over all I'm leaning towards **Option 1** as it seems simpler from an end user perspective, just thought I would raise it before I make decisions in code.    I'm sure there's other things to consider that I have not thought of yet, the above is just what immediately jumped out at me when I started working on it.

fbogsany:
TBH I had not considered the `InstrumentationLibrary` issue. Thanks for pointing that out @robertlaurin. 🤔

fbogsany:
I feel like **Option 1** is the way to go, with the span being "owned" by the Rack instrumentation, however the Rack instrumentation has a bunch of config options. I think we need to _allow_ the user to configure those, but not _require_ them to. I'm not yet clear on how to do that - it may require changes to the `Configurator`.

Hmm, maybe I'm leaning toward **Option 2** 😄. I'm not sure "they need to know about each other to add the middleware" or what order dependencies there might be in configuration. There will be a single Rack middleware and it will stick its span into the context. The Rails `ActionController::Metal` instrumentation will retrieve the span from the context and decorate it - if it doesn't find it in the context, then there is nothing to decorate. Alternatively, if it can't find the Rack span in the context, it could create its own span?

robertlaurin:
>. I think we need to allow the user to configure those, but not require them to. I'm not yet clear on how to do that - it may require changes to the Configurator.

I'm thinking that the rails config flow would pass in its own `Instrumentation.instance` into the rack middleware and amend the middleware to favour the passed in instrumentation instance over trying to reach for its own (which will not be present).  This means that the spans would be tagged as instrumented as rails instrumentation library.  This also doesn't require any changes to the configurator.

fbogsany:
Does that mean all the Rack config options would instead be set in the Rails config hash?

ericmustin:
@robertlaurin , nice work here.

So, I figured I could try to pass along what i've learned here on the patching approach, following up on the discussion between @mwear and @fbogsany . I spoke with some colleagues on the tradeoffs of using `AS Notification process_action.action_controller` vs monkeypatching `Metal`. The main drawback of `process_action.action_controller` is that, I believe, it won't include any before/after callbacks among it's timing details, such as those [here](http://github.com/rails/rails/blob/7547a6b07399f7ea85d117a3b16eb46c471b6067/actionpack/lib/abstract_controller/callbacks.rb#L9-L20).  Patching `Metal` would capture those timing details.

Some more minor areas are that we'd have to be sure to patch `Metal` and not `Base`, as `API` Controllers inherit directly from `Metal`. Additionally, from my investigation (I did what i could given bandwidth, but the rails codebase is still not something i've studied as closely as I'd like to) there's the possibility that someone is running a very slim `API` Controller that does not include the `Instrumentation` module, as there's [some helpers that allow users to remove it](https://github.com/rails/rails/blob/bd28db1400609cba0a3e7f053d07159588d27d35/actionpack/lib/action_controller/api.rb#L104). This one is extremely minor but worth mentioning.

So, long story short, i think monkeypatching `Metal` is a more robust approach here vs Active Support Notifications. While we wouldn't get separate before/after spans OOTB for free, we'd at least accurately track the amount of time Rails is _"doing work associated with this this controller"_ which, i _think_ is what we want here.

robertlaurin:
> Does that mean all the Rack config options would instead be set in the Rails config hash?

Yes because there would be no explicit call to the rack config/install hook (or shouldn't be afaik).  In option 1, I imagined all config would go through rails and be delegated to whatever underlying piece needs it.

johnnyshields:
@robertlaurin I'd like to clarify the scope here. Which of the following will be covered by this PR:
* action_controller (yes)
* action_dispatch
* action_view
* active_job
* active_job_enqueue
* active_model_serializers

robertlaurin:
> @robertlaurin I'd like to clarify the scope here. Which of the following will be covered by this PR?
> 
> * action_controller (yes)
> * action_dispatch
> * action_view
> * active_job
> * active_job_enqueue
> * active_model_serializers

This PR is scoped to trace a request and response to a Rails app.

You have action controller and action dispatch as separate bullet points, is that meant to imply that you think they should traced as distinct units?  

That was not my intention, as it stands rack middleware will start and finish the span, while a patch to metal will append additional controller information to the span.

I will not be including anything beyond that as part of this PR.  The goal of this PR is to establish a starting point for our rails instrumentation so that it can be added to while still providing some basic value to anyone who adopts it.

> Perhaps each of these should be it's own gem?

This is an interesting idea, but I would prefer start with this being a single gem that could later be broken apart if it makes sense to do so.

johnnyshields:
@robertlaurin perhaps this instrumenter should be called "ActionController" instead of "Rails"? (or perhaps "Rails Metal" if you're going that approach).

I think it's very important to also instrument the Views and Serializers, as that can tell you a lot about where you are spending CPU time. Those would need to use ActiveSupport Notifications.

robertlaurin:
> I think it's very important to also instrument the Views and Serializers, as that can tell you a lot about where you are spending CPU time. Those would need to use ActiveSupport Notifications.

Agree that it's valuable, just not doing it as part of this PR.  I'll be doing follow up PRs to improve on the rails instrumentation.

johnnyshields:
Would propose ActionView/ActiveModelSerializers they be done as their own instrumenters so users can choose whether they want them on/off

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -15,6 +15,8 @@ module Middlewares
         # TracerMiddleware propagates context and instruments Rack requests
         # by way of its middleware system
         class TracerMiddleware # rubocop:disable Metrics/ClassLength
+          THREAD_KEY = :__opentelemetry_rack_span__
```
fbogsany, line 18: Why not use `OpenTelemetry::Context` here instead of explicit thread locals? It'd look something like https://github.com/open-telemetry/opentelemetry-ruby/blob/7bf30855fa5353701457ecac183fc495886cbec3/api/lib/opentelemetry/trace.rb#L14-L16 and https://github.com/open-telemetry/opentelemetry-ruby/blob/7bf30855fa5353701457ecac183fc495886cbec3/api/lib/opentelemetry/trace.rb#L46-L73
probably on the `OpenTelemetry::Instrumentation::Rack` module.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -63,13 +65,15 @@ def call(env)
               tracer.in_span(request_span_name,
                              attributes: request_span_attributes(env: env),
                              kind: request_span_kind) do |request_span|
+                Thread.current[THREAD_KEY] = request_span
```
fbogsany, line 68: Then here you'd:
```suggestion
                OpenTelemetry::Instrumentation::Rack.with_span(request_span) do ...
```

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
```diff
@@ -0,0 +1,26 @@
...
+  module Instrumentation
+    module Rails
+      module Patches
+        module ActionController
+          # Module to prepend to ActionController::Metal for instrumentation
+          module Metal
+            THREAD_KEY = :__opentelemetry_rack_span__
+
+            def dispatch(name, request, response)
+              rack_span = Thread.current[THREAD_KEY]
```
fbogsany, line 17:
```suggestion
              rack_span = OpenTelemetry::Instrumentation::Rack.current_span
```

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    module Rails
+      module Patches
+        module ActionController
+          # Module to prepend to ActionController::Metal for instrumentation
+          module Metal
+            THREAD_KEY = :__opentelemetry_rack_span__
+
```
fbogsany, line 15:
```suggestion
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: This is much better 😄

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
```diff
@@ -0,0 +1,24 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module Rails
+      module Patches
+        module ActionController
+          # Module to prepend to ActionController::Metal for instrumentation
+          module Metal
+            def dispatch(name, request, response)
+              rack_span = OpenTelemetry::Instrumentation::Rack.current_span
+              rack_span.name = "#{self.class.name}##{name}" if rack_span && !request.env['action_dispatch.exception']
```
fbogsany, line 16:
```suggestion
              rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
```

fbogsany (approved):
LGTM, thanks @robertlaurin! This is a great starting point for Rails instrumentation ❤️. Let's wait to hear from @ericmustin or @mwear before we merge.

ericmustin (approved):
lgtm

mwear (approved):
Nice work @robertlaurin!

johnnyshields:
Want to again voice my opinion that this should be called `ActionController` instrumentation rather than `Rails`--unless we have plans to include ActionView, ActiveModelSerializers, etc. into the same gem.

fbogsany:
I think this is a good start for Rails instrumentation. We can debate how we want to evolve this, and once that becomes clear, we can consider renaming the gem. For now, I'm going to merge this so we can keep the momentum.

fbogsany:
Before I merge @robertlaurin, can you please add the gem to the `gems` list in https://github.com/open-telemetry/opentelemetry-ruby/blob/master/.toys/.data/releases.yml

fbogsany (approved):

PR #460, opened 2020-10-28
fix: signal at batch_size

fbogsany:
... instead of half max queue size. This makes the trigger independent of the queue size. We're running with this in production and it has helped significantly reduce dropped spans.


robertlaurin (approved):

ericmustin (approved):
👍 lgtm. I think there are some slight tradeoffs here (more network traffic potentially) but they're well worth it.

PR #463, opened 2020-10-31
fix: BatchSpanProcessor should not spawn a thread during app boot for forking webservers

### LINKED ISSUE #462: OTEL BatchSpanProcessor should not spawn a thread during Puma/Unicorn boot
When using a multi-process webserver such as Puma or Unicorn, no threads should be started during the boot process. These threads won't survive to the child workers.

Here's my config:

```
# /config/initializers/opentelemetry.rb

    exporter = OpenTelemetry::Exporter::OTLP::Exporter.new(insecure: true)
    processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(exporter: exporter)
    c.use_all
  end
```

Puma complains `WARNING: Detected 1 Thread(s) started in app boot` when I use the OTEL ruby agent. See example log below from latest Puma:

```
$ WEB_CONCURRENCY=2 bundle exec puma

[19991] Puma starting in cluster mode...
[19991] * Version 5.0.2 (ruby 2.6.6-p146), codename: Spoony Bard
[19991] * Min threads: 5, max threads: 5
[19991] * Environment: development
[19991] * Process workers: 2
[19991] * Preloading application
I, [2020-11-01T00:14:33.022327 #19991]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::ConcurrentRuby was successfully installed
I, [2020-11-01T00:14:33.024353 #19991]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Dalli was successfully installed
I, [2020-11-01T00:14:33.025486 #19991]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Excon was successfully installed
[19991] * Listening on http://0.0.0.0:9292
[19991] ! WARNING: Detected 1 Thread(s) started in app boot:
[19991] ! #<Thread:0x00007ff7dfb90080@/home/shields/.rvm/gems/ruby-2.6.6/gems/opentelemetry-sdk-0.7.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:147 sleep> - /home/shields/.rvm/gems/ruby-2.6.6/gems/opentelemetry-sdk-0.7.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:131:in `sleep'
[19991] Use Ctrl-C to stop
[19991] * Starting control server on http://0.0.0.0:9293
[19991] - Worker 0 (pid: 20003) booted, phase: 0
[19991] - Worker 1 (pid: 20007) booted, phase: 0
```

This issue has been fixed in other telemetry libs, see for example here:
https://github.com/newrelic/newrelic-ruby-agent/commit/57cd9711003456288d6fe81aeda637173416d075

Here's a bit more info:
https://github.com/puma/puma/issues/614

johnnyshields:
BatchSpanProcessor should not spawn a thread during app boot for forking webservers (passenger puma rainbows unicorn)

I have added code which detects the webserver. You can also set it explicity with OTEL_DISPATCHER env var.

Fixes #462

TODO:
* [ ] determine where to put DispatcherDetector (i.e. which namespace)
* [ ] add tests to DispatcherDetector after approach is confirmed by code reviewer

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -51,6 +51,7 @@ module OpenTelemetry
             raise ArgumentError if max_export_batch_size > max_queue_size
 
             @exporter = exporter
+            @dispatcher = DispatcherDetector.dispatcher
             @exporter_timeout_seconds = exporter_timeout_millis / 1000.0
             @mutex = Mutex.new
             @condition = ConditionVariable.new
@@ -61,7 +62,7 @@ module OpenTelemetry
             @spans = []
             @pid = nil
             @thread = nil
-            reset_on_fork
+            reset_on_fork(restart_thread: !forking_dispatcher?)
           end
 
           # does nothing for this processor
@@ -173,6 +174,10 @@ module OpenTelemetry
             @mutex.synchronize do
               yield
             end
+          end
+
+          def forking_dispatcher?
+            %i[passenger puma rainbows unicorn].include?(@dispatcher)
           end
         end
       end
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -47,11 +47,11 @@ module OpenTelemetry
                          exporter_timeout_millis: Float(ENV.fetch('OTEL_BSP_EXPORT_TIMEOUT_MILLIS', 30_000)),
                          schedule_delay_millis: Float(ENV.fetch('OTEL_BSP_SCHEDULE_DELAY_MILLIS', 5_000)),
                          max_queue_size: Integer(ENV.fetch('OTEL_BSP_MAX_QUEUE_SIZE', 2048)),
-                         max_export_batch_size: Integer(ENV.fetch('OTEL_BSP_MAX_EXPORT_BATCH_SIZE', 512)))
+                         max_export_batch_size: Integer(ENV.fetch('OTEL_BSP_MAX_EXPORT_BATCH_SIZE', 512)),
+                         start_thread_on_boot: OpenTelemetry::SDK::Internal.to_boolean(ENV['OTEL_START_EXPORT_THREAD_ON_BOOT']))
             raise ArgumentError if max_export_batch_size > max_queue_size
 
             @exporter = exporter
-            @dispatcher = DispatcherDetector.dispatcher
             @exporter_timeout_seconds = exporter_timeout_millis / 1000.0
             @mutex = Mutex.new
             @condition = ConditionVariable.new
@@ -62,7 +62,8 @@ module OpenTelemetry
             @spans = []
             @pid = nil
             @thread = nil
-            reset_on_fork(restart_thread: !forking_dispatcher?)
+            start_thread_on_boot = !forking_dispatcher? if start_thread_on_boot.nil?
+            reset_on_fork(restart_thread: start_thread_on_boot)
           end
 
           # does nothing for this processor
@@ -177,7 +178,7 @@ module OpenTelemetry
           end
 
           def forking_dispatcher?
-            %i[passenger puma rainbows unicorn].include?(@dispatcher)
+            %i[passenger puma rainbows unicorn].include?(DispatcherDetector.dispatcher)
           end
         end
       end
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -16,6 +16,16 @@ module OpenTelemetry
         value.is_a?(TrueClass) || value.is_a?(FalseClass)
       end
 
+      def to_boolean(value)
+        return nil if value.nil?
+        return value if boolean?(value)
+
+        case value.to_s
+        when /\A(true|t|yes|y|1)\z/i then true
+        when /\A(false|f|no|n|0)\z/i then false
+        end
+      end
+
       def valid_key?(key)
         key.instance_of?(String)
       end
--- a/sdk/lib/opentelemetry/sdk/trace/export.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export.rb
@@ -27,6 +27,7 @@ module OpenTelemetry
   end
 end
 
+require 'opentelemetry/sdk/trace/export/dispatcher_detector'
 require 'opentelemetry/sdk/trace/export/batch_span_processor'
 require 'opentelemetry/sdk/trace/export/console_span_exporter'
 require 'opentelemetry/sdk/trace/export/in_memory_span_exporter'
--- a/sdk/lib/opentelemetry/sdk/trace/export/dispatcher_detector.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/dispatcher_detector.rb
@@ -66,7 +66,7 @@ module OpenTelemetry
           end
 
           def detect_glassfish
-            defined?(::JRuby) && defined?(::JRuby::Rack::VERSION) && defined?(::GlassFish::Server)
+            defined?(::JRuby::Rack::VERSION) && defined?(::GlassFish::Server)
           end
 
           def detect_litespeed
@@ -74,7 +74,7 @@ module OpenTelemetry
           end
 
           def detect_mongrel
-            defined?(::Mongrel) && defined?(::Mongrel::HttpServer)
+            defined?(::Mongrel::HttpServer)
           end
 
           def detect_passenger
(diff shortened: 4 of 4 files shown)


fbogsany:
@johnnyshields can you please sign the CLA? This is blocking all your PRs at the moment.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -174,6 +175,10 @@ def lock
               yield
             end
           end
+
+          def forking_dispatcher?
+            %i[passenger puma rainbows unicorn].include?(@dispatcher)
```
fbogsany, line 180: This is brittle, and this knowledge should not be in the SDK. If the functionality is necessary (and I'm not really convinced it is), we should find a way to configure this in the instrumentation for each of these runtimes rather than hardcoding a list in the SDK.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -61,7 +62,7 @@ def initialize(exporter:,
             @spans = []
             @pid = nil
             @thread = nil
-            reset_on_fork
+            reset_on_fork(restart_thread: !forking_dispatcher?)
```
fbogsany, line 65: Note that simply removing this line should lazily start the thread the first time a span is finished.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
johnnyshields: This is the exact way it is implemented in other telemetry libs. It's "brittle" only in the sense that if someone creates a new forking ruby server we'd need to add it to the list; it's not brittle in the sense that it will spontaneously break for these specific widely-used servers. As far as I'm aware there is no way to pre-emptively know that a process is going to fork, unless we [violate the laws of causality](https://www.youtube.com/watch?v=AZGcmvrTX9M)

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
johnnyshields: Yep, I'm aware and this is intentional. In general it is desirable to start the BatchSpanProcessor thread at boot time **except** for forking web servers.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: > It's "brittle" only in the sense that if someone releases a new forking ruby server we'd need to add it to the list

This is exactly the problem. For example, a quick search turns up https://github.com/bdurand/sidekiq-process_manager for Sidekiq. We have no idea what pre-forking runtimes exist as private repos. It should not be necessary to fork OpenTelemetry Ruby or to monkey-patch the SDK in order to support additional pre-forking runtimes.

johnnyshields: "sidekiq-process_manager" is an unofficial hack gem which is not part of the official Sidekiq release. For these non-standard use cases, I suppose we could add an env var such as OTEL_FORKING_DISPATCHER to force a behavior one way or the other.

In such cases there is no hard issue with spawning the thread pre-fork, aside from wasting a bit of CPU and possibly seeing a annoying warning log.

johnnyshields: Is it fair to say the approach should be:
1. Add an env var (`OTEL_FORKING_DISPATCHER`) to explicitly control behavior
2. Fallback to detecting the dispatcher runtime as this PR does, to "do the right thing" automatically in 95% of cases

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: If you want the option of eagerly or lazily starting the thread, it seems like the simplest change is to pass a `start_export_thread` argument to `initialize`, then:
```suggestion
            reset_on_fork(restart_thread: start_export_thread)
```
We can bikeshed the name of the argument, but this seems a lot less brittle and simpler than the `DispatcherDetector`.

johnnyshields:
@fbogsany please check how I've set it up now. @dazuma @ericmustin your comments are also welcome.

I think we have a philosophical decision to make here. Do we want the OTEL SDK to do auto-detection to "do the right thing" in 95% of cases? Or do we want to take a "configuration-over-convention" approach?

Personally I think there should be an aspect of "user-service" in the SDK, which is the approach other widely-used telemetry vendors have taken in their SDK libs...

fbogsany:
My strong preference is to favour configuration. The span processor has to be configured in code at present, so asking the user to decide whether to start the `BatchSpanProcessor` export thread eagerly seems pretty reasonable.

Our approach to "auto-configuration" at Shopify is to provide a wrapper gem that pulls in "blessed" auto-instrumentation gems and sets up the span processor and exporter with sane defaults according to the environment (dev, prod, CI). Our apps then add just this gem to their `Gemfile`.

ericmustin:
@johnnyshields @fbogsany . First off let me just say that I think this is very thoughtful, well done work, and very much appreciated. And what this PR addresses is, I agree, useful. That being said, I do think this has a little more brittleness than i'd prefer, and I think the maintenance burden here feels like a long tail of edge cases. The `configuration-over-convention` approach seems to solve a lot of that brittleness (probably not the right word here because I think this is very well written code) while preserving the original intention of your PR, the tradeoff being it requires more thought from a user.

But, i'm ok with that? It's not unreasonable that user should be familiar with their webservers behavior. I agree with you that widely used telemetry vendors for Ruby optimize for "up and running in seconds" far more than OpenTelemetry-Ruby, but, that's sort've the point, vendors optimize for some things that i'm not sure opentelemetry-ruby-sdk should care as strongly about? And given that users are already having to modify application code to set the batch span processor, changing this config doesn't sound unreasonable to me.

Just my two cents, i am certainly not the cosmos here.

johnnyshields:
We should make it easy for people to adopt this gem. Should be aiming for 5,000 Github stars in 12 months--we're at 80 now.

I'd venture to say that Puma, Unicorn, and Passenger together have at least 80%+ share of Ruby webservers. The gem should know to "do the right thing" for 80% of the users.

I do not consider this code "brittle", because it only "enhances" behavior for the webservers which benefit from it, and otherwise gracefully degrades to spawn-thread-on-boot.

fbogsany:
> Should be aiming for 5,000 Github stars in 12 months--we're at 80 now.

I don't think this is an explicit goal for us.

johnnyshields:
Sure--I'd like to encourage the maintainers to think about what gets developers to actually adopt software libraries. One should be able to add the gem and with minimal effort start tracing--that's the selling point of NewRelic, Skylight, etc.

robertlaurin:
To over simplify the PR (maybe a bit too much), this appears to be whether or not we should embed configuration inside the SDK or not.  

So far the SDK tries not assume anything about the application that consumes it and instead it exposes the ability to configure it.  

Any library specific instrumentation is carved off into its own gem, and we leave that as something we support but is ultimately optional.

I would like to see us continue that approach going forward, and avoid embedding any config directly inside of the SDK.

johnnyshields:
> SDK tries not assume anything about the application that consumes it and instead it exposes the ability to configure it
While everything should be open to configuration, here I'm propose the SDK should do a modest amount of "auto-configuration" based on what it can easily detect to handle 80%+ use cases. This is what other telemetry libs do.

> Any library specific instrumentation is carved off into its own gem
This PR isn't about instrumentation, it's about running the exporter correctly given the runtime we're operating in.

> avoid embedding any config directly inside of the SDK.
So we're going "configuration over convention". Great...

mwear (commented):
I'd like chime in and echo some of the sentiments that have already been expressed here. First off, thank you so much for caring about this and trying to make this project better. We certainly appreciate your contributions and enthusiasm!

I can see this approach was heavily inspired by the work in the New Relic Ruby Agent: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/local_environment.rb. Given the history of that software, it is probably a pretty well vetted approach. While I do expect this does work for the majority of use cases, there are trade offs to be considered for auto detection of this sort. For one, it's great when it works, but can be very frustrating and confusing when it doesn't. There are enough edge cases where this won't work out of the box. Additionally, as others have pointed out, this code does carry with it a long term maintenance burden.

The configuration option is pretty straightforward to use, and to maintain, at the slight penalty of having to know it exists and use it. The simplicity of this approach is certainly appealing.

The intention of the OpenTelemetry SDK package is to provide all or most of the functionality that an application is likely to need, but it has few opinions about how you use it. It provides the building blocks, but expects that users and organizations will cherry pick their components and wrap the SDK to provide their opinionated configuration. I could see this dispatcher detection being an extension that you could use in your own wrapper, or provide to the community if there is wider interest. 

We all seem to be in agreement that at a bare minimum, configuration should be an option. I'd recommend that we start there and make incremental improvements as needed.

johnnyshields:
Closing in favor of https://github.com/open-telemetry/opentelemetry-ruby/pull/472

PR #465, opened 2020-11-02
fix: add error_handler to configurator

### LINKED ISSUE #464: SDK Exporter/BatchProcessor should not hard-fail in an initializer
If you put a bad URL in to initialize an `OpenTelemetry::Exporter::OTLP::Exporter`, it fails with:

```
Unable to load application: ArgumentError: invalid url for OTLP::Exporter google.com:55681
```

The recommended Rails initializer code doesn't have any `rescue` block. We were a bit surprised when our app hard failed to boot with an invalid url. We've added the rescue block to the initializer, but I think it would be nicer if maybe OpenTelemetry::SDK.configure could rescue it's block and log an error `"OpenTelemetry failed to configure: ArgumentError: invalid url ...`

```
OpenTelemetry::SDK.configure do |c| # rescue this block
    ...
    c.use_all
 end
```

fbogsany:
fixes #464 

This is a localized fix. I'd like to discuss whether we should have an `error_handler` just for `OpenTelemetry::SDK.configure` or whether we should try to address the "strict mode" thing again.

The spec [states](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md#configuring-error-handlers):
> SDK implementations MUST allow end users to change the library's default error handling behavior for relevant errors. Application developers may want to run with strict error handling in a staging environment to catch invalid uses of the API, or malformed config. Note that configuring a custom error handler in this way is the only exception to the basic error handling principles outlined above. The mechanism by which end users set or register a custom error handler should follow language-specific conventions.

It provides examples for Go and for Java. Both are global error handlers rather than purely config-related. Java uses the logger mechanism. I do not know how to make that work in Ruby, or what an idiomatic approach to this in Ruby might look like.

### DIFF
--- a/sdk/lib/opentelemetry/sdk.rb
+++ b/sdk/lib/opentelemetry/sdk.rb
@@ -56,6 +56,13 @@ module OpenTelemetry
       configurator = Configurator.new
       yield configurator if block_given?
       configurator.configure
+    rescue StandardError => e
+      if configurator.error_handler
+        configurator.error_handler.call(e)
+      else
+        OpenTelemetry.logger.error("unexpected error in span.on_finish - #{e}")
+        raise
+      end
     end
   end
 end
--- a/sdk/lib/opentelemetry/sdk/configurator.rb
+++ b/sdk/lib/opentelemetry/sdk/configurator.rb
@@ -15,10 +15,18 @@ module OpenTelemetry
 
       private_constant :USE_MODE_UNSPECIFIED, :USE_MODE_ONE, :USE_MODE_ALL
 
+      # An optional handler for configuration errors. If no error_handler is
+      # configured, configuration errors will escape the OpenTelemetry.sdk.configure
+      # method.
+      #
+      # Error handlers are callable and accept a single (exception object) argument.
+      attr_accessor :error_handler
+
       attr_writer :logger, :http_extractors, :http_injectors, :text_map_extractors,
                   :text_map_injectors
 
       def initialize
+        @error_handler = nil
         @instrumentation_names = []
         @instrumentation_config_map = {}
         @http_extractors = nil
--- a/sdk/test/opentelemetry/sdk/configurator_test.rb
+++ b/sdk/test/opentelemetry/sdk/configurator_test.rb
@@ -23,7 +23,7 @@ describe OpenTelemetry::SDK::Configurator do
 
   describe '#error_handler' do
     it 'raises configuration errors by default' do
-      _ { OpenTelemetry::SDK.configure { |c| raise 'hell' } }.must_raise(StandardError)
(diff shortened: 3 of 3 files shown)


mwear (commented):
I think we should consider making this error handler more widely available. I found the original issue: https://github.com/open-telemetry/opentelemetry-specification/issues/696 and the PR that accompanied it: https://github.com/open-telemetry/opentelemetry-specification/pull/757.

Reading through both, there was a strong desire to have a user provided global error handler that SDK components and potentially, the API could call into when they encounter an error. The default handler was expected to log a message at error level. There were debates in the [spec issue](https://github.com/open-telemetry/opentelemetry-specification/issues/696 ) about where the global error handler belongs (in the SDK or API), but the tl;dr appears to be that languages can decide.

In any case, I think we should have an error handler available at either `OpenTelemetry::SDK.error_handler` or `OpenTelemetry.error_handler`. It should default to logging a message at `error` level, but be user overridable to do something else. It makes sense to be able to provide the override in the configurator.

FWIW, there is prior art for this in Go, Python and JS. Go has their error handler as part of the API and Python and JS have it as part of the SDK.

fbogsany:
Sounds good. Thanks @mwear. I’ll implement this global handler. 

Do we want to provide a ConfigurationError exception type for config errors, at least, so users can choose to raise for those errors while ignoring/logging API errors?

mwear:
> Sounds good. Thanks @mwear. I’ll implement this global handler.
> 
> Do we want to provide a ConfigurationError exception type for config errors, at least, so users can choose to raise for those errors while ignoring/logging API errors?

👍 That would be useful.

johnnyshields:
IMHO there is a difference between handling errors for spans vs. boot-time configuration. I read the OTEL spec you posted above to be more relevant to spans. Boot-time configuration--at least in production--it is generally undesirable to hard-fail supposing we give a bad OTEL endpoint url for example. This has potential to bring down a prod system with a simple misconfig--I'd much rather lose OTEL tracing but still boot otherwise!

fbogsany:
The spec is pretty clear:
> The API or SDK may fail fast and cause the application to fail on initialization, e.g. **_because of a bad user config_** or environment, but MUST NOT cause the application to fail later at run time, e.g. due to dynamic config settings received from the Collector.

This is `may` rather than `must`, and the configurable error handler is intended to allow overriding that behaviour.

johnnyshields:
IMHO an error handler for configuration is overkill. I've solved the issue on my end by wrapping the initializer code in a `begin / rescue` block.

Using the same error handler for both initialization and runtime errors feels wrong.

A simpler approach here might be:

```
OpenTelemetry::SDK.configure(rescue_errors: true) do |c|
   ...
end
```

fbogsany:
Replaced with #521

Issue #109, opened 2019-09-26
API? metadata helpers for semantic conventions


fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md

alolita:
This link is dead -> https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md Can you repost the right link? Thanks.

xukaren:
Hi, could you please specify the issue a bit further? I'd love to help, but I am new to this :) Thanks!

fbogsany:
Honestly, I wouldn't pick this issue just yet. The observation here is that a lot of instrumentation gems will record the same set of attributes (with different values). For example, MySQL, Redis and Memcached are all datastores and will record this set of attributes: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/database.md. It is easy to get this wrong, e.g. by making a typo in an attribute name or forgetting an attribute, so the idea is that a common set of helper methods might make it easier to get this right.

I don't think this is a high priority, TBH, because we haven't implemented enough instrumentation to know what a helpful abstraction would look like.

fbogsany:
https://pkg.go.dev/go.opentelemetry.io/otel@v0.12.0/semconv is roughly what I had in mind.

ahayworth:
Some time ago (last year?) we implemented a basic `semconv` package to fulfill these goals.

PR #476, opened 2020-11-18
feat: ruby-kafka instrumentation

robertlaurin:
Todo:
- [x] Getting kafka working in ci
- [x] Add readme
- [x] Add examples
- [x] Configure appraisals and minimum version supported
- [x] Test with consumer


instrumentation/ruby_kafka/.yardopts
```diff
@@ -0,0 +1,9 @@
+--no-private
+--title=OpenTelemetry Rack Instrumentation
```
fbogsany, line 2:
```suggestion
--title=OpenTelemetry RubyKafka Instrumentation
```

instrumentation/ruby_kafka/opentelemetry-instrumentation-ruby_kafka.gemspec
```diff
@@ -0,0 +1,47 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'activesupport'
```
robertlaurin, line 28: Does it make sense to make `activesupport` a dependency here or let the consumer of the gem decide?

instrumentation/ruby_kafka/Appraisals
```diff
@@ -0,0 +1,21 @@
...
+
+appraise 'ruby-kafka-1.2.0' do
+  gem 'ruby-kafka', '~> 1.2.0'
+end
+
+appraise 'ruby-kafka-1.1.0' do
+  gem 'ruby-kafka', '~> 1.1.0'
+end
+
+appraise 'ruby-kafka-1.0.0' do
```
robertlaurin, line 19: I don't have any strong opinions on what version we should be supporting.  Mostly just picked the first major release and worked my way up.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/connection/request.rb
```diff
@@ -0,0 +1,40 @@
...
+        module Connection
+          # The Request module contains the instrumentation for connection request event
+          module Request
+            extend self
+
+            EVENT_NAME = 'request.connection.kafka'
+            SPAN_NAME = 'kafka.connection.request'
+
+            def subscribe!
+              ::ActiveSupport::Notifications.subscribe(EVENT_NAME) do |_name, start, finish, _id, payload|
```
robertlaurin, line 20: There's a lot of duplicate code ahead for all of the notifications, this was intentional as I'm not overly familiar with kafka and wanted to keep things as simple as possible.  Open to feedback on we might want to clean this up a bit.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/consumer/process_message.rb
```diff
@@ -0,0 +1,45 @@
...
+        module Consumer
+          # The ProcessMessage module contains the instrumentation for consumer process message event
+          module ProcessMessage
+            extend self
+
+            EVENT_NAME = 'process_message.consumer.kafka'
+            SPAN_NAME = 'process'
+
+            def subscribe!
+              ::ActiveSupport::Notifications.subscribe(EVENT_NAME) do |_name, start, finish, _id, payload|
```
robertlaurin, line 20: This is a really tricky loop to instrument, with the way this is setup we'll generate a span for each processed message but the trace won't propagate further as we're creating the span after the fact so to speak.

Would love to improve this, just unclear what is the way forward here.

The code in question.
https://github.com/zendesk/ruby-kafka/blob/4e63e1ddd2da37f19703ba7a01b13dbc7c8f9a3f/lib/kafka/consumer.rb#L243-L254

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/consumer/process_batch.rb
```diff
@@ -0,0 +1,44 @@
...
+        module Consumer
+          # The ProcessBatch module contains the instrumentation for consumer process batch event
+          module ProcessBatch
+            extend self
+
+            EVENT_NAME = 'process_batch.consumer.kafka'
+            SPAN_NAME = 'kafka.consumer.process_batch'
+
+            def subscribe!
+              ::ActiveSupport::Notifications.subscribe(EVENT_NAME) do |_name, start, finish, _id, payload|
```
robertlaurin, line 20: This is the same problem as the `consumer.each_message` just at a coarser level.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module RubyKafka
+      module Patches
+        # The Client module contains the instrumentation patch the Producer#deliver_message method
+        module Client
+          def deliver_message(value, key: nil, headers: {}, topic:, partition: nil, partition_key: nil, retries: 1)
```
robertlaurin, line 13: This method is not actually recommended in production systems and I'm not sure its worth instrumenting, I did it to start just for the example and my own initial testing.

robertlaurin (commented):
Moving this out of draft to get some reviews in, I don't feel that this is 100% yet but I think having a review will speed things up a bit.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/consumer/process_batch.rb
robertlaurin: No longer relevant

instrumentation/ruby_kafka/Appraisals
fbogsany: 🤷 which versions are supported? Generally, we prefer to only support ... supported gem versions.

instrumentation/ruby_kafka/README.md
```diff
@@ -0,0 +1,53 @@
+# OpenTelemetry RubyKafka Instrumentation
+
+The RubyKafka instrumentation is a community-maintained instrumentation for the [RubyKafka][ruby_kafka-home] web server interface.
```
fbogsany, line 3: > web server interface

Umm...

instrumentation/ruby_kafka/opentelemetry-instrumentation-ruby_kafka.gemspec
fbogsany: Good question. I think having it here breaks assumptions that `opentelemetry-instrumentation-all` only pulls in OTel instrumentation. I'd prefer to check for it dynamically and, only if RubyKafka is present, warn that we can't instrument without it.

instrumentation/ruby_kafka/README.md
robertlaurin: Copy paste 🙈

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,51 @@
...
+
+# Copyright 2020 OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module RubyKafka
+      module Patches
+        # The Client module contains the instrumentation patch the Producer#deliver_message method
```
fbogsany, line 11:
```suggestion
        # The Client module contains the instrumentation patch the Client#deliver_message and Client#each_message methods.
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,51 @@
...
+        module Client
+          def deliver_message(value, key: nil, headers: {}, topic:, partition: nil, partition_key: nil, retries: 1)
+            attributes = {
+              'messaging.system' => 'kafka',
+              'messaging.destination' => topic
```
fbogsany, line 16:
```suggestion
              'messaging.destination' => topic,
              'messaging.destination_kind' => 'topic'
```

It is also strongly recommended to add networking attributes.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,51 @@
...
+              super
+            end
+          end
+
+          def each_message(topic:, start_from_beginning: true, max_wait_time: 5, min_bytes: 1, max_bytes: 1_048_576, &block)
+            super do |message|
+              attributes = { 'messaging.system' => 'kafka' }
+              attributes['messaging.destination'] = message.topic
+              attributes['messaging.kafka.message_key'] = message.key if message.key
+              attributes['messaging.kafka.partition'] = message.partition
```
fbogsany, line 33:
```suggestion
              attributes = {
                'messaging.system' => 'kafka',
                'messaging.destination' => message.topic,
                'messaging.destination_kind' => 'topic',
                'messaging.kafka.partition' => message.partition
              }
              attributes['messaging.kafka.message_key'] = message.key if message.key
```

+ network attributes, if we have them here.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,54 @@
...
+
+          def each_batch(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true)
+            super do |batch|
+              attributes = { 'messaging.system' => 'kafka' }
+              attributes['messaging.destination'] = batch.topic
+              attributes['messaging.kafka.partition'] = batch.partition
+
+              attributes['offset_lag'] = batch.offset_lag
+              attributes['highwater_mark_offset'] = batch.highwater_mark_offset
+              attributes['message_count'] = batch.messages.count
```
fbogsany, line 37: Do all these (and `offset`) come from Datadog's instrumentation?

robertlaurin: Yes, any attribute not prefixed with `messaging` was pulled from datadogs instrumentation.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
robertlaurin: >It is also strongly recommended to add networking attributes.
Regarding networking attributes, I have access to the brokers from here.
```rb
[16] pry(#<Kafka::Client>)> brokers
=> [#<Kafka::BrokerInfo:0x00007faf231fa408 @host="kafka", @node_id=1, @port=9092>]
```

instrumentation/ruby_kafka/opentelemetry-instrumentation-ruby_kafka.gemspec
robertlaurin: Removed as a dependency here https://github.com/open-telemetry/opentelemetry-ruby/pull/476/commits/fed881efed4410fff6a4f0dae4bcf44f981314a2

I think it makes sense to still apply patches regardless of active support being present, as we don't need it for that level of instrumentation.

.toys/.data/releases.yml
```diff
@@ -114,6 +114,10 @@ gems:
     directory: instrumentation/restclient
     version_constant: [OpenTelemetry, Instrumentation, RestClient, VERSION]
 
+  - name: opentelemetry-instrumentation-ruby_kafka
```
johnnyshields, line 117: Why not just name this `opentelemetry-instrumentation-kafka` (no `ruby_`)

Otherwise we should do ruby_elasticsearch, ruby_on_rails, mongo_ruby_driver, etc.

robertlaurin: This instrumentation is for https://rubygems.org/gems/ruby-kafka/versions/1.3.0

Not to be confused with https://rubygems.org/gems/kafka/versions/0.5.2

I don't like this naming either, but I'm attempting to prevent future ambiguity.

johnnyshields: d'oh fair point! thanks

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
robertlaurin: Should we drop these?

fbogsany: I don't know whether they're useful. We could hide them behind a config option. If we keep them (whether optional or not), we should prefix them with `messaging.kafka.`.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
fbogsany: Maybe ask @d1egoaz ?

d1egoaz: it's hard to tell here what specific broker will receive the message, as messages for topics go to different partitions  depending if there is a key or no (random), also, it depends on the client acks definition, as per configured in kafka, `acks=all` means that a message is sent to the broker leader for the partition, and then it waits until it's in sync with two followers, the broker leader can also change.

That's why we decided to use this parameter, and filling it with the current target kafka cluster.

https://github.com/Shopify/shopify-tracing/blob/master/lib/shopify/tracing/adapters/kafka_client_ruby_adapter.rb#L13`kk

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/producer/deliver_messages.rb
```diff
@@ -0,0 +1,41 @@
...
+          # The DeliverMessages module contains the instrumentation for producer deliver messages event
+          module DeliverMessages
+            extend self
+
+            EVENT_NAME = 'deliver_messages.producer.kafka'
+            SPAN_NAME = 'kafka.producer.deliver_messages'
+
+            def subscribe!
+              ::ActiveSupport::Notifications.subscribe('deliver_messages.producer.kafka') do |_name, start, finish, _id, payload|
+                attributes = { 'messaging.system' => 'kafka' }
```
ericmustin, line 21: small nit but, as i think you're aware, there's a bunch of places in this instrumentation where we are frequently allocating strings that would be easier to reason about as constants.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/connection/request.rb
ericmustin: I generally am fine with a lot of this repetitive code, I think eventually the goal here should be to abstract things into a general active support base instrumentation module, which is the approach a few vendors have taken. at this point though i don't think it's really worth that much to invest in determining that pattern, so the repetition is fine with me

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
ericmustin: fine to remove it if it isn't going to provide any value to users

ericmustin (commented):
I'm happy to approve this,  i think this looks ready to ship to me, setting aside a few unresolved questions around span attributes. nice work here

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/producer/deliver_messages.rb
johnnyshields: strings are frozen so we aren't allocating them each time.

IMHO constants makes things much harder to deal with for actual coding, you have to constantly refer to your constants file (pun intended)

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
johnnyshields: or just prefix them with `kafka.` to indicate they are not part of the official schema?

instrumentation/ruby_kafka/Appraisals
robertlaurin: I didn't see a formal statement about their supported versions, and I've added support to cover all the minor revisions of the latest major version (1).  I think this is sufficient for now.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
robertlaurin: I'm going to leave it if only to serve as a simple way of demonstrating/experimenting with the instrumentation.  I also don't want to make any assumptions on how frequently the community may be actually using these features, despite the warning.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
robertlaurin: I've added the `messaging.kafka` prefix so that things remain consistent.  These attributes were lifted from ddtrace, and while I know it's not great to make assumptions, I'm going to err to side of believing these were added because they provide value and we too should include them.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/producer/deliver_messages.rb
robertlaurin: Just for further clarification you're referring to the attribute keys, correct?  For general developer ergonomics it's nice just be able to plunk in a string, but I guess "magic strings" can become a problem as the codebase grows.

FWIW, I'm fine leaving it as is or referencing a constants file if anyone has some really strong opinions on this.  

In lieu of a stronger opinion I will just leave it.

fbogsany: There's some desire to add helpers for semantic conventions (see #109), but it isn't clear what the API for those helpers should look like. `OpenTelemetry::SDK::Resources::Constants` is one approach, and it is decidedly unpleasant to use. For now, I think the string literals are fine, but I really do hope that at some point someone will be inspired to define a reasonably pleasant set of semantic convention helpers and refactor all the auto instrumentation in terms of those helpers.

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/events/connection/request_test.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# Copyright 2020 OpenTelemetry Authors
```
fbogsany, line 3:
```suggestion
# Copyright The OpenTelemetry Authors
```
Can we fix copyright notices throughout this PR, or do you plan to do that with #512?

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+      module Patches
+        # The Consumer module contains the instrumentation patch for the Consumer class
+        module Consumer
+          def each_message(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true)
+            super do |message|
+              attributes = {
+                'messaging.system' => 'kafka',
+                'messaging.destination' => message.topic,
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => message.partition
```
fbogsany, line 19:
```suggestion
                'messaging.kafka.partition' => message.partition,
                'messaging.kafka.offset' => message.offset
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+              }
+
+              attributes['messaging.kafka.message_key'] = message.key if message.key
+              attributes['offset'] = message.offset
```
fbogsany, line 23:
```suggestion
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+            end
+          end
+
+          def each_batch(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true)
+            super do |batch|
+              attributes = {
+                'messaging.system' => 'kafka',
+                'messaging.destination' => batch.topic,
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => batch.partition
```
fbogsany, line 38:
```suggestion
                'messaging.kafka.partition' => batch.partition,
                'messaging.kafka.offset_lag' => batch.offset_lag,
                'messaging.kafka.highwater_mark_offset' => batch.highwater_mark_offset,
                'messaging.kafka.message_count' => batch.messages.count
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+              }
+
+              attributes['offset_lag'] = batch.offset_lag
+              attributes['highwater_mark_offset'] = batch.highwater_mark_offset
+              attributes['message_count'] = batch.messages.count
```
fbogsany, line 43:
```suggestion
```

fbogsany:
Not blockers:
1. I find all the little event subscriber modules unnecessarily wordy. It seems like we could perform all the subscriptions in `OpenTelemetry::Instrumentation::RubyKafka::Instrumentation.subscribe` without all the additional modules. 🤷 
2. Should all the events really be spans? Might they be better as ... events?

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,56 @@
...
+            attributes = {
+              'messaging.system' => 'kafka',
+              'messaging.destination' => topic,
+              'messaging.destination_kind' => 'topic'
+            }
+
+            attributes['messaging.kafka.message_key'] = key if key
+            attributes['messaging.kafka.partition'] = partition if partition
+
+            tracer.in_span('send', attributes: attributes, kind: :producer) do
```
mwear, line 23: Looking at the spec, it appears the span name should be of the form `<destination name> <operation name>`.  In the case of this span, it should be `"#{topic} send"`.

See: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#span-name

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/client.rb
```diff
@@ -0,0 +1,56 @@
...
+                'messaging.system' => 'kafka',
+                'messaging.destination' => message.topic,
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => message.partition
+              }
+
+              attributes['messaging.kafka.message_key'] = message.key if message.key
+
+              parent_context = OpenTelemetry.propagation.text.extract(message.headers)
+              tracer.in_span('process', with_parent: parent_context, attributes: attributes, kind: :consumer) do
```
mwear, line 41: Similarly, this span should be named `"#{topic} process"`.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+                'messaging.destination' => message.topic,
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => message.partition
+              }
+
+              attributes['messaging.kafka.message_key'] = message.key if message.key
+              attributes['offset'] = message.offset
+
+              parent_context = OpenTelemetry.propagation.text.extract(message.headers)
+              tracer.in_span('process', with_parent: parent_context, attributes: attributes, kind: :consumer) do
```
mwear, line 26:
```suggestion
              tracer.in_span("#{message.topic} process", with_parent: parent_context, attributes: attributes, kind: :consumer) do
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+                'messaging.destination' => batch.topic,
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => batch.partition
+              }
+
+              attributes['offset_lag'] = batch.offset_lag
+              attributes['highwater_mark_offset'] = batch.highwater_mark_offset
+              attributes['message_count'] = batch.messages.count
+
+              tracer.in_span('process', attributes: attributes, kind: :consumer) do
```
mwear, line 45:
```suggestion
              tracer.in_span("#{batch.topic} process", attributes: attributes, kind: :consumer) do
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/producer.rb
```diff
@@ -0,0 +1,35 @@
...
+        # The Producer module contains the instrumentation patch the Producer#produce method
+        module Producer
+          def produce(value, key: nil, headers: {}, topic:, partition: nil, partition_key: nil, create_time: Time.now)
+            attributes = {
+              'messaging.system' => 'kafka',
+              'messaging.destination' => topic,
+              'messaging.destination_kind' => 'topic'
+            }
+
+            tracer.in_span('send', attributes: attributes, kind: :producer) do
```
mwear, line 20:
```suggestion
            tracer.in_span("#{topic} send", attributes: attributes, kind: :producer) do
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,60 @@
...
+              attributes['offset'] = message.offset
+
+              parent_context = OpenTelemetry.propagation.text.extract(message.headers)
+              tracer.in_span('process', with_parent: parent_context, attributes: attributes, kind: :consumer) do
+                yield message
+              end
+            end
+          end
+
+          def each_batch(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true)
```
mwear, line 32: According the semantic conventions, batch processing should add links: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-processing

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/events/connection/request_test.rb
robertlaurin: I'll fix up any leftover copy pasted cruft I can find here.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/events/producer/deliver_messages.rb
johnnyshields: +1 for using inline strings and not constants.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
fbogsany: So this should iterate over `batch.messages` and add a link for each span context extracted? How does this method interact with `each_message` above, if at all?

fbogsany: It seems like we should instrument [`fetch_batches`](https://github.com/zendesk/ruby-kafka/blob/0a07531e3a4214a0f8b8b023fa1eff479c535d3f/lib/kafka/consumer.rb#L542-L578) to create a single receive span. I'm unsure how we're supposed to represent `each_batch`, though. Effectively, `KafkaRuby` has a batch receive (with `fetch_batches`) and then either a batch process (`each_batch`) or a single process of each message in a batch (`each_message`). The second case is equivalent to the Batch Receive example in the spec, but the first case is different from the Batch Process example.

If we instrument `fetch_batches`, we would need to propagate the span somehow to the instrumentation of `each_message` and `each_batch`.

robertlaurin:
Updated to use links in this commit https://github.com/open-telemetry/opentelemetry-ruby/pull/476/commits/8a1ef1b15af53b1e636e4b35f01a84706f23f237

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
```diff
@@ -0,0 +1,135 @@
...
+        consumer.each_message do |_msg|
+          @counter += 1
+          raise 'oops' if @counter >= 2
+        end
+      rescue StandardError # rubocop:disable Lint/HandleExceptions
+      end
+
+      process_spans = spans.select { |s| s.name == "#{topic} process" }
+
+      # First pair for senda and process spans
```
fbogsany, line 61:
```suggestion
      # First pair for send and process spans
```

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
```diff
@@ -0,0 +1,135 @@
...
+    kafka.close
+  end
+
+  describe '#each_message' do
+    it 'traces each_meassage call' do
+      kafka.deliver_message('hello', topic: topic)
+      kafka.deliver_message('hello2', topic: topic)
+
+      begin
+        @counter = 0
```
fbogsany, line 51: What is this counter for? It seems like a hold over from an earlier iteration of the code.

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
```diff
@@ -0,0 +1,135 @@
...
+        consumer.each_message do |_msg|
+          @counter += 1
+          raise 'oops' if @counter >= 2
```
fbogsany, line 54: Oh, I guess it is just to terminate the `each_message` loop. Does it need to be a member variable? Wouldn't a local be sufficient?

robertlaurin: A local is sufficient, I think there was a reason why it was this way before but it's no longer valid.  Updated!

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,68 @@
...
+                'messaging.kafka.offset_lag' => batch.offset_lag,
+                'messaging.kafka.highwater_mark_offset' => batch.highwater_mark_offset,
+                'messaging.kafka.message_count' => batch.messages.count
+              }
+
+              links = batch.messages.map do |message|
+                OpenTelemetry::Trace::Link.new(
+                  OpenTelemetry.propagation.text.extract(message.headers)
+                )
+              end
```
fbogsany, line 51: I'm not sure how the tests are working 🤔 . The `Link` initializer takes a `SpanContext`, but the propagator's `extract` returns a `Context` that *might* contain a `Span` that wraps a `SpanContext`. We need to dig through an unfortunate number of layers to get to a `SpanContext` that might not be there.
```suggestion
              links = batch.messages.map do |message|
                span_context = OpenTelemetry::Trace.current_span(OpenTelemetry.propagation.text.extract(message.headers)).context
                OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
              end
              links.compact!
```

robertlaurin: My tests might be doing the wrong thing, for more isolated view on what I did with links this commit doesn't have the extra noise. https://github.com/open-telemetry/opentelemetry-ruby/pull/476/commits/8a1ef1b15af53b1e636e4b35f01a84706f23f237

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,68 @@
...
+                'messaging.destination_kind' => 'topic',
+                'messaging.kafka.partition' => message.partition,
+                'messaging.kafka.offset' => message.offset
+              }
+
+              attributes['messaging.kafka.message_key'] = message.key if message.key
+
+              link = OpenTelemetry::Trace::Link.new(
+                OpenTelemetry.propagation.text.extract(message.headers)
+              )
```
fbogsany, line 27:
```suggestion
              span_context = OpenTelemetry::Trace.current_span(OpenTelemetry.propagation.text.extract(message.headers)).context
              link = OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
              links = [link] if link
```

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,68 @@
...
+
+              tracer.in_span("#{message.topic} process", links: [link], attributes: attributes, kind: :consumer) do
```
fbogsany, line 29:
```suggestion
              tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer) do
```

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
```diff
@@ -0,0 +1,133 @@
...
+
+      # First pair for send and process spans
+      first_process_span = process_spans[0]
+      _(first_process_span.name).must_equal("#{topic} process")
+      _(first_process_span.kind).must_equal(:consumer)
+      _(first_process_span.attributes['messaging.destination']).must_equal(topic)
+      _(first_process_span.attributes['messaging.kafka.partition']).must_equal(0)
+
+      first_process_span_link = first_process_span.links[0]
+      linked_span_context = OpenTelemetry::Trace.current_span(first_process_span_link.span_context).context
```
fbogsany, line 69: Oh, I see - this is treating the `link.span_context` as a `Context` instead of a `SpanContext`. This (and other examples below) should look like:
```suggestion
      linked_span_context = first_process_span_link.span_context
```

fbogsany: Kinda feels like we should be more strict about type-checking the `span_context` argument in the `Link` initializer.

robertlaurin: Yeah I agree, it's a bit too easy to do the wrong thing if you're not entirely familiar with that part of the code.

fbogsany (approved):
LGTM, other than appeasing Rubocop.

Thanks @robertlaurin! I know it was challenging to get this over the finish line.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -0,0 +1,68 @@
...
+              link = OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
+              links = [link] if link
+
+              tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer) do
+                yield message
+              end
+            end
+          end
+
+          def each_batch(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true) # # rubocop:disable Metrics/AbcSize
```
fbogsany, line 35:
```suggestion
          def each_batch(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automatically_mark_as_processed: true) # rubocop:disable Metrics/AbcSize
```

PR #477, opened 2020-11-20
feat: Active Model Serializers instrumentation

johnnyshields:
Active Model Serializers is the standard Rails JSON serialization library used with Rails API mode.

Ported from Datadog. I've dropped support for AMS 0.9.0 and am only supporting AMS 0.10.0+

There doesn't seem to be any semantic standards for serializers (which I guess is a subset of view/template rendering, which also doesn't exist)


instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/event_handler.rb
```diff
@@ -0,0 +1,50 @@
...
+
+        def span_name(payload)
+          "#{demodulize(payload[:serializer].name)} render"
+        end
+
+        def build_attributes(payload)
+          {
+            'serializer.name' => payload[:serializer].name,
+            'serializer.renderer' => 'active_model_serializers',
+            'serializer.format' => payload[:adapter]&.class&.name
```
fbogsany, line 34: We should add a default here.

fbogsany (approved):

fbogsany:
Thanks @johnnyshields!

Issue #469, opened 2020-11-03
Allow samplers to modify tracestate


fbogsany:
Required by the spec compliance matrix.

PR #487, opened 2020-11-30
fix!: allow samplers to modify tracestate

fbogsany:
Fixes: #469 

A few notes:
- We now have to allocate an object for the `Result` on every call to `should_sample?`. This is unavoidable because the result has to wrap the tracestate from the parent context.
- I'll implement `Tracestate` in the API in a separate PR.
- This is a breaking change because it adds an extra required argument to `Result#initialize`.


robertlaurin (approved):
LGTM

PR #490, opened 2020-12-02
Fix Jaeger exporter with nil values

drogus:
When there's a `nil` value in the span data attributes Jaeger exporter will fail with a following error: `unexpected error in Jaeger::CollectorExporter#export - Unknown key given to OpenTelemetry::Exporter::Jaeger::Thrift::Tag.new:`. It's unexpected and hard to track down as the error does not even give the key name. It's also quite common in Ruby apps to not check for `nil` values when setting them as an attribute - a common practice is to either conver them to empty strings or ignore them.

This PR ignores `nil` attributes, but an alternative (empty strings) could also be implemented.

### DIFF
--- a/exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb
+++ b/exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb
@@ -43,9 +43,13 @@ module OpenTelemetry
         end
 
         def encoded_tags(attributes)
-          attributes&.map do |key, value|
-            encoded_tag(key, value)
-          end || EMPTY_ARRAY
+          if attributes
+            attributes.find_all { |_key, value| value }.map do |key, value|
+              encoded_tag(key, value)
+            end
+          else
+            EMPTY_ARRAY
+          end
         end
 
         def encoded_tag(key, value)
--- a/exporter/jaeger/test/opentelemetry/exporters/jaeger/encoder_test.rb
+++ b/exporter/jaeger/test/opentelemetry/exporters/jaeger/encoder_test.rb
@@ -85,6 +85,14 @@ describe OpenTelemetry::Exporter::Jaeger::Encoder do
     )
   end
 
+  it 'ignores nil values' do
+    attributes = { 'a_nil' => nil }
+    span_data = create_span_data(attributes: attributes)
+    encoded_span = Encoder.encoded_span(span_data)
+    _(encoded_span.tags.length).must_equal(0)
+  end
+
+
   describe 'instrumentation library' do
     it 'encodes library and version when set' do
       lib = OpenTelemetry::SDK::InstrumentationLibrary.new('mylib', '0.1.0')


ericmustin (commented):
@drogus thanks for the PR here.

I think this generally lgtm, `nil` values generally shouldnt be allowed in span attributes was my understanding though, https://github.com/open-telemetry/opentelemetry-specification/blob/82865fae64e7b30ee59906d7c4d25a48fe446563/specification/common/common.md#attributes, so i'm curious how we've got to this state. 

That being said I think it's good to be defensive here. 

I think once you sign the CLA i can approve

fbogsany:
It happens because we're not checking parameters here: https://github.com/open-telemetry/opentelemetry-ruby/blob/659be35520552779e605915094a7b5b0ad376ceb/sdk/lib/opentelemetry/sdk/trace/span.rb#L69-L82 or here: https://github.com/open-telemetry/opentelemetry-ruby/blob/659be35520552779e605915094a7b5b0ad376ceb/sdk/lib/opentelemetry/sdk/trace/span.rb#L272-L280. #207 is a slightly stagnant PR attempting to address this.

drogus:
`I think this generally lgtm, nil values generally shouldnt be allowed in span attributes was my understanding though,`

If that's the case I think a better fix would be to make sure that they're disallowed in the spans, cause my fix is just for the Jaeger exporter. If another exporter also makes the assumption that `nil` are not allowed it might happen again.

#207 looks interesting, but I see it's quite old, so would it be fine if I created a smaller PR that returns early from `set_attribute` if a passed value is `nil`? Not sure if it should do anything else, though. Like - should we raise, warn, log it as a debug message? I'm fine with either I think. Ignoring a `nil` value would be probably the easiest from the developer's perspective, but could also make it easier for the developer to introduce a bug in their code (like sending a wrong thing, which is `nil`). So I think my preference would be to ignore and log to `DEBUG`, but again, I'm fine with any solution that will disallow `nil`s.

fbogsany:
> So I think my preference would be to ignore and log to `DEBUG`

👍 I think that's good for now. I will be working to improve our error handling story in #465 and we can decide then whether we want to allow devs to optionally `raise` (e.g. to catch API usage errors in dev).

drogus:
Sorry for no activity here. I'll work on a fix, but there's also another problem that I've found - it will also fail when the value can't be mapped to one of the thrift values (like any Ruby class other than `String` and other simple "types"). For now I fixed it on the application level, but I'll also think on a fix for the library, because it's also hard to debug.

fbogsany:
> Sorry for no activity here. I'll work on a fix, but there's also another problem that I've found - it will also fail when the value can't be mapped to one of the thrift values (like any Ruby class other than `String` and other simple "types"). For now I fixed it on the application level, but I'll also think on a fix for the library, because it's also hard to debug.

It is supposed to only allow:
1. Array of (only one of) String, Float, Integer, boolean
2. String
3. Float
4. Integer
5. boolean

We should log and ignore any other values.

genebean:
I just ran into this while instrumenting some code and was curious if there has been any progress since the last update.

PR #498, opened 2020-12-08
fix: active_model_serializers installer

robertlaurin:
Small fix for the `active_model_serializers` instrumentation.

### DIFF
--- a/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
+++ b/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
@@ -18,14 +18,14 @@ module OpenTelemetry
 
         present do
           !defined?(::ActiveSupport::Notifications).nil? &&
-            !defined?(::ActiveModel::Serializers).nil? &&
+            !defined?(::ActiveModelSerializers).nil? &&
             gem_version >= MINIMUM_VERSION
         end
 
         private
 
         def gem_version
-          Gem.loaded_specs['active_model_serializers']&.version
+          Gem.loaded_specs['active_model_serializers'].version
         end
 
         def require_dependencies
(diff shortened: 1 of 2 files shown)


instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -19,13 +19,14 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         present do
           !defined?(::ActiveSupport::Notifications).nil? &&
             !defined?(::ActiveModel::Serializers).nil? &&
```
robertlaurin, line 21: ActiveModel actually defines this namespace without the presence of the `active_model_serializers` gem 

https://github.com/rails/rails/blob/master/activemodel/lib/active_model.rb#L62

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -19,13 +19,14 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
...
+            !defined?(::ActiveModelSerializers).nil? &&
```
robertlaurin, line 22: This is the error that comes up on a rails app without `active_model_serializers`
```
WARN -- : Instrumentation: OpenTelemetry::Instrumentation::ActiveModelSerializers unhandled exception during install undefined method `>=' for nil:NilClass
```

robertlaurin: Hey @johnnyshields does this check look right?  I haven't spent any time with this gem before.

johnnyshields: @robertlaurin yep this change is correct

johnnyshields: you can actually remove `!defined?(::ActiveModel::Serializers).nil?` if you want, it's redundant.

robertlaurin: Will do, thanks for looking it over.

fbogsany (approved):

PR #504, opened 2020-12-09
fix: rails tests

robertlaurin:
Fixes appraisals to use the versions originally intended to be tested, also adds rails `6.1` to the appraisals.

### DIFF
--- a/instrumentation/rails/Appraisals
+++ b/instrumentation/rails/Appraisals
@@ -5,9 +5,9 @@
 # SPDX-License-Identifier: Apache-2.0
 
 appraise 'rails-5.2' do
-  gem 'rails', '~> 5.2'
+  gem 'rails', '~> 5.2.0'
 end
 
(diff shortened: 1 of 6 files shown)


instrumentation/rails/Appraisals
```diff
@@ -5,9 +5,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 appraise 'rails-5.2' do
-  gem 'rails', '~> 5.2'
+  gem 'rails', '~> 5.2.0'
 end
 
 appraise 'rails-6.0' do
-  gem 'rails', '~> 6.0'
+  gem 'rails', '~> 6.0.0'
```
robertlaurin, line 12: It was using `6.1` instead of `6.0` which has different config support.  Which is why the tests are exploding on the repo right now.

fbogsany (approved):

PR #508, opened 2020-12-10
feat: instrument http client gem

robertlaurin:
Instrumentation for https://github.com/nahi/httpclient

The tests are almost exact copies of what we have for `net::http`.


fbogsany:
https://github.com/nahi/httpclient/blob/4658227a46f7caa633ef8036f073bbd1f0a955a2/lib/httpclient.rb#L1019-L1021

🤣

fbogsany:
A few comments:
1. I wonder if we should instrument a level deeper, by monkey-patching `do_get_block` and `do_get_stream`? WebMock, at least, appears to hook in at that level.
2. (How) should we instrument asynchronous requests?
3. I'm not sure we should generate a span for `connect` (this applies to `Net::HTTP` as well). I look to [JBD's examples](https://medium.com/google-cloud/tracing-google-cloud-8f0c8ba8181c) of Go networking event instrumentation, where a request has a span, but `ConnectStart` and `ConnectDone` are events attached to the request span. There's a few arguments for this: a `client` request will often have a corresponding `server` span, but a `connect` usually will not; `connect` is an additional level of detail that may help us understand why one request was slower than another, so it helps to have it directly attached to the request span; on the other hand, we may not want that level of detail in general, so we might want to strip it out before exporting a span, and that is easier to do with a processor that removes events.

Of course, I may be reading too much into JBDs excellent screenshots.

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/session.rb
```diff
@@ -0,0 +1,32 @@
...
+        # Module to prepend to HTTPClient::Session for instrumentation
+        module Session
+          def connect
+            site = @proxy || @dest
+            url = site.addr
+
+            attributes = OpenTelemetry::Common::HTTP::ClientContext.attributes.merge('http.url' => url)
+            tracer.in_span('HTTP CONNECT', attributes: attributes, kind: :client) do
+              super
+            end
```
fbogsany, line 20: There's some weirdness with this instrumentation:
1. We're not injecting context. On one hand, this is probably OK because the server side probably isn't tracing connection establishment, on the other hand, why are we calling this a `client` span?
2. *If* the server *is* instrumenting connection establishment, then the `server` span(s) will be associated with the wrong `client` span, because a request span is the parent of this span, and it *does* inject context.

robertlaurin:
> A few comments:
> 
> 1. I wonder if we should instrument a level deeper, by monkey-patching `do_get_block` and `do_get_stream`? WebMock, at least, appears to hook in at that level.

The motivation isn't obvious to me, what would be the value of this change?  I see that `do_get_block` is a private method only used in `do_request`.  

> 2. (How) should we instrument asynchronous requests?

For the `how` that's a good question, do we have any prior art of instrumenting async operations like this?  This might be the reason to trace a level down that I wasn't seeing.  If we patched `do_request_async` just to inject the trace headers before it created a new thread, we could then extract them in `do_get_stream`, I think that might work?   This is something I would prefer to break this off into smaller followup PR, I don't see a reason why we shouldn't do this.

> 3. I'm not sure we should generate a span for `connect` (this applies to `Net::HTTP` as well). I look to [JBD's examples](https://medium.com/google-cloud/tracing-google-cloud-8f0c8ba8181c) of Go networking event instrumentation, where a request has a span, but `ConnectStart` and `ConnectDone` are events attached to the request span. There's a few arguments for this: a `client` request will often have a corresponding `server` span, but a `connect` usually will not; `connect` is an additional level of detail that may help us understand why one request was slower than another, so it helps to have it directly attached to the request span; on the other hand, we may not want that level of detail in general, so we might want to strip it out before exporting a span, and that is easier to do with a processor that removes events.

In the case of `Net::HTTP` and generating a span for `connect`, I do think we should continue to trace it and not have it as an event.  We have examples from our internal tracing gem where `connect` requests are failing (timing out) and the trace is capturing valuable information that I'm not sure would fit well into an event.  I also have an example where the parent of the connect span is not an http request, so it would look odd having an http connect event on a seeming unrelated span.

I believe tracing [connect](https://github.com/nahi/httpclient/blob/4d60d8bdba79719b7729fca4e4a91b8959b0c1d0/lib/httpclient/session.rb#L773) in this gem could surface similar value.  

> Of course, I may be reading too much into JBDs excellent screenshots.

No sarcasm intended, they do make for a compelling argument in making these events.

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/session.rb
robertlaurin: This is a shameless port of our internal instrumentation of this gem, weirdness and all 😄 .

fbogsany:
> I also have an example where the parent of the connect span is not an http request, so it would look odd having an http connect event on a seeming unrelated span.

That's an excellent argument for keeping it as a span - thanks for that.

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/session.rb
fbogsany: meh, ok 😉

robertlaurin: Removed client span type from connect span.

fbogsany:
> > 1. I wonder if we should instrument a level deeper, by monkey-patching `do_get_block` and `do_get_stream`? WebMock, at least, appears to hook in at that level.
> 
> The motivation isn't obvious to me, what would be the value of this change? I see that `do_get_block` is a private method only used in `do_request`.

`do_request` includes retries, so there are odd edge cases where a `kind=client` span created at that level will have multiple `kind=server` child spans. `do_get_block` and `do_get_stream` are lower level methods that only handle a single request-response for sync and async requests respectively.

I agree we can focus on the sync request path in this PR and defer the async path to another PR.

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/client.rb
```diff
@@ -0,0 +1,56 @@
...
+            uri = req.header.request_uri
+            url = "#{uri.scheme}://#{uri.host}"
+            request_method = req.header.request_method
+            attributes = {
+              'http.method' => request_method,
+              'http.scheme' => uri.scheme,
+              'http.target' => uri.path,
+              'http.url' => url,
+              'peer.hostname' => uri.host,
+              'peer.port' => uri.port
```
fbogsany, line 25: is `uri.port` always non-nil?

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/client.rb
```diff
@@ -0,0 +1,56 @@
...
+                annotate_span_with_response!(span, response)
+                conn.push response
+              end
+            end
+          end
+
+          def annotate_span_with_response!(span, response)
+            return unless response&.status_code
+
+            status_code = response&.status_code.to_i
```
fbogsany, line 41:
```suggestion
            status_code = response.status_code.to_i
```
No need to be lonely - you already tested it above, so we know it is non-`nil`.

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/client.rb
fbogsany: Based on the tests, I think the answer is "yes" - I withdraw the question 😄

fbogsany (approved):

PR #510, opened 2020-12-10
feat: metrics reporting from trace export

fbogsany:
This enables reporting of metrics from the trace export pipeline, including the OTLP exporter and the BatchSpanProcessor. We need this mechanism to allow reporting metrics that are useful for monitoring trace export from applications in production, and we can't create a binding to the (currently) unstable Metrics API from the stable SDK.

The interface to metrics reporting is defined by the `OpenTelemetry::SDK::Trace::Export::MetricsReporter`, which also provides the default empty implementation. I tried to stick with concepts that would map cleanly to the current spec for OTel Metrics, and also to StatsD. It's slightly weird in that we have `record_value`, which is a synchronous operation in OTel Metrics that maps to StatsD `distribution` (or `histogram` if you prefer that), and also `observe_value`, which is asynchronous in OTel Metrics and maps to StatsD `gauge`. The weirdness comes about because we're calling both synchronously.

The metrics I defined include:
* `otel.otlp_exporter.request_duration`
* `otel.otlp_exporter.failure` ("soft" failure - request will be retried)
* `otel.bsp.buffer_utilization` (a snapshot of "fullness" of the BSP buffer)
* `otel.bsp.export.success`
* `otel.bsp.export.failure` (hard failure - request will not be retried)
* `otel.bsp.exported_spans`
* `otel.bsp.dropped_spans`

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -183,7 +183,7 @@ module OpenTelemetry
           def report_result(result_code, batch)
             if result_code == SUCCESS
               @metrics_reporter.add_to_counter('otel.bsp.export.success')
-              @metrics_reporter.add_to_counter('otel.bsp.export.spans', increment: batch.size)
+              @metrics_reporter.add_to_counter('otel.bsp.exported_spans', increment: batch.size)
             else
               OpenTelemetry.logger.error("Unable to export #{batch.size} spans")
               @metrics_reporter.add_to_counter('otel.bsp.export.failure')
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -192,7 +192,7 @@ module OpenTelemetry
           end
 
           def report_dropped_spans(count, reason:)
-            @metrics_reporter.add_to_counter('otel.bsp.dropped_span', increment: count, labels: { 'reason' => reason })
+            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: count, labels: { 'reason' => reason })
           end
 
           def fetch_batch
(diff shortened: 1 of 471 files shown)


ericmustin:
@fbogsany on first pass this looks good, I think this is super valuable but curious if there's any specification guidance around this? I maybe have some nits on the naming conventions but otherwise lgtm

fbogsany:
> @fbogsany on first pass this looks good, I think this is super valuable but curious if there's any specification guidance around this? I maybe have some nits on the naming conventions but otherwise lgtm

There's no guidance in the spec yet. I was asked awhile back to add some, but they're considered post-GA at this point. Feel free to bikeshed the naming conventions (though also keep in mind that an implementation can map these to whatever makes sense in a given production environment).

robertlaurin (approved):
🎉

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -173,7 +181,18 @@ def export_batch(batch, timeout: @exporter_timeout_seconds)
...
+              @metrics_reporter.add_to_counter('otel.bsp.export.spans', increment: batch.size)
+            else
+              OpenTelemetry.logger.error("Unable to export #{batch.size} spans")
+              @metrics_reporter.add_to_counter('otel.bsp.export.failure')
+              report_dropped_spans(batch.size, reason: 'export-failure')
+            end
+          end
+
+          def report_dropped_spans(count, reason:)
+            @metrics_reporter.add_to_counter('otel.bsp.dropped_span', increment: count, labels: { 'reason' => reason })
```
ericmustin, line 195: nit: should this be plural?

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -173,7 +181,18 @@ def export_batch(batch, timeout: @exporter_timeout_seconds)
           end
 
           def report_result(result_code, batch)
-            OpenTelemetry.logger.error("Unable to export #{batch.size} spans") unless result_code == SUCCESS
+            if result_code == SUCCESS
+              @metrics_reporter.add_to_counter('otel.bsp.export.success')
+              @metrics_reporter.add_to_counter('otel.bsp.export.spans', increment: batch.size)
```
ericmustin, line 186: very small nit, should we have `otel.bsp.export.spans and `otel.bsp.dropped_span` have the same namespace? so, either `otel.bsp.export.dropped_span` or `otel.bsp.spans`? non blocking imo

ericmustin (approved):
very small questions or thoughts but lgtm, nice work

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: Hmm, so initially I 👍 this suggestion, but upon reflection I don't think they should share a namespace. `otel.bsp.export.spans` is intended to communicate the number of spans successfully exported. `otel.bsp.dropped_spans` is a count of spans dropped, which might not be due to a failed export (we have 3 reasons at least for dropping spans: termination timeout, full buffer, failed export).

fbogsany: I could see renaming `otel.bsp.export.spans` -> `otel.bsp.exported_spans`, so back to 👍 . I'll do that.

PR #521, opened 2020-12-17
feat: Structured error handling

### LINKED ISSUE #203: Invalid attributes on Spans, Events, Links, and Resources
We've discussed this before and temporarily went with the idea of a strict mode, that was never finished and the [Exception handing spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/error-handling.md) discourages an OTel implementation from raising unhandled exceptions. See: #14 and #126.

We are in the situation where invalid attributes on spans, links, events and resources can make it into the export pipeline where currently it breaks things. In order to free the export pipeline from having to validate all data, we should ensure that attributes that make to the pipeline are valid.

I'm proposing the following strategy for attribute validation:

* When possible, validate attributes only on telemetry that is going to be exported
* Avoid excessive runtime checks on data that won't be exported
* Valid attributes should be kept, while invalid attributes should be dropped and a message can optionally be logged.
* Invalid attributes have the potential to generate lots of logging data, so we should be conservative in logging, if we go that route

### LINKED ISSUE #464: SDK Exporter/BatchProcessor should not hard-fail in an initializer

fbogsany:
Replaces #207 
Replaces #465 
Fixes #464 
Fixes #203 (maybe?)

This introduces a configurable error handling mechanism to the API, handles configuration errors by wrapping them in a `ConfigurationError` and passing it to the error handler, and validates attribute types when finishing an SDK span.

### TODO
- [x] tests
- [x] ~~decide whether we should have a distinguished error type for attribute validation errors, to allow an error handler to ignore or otherwise control logging volume~~ punting this to a later PR

### DIFF
--- a/api/lib/opentelemetry.rb
+++ b/api/lib/opentelemetry.rb
@@ -22,11 +22,25 @@ require 'opentelemetry/version'
 module OpenTelemetry
   extend self
 
-  attr_writer :tracer_provider, :meter_provider, :baggage, :logger
+  attr_writer :tracer_provider, :meter_provider, :baggage, :logger, :error_handler
 
   # @return [Object, Logger] configured Logger or a default STDOUT Logger.
   def logger
     @logger ||= Logger.new(STDOUT, level: ENV['OTEL_LOG_LEVEL'] || Logger::INFO)
+  end
+
+  # @return [Callable] configured error handler or a default that logs the
+  #   exception and message at ERROR level.
+  def error_handler
+    @error_handler ||= ->(exception: nil, message: nil) { logger.error("OpenTelemetry error: #{[message, exception&.message].compact.join(' - ')}") }
+  end
+
+  # Handles an error by calling the configured error_handler.
+  #
+  # @param [optional Exception] exception The exception to be handled
+  # @param [optional String] message An error message.
+  def handle_error(exception: nil, message: nil)
+    error_handler.call(exception: exception, message: message)
   end
 
   # @return [Object, Trace::TracerProvider] registered tracer provider or a
--- a/api/lib/opentelemetry/instrumentation/registry.rb
+++ b/api/lib/opentelemetry/instrumentation/registry.rb
@@ -79,8 +79,7 @@ module OpenTelemetry
           OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install"
         end
       rescue => e # rubocop:disable Style/RescueStandardError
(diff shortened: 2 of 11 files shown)


api/lib/opentelemetry.rb
```diff
@@ -22,13 +22,27 @@
...
+  def error_handler
+    @error_handler ||= ->(exception: nil, message: nil) { logger.error("OpenTelemetry error: #{[message, exception&.message].compact.join(' - ')}") }
+  end
+
+  # Handles an error by calling the configured error_handler.
+  #
+  # @param [optional Exception] exception The exception to be handled
+  # @param [optional String] message An error message.
+  def handle_error(exception: nil, message: nil)
+    error_handler.call(exception: exception, message: message)
```
robertlaurin, line 43: I like this, it provides a simple flexible interface to let users to configure how they want to handle exceptions, or non-exceptional errors (like invalid attributes)

api/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -79,8 +79,7 @@ def install_instrumentation(instrumentation, config)
           OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install"
         end
       rescue => e # rubocop:disable Style/RescueStandardError
-        OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} unhandled exception " \
-                                  "during install #{e}: #{e.backtrace}"
+        OpenTelemetry.handle_error(exception: e, message: "Instrumentation: #{instrumentation.name} unhandled exception during install: #{e.backtrace}")
```
robertlaurin, line 82: ❤️

robertlaurin (approved):
This seems like a good approach to me!

robertlaurin (approved):

PR #527, opened 2020-12-22
feat: instrument graphql

robertlaurin:
Continuation of the work done here https://github.com/open-telemetry/opentelemetry-ruby/pull/410


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+            'analyze_multiplex' => 'graphql.analyze_multiplex',
+            'execute_query' => 'graphql.execute_query',
+            'execute_query_lazy' => 'graphql.execute_query_lazy',
+            'execute_multiplex' => 'graphql.execute_multiplex'
+          }
+
+          def platform_trace(platform_key, key, data)
+            return yield if platform_key.nil?
+
+            tracer.in_span(platform_key || 'test', attributes: attributes_for(key, data)) do |_span|
```
fbogsany, line 30:
```suggestion
            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |_span|
```

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+            GraphQL::Instrumentation.instance.config
+          end
+
+          def attributes_for(key, data)
+            case key
+            when 'execute_query'
+              {
+                'selected_operation_name' => data[:query].selected_operation_name,
+                'selected_operation_type' => data[:query].selected_operation.operation_type,
+                'query_string' => data[:query].query_string
```
fbogsany, line 69: Is there any chance of these being `nil` or are they always defined for `execute_query`?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -0,0 +1,79 @@
...
+          end
+
+          def platform_authorized_key(type)
+            return unless config[:enable_platform_authorized_key]
+
+            "#{type.graphql_name}.authorized"
+          end
+
+          def platform_resolve_type_key(type)
+            return unless config[:enable_platform_resolve_type_key]
```
fbogsany, line 48: These config options should be documented in the README and/or with the `...::Instrumentation` class docs.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
robertlaurin: Selected operation name might be, I'll investigate and circle back.

fbogsany: Thanks - keep in mind that `nil` attribute values are not permitted.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
robertlaurin: Updated.

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
robertlaurin: Added tests, and updated to handle when `selected_operation_name` is nil.

instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb
```diff
@@ -0,0 +1,239 @@
...
+  after do
+    # Force re-install of instrumentation
+    instrumentation.instance_variable_set(:@installed, false)
+
+    # Reset various instance variables to clear state between tests
+    ::GraphQL::Schema.instance_variable_set(:@own_tracers, [])
+
+    # Reseting @graphql_definition is needed for tests running against version `1.9.x`
+    SomeOtherGraphQLAppSchema.remove_instance_variable(:@graphql_definition) if SomeOtherGraphQLAppSchema.instance_variable_defined?(:@graphql_definition)
+    SomeGraphQLAppSchema.remove_instance_variable(:@graphql_definition) if SomeGraphQLAppSchema.instance_variable_defined?(:@graphql_definition)
```
robertlaurin, line 45: There was some fairly tricky state leaking between tests, particularly for `1.9.x` version of GraphQL.  This is required to prevent duplicate tracers executing, as well as running the tests that would not otherwise have a tracer added.  This came up when adding the config options for `schemas`.

instrumentation/graphql/README.md
```diff
@@ -22,6 +22,28 @@ OpenTelemetry::SDK.configure do |c|
...
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::GraphQL', {
+    # If this option is not provided the default behaviour is to instrument all
+    # schemas that extend GraphQL::Schema
+    schemas: [MyAppSchema],
+
+    # The following fields are disabled by default as they generate
+    # large traces, but can be enabled using the following keys.
```
fbogsany, line 35:
```suggestion
    # The following options are disabled by default as they generate
    # large traces, but can be enabled using the following keys to add
    # spans that provide more detail.
```

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/instrumentation.rb
```diff
@@ -11,6 +11,19 @@ module Instrumentation
...
+
+        # The enable_platform_authorized key expects a boolean value,
+        # and enables the tracing of "authorized", "authorized_lazy"
+
+        # The enable_platform_resolve_type key expects a boolean value,
+        # and enables the tracing of "resolve_type", "resolve_type_lazy"
+
+        # The schemas key expects an array of Schemas, and is used to specify
+        # which schemas are to be instrumented. If this value is not supplied
+        # the default behaviour is to instrument all schemas.
```
fbogsany, line 26:
```suggestion
        ## Supported configuration keys for the install config hash:
        #
        # The enable_platform_field key expects a boolean value,
        # and enables the tracing of "execute_field" and "execute_field_lazy".
        #
        # The enable_platform_authorized key expects a boolean value,
        # and enables the tracing of "authorized" and "authorized_lazy".
        #
        # The enable_platform_resolve_type key expects a boolean value,
        # and enables the tracing of "resolve_type" and "resolve_type_lazy".
        #
        # The schemas key expects an array of Schemas, and is used to specify
        # which schemas are to be instrumented. If this value is not supplied
        # the default behaviour is to instrument all schemas.
```

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
fbogsany: What do you need the `|| 'test'` for here? `platform_key` will never be `nil` (but I supposed it could be `false`).

robertlaurin: 🤕 that was placeholder that was accidentally re-added during a rebase. Gone now.

fbogsany (approved):

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -0,0 +1,77 @@
...
+require 'opentelemetry'
+
+module OpenTelemetry
+  module Instrumentation
+    module GraphQL
+      module Tracers
+        # GraphQLTracer contains the OpenTelemetry tracer implementation compatible with
+        # the GraphQL tracer API
+        class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
+          self.platform_keys = {
```
ericmustin, line 16: Where did we ever land with handling `execute_field` and `execute_field_lazy`? https://github.com/open-telemetry/opentelemetry-ruby/pull/410/files#r492104122 There was a suggestion at one point to handle these by adding them as events to current span context. Is there any prior art here in other graphql language instrumentations on what approach to take? fine to skip but just wanted to see where this stood

ericmustin: Ah, so we add them as spans but its set to false via config by default? `https://github.com/open-telemetry/opentelemetry-ruby/pull/527/files#diff-d66041e54fddd1da2dabb3f3b3d15692430f451a6ff416e6fa80d12b53ef588aR38` I'd be curious to see what js/python/etc do here

instrumentation/graphql/README.md
```diff
@@ -0,0 +1,76 @@
...
+    # The following options are disabled by default as they generate
+    # large traces, but can be enabled using the following keys to add
+    # spans that provide more detail.
+    # Further details about what these keys correspond can be found on the [platform_tracing.rb class](https://github.com/rmosolgo/graphql-ruby/blob/1.10.x/lib/graphql/tracing/platform_tracing.rb#L28-L73).
+    # enable_platform_field maps to the execute_field and execute_field_lazy keys
+    enable_platform_field: true,
+    # enable_platform_authorized maps to the authorized and authorized_lazy keys
+    enable_platform_authorized: true,
+    # enable_platform_resolve_type maps to the resolve_type and resolve_type_lazy keys
+    enable_platform_resolve_type: true
```
ericmustin, line 43: very small nit but maybe we should leave these commented out so we avoid ppl copy/pasting this snippet in and enabling stuff that, for the most part, we want disabled

ericmustin (approved):
lgtm, i had some minor feedback around config options/defaults and whether there was any interest in moving those "optionally on" fields to be events instead, but non blocking imo

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
robertlaurin: >There was a suggestion at one point to handle these by adding them as events to current span context. Is there any prior art here in other graphql language instrumentations on what approach to take? fine to skip but just wanted to see where this stood

I would take the position that the `execute_field` fields should not be events, as there is valuable timing that could be captured here.  Example would be if you had a query that had an `order` object and you retrieved `order { total }` the total may be a slow calculation and we would like to be able to surface that timing, I don't think an event is the right choice there.

The same is true for the other two though I'm not sure how interesting an `authorization` computation would be, but I'm sure there an exists a use case out there where the timing would be valuable.  

The only one I could maybe see possibly an event is the type resolution field, but again I could make the same argument for that as I would the `authorized` field, someone may have some interestingly slow type resolution code added, in which case we'd want to capture that timing.

ericmustin: @robertlaurin alrighty, makes sense to me then with the approach taken here.

robertlaurin: >Ah, so we add them as spans but its set to false via config by default?

Yeah quiet by default, I think that's the better user experience.  It's also to be mindful of vendor limits when submitting traces.  

> I'd be curious to see what js/python/etc do here

Admittedly I haven't looked, the future improvement I'd like to do in this regard is allowing a mechanism to enable tracing these fields on per request basis.  Maybe through the use of GraphQL contexts, or otel context propagation.

instrumentation/graphql/README.md
robertlaurin: I see what you mean, I wanted to demonstrate how they could turn it on but I guess I can just as easily do that by provided a false value instead.  I'll give it an update.

fbogsany (approved):

PR #529, opened 2020-12-23
feat: Pluggable ID generation

### LINKED ISSUE #528: SDK Trace & Span ID generation should be pluggable
The generation code currently lives in the API, in `OpenTelemetry::Trace`. We need to relocate it to somewhere that allows the SDK to replace it, if requested. It isn't immediately clear where to place it. It is called from the `SpanContext` constructor, which has no knowledge of the `TracerProvider`, for example.

fbogsany:
Fixes #528 

This adds support to the SDK for pluggable trace ID and span ID generation. The required interface is:
```ruby
generate_span_id -> String  # An 8-byte string with at least one non-zero byte.
generate_trace_id -> String # A 16-byte string with at least one non-zero byte.
```

This does not alter the behaviour of the API, and it defaults to using the API's ID generators.


robertlaurin (approved):
Looks good, maybe adding a short test using a deterministic generator?

PR #532, opened 2020-12-24
fix: remove incorrect TODO

fbogsany:
This fixes a TODO in the codebase for a perf improvement that only works on Ruby 2.5 and above. We removed support for Ruby 2.4 awhile back, so we can use this enhancement now.


ericmustin (approved):

fbogsany:
Looks like this variant of `shift` only applies to `Array`, not to `Hash`. I'll delete the comment instead.

ericmustin:
err, approved too soon, some trimming related tests are failing it looks like

ericmustin:
lgtm (removing the TODO, that is)

robertlaurin (approved):

PR #536, opened 2020-12-26
"all" gem is missing ruby_kafka

johnnyshields:
It would be nice if the release process could automatically update the `all` instrumentation gem so that this is not missed in the future.

Please cut a release (0.12.1?) after merging.


johnnyshields:
@fbogsany please merge.

robertlaurin (approved):
>It would be nice if the release process could automatically update the all instrumentation gem so that this is not missed in the future.

Yeah I agree, I added an instrumentation boilerplate generator that automatically adds to the `all` gem to alleviate this, but the ruby kafka pr predated it.

fbogsany (changes requested):
This also needs `require 'opentelemetry-instrumentation-ruby_kafka'` in `instrumentation/all/lib/opentelemetry/instrumentation/all.rb`

johnnyshields:
@fbogsany done

mwear (approved):

fbogsany (approved):

PR #546, opened 2021-01-11
fix: Add minimum gem version for ruby-kafka

johnnyshields:
I found this out the hard way :)

The method API of the monkey patched changed in ruby-kafka changed in 0.7.0+ (`headers:` kwarg was added). If you try to use 0.6.8 or earlier, chaos ensues.


robertlaurin (commented):
>I found this out the hard way :)

:( 

I think it probably makes sense for us to start adding this minimum version check for all instrumented gems using the oldest version defined in our appraisals.  In this case we would add an [appraisal](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/ruby_kafka/Appraisals#L21) for `0.7.0` up to the most recent version.  

This change looks good to me, but we should add appraisals to fill the gap though.  It looks like theres some issues with ruby 3.0 and this instrumentation that I need to look at.  So this is blocked for the moment, I will be working on it this week and circle back to this PR once it is done.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/instrumentation.rb
```diff
@@ -10,17 +10,23 @@ module RubyKafka
...
+        MINIMUM_VERSION = Gem::Version.new('0.7.0')
+
         install do |_config|
           require_patches
           patch
         end
 
         present do
-          defined?(::Kafka)
+          defined?(::Kafka) && gem_version >= MINIMUM_VERSION
```
mwear, line 21: There is a separate hook where this check would be more appropriate: 
https://github.com/open-telemetry/opentelemetry-ruby/blob/b61ea9601accfa2ad92f3c791fb16cc5221cc9b4/api/lib/opentelemetry/instrumentation/base.rb#L122-L130

mwear (changes requested):
One comment, otherwise LGTM.

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/instrumentation.rb
johnnyshields: This is done

johnnyshields:
> add an [appraisal](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/ruby_kafka/Appraisals#L21) for `0.7.0` up to the most recent version.

I've done this. Actually 1.0.0+ are already there, and 0.7.0 is the last in the 0.x series so I've added it (refer to code changes)

johnnyshields:
@robertlaurin it looks like there is some unrelated spec failure here:

```
  1) Failure:
OpenTelemetry::Instrumentation::RubyKafka::Patches::Consumer::#each_message#test_0001_traces each_meassage call [/__w/opentelemetry-ruby/opentelemetry-ruby/instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb:73]:
Expected # encoding: ASCII-8BIT
#    valid: true
"k\xD7\xBFu\xFF\xB7X\x92\xC3p\xB4\x9Cl\vH\xEA" to not be equal to # encoding: ASCII-8BIT
#    valid: true
"k\xD7\xBFu\xFF\xB7X\x92\xC3p\xB4\x9Cl\vH\xEA".
```

johnnyshields:
Ready for merge

robertlaurin (approved):
This lgtm, thanks.

fbogsany (approved):

PR #547, opened 2021-01-12
Adjust Sidekiq middlewares to match semantic conventions

BlakeMesdag:
Adds `messaging.system` and `messaging.destination_type` for Sidekiq spans, as well as adjusting the span names to match the conventions here: https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md

Another option is to use the job class in the span name along with the operation name to imbue more meaning. I've found that the using the queue name in practice is quite generic and requires the use of additional attributes for job class to do useful analysis.


instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/server/tracer_middleware.rb
```diff
@@ -15,10 +15,12 @@ class TracerMiddleware
             def call(_worker, msg, _queue)
               parent_context = OpenTelemetry.propagation.text.extract(msg)
               tracer.in_span(
-                msg['wrapped']&.to_s || msg['class'],
+                "#{job['queue']} process",
```
robertlaurin, line 18:
```suggestion
                "#{msg['queue']} process",
```

fbogsany:
> Another option is to use the job class in the span name along with the operation name to imbue more meaning. I've found that the using the queue name in practice is quite generic and requires the use of additional attributes for job class to do useful analysis.

I think we should add the job class as an attribute `messaging.sidekiq.job_class`.

I thought about adding a configuration option to use the job class instead of the queue name in the span name, but I think just adding the job class attribute provides enough information for a span processor to change the span name in the OTel collector. There's a larger discussion to be had around deployments without the collector.

fbogsany:
Builds are blocked on a Fastly CDN issue that affects requests to rubygems: https://status.fastly.com/incidents/1rbr326kjsmt

fbogsany:
We discussed this on today's call. There was general agreement that:
1. We should default to adhering to the semantic conventions
2. The messaging semantic conventions aren't a perfect fit for a job queuing system - they're the closest for now, but we should look into creating semantic conventions for job queues (e.g. Sidekiq, Resque in Ruby, Celery? in Python)
3. We should have attributes for the job queue and the job class
4. We should provide a config option to use the job class name as the span name rather than `#{queue} send` and `#{queue} process`

The config option is the thing we need before 👍

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -29,6 +32,18 @@ def call(_worker_class, job, _queue, _redis_pool)
 
             private
 
+            def span_name(job)
+              if config[:enable_job_class_span_names]
+                job['wrapped']&.to_s || job['class']
```
fbogsany, line 37:
```suggestion
                "#{job['wrapped']&.to_s || job['class']} enqueue"
```
@robertlaurin and I looked at some traces with the job class as the span name, and it is a little confusing to see the same span name for the producer and consumer. It is much more readable to see `Foo enqueue` and `Foo process`, and "enqueue fits the domain language better [than send]" (quoting @robertlaurin).

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
```diff
@@ -14,10 +14,13 @@ module Client
...
               tracer.in_span(
-                job['wrapped']&.to_s || job['class'],
+                span_name(job),
                 attributes: {
+                  'messaging.system' => 'sidekiq',
+                  'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
                   'messaging.message_id' => job['jid'],
-                  'messaging.destination' => job['queue']
+                  'messaging.destination' => job['queue'],
+                  'messaging.destination_type' => 'queue'
```
fbogsany, line 23:
```suggestion
                  'messaging.destination_kind' => 'queue'
```
Not sure why we had `destination_type`. Please do re-check the semantic conventions in case my eyes are playing tricks on me.

fbogsany:
Trying to make the case for batch job system semantic conventions in https://github.com/open-telemetry/opentelemetry-specification/issues/1347

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
BlakeMesdag: Woops, you're right, conventions call for `_kind` not `_type`

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
BlakeMesdag: Adjusted, went with `enqueue` only for job class named spans until the conventions are updated for job systems.

robertlaurin (approved):

fbogsany (approved):

Issue #549, opened 2021-01-12
Baggage not propagated for consumers who use ruby_kafka instrumentation


Asafb26:
currently, there is a different logic between:
```ruby
parent_context = OpenTelemetry.propagation.text.extract({traceparent: ..., baggage: my_baggage})
tracer.in_span("test", with_parent: parent_context, attributes: attributes, kind: :consumer) do
  puts OpenTelemetry.baggage.values
end

#prints nil
```
and
```ruby
parent_context = OpenTelemetry.propagation.text.extract({traceparent: ..., baggage: my_baggage})
OpenTelemetry::Context.with_current(parent_context) do
  tracer.in_span("test", attributes: attributes, kind: :consumer) do
    puts OpenTelemetry.baggage.values
  end
end

#prints the content of my_baggage
```

The first keeps the current context active while use the parent span_context data, while the second replaces the current context, yet use the parent span_context data (from the replaced context).

The issue starts with the fact that baggage is not a part of span_context, so for proper propagation of baggage data - we must use the second approach.

fbogsany:
This difference seems really unfortunate. Are we correctly implementing the spec here in `tracer.in_span(..., with_parent: ...)`? If not, we should fix that. If we are, I wonder if this difference in behaviour should be raised as a spec issue?

Asafb26:
AFAIK most instrumentation uses the second approach, so in my perspective - if it's not an issue, I can align the implementation of ruby_kafka with the rest.
regarding the specs - the baggage is not a part of the span_context, which I found it bit weird, but yet confirms that we are aligned to the specs

fbogsany:
Fixed in #550

PR #550, opened 2021-01-12
fix: ruby_kafka baggage propagation

Asafb26:
See associated issue #549


instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -22,12 +22,15 @@ def each_message(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automati
...
-              span_context = OpenTelemetry::Trace.current_span(OpenTelemetry.propagation.text.extract(message.headers)).context
+              parent_context = OpenTelemetry.propagation.text.extract(message.headers)
+              span_context = OpenTelemetry::Trace.current_span(parent_context).context
               link = OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
               links = [link] if link
 
-              tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer) do
-                yield message
+              OpenTelemetry::Context.with_current(parent_context) do
+                tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer) do
```
fbogsany, line 31: This isn't the same thing. This will set the parent of the new span to the extracted span context in addition to linking to it, whereas what we're trying to do here is create a span with the current local context as a parent (I'm not sure that's right - I think we want the new span to be a root span), and a link to the extracted span context.

Asafb26: Got it, but how should we propagate the baggage without it? should we need to change the specs for that?

fbogsany: I think this will use the current span context as parent, the extracted span context as a link, and the rest of the extracted context as the current context:
```ruby
              outer_context = OpenTelemetry::Context.current
              extracted_context = OpenTelemetry.propagation.text.extract(message.headers)
              span_context = OpenTelemetry::Trace.current_span(extracted_context).context
              link = OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
              links = [link] if link

                tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer, with_parent: outer_context) do
                  yield message
                end
              end
```

If we explicitly wanted to create a root span (I don't think we do, on reflection), the `in_span` call would be:
```ruby
                tracer.in_span("#{message.topic} process", links: links, attributes: attributes, kind: :consumer, with_parent: OpenTelemetry::Context.empty) do
                  yield message
                end
```

fbogsany: @mwear ^ discussed on today's call.

Asafb26: in this case, the trace_id of the consumer will be different from the producer's trace_id, right? shouldn't we trace our entire flow under a single parent trace_id?

Asafb26: and what about batch consumption?
https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb#L48

the baggage has no meaning in batch operation, should we keep that flow untouched?

fbogsany: > the baggage has no meaning in batch operation, should we keep that flow untouched?

Agreed, baggage has no meaning there - we should not update that path.

fbogsany: There are 3 Kafka examples provided in the spec's semantic conventions:

1. When [receiving a single message](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#apache-kafka-example) we continue the trace.
2. When [receiving a batch](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-receiving) and processing messages individually, a new trace is started for the `receive` operation.
3. When [processing a batch](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/messaging.md#batch-processing) of messages received individually, each `receive` operation continues the trace from its message, but the `process` operation starts a new trace.

In `ruby-kafka`, there is only batch `receive` followed by either individual `process`ing of messages or batch `process`ing. This means we can implement 2. directly, although we don't yet instrument `receive`. Since we're only instrumenting `process`, we assume that there may be a parent `receive` span already in the local context and we don't continue the trace for the message we're extracting.

fbogsany: As stated in #551, "Kafka is hard". This seems especially true of Ruby Kafka 😞.

Asafb26: Really hard 😕 
I've got your point, the existence of batch receive breaks the current trace scope.
In my production needs, we don't really care about the batch receive operation, but more than a "single flow" per producer, I mean that for tracing perspective - I prefer to omit the receive operation entirely for better readability of the entire trace.
is there any point to make it configurable? or should we fix it to 2 and live with it?

fbogsany: > is there any point to make it configurable? or should we fix it to 2 and live with it?

It's worth a try. I also concluded that the receive span was pretty useless, but I can see the value in configuring whether or not to continue the trace. 🤔 

Maybe, if we opt to **not** instrument `receive`, we can treat `each_message` as 1. and treat `each_batch` as 3. That seems more in line with user expectations, even if it seems to deviate slightly from the letter of the spec (or at least the examples). Does that seem reasonable to you?

(cc @robertlaurin and @d1egoaz for opinions too)

For those who don't want to read the thread, this means:
1. `each_message` will create a `... process` span per message with both a link and parent referring to the `... send` span for that message, and continue the trace.
2. `each_batch` will create a single `... process` span for the batch with links to each of the `... send` spans for messages in the batch. The `... process` span will be the root span of a new trace.

robertlaurin: To clarify further:
- We would leave `each_batch` as is.
- For `each_message` we could/would add a configuration option that allows for continuing the trace from the `... send` span.
- Links would continue to exist as they are regardless of the configuration option.

If I'm following correctly this seems completely reasonable from my perspective.

fbogsany: > * For `each_message` we could/would add a configuration option that allows for continuing the trace from the `... send` span that created the message.

I'm on the fence about the config option. I wonder if we should default to continuing the trace, and add a config option to start a new trace instead.

robertlaurin: If we don't currently have a need for starting a new trace then we can skip adding the config option.  We can always add it later *if* there is demand for it.

Asafb26: agree, if we decide that the default behavior is trace continuation, we can add the new trace logic in a future config option/once we add `... receive` instrumentation

instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/patches/consumer.rb
```diff
@@ -22,12 +22,14 @@ def each_message(min_bytes: 1, max_bytes: 10_485_760, max_wait_time: 1, automati
 
               attributes['messaging.kafka.message_key'] = message.key if message.key
 
-              span_context = OpenTelemetry::Trace.current_span(OpenTelemetry.propagation.text.extract(message.headers)).context
-              link = OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
-              links = [link] if link
+              parent_context = OpenTelemetry.propagation.text.extract(message.headers)
+              span_context = OpenTelemetry::Trace.current_span(parent_context).context
+              links = [OpenTelemetry::Trace::Link.new(span_context)] if span_context.valid?
```
Asafb26, line 27: changed to one-liner  due to assignment-branch-condition limitation (18.28/18)

fbogsany (approved):

robertlaurin (approved):

PR #551, opened 2021-01-12
fix: set context in in_span

fbogsany:
# THIS IS NOT INTENDED TO BE MERGED AS IS - IT IS INTENDED FOR DISCUSSION OF SEMANTICS ONLY

#549 shows two code examples that behave differently. At first glance, it seems like they should behave the same way, and this PR attempts to "fix" the `Tracer.in_span` helper to make them behave the same way.

I want to make it clear that, after close reading of the spec, I do **NOT** think we should merge this. I think the current behaviour is correct, even if it isn't immediately intuitive.

`in_span` is a helper that we've added. The spec [states](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md#span-creation):
> `Span` creation MUST NOT set the newly created `Span` as the currently active `Span` by default, but this functionality MAY be offered additionally as a separate operation.

This is what `in_span` implements. The spec says nothing about this optional operation setting an outer Context, and I don't think it should. It should simply behave as described: create a span as if `start_span` were called with the same parameters and then make it active for the duration of the provided block, using `Trace.with_span`.

As discussed in today's call, the behaviour we're trying to achieve in #549 doesn't seem like a common case, and largely falls under the heading of "Kafka is hard".

cc @Asafb26

### DIFF
--- a/api/lib/opentelemetry/trace/tracer.rb
+++ b/api/lib/opentelemetry/trace/tracer.rb
@@ -25,8 +25,15 @@ module OpenTelemetry
       #   span to the block.
       def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, with_parent: nil)
         span = nil
-        span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, with_parent: with_parent)
-        Trace.with_span(span) { |s, c| yield s, c }
+        if with_parent.nil?
+          span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, with_parent: with_parent)
+          Trace.with_span(span) { |s, c| yield s, c }
+        else
+          OpenTelemetry::Context.with_current(with_parent) do
+            span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind)
+            Trace.with_span(span) { |s, c| yield s, c }
+          end
+        end
       rescue Exception => e # rubocop:disable Lint/RescueException
         span&.record_exception(e)
         span&.status = Status.new(Status::ERROR,


api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -25,8 +25,15 @@ class Tracer
       #   span to the block.
       def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, with_parent: nil)
         span = nil
-        span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, with_parent: with_parent)
-        Trace.with_span(span) { |s, c| yield s, c }
+        if with_parent.nil?
+          span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind, with_parent: with_parent)
```
johnnyshields, line 29: If `with_parent` is nil, why pass it here?

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -25,8 +25,15 @@ class Tracer
...
+          Trace.with_span(span) { |s, c| yield s, c }
+        else
+          OpenTelemetry::Context.with_current(with_parent) do
```
johnnyshields, line 32: Another way to code this would be to make `OpenTelemetry::Context.with_current` yield and return if `with_parent` is nil. That would eliminate the `if` statement in this method.

mwear:
I think that we probably want this behavior. I mentioned this on #630, but without this change our Rack instrumentation does not automatically pick up baggage. We could treat this as a bug in the Rack instrumentation, and not use `in_span`, but this is now a second instance where we've run into this issue.

fbogsany:
I think we should fix the Rack instrumentation, and probably other instrumentation that propagates trace context (Sidekiq?). My concern with changing behaviour of `in_span` now is that there is an active project in the Specification SIG aiming to define a convenience API for OpenTelemetry, and this falls squarely in that domain. Changing instrumentation and recommendations now will not break later if something like `in_span` is specified as in this PR. Conversely, changing `in_span` now **will** cause breakage in instrumentation and (more importantly) client code if the spec ends up reflecting our current implementation.

fbogsany:
Replaced by #729

PR #556, opened 2021-01-14
fix: use port 443 for secure connections

jfeltesse-mdsol:
When the `insecure` flag it set to `false` and the port is not specified in the endpoint parameter, the http client will end up using SSL but on the port 80.

Please let me know if I need to update other files like changelog, version etc.

I've not signed the CNCF CLA yet because I'm not sure how we ought to handle this in our company. Opening the PR now to get feedback in the meantime.


fbogsany:
FYI the spec has changed to remove the insecure flag: https://github.com/open-telemetry/opentelemetry-specification/pull/1234

jfeltesse-mdsol:
Oh ok so it means the exporter and its tests here should be updated to reflect this then? Is there any blocker?

fbogsany:
> Oh ok so it means the exporter and its tests here should be updated to reflect this then?

👍 

> Is there any blocker?

I don't think there's any blocker.

jfeltesse-mdsol:
Closing then. Thanks for the pointers!

Issue #470, opened 2020-11-03
Implement Jaeger Propagator


fbogsany:

mwear:
@fhwang showed interest in this on gitter. If you comment here, I can assign this to you.

fhwang:
🙋

mwear:
Closing as #564 has merged.

PR #564, opened 2021-01-20
feat: Implement Jaeger Propagator.

fhwang:
I think this is ready for review now. I started by mimicking the B3 propagator and changing where needed. A few design notes:

- The Jaeger format supports baggage using keys like `uberctx-{baggage key}` which then begs the question as to how to customize the Rack extractor and injector. I chose to go with a design that allowed optional blocks for handling keys.
- Regarding testing this baggage functionality, I opted to include SDK as a development dependency. I spent some time trying to do some surface-level mocking with Mocha, but it ended up being pretty tangled-up. I think the ideal way to test this is to have something that is a functional baggage implementation, which seems like a good use of the SDK. If we consider this a risk I can make another pass on using mocking however.


mwear (commented):
@fhwang sorry for the turnaround time on this review. I took a look over this last week and realized that some of the issues you were running into processing baggage headers with Rack could likely be improved with some changes to our Getter implementation.

I started experimenting with that and opened #589. With the Getter changes, we are able to write the injectors and extractors in a carrier agnostic way. We no longer need to specify Rack specific keys, or maintain Rack specific injector / extractor instances. If we need to extract from a Rack based carrier we just pass in a RackEnvGetter.

This generally looks to be on the right track. If you get a chance, can you rebase this, take a look at the updated B3 injectors / extractor, and follow the same pattern?

fhwang:
No problem. I'll take a look at incorporating #589 in the next day or so.

fhwang:
Closes #470.

propagator/jaeger/README.md
```diff
@@ -0,0 +1,45 @@
+# opentelemetry-propagator-jaeger
+
+The `opentelemetry-propagator-jaeger` gem contains injectors and extractors for the
+[Jaeger native propagation format][jaeger-spec]  They can be used in
+conjuction with each other as well as with injectors and extractors for
```
fbogsany, line 5:
```suggestion
conjunction with each other as well as with injectors and extractors for
```

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,105 @@
...
+# The OpenTelemetry module provides global accessors for telemetry objects.
+# See the documentation for the `opentelemetry-api` gem for details.
+module OpenTelemetry
+  # Namespace for OpenTelemetry propagator extension libraries
+  module Propagator
+    # Namespace for OpenTelemetry Jaeger propagation
+    module Jaeger
+      # Extracts context from carriers
+      class TextMapExtractor
+        TRACE_SPAN_IDENTITY_REGEX = /\A(?<trace_id>(?:[0-9a-f]{16}){1,2}):(?<span_id>[0-9a-f]{16}):[0-9a-f]{1,16}:(?<sampling_flags>[0-9a-f]{1,2})\z/.freeze
```
mwear, line 20: The Jaeger spec (https://www.jaegertracing.io/docs/1.21/client-libraries/#propagation-format) seems to allow trace ids to be lengths other than 16 or 32 hex chars.

trace id:

> Receivers MUST accept hex-strings shorter than 32 characters and 0-pad them on the left

span id:

> Receivers MUST accept hex-strings shorter than 16 characters and 0-pad them on the left

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,105 @@
...
+        SAMPLED_VALUES = %w[1 3 5 7].freeze
```
mwear, line 21: You shouldn't need to enumerate these values. Just parse the flags, convert to an integer, and use a bitwise & to test for sampled or debug.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,105 @@
...
+            span_id: Array(match['span_id']).pack('H*'),
+            trace_flags: to_trace_flags(match['sampling_flags']),
+            remote: true
+          )
+          Trace::Span.new(span_context: span_context)
+        end
+
+        def set_baggage(carrier, context, getter)
+          getter.keys(carrier).each do |carrier_key|
+            baggage_key = (carrier_key =~ /^uberctx-(.*)/) && Regexp.last_match(1).downcase.gsub(/_/, '-')
```
mwear, line 76: Usually, if you can get away with `String#start_with?` it's more efficient than a regex. I think you can use that here. Is there conversion of '_' to '-' necessary? I believe the RackGetter handles this already.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+module OpenTelemetry
+  # Namespace for OpenTelemetry propagator extension libraries
+  module Propagator
+    # Namespace for OpenTelemetry Jaeger propagation
+    module Jaeger
+      # Injects context into carriers
+      class TextMapInjector
+        DEFAULT_FLAG_BIT = 0b0
```
mwear, line 13: Any reason not to represent these as base16 like the spec: https://www.jaegertracing.io/docs/1.21/client-libraries/#propagation-format

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,105 @@
...
+            next unless baggage_key
+
+            value = getter.get(carrier, carrier_key)
```
mwear, line 79: I believe that baggage values should be url decoded on extract: https://www.jaegertracing.io/docs/1.21/client-libraries/#value-encoding

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          flags = to_flags(context, span_context)
+          trace_span_identity_value = [
+            span_context.hex_trace_id, span_context.hex_span_id, '0', flags
+          ].join(':')
+          setter ||= @default_setter
+          setter.set(carrier, IDENTITY_KEY, trace_span_identity_value)
+          OpenTelemetry.baggage.values(context: context).each do |key, value|
+            baggage_key = 'uberctx-' + key
+            setter.set(carrier, baggage_key, value)
```
mwear, line 50: Similarly, I think that baggage values should be url encoded on inject (https://www.jaegertracing.io/docs/1.21/client-libraries/#value-encoding)

api/lib/opentelemetry/context/propagation/rack_env_setter.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  class Context
+    module Propagation
+      # RackEnvSetter provides common methods for setting keys and values to
+      # a rack environment. It abstracts key transformation, turning for
+      # example +traceparent+ to +HTTP_TRACEPARENT+.
+      class RackEnvSetter
```
mwear, line 13: It seems like this is only used it tests. I fear it might cause confusion for users and I don't know that there is much of a real world use case for it. My preference would be to omit it and create the rack formatted keys by hand in tests.

mwear (changes requested):

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
fhwang: That part of the spec says to only URL-encode when using the `HTTP_HEADERS` propagation format wrt OpenTracing. I'm not completely clear how we'd map this part of the spec to OpenTelemetry, but maybe we should just use the `TEXT_MAP` format instead?

fhwang:
Thanks for the helpful review @mwear . I believe I've addressed all the issues, with the exception of URL encoding / decoding. I'm submitting these changes as standalone commits to ease review but once I have approval I was thinking of squash-merging to keep `main` clean.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,113 @@
...
+# The OpenTelemetry module provides global accessors for telemetry objects.
+# See the documentation for the `opentelemetry-api` gem for details.
+module OpenTelemetry
+  # Namespace for OpenTelemetry propagator extension libraries
+  module Propagator
+    # Namespace for OpenTelemetry Jaeger propagation
+    module Jaeger
+      # Extracts context from carriers
+      class TextMapExtractor < Operation
+        TRACE_SPAN_IDENTITY_REGEX = /\A(?<trace_id>(?:[0-9a-f]{2}){1,16}):(?<span_id>([0-9a-f]{2}){1,8}):[0-9a-f]{1,16}:(?<sampling_flags>[0-9a-f]{1,2})\z/.freeze
```
mwear, line 20: This regex enforces that the trace id and span id have an even number of hex digits. I don't think that is mandated by the Jaeger spec and we should probably allow for an odd number of digits.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/operation.rb
```diff
@@ -0,0 +1,48 @@
...
+# The OpenTelemetry module provides global accessors for telemetry objects.
+# See the documentation for the `opentelemetry-api` gem for details.
+module OpenTelemetry
+  # Namespace for OpenTelemetry propagator extension libraries
+  module Propagator
+    # Namespace for OpenTelemetry Jaeger propagation
+    module Jaeger
+      # Common superclass for classes that perform propagator operations, i.e.
+      # TextMapExtractor and TextMapInjector
+      class Operation
```
mwear, line 20: There's really no reason for this class. You can put these constants on the Jaeger module or in the injector / extractor depending on whether they need to be shared in both places.

mwear (changes requested):

propagator/jaeger/lib/opentelemetry/propagator/jaeger.rb
```diff
@@ -0,0 +1,61 @@
...
+
+      DEBUG_CONTEXT_KEY = Context.create_key('jaeger-debug-key')
+      private_constant :DEBUG_CONTEXT_KEY
+
+      TEXT_MAP_EXTRACTOR = TextMapExtractor.new
+      TEXT_MAP_INJECTOR = TextMapInjector.new
+
+      private_constant :TEXT_MAP_INJECTOR, :TEXT_MAP_EXTRACTOR
+
+      IDENTITY_KEY = 'uber-trace-id'
```
mwear, line 31: This and the constants below can be private

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,117 @@
...
+        #   respond to [] and []=.
+        # @yield [Carrier, String] if an optional getter is provided, extract
+        #   will yield the carrier and the header key to the getter.
+        # @return [Context] Updated context with active span derived from the
+        #   header, or the original context if parsing fails.
+        def extract(carrier, context, getter = nil)
+          getter ||= @default_getter
+          header = getter.get(carrier, IDENTITY_KEY)
+          return context unless (match = header.match(TRACE_SPAN_IDENTITY_REGEX))
+          return context if match['trace_id'] =~ /^0+$/
```
mwear, line 54: This regex can be a constant

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+
+        def set_baggage(carrier, context, getter)
+          baggage_key_prefix = 'uberctx-'
+          getter.keys(carrier).each do |carrier_key|
+            baggage_key = carrier_key.start_with?(baggage_key_prefix) && carrier_key[baggage_key_prefix.length..-1]
+            next unless baggage_key
+
+            raw_value = getter.get(carrier, carrier_key)
+            value = URI.decode_www_form_component(raw_value)
```
mwear, line 84: CGI.escape and CGI.unescape are what we should be using here
```suggestion
            value = CGI.escape(raw_value)
```

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,117 @@
...
+          if (sampling_flags & SAMPLED_FLAG_BIT) != 0
+            Trace::TraceFlags::SAMPLED
+          else
+            Trace::TraceFlags::DEFAULT
+          end
+        end
+
+        def to_span_id(span_id_str)
+          zero_pad_count = (16 - span_id_str.length) % 16
+          to_zero_padded_id(span_id_str, zero_pad_count)
```
mwear, line 102: I think this can be simplified to:
```suggestion
          span_id_str.rjust(16, '0')
```

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,117 @@
...
+        end
+
+        def to_trace_id(trace_id_str)
+          zero_pad_count = (16 - (trace_id_str.length % 16)) % 16
+          to_zero_padded_id(trace_id_str, zero_pad_count)
```
mwear, line 107: Similarly, I think this can be:
```suggestion
         trace_id.rjust(32, '0')
```

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+
+          flags = to_flags(context, span_context)
+          trace_span_identity_value = [
+            span_context.hex_trace_id, span_context.hex_span_id, '0', flags
+          ].join(':')
+          setter ||= @default_setter
+          setter.set(carrier, IDENTITY_KEY, trace_span_identity_value)
+          OpenTelemetry.baggage.values(context: context).each do |key, value|
+            baggage_key = 'uberctx-' + key
+            encoded_value = ERB::Util.url_encode(value)
```
mwear, line 49:
```suggestion
            encoded_value = CGI.escape(value)
```

mwear (changes requested):

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
fhwang: The current implementation emits 64-bit trace IDs when it receives 64-bit trace IDs; using `rjust(32, '0')` would force them to zero-padded 128-bit trace IDs. Both would seem to match the Jaeger spec, but given that the spec says "Clients in some languages support 128-bit, migration pending", that would imply to me that some clients **do not** support 128-bit. Meaning that it might be marginally safer to stick with 64-bit trace IDs in case there are legacy clients downstream.

Or is that gold-plating? Happy to just change it too.

propagator/jaeger/lib/opentelemetry/propagator/jaeger.rb
fhwang: These are private now.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
mwear: The official go jaeger client library uses [url.QueryEscape](https://golang.org/pkg/net/url/#QueryEscape) which works the same as CGI.escape, see: https://github.com/jaegertracing/jaeger-client-go/blob/master/propagation.go#L76-L92. The Ruby Jaeger client also uses `CGI.escape`, see: https://github.com/salemove/jaeger-client-ruby/blob/master/lib/jaeger/injectors.rb#L28.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
mwear: The [opentelemetry spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#spancontext) says:

> TraceId A valid trace identifier is a 16-byte array with at least one non-zero byte.

So, I think when extracting we can safely zero pad to 32 hex chars.

fhwang: Sounds good, I'll zero-pad to 32 hex chars.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
fhwang: 👍 works for me, I've changed these to CGI escape and unescape.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+        #
+        # @param [Context] context The active Context.
+        # @param [optional Callable] setter An optional callable that takes a
+        #   carrier and a key and a value and assigns the key-value pair in the
+        #   carrier. If omitted the default setter will be used which expects
+        #   the carrier to respond to [] and []=.
+        # @yield [Carrier, String, String] if an optional setter is provided,
+        #   inject will yield carrier, header key, header value to the setter.
+        # @return [Object] the carrier with context injected
+        def inject(carrier, context, &setter)
```
mwear, line 37:
```suggestion
        def inject(carrier, context, setter = nil)
```

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+        end
+
+        # Set the span context on the supplied carrier.
+        #
+        # @param [Context] context The active Context.
+        # @param [optional Callable] setter An optional callable that takes a
+        #   carrier and a key and a value and assigns the key-value pair in the
+        #   carrier. If omitted the default setter will be used which expects
+        #   the carrier to respond to [] and []=.
+        # @yield [Carrier, String, String] if an optional setter is provided,
```
mwear, line 34: Docs are out of date here too. The setter is no longer a block. Take a look at the docs for other injectors.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,111 @@
...
+        # Jaeger cannot be extracted from the carrier.
+        #
+        # @param [Carrier] carrier The carrier to get the header from.
+        # @param [Context] context The context to be updated with extracted
+        #   context
+        # @param [optional Callable] getter An optional callable that takes a
+        #   carrier and a key and returns the value associated with the key. If
+        #   omitted the default getter will be used which expects the carrier to
+        #   respond to [] and []=.
+        # @yield [Carrier, String] if an optional getter is provided, extract
```
mwear, line 47: Docs are out of date. The getter is no longer a block. Take a look at the docs for other extractors.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_injector.rb
```diff
@@ -0,0 +1,71 @@
...
+            encoded_value = CGI.escape(value)
+            setter.set(carrier, baggage_key, encoded_value)
+          end
+          carrier
+        end
+
+        private
+
+        def to_flags(context, span_context)
+          if span_context.trace_flags == TraceFlags::SAMPLED
```
mwear, line 58: This constant isn't found and needs further qualification. Its full path is `OpenTelemetry::Trace::TraceFlags`, but this will do:
```suggestion
          if span_context.trace_flags == Trace::TraceFlags::SAMPLED
```

mwear (changes requested):
This is almost here. I did an end to end test earlier and found a couple of issues. Two are related directly to this PR and one is an issue with the RackEnvGetter. I'll open another PR to fix it. Also, the docs are slightly out of date and also noted in the comments.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_extractor.rb
```diff
@@ -0,0 +1,111 @@
...
+
+        def set_baggage(carrier, context, getter)
+          baggage_key_prefix = 'uberctx-'
+          getter.keys(carrier).each do |carrier_key|
+            baggage_key = carrier_key.start_with?(baggage_key_prefix) && carrier_key[baggage_key_prefix.length..-1]
+            next unless baggage_key
+
+            raw_value = getter.get(carrier, carrier_key)
+            value = CGI.unescape(raw_value)
+            context = OpenTelemetry.baggage.set_value(
```
mwear, line 86: This will create a new baggage object for each value set. You can more efficiently make multiple modifications to baggage by using the `build` method. Basically, I think you can wrap this whole method in a:

```ruby
OpenTelemetry.baggage.build(context) do |b|
 ...
  b.set_value(baggage_key, value)
  ...
end
```

fhwang:
@mwear Ready for another review. Thanks!

mwear (approved):
Looks good! Thanks @fhwang!

PR #565, opened 2021-01-20
Add caller location where Faraday was called. (Draft: for Discussion)

mjobin-mdsol:
I am interested to see the location where a call is made within the metadata.

I thought of doing something like this. which would work for apps calling Faraday directly. But I think most people wraps Faraday into their own client.

would there be a better way to report the location within the application? 

would removing all lines that contains `gems` from the `caller` array be an option?

### DIFF
--- a/api/lib/opentelemetry/instrumentation/caller_locations.rb
+++ b/api/lib/opentelemetry/instrumentation/caller_locations.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The instrumentation CallerLocations contains a reusable function to add caller information to instrumentations
+    class CallerLocations
+      def self.detect_caller_location
+        if ENV['OTEL_EXPORTER_ADD_CALLER_LOCATION']
+          cleaned_up_trace = caller_locations.delete_if { |loc| loc.absolute_path.match('/.rvm/gems/') }
+          non_gem_first_line = cleaned_up_trace.last
+          {
+            'code.lineno' => non_gem_first_line.lineno,
+            'code.filepath' => non_gem_first_line.absolute_path,
+            'code.function' => non_gem_first_line.base_label
+          }
+        else
+          {}
+        end
+      end
+    end
+  end
+end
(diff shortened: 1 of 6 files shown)


johnnyshields:
Why would such functionality be specific to Faraday? It would be interesting to be able to add this for all traces...

fbogsany:
> Why would such functionality be specific to Faraday? It would be interesting to be able to add this for all traces...

There is a non-trivial overhead to calling `caller` on every span. We certainly wouldn't want to do this by default for every span. It may be worthwhile as an opt-in capability either for specific auto-instrumentation or globally for `kind=client` spans.

The spec contains [semantic conventions for source code](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/semantic_conventions/span-general.md#source-code-attributes). If we add this capability, we should follow those conventions.

mjobin-mdsol:
Thank you both for your replies,

@johnnyshields you're right this isn't specific to Faraday. I wanted to have a first go a possible implementation to open up the discussion. 

@fbogsany Good call, maybe this is something that could be enabled via an ENV var?

api/lib/opentelemetry/instrumentation/caller_locations.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    # The instrumentation CallerLocations contains a reusable function to add caller information to instrumentations
+    class CallerLocations
+      def self.detect_caller_location
+        if ENV['OTEL_EXPORTER_ADD_CALLER_LOCATION']
+          cleaned_up_trace = caller_locations.delete_if {|loc| loc.absolute_path.match('/.rvm/gems/')}
```
mjobin-mdsol, line 13: I'm not too sure what is the smarter/better way to do this, but I thought `'/.rvm/gems/'` could be an ENV var as well.

instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
```diff
@@ -80,6 +88,9 @@
       _(response.env.request_headers['Traceparent']).must_equal(
         "00-#{span.hex_trace_id}-#{span.hex_span_id}-01"
       )
+      _(span.attributes['code.filepath']).must_equal __FILE__
+      _(span.attributes['code.lineno']).must_equal 82
+      _(span.attributes['code.function']).must_equal '<top (required)>'
```
mjobin-mdsol, line 93: this is because the call is happening directly inside an `Rspec#it` block

mjobin-mdsol:
@fbogsany @johnnyshields updated my draft

mjobin-mdsol:
waiting on another round of comments before I fix the tests.
I would need to add an ENV var to Github Actions if I proceed with what I am currently thinking and mentioned in my above comment.

mwear:
I'd be curious to know what the use case is. Everything in observability comes at a cost to the application, and calling `caller` for each span comes at a fairly high cost. Adding configuration can mitigate some of the cost, but adds complexity and a long term maintenance burden.

PR #566, opened 2021-01-20
feat: bunny instrumentation

johanneswuerbach:
Basic instrumentation for [bunny](http://rubybunny.info/) based on the existing [ruby_kafka](https://github.com/open-telemetry/opentelemetry-ruby/tree/master/instrumentation/ruby_kafka) instrumentation.

The instrumentation provides:

* context propagation through rabbitmq stored in the message header
* span generation for publishing, manual consumption (pop) and consumption using a consumer (subscibe)

This basic instrumentation should also be sufficient when using higher level libraries based on bunny like https://github.com/jondot/sneakers


instrumentation/bunny/test/test_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+  attempts = 0
+  while attempts < max_attempts
+    return if yield
+
+    attempts += 1
+    raise error_message if attempts >= max_attempts
+
+    sleep retry_delay
+  end
+end
```
fbogsany, line 33: I don't see this used in the tests.
```suggestion
```

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/queue.rb
```diff
@@ -0,0 +1,36 @@
...
+                OpenTelemetry::Instrumentation::Bunny::PatchHelpers.with_process_span(tracer, delivery_info, properties) do
+                  yield delivery_info, properties, payload
+                end
+              else
+                OpenTelemetry::Instrumentation::Bunny::PatchHelpers.with_receive_span(tracer, delivery_info, properties) do
+                  [delivery_info, properties, payload]
+                end
+              end
+            end
+          end
```
fbogsany, line 25: Looking at the code for `Queue#pop`, I don't think this is quite correct. I think we want to instrument `process` here only if a block is provided, and we should instrument `Channel#basic_get` to generate the `receive` span. This will result in `enqueue -> receive` in all cases, and `enqueue -> receive -> process` if a user calls `queue.pop { ... }`.

The challenge here is how to propagate the `span_id` from the `receive` span to use as the parent of the `process` span. I'd love to use `OpenTelemetry::Context` for this, but we should really only use this in a block-structured fashion. One option is to set a thread-local variable to record the last `receive` span ID. An alternative is to mutate the `properties[:header]` by injecting the `receive` span context.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
```diff
@@ -0,0 +1,63 @@
...
+          [exchange, routing_key].compact.join('.')
+        end
+
+        def self.basic_attributes(exchange, routing_key)
+          attributes = {
+            'messaging.system' => 'rabbitmq',
+            'messaging.destination' => exchange,
+            'messaging.destination_kind' => 'topic',
+            'messaging.protocol' => 'AMQP',
+            'messaging.protocol_version' => '0.9.1'
```
fbogsany, line 55:
```suggestion
            'messaging.protocol_version' => ::Bunny.protocol_version
```

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
```diff
@@ -0,0 +1,63 @@
...
+
+        def self.destination_name(exchange, routing_key)
+          [exchange, routing_key].compact.join('.')
+        end
+
+        def self.basic_attributes(exchange, routing_key)
+          attributes = {
+            'messaging.system' => 'rabbitmq',
+            'messaging.destination' => exchange,
+            'messaging.destination_kind' => 'topic',
```
fbogsany, line 53: I think this should typically be `queue` for RabbitMQ. If a topic exchange is used, then this should be `topic`, but that doesn't look like the default use-case.

johanneswuerbach: I'm not entirely sure as even the default exchange model used in the example is topic based under the hood https://www.rabbitmq.com/tutorials/amqp-concepts.html#exchange-default (we also only use topic exchanges at my company)

I guess this could be set based on the exchange type https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/exchange.rb#L22 `queue` for direct, `topic` otherwise?

fbogsany: > I guess this could be set based on the exchange type https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/exchange.rb#L22 `queue` for direct, `topic` otherwise?

👍

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/queue.rb
johanneswuerbach: That makes sense, I used header propagation as with consumers the thread receiving the message is different from the thread processing the message.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
johanneswuerbach: Fixed.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
johanneswuerbach: Fixed.

instrumentation/bunny/Gemfile
```diff
@@ -0,0 +1,19 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+gemspec
+
+gem 'opentelemetry-api', path: '../../api'
+
+group :test do
+  gem 'activesupport'
```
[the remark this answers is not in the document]
johanneswuerbach: Replaced with secure random from core.

johanneswuerbach:
Thanks for the feedback, I pushed a round of fixes which should address all your comments.

johanneswuerbach:
@fbogsany rebased and adopted to latest main changes. Let me know if you have any other comments.

mwear (approved):
LGTM. Did you want to take a final look @fbogsany or @robertlaurin?

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
```diff
@@ -0,0 +1,51 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Bunny
+      module Patches
+        # The Channel module contains the instrumentation patch the Channel#basic_get, Channel#basic_publish and Channel#handle_frameset methods
+        module Channel
+          def basic_get(queue, opts = { manual_ack: false })
+            delivery_info, properties, payload = super
```
robertlaurin, line 14: If `super` [raises](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/channel.rb#L596) here do we not want this event captured in the trace?

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
```diff
@@ -0,0 +1,51 @@
...
+          end
+
+          # This method is called when rabbitmq pushes messages to subscribed consumers
+          def handle_frameset(basic_deliver, properties, content)
+            OpenTelemetry::Instrumentation::Bunny::PatchHelpers.with_receive_span(self, tracer, basic_deliver, properties) do
+              properties[:headers] ||= {}
+              OpenTelemetry.propagation.inject(properties[:headers])
+            end
+
+            super
```
robertlaurin, line 39: Similar to the above, it looks like a [uncaught_exception_handler](https://github.com/ruby-amqp/bunny/blob/master/lib/bunny/channel.rb#L1708) could be provided that raises the exception, in which case we would not capture that event here either.

Not being familiar with rabbitmq, is this call to `super` intentionally not being wrapped with the trace block for any particular reason?

instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patches/queue_test.rb
```diff
@@ -0,0 +1,64 @@
...
+describe OpenTelemetry::Instrumentation::Bunny::Patches::Queue do
+  let(:instrumentation) { OpenTelemetry::Instrumentation::Bunny::Instrumentation.instance }
+  let(:exporter) { EXPORTER }
+  let(:spans) { exporter.finished_spans }
+
+  let(:url) { ENV.fetch(' RABBITMQ_URL') { 'amqp://guest:guest@rabbitmq:5672' } }
+  let(:bunny) { Bunny.new(url) }
+  let(:topic) { "topic-#{SecureRandom.uuid}" }
+  let(:channel) { bunny.create_channel }
+  # let(:exchange) { channel.topic(topic, :auto_delete => true) }
```
robertlaurin, line 22: Can probably remove this right?

robertlaurin (commented):
Looks good, just a couple questions.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
```diff
@@ -0,0 +1,51 @@
...
+          end
```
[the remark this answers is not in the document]
fbogsany: Hmm, this is interesting because you're also wrapping the `Consumer.call`. Maybe we should just not wrap `handle_frameset`? We don't *have* to have a `receive` span in every case - the `process` span is much more interesting.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/queue.rb
```diff
@@ -0,0 +1,32 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Bunny
+      module Patches
+        # The Queue module contains the instrumentation patch the Queue#pop method.
+        module Queue
```
fbogsany, line 12: Is `Queue` typically used directly by clients or is it only used internally? It feels like we're going to end up with nested `process` spans in some cases, but I may be misunderstanding the gem.

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/queue.rb
```diff
@@ -0,0 +1,32 @@
...
+          def pop(opts = { manual_ack: false }, &block)
+            super do |delivery_info, properties, payload|
+              return unless block
```
fbogsany, line 15: I might be mistaken, but I think you want:
```suggestion
            return super unless block
            super do |delivery_info, properties, payload|
```

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
```diff
@@ -0,0 +1,83 @@
...
+        end
+
+        def self.with_process_span(channel, tracer, delivery_info, properties, &block)
+          destination = destination_name(delivery_info[:exchange], delivery_info[:routing_key])
+          parent_context, links = extract_context(properties[:headers])
+
+          OpenTelemetry::Context.with_current(parent_context) do
+            tracer.in_span("#{destination} process", links: links, kind: :consumer, &block)
+          end
+        end
```
fbogsany, line 41: There are some corner cases where `delivery_info` and `properties` can both be `nil` in both these methods. We should make sure we handle those safely.

fbogsany (changes requested):

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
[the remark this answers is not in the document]
johanneswuerbach: The handle_frameset submits the message to a worker pool, so there is a time between receiving the message and eventually processing it in a consumer thread.

Should that time be part of process or receive or not be tracked at all?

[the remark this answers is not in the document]
johanneswuerbach: The handle_frameset submits the message to a worker pool, so there is a time between receiving the message and eventually processing it in a consumer thread.

Should that time be part of process or receive or not be tracked at all?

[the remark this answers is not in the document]
johanneswuerbach: The handle_frameset submits the message to a worker pool, so there is a time between receiving the message and eventually processing it in a consumer thread.

Should that time be part of process or receive or not be tracked at all?

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/queue.rb
johanneswuerbach: `.pop` can be used by clients http://rubybunny.info/articles/queues.html#fetching_messages_when_needed_pull_api

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patches/channel.rb
johanneswuerbach: `super` only enqueues the message into the internal worker pool in this case and the actual consumer call is wrapped into a process span, which should handle the exception in this case.

johanneswuerbach:
Thank you for the all the feedback, I made a couple of changes to be more spec compliant around receive vs. process.

Now the push & pull api are modelled as https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#batch-receiving.

`basic_get`, used by `Queue.pop` starts a receive span when the receiving is started and in queue pop we link that span to the process span if a block is defined.

`handle_frameset` is doing the same before submitting the message into internal consumer pool, where it will eventually be processed by a consumer. Technically the receive should already wrap https://github.com/ruby-amqp/bunny/blob/0b5a6f2778f64f3c4bdf8b3063c6bdcbecfc8123/lib/bunny/reader_loop.rb#L77, but I wasn't sure how to integrate there so receive now only starts once the message was read from the socket.

johanneswuerbach:
@fbogsany could I get another review? :-)

fbogsany:
🤦 sorry! 👀 now.

johanneswuerbach:
@fbogsany friendly ping :-)

fbogsany:
🤦 👀

fbogsany (approved):

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Bunny
+      VERSION = '0.16.0'
```
fbogsany, line 10:
```suggestion
      VERSION = '0.17.0'
```

fbogsany:
Thanks @johanneswuerbach - sorry this took so long to review and merge 😞

johanneswuerbach:
No worries, we are all busy people. Thank you for working on this 🙇

instrumentation/bunny/lib/opentelemetry/instrumentation/bunny/patch_helpers.rb
```diff
@@ -0,0 +1,94 @@
...
+        def self.destination_name(exchange, routing_key)
+          [exchange, routing_key].compact.join('.')
+        end
+
+        def self.extract_context(properties)
+          # use the receive span as parent context
+          parent_context = OpenTelemetry.propagation.extract(properties[:tracer_receive_headers])
+
+          # link to the producer context
+          producer_context = OpenTelemetry.propagation.extract(properties[:headers])
```
indrekj, line 40: @fbogsany shouldn't this be `.extract(properties[:headers] || {})`?

If I have two services A & B. The first service is instrumented but the second one is not, then I get this exception in service A:
```
Uncaught exception from consumer #<Bunny::Consumer:70980 @channel_id=2 @queue=amq.gen-p2ZzQp6h4r6QWdCBIC4Pgg> @consumer_tag=bunny-1629126993000-738472400850>: #<NoMethodError: undefined method `[]' for nil:NilClass> @ /home/indrek/.rvm/gems/ruby-2.7.2@admin/gems/opentelemetry-api-1.0.0.rc2/lib/opentelemetry/context/propagation/text_map_getter.rb:16:in `get'
```

I think that's because `properties[:headers]` is `nil` when receiving a message from a service that is not using OTEL.

johanneswuerbach: Yes, that seems possible. Could you create a separate issue for this?

PR #568, opened 2021-01-21
fix: conditionally append path to collector endpoint

mwear:
A recent change to the otlp exporter unconditionally appends `/v1/trace` to the provided endpoint. This prevents the possibility of exporting spans to any other path using in-code configuration. This was likely due to a recent change in the spec that reads:

> Target to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, and MAY contain a port and path. A scheme of https indicates a secure connection. When using OTEL_EXPORTER_OTLP_ENDPOINT with OTLP/HTTP, exporters SHOULD follow the collector convention of appending the version and signal to the path (e.g. v1/traces or v1/metrics). The per-signal endpoint configuration options take precedence and can be used to override this behavior.

There is a general `OTLP_EXPORTER_OTLP_ENDPOINT` configuration that is meant to be used for both spans and metrics. OTLP/HTTP requires a path, so when the general configuration is used, a path must be appended to it. However, more specific configuration such as `OTLP_EXPORTER_OTLP_SPAN_ENDPOINT` should take precedence and a path should not be appended. The OTLP exporter that we have is a trace exporter and the `endpoint` argument is a span endpoint, and should not have `/v1/trace` appended to it by default. This PR makes that change, but still respects the more general configuration, which is currently only settable by environment variable.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -31,7 +31,7 @@ module OpenTelemetry
         WRITE_TIMEOUT_SUPPORTED = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
         private_constant(:KEEP_ALIVE_TIMEOUT, :RETRY_COUNT, :WRITE_TIMEOUT_SUPPORTED)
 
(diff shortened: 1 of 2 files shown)


fbogsany (approved):

ericmustin:
just to be clear here:
- don't suppply path for otlp_endpoint, 
- do supply path for span_endpoint,
- span takes precedence over otlp

do i have this right?

jfeltesse-mdsol:
Shouldn't the default path be `/v1/traces`? https://github.com/open-telemetry/opentelemetry-collector/pull/1979

fbogsany:
> Shouldn't the default path be `/v1/traces`? [open-telemetry/opentelemetry-collector#1979](https://github.com/open-telemetry/opentelemetry-collector/pull/1979)

Fixed in #570

PR #573, opened 2021-01-22
fix: Include http.status_text only if reason_phrase is present in the response

acronin-mdsol:
We noticed the error `OpenTelemetry error: invalid attribute value type NilClass` appearing frequently in our logs after the addition of Faraday instrumentation to our project.  This fix will skip the addition of attribute if it is nil in the response.

### DIFF
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
@@ -50,7 +50,7 @@ module OpenTelemetry
 
           def trace_response(span, response)
             span.set_attribute('http.status_code', response.status)
-            span.set_attribute('http.status_text', response.reason_phrase)
+            span.set_attribute('http.status_text', response.reason_phrase) if response.reason_phrase
             span.status = OpenTelemetry::Trace::Status.http_to_status(
               response.status
             )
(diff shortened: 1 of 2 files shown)


robertlaurin (commented):
Hey, thanks for the patch.

We should add a test here to prevent future regressions around this attribute, then it's ✅ from me.

You will also need to sign the CLA before we can merge in any changes.

fbogsany (approved):

acronin-mdsol:
@robertlaurin I'm having some trouble getting a quick test to work for this (probably something silly). Before I dig in, are you opposed to testing private methods using `send` ? If thats acceptable, I'll push up my current test. If not, I'll keep going

robertlaurin:
>I'm having some trouble getting a quick test to work for this (probably something silly). Before I dig in, are you opposed to testing private methods using send ? If thats acceptable, I'll push up my current test. If not, I'll keep going

I've definitely made use of `send` in tests myself, try to avoid it when possible.

I just ran the tests locally for faraday and noticed that its emitting the logs there.  I think instead of testing that the logs are not emitted which can be tricky. 

What if we test the positive scenario that asserts this field gets set when it is present?

```rb
    it 'it sets the reason phrase' do
      stub_request(:any, 'example.com').to_return(status: [500, "Internal Server Error"])
      ::Faraday.new('http://example.com').get('/')
      _(span.attributes['http.status_text']).must_equal 'Internal Server Error'
    end
```

After that we can call it good, don't want to be too nitpicky over a one liner such as this one.

acronin-mdsol:
Perfect, makes sense. Thanks for that

Now just waiting to hear back about signing the CLA.

robertlaurin (approved):

fbogsany:
@acronin-mdsol have you made any progress with signing the CLA? We'd like to get this fix merged, and we'd like you to get the credit for it. If you are unable to sign the CLA, could you please let us know so we can fix this in another PR?

acronin-mdsol:
hi @fbogsany sorry for the delay on this. I am just going to close this PR and open one from my personal account to sign the CLA is that's alright.

PR #574, opened 2021-01-22
feat: Instrument sidekiq background work

robertlaurin:
This PR instruments some very noisy spans and disables them by default.

### DIFF
--- a/instrumentation/sidekiq/.rubocop.yml
+++ b/instrumentation/sidekiq/.rubocop.yml
@@ -14,6 +14,10 @@ Metrics/MethodLength:
   Max: 20
 Metrics/ParameterLists:
   Enabled: false
+Style/AsciiComments:
+  Enabled: false
+Naming/AsciiIdentifiers:
+  Enabled: false
 Naming/FileName:
   Exclude:
     - "lib/opentelemetry-instrumentation-sidekiq.rb"
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
@@ -10,6 +10,8 @@ module OpenTelemetry
       module Patches
         # The Launcher module contains the instrumentation for the Sidekiq heartbeat
         module Launcher
+          private unless $TESTING
+
           def ❤ # rubocop:disable Naming/MethodName
             if config[:trace_launcher_heartbeat]
               tracer.in_span('Sidekiq::Launcher#heartbeat') { super }
@@ -17,8 +19,6 @@ module OpenTelemetry
               untraced { super }
             end
           end
-
-          private
 
           def tracer
             Sidekiq::Instrumentation.instance.tracer
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
@@ -10,7 +10,7 @@ module OpenTelemetry
       module Patches
         # The Launcher module contains the instrumentation for the Sidekiq heartbeat
         module Launcher
-          private unless $TESTING
+          private
 
           def ❤ # rubocop:disable Naming/MethodName
             if config[:trace_launcher_heartbeat]
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
@@ -1,0 +1,37 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Sidekiq
+      module Patches
+        module Launcher
+          def ❤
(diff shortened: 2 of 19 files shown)


instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
```diff
@@ -0,0 +1,37 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Sidekiq
+      module Patches
+        module Launcher
+          def ❤
```
robertlaurin, line 12: The method signature changes in earlier versions of `sidekiq 4.x.x`, I bumped the appraisal to be the latest `4.2.x` release. 

https://github.com/mperham/sidekiq/blob/v4.1.3/lib/sidekiq/launcher.rb#L74

instrumentation/sidekiq/test/opentelemetry/instrumentation/sidekiq/patches/launcher_test.rb
```diff
@@ -0,0 +1,53 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+require 'opentelemetry-instrumentation-redis'
+require 'fakeredis/minitest'
```
robertlaurin, line 10: Including the redis instrumentation and the fakeredis allows for testing with a bit more confidence.

instrumentation/sidekiq/.rubocop.yml
```diff
@@ -14,6 +14,10 @@ Metrics/MethodLength:
   Max: 20
 Metrics/ParameterLists:
   Enabled: false
+Style/AsciiComments:
```
robertlaurin, line 17: Disabled these for the unicode heart named method.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
```diff
@@ -0,0 +1,38 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Sidekiq
+      module Patches
+        # The Launcher module contains the instrumentation for the Sidekiq heartbeat
+        module Launcher
+          def ❤ # rubocop:disable Naming/MethodName
+            if config[:trace_launcher_heartbeat]
```
robertlaurin, line 14: - Do we want to establish some sort of naming convention for config options?
- I'm in the opinion that this should be off by default to provide a better use experience, this will generate a non trivial amount of spans.

robertlaurin:
I think this PR has some high level overlap with what is going on in [this issue](https://github.com/open-telemetry/opentelemetry-ruby/issues).  

In the case of the issue it's likely delayed job checked the database which has the mysql calls instrumented, but there is no parent span.  And in the case of this PR redis is instrumented but sidekiq is doing background work that is not instrumented, resulting in a lot of redis spans being generated with little to no additional context.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/launcher.rb
fbogsany: > * Do we want to establish some sort of naming convention for config options?

We don't have many config options yet, so you're in early & get to establish the conventions 😄 

> * I'm in the opinion that this should be off by default to provide a better use experience, this will generate a non trivial amount of spans.

👍

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/poller.rb
```diff
@@ -0,0 +1,46 @@
...
+            Sidekiq::Instrumentation.instance.tracer
+          end
+
+          def config
+            Sidekiq::Instrumentation.instance.config
+          end
+
+          def untraced
+            OpenTelemetry::Trace.with_span(OpenTelemetry::Trace::Span.new) { yield }
+          end
```
fbogsany, line 41: Is it worth extracting this to a common place?

robertlaurin: I think we could push this into the common util gem?

instrumentation/sidekiq/test/helpers/mock_loader.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'sidekiq/cli'
+require 'sidekiq/launcher'
+
+class MockLoader
```
robertlaurin, line 10: This class is mimicking some of the behaviour found here https://github.com/mperham/sidekiq/blob/master/bin/sidekiqload

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/poller.rb
robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/pull/580

mwear (approved):
LGTM

fbogsany (approved):

PR #580, opened 2021-01-27
feat: add untraced wrapper to common utils

robertlaurin:
Makes the `untraced` method available as part of common utils.


ericmustin (approved):
seems reasonable to me

mwear (approved):

PR #589, opened 2021-01-31
feat!: replace getter and setter callables and remove rack specific propagators

mwear:
When looking over #564 earlier this week, I realized that we do not have a general way to iterate over a carrier's keys. We haven't missed this, because before the Jaeger propagator we didn't have this need. When we originally implemented the propagation API, the getters and setters passed to inject and extract were specified as functions. The spec has since evolved to recommend that they be implemented as objects, especially the getter, with `get` and `keys`methods. The setter is less useful, but not useless, and has a `set` method.

Below are links to the spec for more details
* [getter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#getter-argument)
* [setter](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#setter-argument)

I started off experimenting with a `TextMapGetter` and a `RackEnvGetter` and found that the getter abstraction is pretty useful. The `RackEnvGetter` provides functionality to access keys in a carrier agnostic way. So, we can use it to lookup `traceparent` instead of `HTTP_TRACEPARENT` for example. With this abstraction, I was able to remove all of the Rack specific injectors and extractors, as I continued the refactor, I was able to remove the `http` and `text` distinction from the top level propagation object. Now, OpenTelemetry.propagation.extract will extract using a `TextMapGetter` by default. If you want to extract from a Rack environment you'd use `OpenTelemetry.propagation.extract(env, Context.current, Propagation.rack_env_getter)`. It's a little more to type, but it's only used a few places. This makes the Ruby context API much closer to the spec and less of a special snowflake.

Summary the changes, this PR:

* Introduces two Getters, with `get` and `keys` methods
  * TextMapGetter
  * RackEnvGetter
* Introduces a TextMapSetter with a `set` method
* Removes all rack specific injectors and extractors
  * This simplifies, but also adds some breaking changes to the configurator
* Removes the `http` and `text` paths from `OpenTelemetry.propagation`. You go directly to `inject` or `extract` now (e.g. `OpenTelemetry.propagation.inject(...)`
  * Updated relevant call sites in instrumentation and examples

This is a pretty big change, but I think it is a good improvement in terms of usability and spec compliance. I also think this will help smooth over some things in the Jaeger propagator PR (#564).


api/lib/opentelemetry/context/propagation/propagator.rb
```diff
@@ -46,14 +46,14 @@ def inject(carrier, context = Context.current, &setter)
...
-        #   which expects the carrier to respond to [] and []=.
+        # @param [optional Getter] getter If the optional getter is provided, it
+        #   will be used to read the header from the carrier, otherwise the default
+        #   getter will be used.
         #
         # @return [Context] a new context updated with state extracted from the
         #   carrier
-        def extract(carrier, context = Context.current, &getter)
-          @extractor.extract(carrier, context, &getter)
+        def extract(carrier, context = Context.current, getter = Context::Propagation.text_map_getter)
```
fbogsany, line 55: Since we're changing the signature here, can we switch to keyword arguments for the optional args? This will simplify things for users of the Rack getter:
```ruby
    context = OpenTelemetry.propagation.extract(env, getter: OpenTelemetry::Context::Propagation.rack_env_getter)
# vs
    context = OpenTelemetry.propagation.extract(
       env,
       OpenTelemetry::Context.current,
       OpenTelemetry::Context::Propagation.rack_env_getter
     )
```
```suggestion
        def extract(carrier, context: Context.current, getter: Context::Propagation.text_map_getter)
```
If we do this, we should make the same change to `inject`.

api/test/opentelemetry/context/propagation/rack_env_getter_test.rb
```diff
@@ -0,0 +1,37 @@
...
+  let(:carrier) { { 'HTTP_TRACEPARENT' => 'tp', 'HTTP_TRACESTATE' => 'ts', 'HTTP_X_SOURCE_ID' => '123' } }
+
+  describe '#get' do
+    it 'reads key from carrier' do
+      _(getter.get(carrier, 'traceparent')).must_equal('tp')
+      _(getter.get(carrier, 'tracestate')).must_equal('ts')
+      _(getter.get(carrier, 'x-source-id')).must_equal('123')
+    end
+
+    it 'returns nil for non-existant key' do
```
fbogsany, line 23:
```suggestion
    it 'returns nil for non-existent key' do
```

fbogsany:
Overall, this is a good change. Since this removes the Rack-specific extractors, maybe we should go a step further and remove the separation of injectors and extractors, and instead unify them in the `Propagator`. One thing we have not implemented yet is the [`Fields`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#fields) operation, which becomes much simpler when we have a B3 Single propagator and a B3 Multi propagator rather than the `Propagator` initialized with an injector and an extractor.

mwear:
> Overall, this is a good change. Since this removes the Rack-specific extractors, maybe we should go a step further and remove the separation of injectors and extractors, and instead unify them in the `Propagator`. One thing we have not implemented yet is the [`Fields`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/api-propagators.md#fields) operation, which becomes much simpler when we have a B3 Single propagator and a B3 Multi propagator rather than the `Propagator` initialized with an injector and an extractor.

We can discuss this, but I'd recommend handling it as a separate PR. When we originally implemented the context prop system the specification had injectors and extractors as separate concerns. Ours are still separate in part because of this, but also because there are some advantages to this design. Namely, an app can have asymmetric formats for inject and extract, also, the order of injection and extraction can be controlled. It might turn out that these advantages are only theoretical and not actually useful in practice. We can still implement fields with separate injectors and extractors, although, I agree it'd be easier for combined propagators. I will be completely honest in that I don't understand the fields use case, and I know that is completely broken for propagators that use header-per-key semantics for baggage (such as Jaeger).

fbogsany:
> I'd recommend handling it as a separate PR.

👍

fbogsany (approved):

fbogsany:
I 👍 this, but I'd like to get #592 merged and a new release cut before we merge this (that's an important bug fix for 0.13.0).

PR #594, opened 2021-02-01
feat: Add untraced endpoints config to rack middleware

robertlaurin:
Adds the ability to specific certain paths to be untraced.  The example used in test is a health check ping that you may not want instrumented.

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -49,9 +49,15 @@ module OpenTelemetry
 
           def initialize(app)
             @app = app
+            @untraced_endpoints = config[:untraced_endpoints].is_a?(Array) ? config[:untraced_endpoints] : []
           end
 
           def call(env) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+            if untraced_request?(env)
+              OpenTelemetry::Common::Utilities.untraced do
+                return @app.call(env)
+              end
+            end
             original_env = env.dup
             extracted_context = OpenTelemetry.propagation.extract(
               env,
@@ -78,6 +84,10 @@ module OpenTelemetry
           end
 
           private
+
+          def untraced_request?(env)
+            @untraced_endpoints.include?(env['PATH_INFO'])
+          end
 
           # return Context with the frontend span as the current span
           def create_frontend_span(env, extracted_context)
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
@@ -23,6 +23,7 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware do
   let(:rack_builder) { Rack::Builder.new }
 
   let(:exporter) { EXPORTER }
+  let(:finished_spans) { exporter.finished_spans }
   let(:first_span) { exporter.finished_spans.first }
 
   let(:default_config) { {} }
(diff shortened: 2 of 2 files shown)


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -49,9 +49,16 @@ def clear_cached_config
 
           def initialize(app)
             @app = app
+            @untraced_endpoints = config[:untraced_endpoints].is_a?(Array) ? config[:untraced_endpoints] : []
```
fbogsany, line 52: Some sort of warning here would be nice, if `config[:untraced_endpoints]` is present and not an array.

Of course, there's a question of whether we want to be so prescriptive about the type -- it seems like anything supporting `.include?` should be fine, e.g. if someone wanted to wrap a regex.

robertlaurin: Anything supporting `include?` seems potentially risky.  If for example someone just passed in `/`, most endpoints would go untraced.

What if we expected a lambda to be provided that accepts the `env`, and that determined if it should be untraced or not?  That seems a bit more flexible as someone could making tracing decisions based off of headers, path, etc...

fbogsany: > If for example someone just passed in `/`, most endpoints would go untraced.

That's not entirely true 😄 
```
irb(main):001:0> '/'.include? '/foo'
=> false
```

There's a trade-off between flexibility and complexity of use. We should at least make the most common use-cases easy (e.g. health checks, for us). Path seems like the most common thing to want to match. I didn't see anything that would let us "duck type" regexes and arrays 🤷 .

I'm not terribly fussed - if you want to punt the more complex use-cases and can support that without changing the interface, I'll 👍

robertlaurin: >That's not entirely true 😄

I may or may not have been thinking of it in reverse `path.include?('/')` 🤦 

>There's a trade-off between flexibility and complexity of use. We should at least make the most common use-cases easy (e.g. health checks, for us).

I agree, definitely want to keep it simple.  Trying to think ahead a little bit I suspect we'll want to support a few use-cases.

- Allow the user to specify one or many endpoints
- Allow the user to supply those endpoints via env var
- Ability to make tracing decisions off of something other than the endpoint (headers)

I think I can come up with a simple interface that supports these scenarios.

robertlaurin: >Some sort of warning here would be nice, if config[:untraced_endpoints] is present and not an array.

Also regarding this comment, I've been thinking of this a little bit more and wondering if we should have some config checking on install/initialization instead of dealing with it at runtime.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -76,6 +83,10 @@ def call(env) # rubocop:disable Metrics/AbcSize
 
           private
 
+          def untraced_endpoint?(path)
```
arielvalentin, line 86: Summarizing a bit of our discussion here from the SIG meting here:

I expressed interest in some flexibility here that would allow users to deal with more complex `PATH_INFO`, e.g.
 
- Allowing for regular expressions to so that users can match more complex resource oriented paths
- A lambda/proc with custom logic

I believe @robertlaurin brought up that in addition to `PATH_INFO` we may in the future want to inspect other parts of the request, e.g. an HTTP Header.

I think the consensus was that starting with supporting string paths would be fine for covering the use case of disabling tracing for static health endpoints. 

Does that sound accurate to y'all @robertlaurin @fbogsany @mwear?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: To keep this interface simple, I'd like to keep this as expecting and only accepting an array of string values.  I'll do the config validation as a separate PR from this one.

fbogsany (approved):

PR #596, opened 2021-02-02
docs: updated gem name to match gemspec

angeliski:
Reading the docs I caught an error in the README, so I did a fix to match the published gem.


robertlaurin (approved):
Thanks for correction!

fbogsany (approved):

PR #599, opened 2021-02-03
feat: Add instrumentation config validation

robertlaurin:
During this weeks SIG I proposed adding a validation step of sorts to create a more structured interface for our instrumentation configuration.

This idea sparked from this comment chain https://github.com/open-telemetry/opentelemetry-ruby/pull/594#discussion_r568753846

### DIFF
--- a/api/lib/opentelemetry/instrumentation/base.rb
+++ b/api/lib/opentelemetry/instrumentation/base.rb
@@ -245,7 +245,7 @@ module OpenTelemetry
       # Invalid configuration values are logged, and replaced by the default.
       #
       # @param [Hash] config The user supplied config hash
-      def config_options(config)
+      def config_options(config) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         @options ||= {}
         config ||= {}
 
--- a/api/lib/opentelemetry/instrumentation/base.rb
+++ b/api/lib/opentelemetry/instrumentation/base.rb
@@ -244,14 +244,13 @@ module OpenTelemetry
       # Unknown configuration keys are not included in the final config hash.
       # Invalid configuration values are logged, and replaced by the default.
       #
-      # @param [Hash] config The user supplied config hash
-      def config_options(config) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+      # @param [Hash] user_config The user supplied configuration hash
+      def config_options(user_config) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         @options ||= {}
-        config ||= {}
-
-        @options.each_with_object({}) do |option, h|
+        user_config ||= {}
+        validated_config = @options.each_with_object({}) do |option, h|
           option_name = option[:name]
-          config_value = config[option_name]
+          config_value = user_config[option_name]
 
           value = if config_value.nil?
                     option[:default]
@@ -270,6 +269,13 @@ module OpenTelemetry
           OpenTelemetry.handle_error(exception: e, message: "Instrumentation #{name} unexpected configuration error")
           h[option_name] = option[:default]
         end
+
+        dropped_config_keys = user_config.keys - validated_config.keys
+        unless dropped_config_keys.empty?
+          OpenTelemetry.logger.warn("Instrumentation #{name} ignored the following unknown configuration options #{dropped_config_keys}")
+        end
+
+        validated_config
       end
 
       # Checks to see if this instrumentation is enabled by env var. By convention, the
--- a/api/lib/opentelemetry/instrumentation/base.rb
+++ b/api/lib/opentelemetry/instrumentation/base.rb
@@ -271,9 +271,7 @@ module OpenTelemetry
         end
 
         dropped_config_keys = user_config.keys - validated_config.keys
-        unless dropped_config_keys.empty?
-          OpenTelemetry.logger.warn("Instrumentation #{name} ignored the following unknown configuration options #{dropped_config_keys}")
-        end
+        OpenTelemetry.logger.warn("Instrumentation #{name} ignored the following unknown configuration options #{dropped_config_keys}") unless dropped_config_keys.empty?
 
         validated_config
       end
--- a/api/lib/opentelemetry/instrumentation/base.rb
+++ b/api/lib/opentelemetry/instrumentation/base.rb
@@ -129,9 +129,12 @@ module OpenTelemetry
           @compatible_blk = blk
         end
 
-        # @param [String] name
-        # @param default The default value
-        # @param [Callable] validate A callable the returns a boolean
+        # The option method is used to define default configuration options
+        # for the instrumentation library. It requires a name, default value,
+        # and a validation callable to be provided.
+        # @param [String] name The name of the configuration option
+        # @param default The default value to be used, or to used if validation fails
+        # @param [Callable] validate A callable that validates the option value and returns a boolean
         def option(name, default:, validate:)
           raise ArgumentError, 'validate must be a callable' unless validate.respond_to?(:call)
 
--- a/api/lib/opentelemetry/instrumentation/base.rb
+++ b/api/lib/opentelemetry/instrumentation/base.rb
@@ -62,6 +62,14 @@ module OpenTelemetry
     class Base # rubocop:disable Metrics/ClassLength
       class << self
         NAME_REGEX = /^(?:(?<namespace>[a-zA-Z0-9_:]+):{2})?(?<classname>[a-zA-Z0-9_]+)$/.freeze
+        VALIDATORS = {
+          array: ->(v) { v.is_a?(Array) },
+          boolean: ->(v) { v.is_a?(TrueClass) || v.is_a?(FalseClass) },
+          callable: ->(v) { v.respond_to?(:call) },
+          integer: ->(v) { v.is_a?(Integer) },
+          string: ->(v) { v.is_a?(String) }
+        }.freeze
+
         private_constant :NAME_REGEX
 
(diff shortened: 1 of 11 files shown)


robertlaurin:
Hey @mwear & @ericmustin wanted to get some feedback on this approach before I keep going with it.  I used the graphql instrumentation as an example of what the configuration options would look like.

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -180,12 +186,31 @@ def install(config = {})
...
+      def config_options(config_hash)
+        config_hash ||= {}
+
+        @options.each_with_object({}) do |option, h|
+          option_name = option[:name]
+          config_hash_value = config_hash[option_name]
+
+          value = if config_hash_value.nil? && option[:allow_nil]
+            nil
+          elsif option[:validate].call(config_hash_value)
```
ericmustin, line 204: In the same way we're providing a validation of config options here, we  shouuld probably validate that this is callable, or a proc, or something to that effect? Wdyt?

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -180,12 +186,31 @@ def install(config = {})
...
+            config_hash_value
```
ericmustin, line 205: So is the idea that the validation proc itself would log a warning? or, should we attempt to log something uniform here?

ericmustin (commented):
I generally think having a `validate` option is reasonable, the only thing i'd say is that we probably want to have some expected failure behavior here, and try to standardize that, rather than letting the validation proc itself be in charge (imo validation proc should just return true/false).

Specifically, i think have a `warn` level log message with
- the integration name, 
- option name, 
- invalid value, 
- and fallback value

Would be appropriate and have enough information to let users fix their inputs without spamming their logs. that's just my two cents tho, my real concern here is that leaving it up to the specific integration's validation proc will create logs at different levels and different amounts of detail over time as integrations get added.


Additionally, from an implementation standpoint, I think we'd need to be defensive and make sure

1. the validate option itself is a proc
2. we catch/swallow any exceptions that get raised within the validation proc.

api/lib/opentelemetry/instrumentation/base.rb
robertlaurin: I think that's a reasonable thing to do, tests would only catch that error if there is test explicitly setting that config option.

api/lib/opentelemetry/instrumentation/base.rb
robertlaurin: >imo validation proc should just return true/false
I agree with what you mentioned above that the proc should only return true/false.  I think that we should only be logging an error/warning/info (to be discussed) in the event that the validation fails on user provided config options.  

When no option is provided and the default is used I think that's a quiet happy path.

robertlaurin: > should we attempt to log something uniform here?

I believe we should be responsible for the logging and not the proc.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -22,6 +22,14 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           defined?(::Rack)
         end
 
+        option :allowed_request_headers,  default: [],    validate: ->(v) { v.is_a?(Array) }
+        option :allowed_response_headers, default: [],    validate: ->(v) { v.is_a?(Array) }
+        option :application,              default: nil,   validate: ->(v) { v.respond_to?(:call) }
+        option :record_frontend_span,     default: false, validate: ->(v) { v.is_a?(TrueClass) || v.is_a?(FalseClass) }
+        option :retain_middleware_names,  default: false, validate: ->(v) { v }
```
fbogsany, line 29: What's the deciding factor for choosing
```ruby
default: false, validate: ->(v) { v.is_a?(TrueClass) || v.is_a?(FalseClass) }
```
vs
```ruby
default: false, validate: ->(v) { v }
```

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -225,6 +239,39 @@ def enabled?(config = nil)
 
       private
 
+      # The config_options method is responsible for validating that the user supplied
+      # config hash is valid.
+      # Unknown configuration keys are not included in the final config hash.
```
fbogsany, line 244: Should we log unknown configuration keys?

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -129,14 +129,27 @@ def compatible(&blk)
           @compatible_blk = blk
         end
 
+        # The option method is used to define default configuration options
+        # for the instrumentation library. It requires a name, default value,
+        # and a validation callable to be provided.
+        # @param [String] name The name of the configuration option
+        # @param default The default value to be used, or to used if validation fails
+        # @param [Callable] validate A callable that validates the option value and returns a boolean
```
fbogsany, line 137: We may want to relax this at some point and make the `validate` Callable optional, but I think it's fine to keep this restrictive for now.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
```diff
@@ -27,6 +27,12 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           gem_version >= MINIMUM_VERSION
         end
 
+        option :enable_job_class_span_names, default: false, validate: ->(v) { v.is_a?(TrueClass) || v.is_a?(FalseClass) }
```
fbogsany, line 30: Is this sufficient or do we need to explicitly test the class?
```suggestion
        option :enable_job_class_span_names, default: false, validate: ->(v) { v == true || v == false }
```

api/lib/opentelemetry/instrumentation/base.rb
robertlaurin: Yeah I think so, it seems like a helpful addition to anyone getting setup with an instrumentation library.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
robertlaurin: Ah, this was just some leftover.  Will make it consistent.

instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb
robertlaurin: I don't think we need to test the class.  I'll update it.

robertlaurin:
Made a few changes based on feedback, the only thing left from my perspective would be potentially adding some common validation helpers for booleans, integers, arrays, etc... 

I'm not sure where they would be best suited to exist, my first thought would be having them be in the `common` gem.

fbogsany (approved):
This LGTM. I'm inclined to 🚢 as is and refine it with helpers later (whether that's relaxing the `validate:` argument to accept symbols or packaging helpers somewhere). If we're going to accept symbols, then the helpers should be in the API gem. Otherwise they should be added to `opentelemetry-common`.

fbogsany:
Hmm, those Rubocop failures are annoying. I don't think we want to encourage ignoring them everywhere, so maybe we should add those helpers now 😄. I like the symbol approach:
```ruby
option :enable_job_class_span_names, default: false, validate: :boolean
```

ericmustin (approved):
🚀

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -59,9 +59,17 @@ module Instrumentation
     # convention for environment variable name is the library name, upcased with
     # '::' replaced by underscores, OPENTELEMETRY shortened to OTEL_{LANG}, and '_ENABLED' appended.
     # For example: OTEL_RUBY_INSTRUMENTATION_SINATRA_ENABLED = false.
-    class Base
+    class Base # rubocop:disable Metrics/ClassLength
       class << self
         NAME_REGEX = /^(?:(?<namespace>[a-zA-Z0-9_:]+):{2})?(?<classname>[a-zA-Z0-9_]+)$/.freeze
+        VALIDATORS = {
+          array: ->(v) { v.is_a?(Array) },
+          boolean: ->(v) { v.is_a?(TrueClass) || v.is_a?(FalseClass) },
```
fbogsany, line 67: _sigh_ I almost think it's worth adding a `# rubocop:disable` to this line, but not quite 🤷

api/lib/opentelemetry/instrumentation/base.rb
fbogsany: This comment should be expanded to include the built-in validators.

api/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -59,9 +59,17 @@ module Instrumentation
...
+          callable: ->(v) { v.respond_to?(:call) },
+          integer: ->(v) { v.is_a?(Integer) },
+          string: ->(v) { v.is_a?(String) }
+        }.freeze
+
         private_constant :NAME_REGEX
```
fbogsany, line 73:
```suggestion
        private_constant :NAME_REGEX, :VALIDATORS
```

robertlaurin:
@fbogsany got those two comments fixed up!

fbogsany (approved):

mwear (approved):
This is a big improvement! Thanks @robertlaurin!

Issue #595, opened 2021-02-02
Opentelemetry SDK - SimpleSpanProcessor error - undefined method export


mat-rumian:
**Description:**
Undefined method export for SimpleSpanProcessor appears.

**Steps:**
I've used [example application](https://github.com/JetBrains/sample_rails_app). 
1. Add in Gemfile:
  gem 'opentelemetry-sdk', '0.13.1'
  gem 'opentelemetry-instrumentation-all', '0.13.0'
  gem 'opentelemetry-exporter-otlp', '0.13.0'
2. Add in config/environment.rb
```
OpenTelemetry::SDK.configure do |c|
  c.use_all
  c.add_span_processor(
    OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
       exporter: OpenTelemetry::Exporter::OTLP::Exporter.new()
    )
  )
end
```
3. Run the [application](https://github.com/JetBrains/sample_rails_app#getting-started) 

**Result:**
```
[...]

PR #600, opened 2021-02-05
fix!: streamline processor pipeline

fbogsany:
This streamlines the SpanProcessor pipeline in the common case (a single non-noop processor), removing the (double) indirection through MultiSpanProcessor when only a single span processor is registered.

It also replaces the required `exporter:` keyword parameter to the BatchSpanProcessor initializer with a positional parameter. This aligns it with the SimpleSpanProcessor initializer and avoids a (probably common) mistake when switching between the BatchSpanProcessor and SimpleSpanProcessor. (e.g. #595).

### DIFF
--- a/exporter/jaeger/README.md
+++ b/exporter/jaeger/README.md
@@ -35,7 +35,7 @@ require 'opentelemetry/exporter/jaeger'
 OpenTelemetry::SDK.configure do |c|
   c.add_span_processor(
     OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(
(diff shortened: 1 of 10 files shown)


robertlaurin (approved):

johnnyshields:
This is a breaking change so please release it as a minor version bump 0.15.0

fbogsany:
> This is a breaking change so please release it as a minor version bump 0.15.0

The `!` in the PR title will trigger the bump automatically - see ["Use conventional commit messages"](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/CONTRIBUTING.md#use-conventional-commit-messages)

PR #601, opened 2021-02-08
feat: add otlp exporter hooks

robertlaurin:
The intention of this method is to allow users of the exporter to either
prepend or inherit from the class so that the exporter send_bytes
method can be traced, or wrapped with any other customer methods.

Two example use cases:
- Tracing the trace pipeline itself, this would allow for the an end to end tracing of the exporter/tracing pipeline
- In some applications we have a requirement of explicitly permitting URLs that are allowed to be communicated via HTTP.  In those applications we will need this hook to wrap the `send_byes` http call with an allow method that permits the network call.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -120,7 +120,7 @@ module OpenTelemetry
         # is to not trace these operations.
         #
         # An example use case would be to prepend a patch, or extend this class
-        # and override this methods behaviour to explicitly trace the. ttp request.
+        # and override this method's behaviour to explicitly trace the HTTP request.
         # This would allow you to trace your export pipeline.
         def around_request
           OpenTelemetry::Common::Utilities.untraced { yield }
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -41,18 +41,18 @@ module OpenTelemetry
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || compression == 'gzip'
           raise ArgumentError, 'headers must be comma-separated k=v pairs or a Hash' unless valid_headers?(headers)
 
-          uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+          @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
                   URI("#{endpoint}/v1/traces")
                 else
                   URI(endpoint)
                 end
 
-          @http = Net::HTTP.new(uri.host, uri.port)
-          @http.use_ssl = uri.scheme == 'https'
+          @http = Net::HTTP.new(@uri.host, @uri.port)
+          @http.use_ssl = @uri.scheme == 'https'
           @http.ca_file = certificate_file unless certificate_file.nil?
           @http.keep_alive_timeout = KEEP_ALIVE_TIMEOUT
 
-          @path = uri.path
+          @path = @uri.path
           @headers = case headers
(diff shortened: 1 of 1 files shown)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -116,11 +117,19 @@ def invalid_url?(url)
           true
         end
 
+        def send_bytes_hook
+          if !@send_bytes_hook.nil?
+            @send_bytes_hook.call { yield }
+          else
+            OpenTelemetry::Common::Utilities.untraced { yield }
+          end
+        end
```
fbogsany, line 126: I'm not sure whether this explicit hook argument is better than
```ruby
def trace_request
  OpenTelemetry::Common::Utilities.untraced { yield }
end
```
and then relying on subclassing or `prepend` to override the behaviour.

fbogsany: Or even just a boolean flag `should_trace_request: false` and then:
```ruby
def trace_request
  if @should_trace_request
    tracer.in_span('OTLP::Exporter.send_bytes') { yield }
  else
    OpenTelemetry::Common::Utilities.untraced { yield }
  end
end
```

fbogsany: The first option is more flexible, since it allows adding attributes.

robertlaurin: I'll take the first suggested approach here.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -41,26 +41,25 @@ def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || compression == 'gzip'
           raise ArgumentError, 'headers must be comma-separated k=v pairs or a Hash' unless valid_headers?(headers)
 
-          uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+          @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
```
robertlaurin, line 44: To allow the trace_request method to be patched to allow network calls, we need a way to access the URI that the exporter has been configured with.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -116,11 +115,15 @@ def invalid_url?(url)
           true
         end
 
+        def trace_request
```
fbogsany, line 118: Small suggestion to broaden expectations for use-cases, and to reduce the cognitive dissonance of `trace_request ... untraced`
```suggestion
        def around_request
```

It would also be really useful to document this method - while it is private, it is expected to be used as an extension point.

robertlaurin: Updated!

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -116,11 +115,22 @@ def invalid_url?(url)
           true
         end
 
+        # The around_request is a private method that provides an extension
+        # point for the exporters network calls. The default behaviour
+        # is to not trace these operations.
+        #
+        # An example use case would be to prepend a patch, or extend this class
+        # and override this methods behaviour to explicitly trace the. ttp request.
```
fbogsany, line 123:
```suggestion
        # and override this method's behaviour to explicitly trace the HTTP request.
```

fbogsany (approved):
Typo in comment - otherwise LGTM, thanks!

mwear (approved):

Issue #403, opened 2020-09-18
Implement OTEL_EXPORTER env var


fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/pull/943

rydzykje:
Will be happy to help with that :)

fbogsany:
Thanks @rydzykje - I've assigned this to you.

PR #604, opened 2021-02-10
fix: OTEL_TRACES_EXPORTER and OTEL_PROPAGATORS

### LINKED ISSUE #412: Implement OTEL_PROPAGATORS environment variable
See: https://github.com/open-telemetry/opentelemetry-specification/pull/962

Propagators to be used as a comma separated list.

Default is "tracecontext,baggage".

> Known values for OTEL_PROPAGATORS are: "tracecontext", "baggage", "b3", "jaeger"

fbogsany:
Fixes #412 and #403

Many of the supported propagators and exporters are required to be implemented external to the SDK, so this requires making assumptions about the structure of the external gems.

The `OTEL_TRACES_EXPORTER` env var is not especially useful for a couple of reasons:
1. The default span processor is not specified and there is no specified environment variable to select it. We defaulted to the `SimpleSpanProcessor` previously, and I've kept that default here, but that is rarely useful with any of the defined exporters.
2. We previously defaulted to `ConsoleSpanExporter`, but the specified default is OTLP, so we instead fall back to `ConsoleSpanExporter` if either an unsupported exporter is requested or a supported but unavailable exporter is requested. In either case, we log a warning, so the out-of-the-box experience is a little noisier.

It may make more sense to wrap any supported exporter in a BSP, and otherwise default to a `ConsoleSpanExporter` wrapped in a `SimpleSpanProcessor` 🤔 - I'll try that... That'll fix 1, but 2 will still be noisy.

### TODO
- [x] tests for new functionality
- [x] change the wrapping of exporters

### DIFF
--- a/sdk/lib/opentelemetry/sdk/configurator.rb
+++ b/sdk/lib/opentelemetry/sdk/configurator.rb
@@ -148,14 +148,41 @@ module OpenTelemetry
       end
 
       def default_span_processor
-        Trace::Export::SimpleSpanProcessor.new(
-          Trace::Export::ConsoleSpanExporter.new
-        )
+        Trace::Export::SimpleSpanProcessor.new(configured_exporter)
       end
 
-      def configure_propagation
-        OpenTelemetry.propagation = create_propagator(@injectors || default_injectors,
-                                                      @extractors || default_extractors)
+      def configured_exporter
+        exporter = ENV.fetch('OTEL_TRACES_EXPORTER', 'otlp')
+        case exporter
+        when 'otlp' then otlp_exporter
+        when 'jaeger' then jaeger_exporter
+        when 'zipkin' then zipkin_exporter
+        else
+          OpenTelemetry.logger.warn "The #{exporter} exporter is unknown and cannot be configured, configuring the console exporter instead"
+          Trace::Export::ConsoleSpanExporter.new
+        end
+      end
+
+      def configure_propagation # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity
+        propagators = ENV.fetch('OTEL_PROPAGATORS', 'tracecontext,baggage').split(',')
+        injectors, extractors = propagators.uniq.collect do |propagator|
+          case propagator
+          when 'tracecontext'
+            [OpenTelemetry::Trace::Propagation::TraceContext.text_map_injector, OpenTelemetry::Trace::Propagation::TraceContext.text_map_extractor]
+          when 'baggage'
+            [OpenTelemetry::Baggage::Propagation.text_map_injector, OpenTelemetry::Baggage::Propagation.text_map_extractor]
+          when 'b3' then b3_single_propagator
+          when 'b3multi' then b3_multi_propagator
+          when 'jaeger' then jaeger_propagator
+          when 'xray' then xray_propagator
+          when 'ottrace' then ot_trace_propagator
+          else
+            OpenTelemetry.logger.warn "The #{propagator} propagator is unknown and cannot be configured"
+            [nil, nil]
+          end
+        end.transpose
+        OpenTelemetry.propagation = create_propagator(@injectors || injectors.compact,
+                                                      @extractors || extractors.compact)
       end
 
       def create_propagator(injectors, extractors)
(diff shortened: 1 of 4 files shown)


sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -148,14 +148,41 @@ def configure_span_processors
       end
 
       def default_span_processor
-        Trace::Export::SimpleSpanProcessor.new(
+        Trace::Export::SimpleSpanProcessor.new(configured_exporter)
```
robertlaurin, line 151: >The default span processor is not specified and there is no specified environment variable to select it. We defaulted to the SimpleSpanProcessor previously, and I've kept that default here, but that is rarely useful with any of the defined exporters.

Does it make sense to just give users a better default then?  I recall at least one user reporting that they went with the simple span processor in production and had issues as a result.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -148,14 +148,41 @@ def configure_span_processors
...
+      end
+
+      def configured_exporter
```
robertlaurin, line 154: nitpick: `exporter_from_env` or some variation of that  is probably a more accurate name.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -166,18 +193,72 @@ def create_propagator(injectors, extractors)
...
+          unavailable_external_propagator 'ottrace'
+        end
+      end
+
+      def unavailable_external_propagator(name, gem_suffix = name)
+        OpenTelemetry.logger.warn "The #{name} propagator cannot be configured - please add opentelemetry-propagator-#{gem_suffix} to your Gemfile"
+        [nil, nil]
+      end
+
+      def otlp_exporter
```
robertlaurin, line 241: Nitpick: I have concerns with the "clutter" associated with adding all these `if defined?(` checking methods to the configurator class.  I would like to see these extracted to some external module, or even common. Maybe something like `OpenTelemetry::Common::ExporterFromEnv.otlp_exporter`, the helper could just as easily exist in the SDK.  I don't have too strong of a preference where.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -148,14 +148,41 @@ def configure_span_processors
...
+
+      def configured_exporter
+        exporter = ENV.fetch('OTEL_TRACES_EXPORTER', 'otlp')
+        case exporter
+        when 'otlp' then otlp_exporter
+        when 'jaeger' then jaeger_exporter
+        when 'zipkin' then zipkin_exporter
+        else
+          OpenTelemetry.logger.warn "The #{exporter} exporter is unknown and cannot be configured, configuring the console exporter instead"
           Trace::Export::ConsoleSpanExporter.new
```
robertlaurin, line 162: I'm wondering if we should not just fail to initialize here, this has some implication that a value was set, but is incorrect.  

I'm considering the situation where an application that relies on the env to configure their tracing exporter has a typo in the env and suddenly all of their logs are flooded with traces.

robertlaurin (commented):
Couple of comments, I understand some of them are around pre-existing behaviour but it might be a good opportunity to discuss changing it.

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: > Does it make sense to just give users a better default then? I recall at least one user reporting that they went with the simple span processor in production and had issues as a result.

Yes, I've chosen to wrap the fallback exporter (the console span exporter) in a SimpleSpanProcessor and wrap all others in a BatchSpanProcessor.

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: I've taken a different approach, based on your suggestions in Slack. 🙏 LMK if this is cleaner or if you still want it extracted.

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: I've opted not to "fail", but instead log and fall back to the NoopSpanProcessor

sdk/lib/opentelemetry/sdk/configurator.rb
ericmustin: ^ i think this is the appropriate approach

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -143,19 +143,42 @@ def install_instrumentation
...
+        injectors, extractors = propagators.uniq.collect do |propagator|
+          case propagator
+          when 'tracecontext'
+            [OpenTelemetry::Trace::Propagation::TraceContext.text_map_injector, OpenTelemetry::Trace::Propagation::TraceContext.text_map_extractor]
+          when 'baggage'
+            [OpenTelemetry::Baggage::Propagation.text_map_injector, OpenTelemetry::Baggage::Propagation.text_map_extractor]
+          when 'b3' then fetch_propagator(propagator, 'OpenTelemetry::Propagator::B3::Single')
+          when 'b3multi' then fetch_propagator(propagator, 'OpenTelemetry::Propagator::B3::Multi', 'b3')
+          when 'jaeger' then fetch_propagator(propagator, 'OpenTelemetry::Propagator::Jaeger')
+          when 'xray' then fetch_propagator(propagator, 'OpenTelemetry::Propagator::XRay')
```
ericmustin, line 173: there's probably a place in the future where this lives in a `-contrib` repo with other cloud provider (or vendor 😅 ) specific formats and we need some different handling here, but for now i think this makes sense

ericmustin (approved):
lgtm

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: Yeah, maybe - for some (political?) reason, this is required by the spec 🤷

robertlaurin (approved):

sdk/test/opentelemetry/sdk/configurator_test.rb
```diff
@@ -197,6 +239,27 @@
           OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor
         )
       end
+
+      it 'can be set by environment variable' do
+        with_env('OTEL_TRACES_EXPORTER' => 'jaeger') do
+          configurator.configure
+        end
+
+        skip 'how do we load this gem for testing?'
```
fbogsany, line 248: Thoughts on this @robertlaurin ?

sdk/test/opentelemetry/sdk/configurator_test.rb
```diff
@@ -237,10 +300,18 @@
   end
 
   def extractors_for(propagator)
-    propagator.instance_variable_get(:@extractors) || propagator.instance_variable_get(:@extractor)
+    if propagator.instance_of? Context::Propagation::CompositePropagator
+      propagator.instance_variable_get(:@extractors)
+    else
+      [propagator.instance_variable_get(:@extractor)]
+    end
```
fbogsany, line 307: This removes some noise in the CI logs.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -143,41 +143,66 @@ def install_instrumentation
...
+      def wrapped_exporter_from_env
+        exporter = ENV.fetch('OTEL_TRACES_EXPORTER', 'otlp')
+        case exporter
+        when 'none' then nil
+        when 'otlp' then fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter')
+        when 'jaeger' then fetch_exporter(exporter, 'OpenTelemetry::Exporter::Jaeger::CollectorExporter')
+        when 'zipkin' then fetch_exporter(exporter, 'OpenTelemetry::Exporter::Zipkin::Exporter')
+        else
+          OpenTelemetry.logger.warn "The #{exporter} exporter is unknown and cannot be configured, spans will not be exported"
+          nil
```
fbogsany, line 159: This (`[wrapped_exporter_from_env].compact` and returning `nil` in the default cases) means we end up with the default NoopSpanProcessor rather than adding another one.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -143,41 +143,66 @@ def install_instrumentation
...
+        OpenTelemetry.propagation = create_propagator(@injectors || injectors.compact,
+                                                      @extractors || extractors.compact)
       end
 
       def create_propagator(injectors, extractors)
         if injectors.size > 1 || extractors.size > 1
           Context::Propagation::CompositePropagator.new(injectors, extractors)
         else
-          Context::Propagation::Propagator.new(injectors, extractors)
+          Context::Propagation::Propagator.new(injectors.first, extractors.first)
```
fbogsany, line 189: This is a bug exposed by new testing here.

sdk/test/opentelemetry/sdk/configurator_test.rb
robertlaurin: I think there's a couple options here:
- We could just include the jaeger exporter as a test dependency in the gemfile.
- We could write our own double for testing
- We could use a [gem like this one](https://github.com/adammck/minitest-stub-const) and stub the constant `OpenTelemetry::Exporter::Jaeger::Collector::Exporter` and have it return a mock 

I think I would reach for requiring the gem as a test dependency and if it gave me any trouble would use the `minitest-stub-const` gem to just stub and return a mock.  I don't think this needs to be more than a shallow integration test to make sure it selects the right thing.

robertlaurin (approved):

sdk/test/opentelemetry/sdk/configurator_test.rb
```diff
@@ -134,6 +135,27 @@
...
+
+        with_env('OTEL_PROPAGATORS' => 'baggage') do
+          configurator.configure
+        end
+
+        _(injectors_for(OpenTelemetry.propagation)).must_equal(expected_injectors)
+      end
+
+      it 'defaults to none with invalid env var' do
+        with_env('OTEL_PROPAGATORS' => 'unladen_swallow') do
```
ericmustin, line 150: lol

PR #607, opened 2021-02-12
fix: add thread error handling to the BSP

robertlaurin:
This adds graceful handling for when creating a new thread raises.  We observed this scenario with an application that was improperly terminating sidekiq which resulted in a `ThreadError`.

mperham/sidekiq#4029
mperham/sidekiq#4738

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -177,8 +177,9 @@ module OpenTelemetry
             @pid = pid
             spans.clear
             @thread = restart_thread ? Thread.new { work } : nil
-          rescue ThreadError
-            @metrics_reporter.add_to_counter('otel.bsp.failure', labels: { 'reason' => 'ThreadError' })
+          rescue ThreadError => error
+            @metrics_reporter.add_to_counter('otel.bsp.error', labels: { 'reason' => 'ThreadError' })
+            OpenTelemetry.logger.error "OpenTelemetry error in BatchSpanProcessor#reset_on_fork #{error.message}"
           end
 
           def export_batch(batch, timeout: @exporter_timeout_seconds)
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -177,6 +177,8 @@ def reset_on_fork(restart_thread: true)
             @pid = pid
             spans.clear
             @thread = restart_thread ? Thread.new { work } : nil
+          rescue ThreadError
+            @metrics_reporter.add_to_counter('otel.bsp.failure', labels: { 'reason' => 'ThreadError' })
```
fbogsany, line 181: Do we also want to log here?

robertlaurin: I think so, did you have any thoughts on the counter name?  Is `otel.bsp.failure` fine?

fbogsany: `otel.bsp.error` might be better

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -177,6 +177,9 @@ def reset_on_fork(restart_thread: true)
             @pid = pid
             spans.clear
             @thread = restart_thread ? Thread.new { work } : nil
+          rescue ThreadError => e
+            @metrics_reporter.add_to_counter('otel.bsp.error', labels: { 'reason' => 'ThreadError' })
+            OpenTelemetry.logger.error "OpenTelemetry error in BatchSpanProcessor#reset_on_fork #{e.message}"
```
mwear, line 182: Are we supposed use `OpenTelemetry.logger.error` or `OpenTelemetry.handle_error` for this purpose (I'm unsure).

fbogsany: Great question - I can't think of a good reason *not* to use `OpenTelemetry.handle_error`.

robertlaurin: Updated, I think it makes sense to let the users decide on how to handle this error.  And it has a sane default of logging it.

fbogsany (approved):

PR #608, opened 2021-02-16
feat: add Net::HTTP#connect tracing

robertlaurin:
Followup to a discussion regarding connect spans https://github.com/open-telemetry/opentelemetry-ruby/pull/508#issuecomment-777152748

### DIFF
--- a/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
+++ b/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
@@ -40,6 +40,24 @@ module OpenTelemetry
 
             private
 
+            def connect
+              if proxy?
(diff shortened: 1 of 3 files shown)


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -40,6 +40,24 @@ def request(req, body = nil, &block)
 
             private
 
+            def connect
+              if proxy?
```
robertlaurin, line 44: Repeating this check to retrieve the hostname and port https://github.com/ruby/ruby/blob/master/lib/net/http.rb#L975-L981

fbogsany (approved):

mwear (approved):
LGTM

PR #611, opened 2021-02-17
fix: do not disorder randomization used by seed when creating spans

aislinncronin:
We noticed that when using `seed` for randomization, the sequence was getting disordered because OTel traces and span IDs are generated using `Random`. This change prevents the disordering.

### DIFF
--- a/api/lib/opentelemetry/trace.rb
+++ b/api/lib/opentelemetry/trace.rb
@@ -21,7 +21,7 @@ module OpenTelemetry
     # the class `Random`.
     #
     # @return [#bytes]
-    RANDOM = Random.respond_to?(:bytes) ? Random : Random::DEFAULT
+    RANDOM = Random.respond_to?(:bytes) ? Random.new : Random::DEFAULT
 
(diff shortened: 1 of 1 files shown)


api/lib/opentelemetry/trace.rb
```diff
@@ -21,7 +21,7 @@ module Trace
     # the class `Random`.
     #
     # @return [#bytes]
-    RANDOM = Random.respond_to?(:bytes) ? Random : Random::DEFAULT
+    RANDOM = Random.respond_to?(:bytes) ? Random.new : Random::DEFAULT
```
fbogsany, line 24: Unfortunately this will break "things that fork" (e.g. pre-forking web servers and pre-forking job servers). The problem is that the **global** random number generator is re-initialized by the Ruby VM on fork but instances are not. This results in an unfortunate situation where new traces generated by sibling processes generate the same sequence of trace IDs.

aislinncronin: aaah I see. thanks for the explanation! will close this 👍

PR #612, opened 2021-02-17
fix: Include http.status_text only if reason_phrase is in the response

aislinncronin:
We noticed the error OpenTelemetry error: invalid attribute value type NilClass appearing frequently in our logs after the addition of Faraday instrumentation to our project. This fix will skip the addition of attribute if it is nil in the response.


fbogsany (approved):
Thank you @aislinncronin !

PR #616, opened 2021-02-18
Configurable obfuscation of sql in mysql2 instrumentation to avoid logging sensitive data

aislinncronin:
This PR adds configurable obfuscation of sql queries for the db.statement attribute in mysql2 instrumentation to avoid logging potentially sensitive data. To enable, set `enable_sql_obfuscation ` to true. The default will be false. 

I noticed there is no README for the mysql2 gem. Would it be best for me to add one (based off of the other gems) and add a section for this configuration option?


fbogsany:
> I noticed there is no README for the mysql2 gem. Would it be best for me to add one (based off of the other gems) and add a section for this configuration option?

Yes please!

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -44,6 +44,18 @@ def query(sql, options = {})
 
           private
 
+          def obfuscate_sql(sql)
+            # if ENV variable is set, removes any string values or numbers inside spaces
+            # (to avoid mathcing dates).
+            return sql unless ENV['OTEL_INSTRUMENTATION_MYSQL2_OBFUSCATE']
```
fbogsany, line 50:
```suggestion
            return sql unless ENV['OTEL_RUBY_INSTRUMENTATION_MYSQL2_OBFUSCATE']
```
The OTel spec [recommends](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/sdk-environment-variables.md#language-specific-environment-variables) that language-specific environment variables follow the pattern:
```
OTEL_{LANGUAGE}_{FEATURE}
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -44,6 +44,18 @@ def query(sql, options = {})
...
+
+            if sql.size > 2000
+              'SQL query too large to remove sensitive data ...'
+            else
+              sql.gsub(/'.+?'/m, '?').gsub(/ \d+ /m, ' ? ')
+            end
```
fbogsany, line 56: It might be preferable to adopt the NewRelic obfuscation rules: https://github.com/newrelic/newrelic-ruby-agent/blob/dev/lib/new_relic/agent/database/obfuscation_helpers.rb

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
fbogsany: Rather than use an env var here (or at least not only an env var), we should add an instrumentation [configuration option](https://github.com/open-telemetry/opentelemetry-ruby/blob/8fbfa71aab426d500e0b3cd9d9ba2a30822e3b7f/api/lib/opentelemetry/instrumentation/base.rb#L140-L154)

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
JordiPolo: Do you mean to use their mysql related rules or bring the file as a shared tool that other instrumentations may use in the future?
Is it proper to copy from that codebase? I do not know if there will be licensing issues or something.

fbogsany: I was proposing to use the mysql related rules. We can generalize later when we add other instrumentation.

That code is under an Apache license, so a) it can be copied, and b) it is compatible with our license.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
aislinncronin: ok replaced the env var with a config option 👍

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -44,6 +54,33 @@ def query(sql, options = {})
...
+              obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
+              obfuscated
+            end
+          end
+
+          def generate_mysql_regex
+            components = %i[single_quotes double_quotes numeric_literals boolean_literals
+                            hexadecimal_literals comments multi_line_comments]
+            Regexp.union(components.map { |component| COMPONENTS_REGEX_MAP[component] })
+          end
```
fbogsany, line 73: This should be cached so we don't have to recompile the regex on every call.

aislinncronin: done 👍

fbogsany (approved):

robertlaurin (approved):
🎉

fbogsany:
@aislinncronin can you please push an empty commit to this branch to kick the EasyCLA check? It gets stuck sometimes 😞

aislinncronin:
@fbogsany that worked ✔️ thanks

fbogsany:
D'oh: I forgot about [Conventional Commit messages](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/CONTRIBUTING.md#use-conventional-commit-messages). @aislinncronin I'll have to remember to add this to the release notes manually. For future PRs, can you please use conventional commit messages? It helps automate some of the release process.

fbogsany:
Also, thank you very much for this contribution - we have been asked for something like this for quite a while.

aislinncronin:
@fbogsany whoops I thought I had been but looks like i missed one. Will do next time 👍

fbogsany:
> @fbogsany whoops I thought I had been but looks like i missed one. Will do next time 👍

I think the PR title needs to have a conventional commit message as well - the title is used for the merge commit message (I realized a second after I hit Merge 😞).

Issue #609, opened 2021-02-16
Add instrumentation info to README


fhwang:
For newbies, answer a few basic questions:

- Which libraries are currently instrumented?
- Which instrumentations are being worked on?
- Which instrumentations have been commonly requested but could use help?
- Where are the instrumentation libraries, both in this repo and as a gem install?

Issue #304, opened 2020-07-23
Add instrumentation packages to the registry


mwear:
https://opentelemetry.io/get-involved/#registry

ericmustin:
@mwear can you assign this to me? Should be able to get to this soon.

genebean:
what all are you looking for with this - just adding entries for things like sinatra and such?

fbogsany:
> what all are you looking for with this - just adding entries for things like sinatra and such?

Yep, exactly that.

fhwang:
@ericmustin How would you feel about me taking this one? I was thinking of making some sort of Rake task to make it easier to automatically generate the Markdown files needed by the OTel registry.

ericmustin:
ah my bad, dropped the ball here.  Yes i think a rake task would be appropriate, feel free to give it a shot. 
![ball drop](https://media.tenor.com/images/72967302d062f1a9f361bb181e589758/tenor.gif)

fhwang:
Oops, I mis-linked these. #628 makes it *easier* to perform this task but it's still something to be run manually for now. I'll make that PR to `opentelemetry.io` now.

PR #619, opened 2021-02-19
docs: Add more documentation around instrumentation libraries.

fhwang:
Not sure if this is the final state of this but it seemed like a good start. After thinking about it I opted against listing all the instrumentation libraries in any READMEs: This seems brittle and I'd imagine most users should be able to navigate Github and/or the repository on their own with a quick pointer. But I'm definitely open to other thoughts if we think there are ways this documentation can be clearer.

### DIFF
--- a/instrumentation/README.md
+++ b/instrumentation/README.md
@@ -1,0 +1,52 @@
+# OpenTelemetry instrumentation libraries
+ 
+[OpenTelemetry](https://opentelemetry.io/) is an open source observability framework, providing a general-purpose API, SDK, and related tools required for the instrumentation of cloud-native software, frameworks, and libraries.
+
+Instrumentation libraries provide pre-built OpenTelemetry instrumentation for popular libraries: This repository contains instrumentation for Rails, Rack, Sinatra, and others.  This way you can start using OpenTelemetry with minimal changes to your application.
+
+## How do I get started?
+
+### Individual instrumentation libraries
+
+Individual instrumentation libraries can be found in subdirectories under `/instrumentation` (with the exception of `all`, see below).
+
+To get started with a single instrumentation library, for example `opentelemetry-instrumentation-rack`:
+
+### 1. Install the gem
+
+```
+gem install opentelemetry-instrumentation-rack
+```
+
+### 2. Configure OpenTelemetry to use the instrumentation
+
+```
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Rack'
+end
+```
+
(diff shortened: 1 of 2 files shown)


fbogsany:
Part of the goal here is to have a table of what we currently support, what we're planning to work on, and what we know people might want but no-one is planning to work on yet. People won't necessarily know how to navigate the repo, and they shouldn't need to unless they plan to contribute.

A way to address "what we currently support" is #304. That doesn't address what we plan and don't plan to work on.

We could make sure we have issues for all the instrumentation we think people want, with ones we plan to implement assigned to the people who intend to implement them and all others marked with `#help-wanted`. Then we can add prominent links to the appropriate issue queries in the README.

instrumentation/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+## License
+
+All OpenTelemetry Ruby instrumentation gems are distributed under the Apache 2.0 license. See [LICENSE][license-github] for more information.
+
+
+[license-github]: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/LICENSE
+[ruby-sig]: https://github.com/open-telemetry/community#ruby-sig
+[community-meetings]: https://github.com/open-telemetry/community#community-meetings
+[ruby-gitter]: https://gitter.im/open-telemetry/opentelemetry-ruby
```
fbogsany, line 50: Gitter is going away: #620

fbogsany:
"Instrumentation wanted" link: https://github.com/open-telemetry/opentelemetry-ruby/labels/instrumentation

"Instrumentation planned" link: https://github.com/open-telemetry/opentelemetry-ruby/issues?q=is%3Aopen+label%3Ainstrumentation+-label%3A%22help+wanted%22+

"Instrumentation help wanted" link: https://github.com/open-telemetry/opentelemetry-ruby/issues?q=is%3Aopen+label%3Ainstrumentation+label%3A%22help+wanted%22

fhwang:
Thanks for the feedback @fbogsany . Just to be clear, do you believe this README should contain a table of what we currently support, in addition to a link to the OTel Registry? Or just the link? I'm inclined to think that the link will be easy enough and will always be up-to-date.

And I'm happy to take on #304 as well. Seems like we ought to be able to generate those registry PRs automatically.

fbogsany:
Just the link is fine - we should only have to update one place for available instrumentation: the registry. Likewise for the "instrumentation we're going to implement" and "instrumentation people might want that we have no plans for yet".

fhwang:
Closes #609.

fbogsany (approved):
This looks great, thanks @fhwang !

PR #625, opened 2021-02-23
Add zipkin exporter

ericmustin:
### Summary 

This PR is ~WIP~ to address https://github.com/open-telemetry/opentelemetry-ruby/issues/602 , by adding a Zipkin Exporter for spec compliance.

Largely based on [the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/cb16422a61219d4fd99b466a70e47cb8af9e26b1/specification/trace/sdk_exporters/zipkin.md) , which is still in progress 😢 , as well as the [OpenTelemetry-JS](https://github.com/open-telemetry/opentelemetry-js/tree/main/packages/opentelemetry-exporter-zipkin) and OpenTelemetry-Collector [Translation](https://github.com/open-telemetry/opentelemetry-collector/blob/ba720d8aa79703cc2d77507c893fa69c2ab8d8a5/translator/trace/zipkin/traces_to_zipkinv2.go) and [Exporter](https://github.com/open-telemetry/opentelemetry-collector/blob/main/exporter/zipkinexporter/zipkin.go) implementations. These implementations all differ in annoyingly subtle ways and have slowed me down a lot here, ~which is why this is a very rough draft PR.~ I also used the [Zipkin Ruby Tracer](https://github.com/openzipkin/zipkin-ruby) as a reference.

~Still very rough and untested but the guts are there, I plan to have this in a bit cleaner state by EOW.~ I think this is ready for review at this point.

### Notes for Reviewers

- The specification and it's various implementation are a little bit all over the place. Some specific portions I am not 100% confident in
1. [Remote Endpoint](https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#remote-endpoint). While the spec says it should (not must) be implemented, it doesn't appear to be implemented in Otel-js. So, I went based on the [implementation in the collector's translation](https://github.com/open-telemetry/opentelemetry-collector/blob/347cfa9ab21d47240128c58c9bafcc0014bc729d/translator/trace/zipkin/traces_to_zipkinv2.go#L284), which doesn't check past `peer.service` for a `remoteEndpoint` `serviceName`, however it does attempt to collect host/port information and set those as tags, as well as determine whether ip addr is ivp6 or ipv4 for the tag name. I'm not sure if this is overkill or not in depth enough since i'm not checking all the `peer.service` fallbacks, but this already isn't a required detail.
2. [Local Endpoint](https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#opentelemetry-to-zipkin-transformation) Same but different as Remote Endpoint, this isn't spec'd out at all but both otel-js and the collector implement it, so i followed the approach the collector took. Tbh This seems like it ought to be required as it's the only way to pass the canonical `service.name` (as zipkin localendpoint `serviceName`), 🤷 
3. [Debug / Shared](https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#opentelemetry-to-zipkin-transformation) I just punted on these, making `debug` always default to false. Neither are specified in the spec and i'm not entirely sure their practical use-case in opentelemetry and whether some of the propagation headers that they are based on are even respected. I figure we can add if ppl are using them.
4. [Error Status](https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#status) The specification says an otel error status should add an `error` tag set to the `description` of the status, but for the life of me I can't find anything detailing this `error` field in the [actual zipkin v2 spec](https://github.com/openzipkin/zipkin-api/blob/master/zipkin2-api.yaml). But in practice this appears to be used [here](https://github.com/openzipkin/zipkin-ruby/blob/7bedb4dd162c4cbeffc7b97dd06c8dbccbfbab62/lib/zipkin-tracer/trace.rb#L259) and [here](https://github.com/open-telemetry/opentelemetry-collector/blob/81c7cc53b7067fbf3db4e1671b13bbe2796eb56e/translator/trace/zipkin/traces_to_zipkinv2.go#L144). However it looks like [some other language zipkin clients](https://github.com/signalfx/signalfx-nodejs-tracing/blob/83ef85a261236bb8e260c83cf7ccd8a90801375b/test/format.spec.js#L164-L166) emit `error: "true"`  in the event there's no description, and frankly i think the `error` tag is some weird v1 holdover. Anyway so i try to set it to `description` and fallback to `true` if needed.
  
- The export type I went with is v2 json over http. That was the most official looking ruby implementation [i found in the wild](https://github.com/openzipkin/zipkin-ruby/tree/9552f9c72cf5f12188996016c77e55d5c0abe91e#http) so I figured it would be good to keep ourselves at least somewhat coupled to another active repository out there. If there's a good case to using the more recent [protobuf]( https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#considerations-for-legacy-v1-format) approach, I'd be open to considering it, tho frankly I have far less professional experience w/protobuf so there's that too.

p.s. I have to test this against a real collector still but i figure it's mainly ready for review


exporter/zipkin/README.md
```diff
@@ -0,0 +1,87 @@
+# opentelemetry-exporter-zipkin
+
+The `opentelemetry-exporter-zipkin` gem provides Zipkin exporters for OpenTelemetry for Ruby. Using `opentelemetry-exporter-zipkin`, an application can configure OpenTelemetry to export collected tracing data to [Zipkin][zipkin-home]. One exporter is included: the `Exporter` exports in HTTP JSONformat over TCP to a Zipkin backend
```
fbogsany, line 3:
```suggestion
The `opentelemetry-exporter-zipkin` gem provides Zipkin exporters for OpenTelemetry for Ruby. Using `opentelemetry-exporter-zipkin`, an application can configure OpenTelemetry to export collected tracing data to [Zipkin][zipkin-home]. One exporter is included: the `Exporter` exports in HTTP JSON format over TCP to a Zipkin backend
```

exporter/zipkin/README.md
```diff
@@ -0,0 +1,87 @@
...
+    pp child_span
+  end
+end
+```
+
+For additional examples, see the [examples on github][examples-github].
+
+## How can I configure the Zipkin exporter?
+
+TODO
```
fbogsany, line 66: ❓

exporter/zipkin/README.md
```diff
@@ -0,0 +1,87 @@
...
+
+[zipkin-home]: https://zipkin.io/
+[opentelemetry-home]: https://opentelemetry.io
+[bundler-home]: https://bundler.io
+[repo-github]: https://github.com/open-telemetry/opentelemetry-ruby
+[license-github]: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/LICENSE
+[examples-github]: https://github.com/open-telemetry/opentelemetry-ruby/tree/main/examples
+[ruby-sig]: https://github.com/open-telemetry/community#ruby-sig
+[community-meetings]: https://github.com/open-telemetry/community#community-meetings
+[ruby-gitter]: https://gitter.im/open-telemetry/opentelemetry-ruby
```
fbogsany, line 87:
```suggestion
[discussions-url]: https://github.com/open-telemetry/opentelemetry-ruby/discussions
```

exporter/zipkin/opentelemetry-exporter-zipkin.gemspec
```diff
@@ -0,0 +1,48 @@
...
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('thrift/gen-rb/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'opentelemetry-api', '~> 0.15.0'
```
fbogsany, line 29: I just realized both this and the other exporters depend on the SDK, not (just) the API.

exporter/zipkin/test/test_helper.rb
```diff
@@ -0,0 +1,45 @@
...
+
+def with_env(new_env)
+  env_to_reset = ENV.select { |k, _| new_env.key?(k) }
+  keys_to_delete = new_env.keys - ENV.keys
+  new_env.each_pair { |k, v| ENV[k] = v }
+  yield
+ensure
+  env_to_reset.each_pair { |k, v| ENV[k] = v }
+  keys_to_delete.each { |k| ENV.delete(k) }
+end
```
fbogsany, line 22: This is the fifth copy of `with_env`. It might be worth extracting it to `opentelemetry-common`.

exporter/zipkin/test/opentelemetry/exporters/zipkin/transformer_test.rb
```diff
@@ -0,0 +1,154 @@
...
+      events,
+      nil,
+      instrumentation_library,
+      OpenTelemetry::Trace.generate_span_id,
+      trace_id,
+      trace_flags,
+      tracestate
+    )
+  end
+end
```
fbogsany, line 154: You have an identical method in `test_helper.rb`.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,192 @@
...
+              end
+            end
+          end
+
+          endpoint
+        end
+
+        def int64(byte_string)
+          int = byte_string.unpack1('Q>')
+          int < (1 << 63) ? int : int - (1 << 64)
```
fbogsany, line 187: Can you just use `q>` here instead of extracting the value as unsigned and converting it to signed?
```suggestion
          byte_string.unpack1('q>')
```

ericmustin: I think so, I was just copying the jaeger exporter basically

exporter/zipkin/test/test_helper.rb
ericmustin: Makes sense to me but not sure to where would we extract this, as it's only for tests.  `common/tests/test_helper.rb` maybe? or would it belong in something like `common/lib/opentelemetry/test_utils/ `? wdyt?

exporter/zipkin/opentelemetry-exporter-zipkin.gemspec
ericmustin: updated the other exporters as well

exporter/zipkin/README.md
ericmustin: updated

exporter/zipkin/test/test_helper.rb
ericmustin: added a small helpers file to `common` but i didn't include it in the `Common` module as I figure test helpers shouldn't be available for use from `opentelemetry/common`...maybe an anti patten, certainly if there's a more 'ruby' way to accomplish this i'm happy to update

ericmustin:
Ah, if i'm reading this correct, i missed a requirement https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md

I think i still need to add v2 protobuf export if i'm understanding the `X` and `*` correctly (this pr currently only has v2 json export)? Do I have that right @mwear @fbogsany @robertlaurin ?

exporter/zipkin/test/test_helper.rb
fbogsany: For now, I think we decided to keep both common test helpers and common ... non-test helpers in `opentelemetry/common`. We can discuss splitting them out later, but so far there hasn't been enough of either to warrant splitting the gem.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
fbogsany: Well then, I suck, don't I? 😆 

Can you check & fix it here, and I'll make the same fix in the Jaeger exporter.

fbogsany:
> Ah, if i'm reading this correct, i missed a requirement https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md
> 
> I think i still need to add v2 protobuf export if i'm understanding the `X` and `*` correctly (this pr currently only has v2 json export)? Do I have that right @mwear @fbogsany @robertlaurin ?

The `X` means optional, so the V1 exporters are optional. The `*` means "one of", so we either need to implement V2 JSON or V2 Protobuf. So this PR is 👌

exporter/zipkin/README.md
```diff
@@ -0,0 +1,93 @@
...
+
+## How can I configure the Zipkin exporter?
+
+The collector exporter can be configured explicitly in code, as shown above, or via environment variables. The configuration parameters, environment variables, and defaults are shown below.
+
+| Parameter   | Environment variable                  | Default                    |
+| ----------- | --------------------------------------| -------------------------- |
+| `endpoint:` | `OTEL_EXPORTER_ZIPKIN_ENDPOINT`       | `"http://localhost:9411"`  |
+| `headers:`  | `OTEL_EXPORTER_ZIPKIN_TRACES_HEADERS` | `nil`                      |
+| `timeoout:` | `OTEL_EXPORTER_ZIPKIN_TRACES_TIMEOUT` | `10`                       |
```
fbogsany, line 72: It's worth adding `OTEL_TRACES_EXPORTER=zipkin` here, to make it clear that no code changes are required.

ericmustin:
- So, I finally ran this locally with a collector and noticed some errors being returned by the zipkin receiver about the zipkin span format, tried to address them, and have pushed up some updates that have this now successfully transmitting spans to a collector with the `zipkin` receiver.

- installed in a [sample app](https://github.com/DataDog/trace-examples/tree/master/opentelemetry) locally and got the following output from the collector: https://gist.github.com/ericmustin/22be2e96b5f8c8edd2e941709f2a0206

  - (^in logging output, the span `Status code` and `Status message` appear not to be getting set correctly, so I have to look into that more still, otherwise seems to work.)

- Also worth noting that I removed the `int64` related translation code as it seems the zipkin receiver accepts the same format we use in the [b3 propagator](https://github.com/open-telemetry/opentelemetry-ruby/blob/4d576d5a680736e2798872262aae81c8819d2a78/propagator/b3/lib/opentelemetry/propagator/b3/multi/text_map_injector.rb#L38-L39)

exporter/zipkin/README.md
ericmustin: updated

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
ericmustin: Upon closer investigation I think the zipkin receiver expects the same format we use for [b3 propagation](https://github.com/open-telemetry/opentelemetry-ruby/blob/4d576d5a680736e2798872262aae81c8819d2a78/propagator/b3/lib/opentelemetry/propagator/b3/multi/text_map_injector.rb#L38), which makes sense. So long story short, i'm actually no longer sure about whether to use `.unpack1('q>')` in the `jaeger` exporter.

mwear:
@ericmustin is there anything to do regarding the status comment above, or is this ready to go?

mwear:
Depending on the merge order, this might be relevant: #657.

ericmustin:
@mwear it could use a review if you have the time, or another pass from @fbogsany . I should be able to look into status code/message more in the next day or so and can add tests, but otherwise it would be good to get feedback

johnduhart:
@jcarres-mdsol @ykitamura-mdsol Since both of you are contributors to zipkin-ruby, maybe we can provide some useful feedback here?

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,188 @@
...
+          start_time = (span_d.start_timestamp.to_f * 1_000_000).to_i
+          duration = (span_d.end_timestamp.to_f * 1_000_000).to_i - start_time
+          tags = {}
+          service_name = DEFAULT_SERVICE_NAME
+          resource&.attribute_enumerator&.select do |key, value|
+            if key == SERVICE_NAME_ATTRIBUTE_KEY
+              service_name = value
+            else
+              tags[key] = value
+            end
```
fbogsany, line 52: The spec doesn't specify a mapping for Resource attributes yet. I don't see any work in progress on that yet. Should we omit Resource attributes other than the service name for now?

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,188 @@
...
+        end
+
+        def add_status_tags(span_data, tags)
+          if span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::ERROR
+            # mark errors if we can, setting error key to description but falling back to 'true'
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#status
+            # https://github.com/openzipkin/zipkin-ruby/blob/7bedb4dd162c4cbeffc7b97dd06c8dbccbfbab62/lib/zipkin-tracer/trace.rb#L259
+            # https://github.com/open-telemetry/opentelemetry-collector/blob/81c7cc53b7067fbf3db4e1671b13bbe2796eb56e/translator/trace/zipkin/traces_to_zipkinv2.go#L144
+            if span_data.status&.description.nil? || span_data.status.description.empty?
+              tags[ERROR_TAG_KEY] = 'true'
```
fbogsany, line 91: Spec says "use an empty string if Description has no value"
```suggestion
              tags[ERROR_TAG_KEY] = ''
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,188 @@
...
+            else
+              tags[ERROR_TAG_KEY] = span_data.status&.description
+              tags[STATUS_CODE_DESCRIPTION] = span_data.status&.description
+            end
+            tags[STATUS_CODE_NAME] = STATUS_ERROR
+          elsif span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::OK
+            tags[STATUS_CODE_NAME] = STATUS_OK
+          else
+            tags[STATUS_CODE_NAME] = STATUS_UNSET
```
fbogsany, line 100: If the status is `UNSET` then this tag should not be set.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,188 @@
...
+          if span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::ERROR
+            # mark errors if we can, setting error key to description but falling back to 'true'
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#status
+            # https://github.com/openzipkin/zipkin-ruby/blob/7bedb4dd162c4cbeffc7b97dd06c8dbccbfbab62/lib/zipkin-tracer/trace.rb#L259
+            # https://github.com/open-telemetry/opentelemetry-collector/blob/81c7cc53b7067fbf3db4e1671b13bbe2796eb56e/translator/trace/zipkin/traces_to_zipkinv2.go#L144
+            if span_data.status&.description.nil? || span_data.status.description.empty?
+              tags[ERROR_TAG_KEY] = 'true'
+            else
+              tags[ERROR_TAG_KEY] = span_data.status&.description
+              tags[STATUS_CODE_DESCRIPTION] = span_data.status&.description
```
fbogsany, line 94: `'otel.status_description'` doesn't seem to be a thing in the spec.
```suggestion
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,188 @@
...
+        def add_status_tags(span_data, tags)
+          if span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::ERROR
+            # mark errors if we can, setting error key to description but falling back to 'true'
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#status
+            # https://github.com/openzipkin/zipkin-ruby/blob/7bedb4dd162c4cbeffc7b97dd06c8dbccbfbab62/lib/zipkin-tracer/trace.rb#L259
+            # https://github.com/open-telemetry/opentelemetry-collector/blob/81c7cc53b7067fbf3db4e1671b13bbe2796eb56e/translator/trace/zipkin/traces_to_zipkinv2.go#L144
+            if span_data.status&.description.nil? || span_data.status.description.empty?
+              tags[ERROR_TAG_KEY] = 'true'
+            else
+              tags[ERROR_TAG_KEY] = span_data.status&.description
```
fbogsany, line 93:
```suggestion
              tags[ERROR_TAG_KEY] = span_data.status.description
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
ericmustin: i think it's something leftover from open census i grabbed from another implementation, will remove.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
ericmustin: removed, although I do think the spec should eventually get updated or clarified. `service.version` and `deployment.environment` are two other resource attributes I'd prefer to see included OOTB eventually.

ericmustin:
> Depending on the merge order, this might be relevant: #657.

@mwear ty for the heads up, i've added `force_flush` and updated `shutdown` methods to match jaeger/otlp.

@fbogsany tried to address all the feedback, if the spec clears up in the future we can always add back some of the details dropped.

I still haven't had a chance to look too much at the `status code` `status message` stuff I thought I had seen locally, but it looks like the tests pass around that behavior. I can dig into it next week, but i don't think it's blocking if nothing stands out to anyone else.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+          zipkin_span
+        end
+
+        def add_il_tags(span_data, tags)
+          tags['otel.library.name'] = span_data.instrumentation_library&.name if span_data.instrumentation_library&.name
+          tags['otel.library.version'] = span_data.instrumentation_library&.version if span_data.instrumentation_library&.version
+        end
+
+        def add_status_tags(span_data, tags)
+          if span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::ERROR
```
fbogsany, line 80:
```suggestion
          if span_data.status&.code == OpenTelemetry::Trace::Status::ERROR
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+            # mark errors if we can, setting error key to description but falling back to an empty string
+            # https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#status
+            # https://github.com/openzipkin/zipkin-ruby/blob/7bedb4dd162c4cbeffc7b97dd06c8dbccbfbab62/lib/zipkin-tracer/trace.rb#L259
+            # https://github.com/open-telemetry/opentelemetry-collector/blob/81c7cc53b7067fbf3db4e1671b13bbe2796eb56e/translator/trace/zipkin/traces_to_zipkinv2.go#L144
+            tags[ERROR_TAG_KEY] = if span_data.status.description.nil? || span_data.status.description.empty?
+                                    ''
+                                  else
+                                    span_data.status.description
+                                  end
```
fbogsany, line 89:
```suggestion
            tags[ERROR_TAG_KEY] = span_data.status.description || ''
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+            tags[STATUS_CODE_NAME] = STATUS_ERROR
+          elsif span_data.status&.code && span_data.status&.code == OpenTelemetry::Trace::Status::OK
```
fbogsany, line 91:
```suggestion
          elsif span_data.status&.code == OpenTelemetry::Trace::Status::OK
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+          # remote endpoint logic https://github.com/open-telemetry/opentelemetry-collector/blob/347cfa9ab21d47240128c58c9bafcc0014bc729d/translator/trace/zipkin/traces_to_zipkinv2.go#L284
+          zipkin_span['remoteEndpoint'] = endpoint_from_tags(tags, nil)
+        end
+
+        def add_annotations(zipkin_span, span_data)
+          # Tried to follow the below
+          # https://github.com/open-telemetry/opentelemetry-collector/blob/347cfa9ab21d47240128c58c9bafcc0014bc729d/translator/trace/zipkin/traces_to_zipkinv2.go#L172
+          # https://github.com/open-telemetry/opentelemetry-collector/blob/347cfa9ab21d47240128c58c9bafcc0014bc729d/translator/trace/zipkin/traces_to_zipkinv2.go#L183-L191
+          # https://github.com/open-telemetry/opentelemetry-specification/blob/cb16422a61219d4fd99b466a70e47cb8af9e26b1/specification/trace/sdk_exporters/zipkin.md#events
+          return if span_data.events.nil? || span_data.events&.size.to_i.zero?
```
fbogsany, line 110:
```suggestion
          return if span_data.events.nil? || span_data.events.empty?
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+
+          events = span_data.events&.map do |event|
```
fbogsany, line 112:
```suggestion
          events = span_data.events.map do |event|
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+            if event.attributes.keys.length.zero?
+              {
+                timestamp: (event.timestamp.to_f * 1_000_000).to_s,
+                value: event.name
+              }
+            else
+              {
+                timestamp: (event.timestamp.to_f * 1_000_000).to_s,
+                value: { "#{event.name}": event.attributes&.transform_values { |val| val.to_s } }.to_json
```
fbogsany, line 121:
```suggestion
                value: { event.name => event.attributes.transform_values(&:to_s) }.to_json
```

AFAICT you don't need to transform the event name to a Symbol.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+
+        DEFAULT_SERVICE_NAME = OpenTelemetry::SDK::Resources::Resource.default.attribute_enumerator.find { |k, _| k == SERVICE_NAME_ATTRIBUTE_KEY }&.last || 'unknown_service'
+        private_constant(:KIND_MAP, :DEFAULT_SERVICE_NAME, :SERVICE_NAME_ATTRIBUTE_KEY, :ERROR_TAG_KEY, :STATUS_CODE_NAME, :STATUS_UNSET, :STATUS_ERROR, :STATUS_OK)
+
+        def to_zipkin_span(span_d, resource)
+          start_time = (span_d.start_timestamp.to_f * 1_000_000).to_i
+          duration = (span_d.end_timestamp.to_f * 1_000_000).to_i - start_time
+          tags = {}
+          service_name = DEFAULT_SERVICE_NAME
+          resource&.attribute_enumerator&.select do |key, value|
```
fbogsany, line 46:
```suggestion
          resource.attribute_enumerator.select do |key, value|
```
`resource` and `resource.attribute_enumerator` should always be non-nil.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+
+          add_conditional_tags(zipkin_span, span_d, tags, service_name)
+          add_annotations(zipkin_span, span_d)
+
+          zipkin_span
+        end
+
+        def add_il_tags(span_data, tags)
+          tags['otel.library.name'] = span_data.instrumentation_library&.name if span_data.instrumentation_library&.name
+          tags['otel.library.version'] = span_data.instrumentation_library&.version if span_data.instrumentation_library&.version
```
fbogsany, line 76:
```suggestion
          tags['otel.library.name'] = span_data.instrumentation_library.name
          tags['otel.library.version'] = span_data.instrumentation_library.version
```
`span_data.instrumentation_library` should always be non-nil.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+
+          zipkin_span[:annotations] = events.map(&:to_h) unless events.empty?
+        end
+
+        def aggregate_span_tags(span_data, tags)
+          # convert attributes to strings
+          # https://github.com/open-telemetry/opentelemetry-specification/blob/84b18b23339dcc0b1a9d48f976a1afd287417602/specification/trace/sdk_exporters/zipkin.md#attribute
+          return tags if span_data[:attributes].nil?
+
+          tags = tags.merge(span_data[:attributes])
```
fbogsany, line 134:
```suggestion
          return tags if span_data.attributes.nil?

          tags = tags.merge(span_data.attributes)
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,180 @@
...
+
+          tags.each_with_object(tags) { |(k, v), m| m[k] = v.to_s }
```
fbogsany, line 136:
```suggestion
          tags.transform_values!(&:to_s)
```

fbogsany (commented):
A bunch of small cleanup things.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,176 @@
...
+          OpenTelemetry::Trace::SpanKind::CLIENT => 'CLIENT',
+          OpenTelemetry::Trace::SpanKind::PRODUCER => 'PRODUCER',
+          OpenTelemetry::Trace::SpanKind::CONSUMER => 'CONSUMER',
+          OpenTelemetry::Trace::SpanKind::INTERNAL => nil
+        }.freeze
+
+        SERVICE_NAME_ATTRIBUTE_KEY = 'service.name'
+        ERROR_TAG_KEY = 'error'
+        STATUS_CODE_NAME = 'otel.status_code'
+        STATUS_UNSET = 'UNSET'
```
fbogsany, line 29:
```suggestion
```

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,176 @@
...
+        STATUS_ERROR = 'ERROR'
+        STATUS_OK = 'OK'
+        ATTRIBUTE_PEER_SERVICE = 'peer.service'
+        ATTRIBUTE_NET_PEER_IP = 'peer.ip'
+        ATTRIBUTE_NET_PEER_PORT = 'peer.port'
+        ATTRIBUTE_NET_HOST_IP = 'host.ip'
+        ATTRIBUTE_NET_HOST_PORT = 'host.port'
+
+        DEFAULT_SERVICE_NAME = OpenTelemetry::SDK::Resources::Resource.default.attribute_enumerator.find { |k, _| k == SERVICE_NAME_ATTRIBUTE_KEY }&.last || 'unknown_service'
+        private_constant(:KIND_MAP, :DEFAULT_SERVICE_NAME, :SERVICE_NAME_ATTRIBUTE_KEY, :ERROR_TAG_KEY, :STATUS_CODE_NAME, :STATUS_UNSET, :STATUS_ERROR, :STATUS_OK)
```
fbogsany, line 39:
```suggestion
        private_constant(:KIND_MAP, :DEFAULT_SERVICE_NAME, :SERVICE_NAME_ATTRIBUTE_KEY, :ERROR_TAG_KEY, :STATUS_CODE_NAME, :STATUS_ERROR, :STATUS_OK)
```

fbogsany: Do you also want to `private_constant` all the `ATTRIBUTE_*` constants?

exporter/zipkin/lib/opentelemetry/exporter/zipkin/transformer.rb
```diff
@@ -0,0 +1,176 @@
...
+            tags[ERROR_TAG_KEY] = span_data.status.description || ''
+            tags[STATUS_CODE_NAME] = STATUS_ERROR
+          elsif span_data.status&.code == OpenTelemetry::Trace::Status::OK
+            tags[STATUS_CODE_NAME] = STATUS_OK
+          end
+        end
+
+        def add_conditional_tags(zipkin_span, span_data, tags, service_name)
+          zipkin_span['tags'] = tags unless tags.empty?
+          zipkin_span['kind'] = KIND_MAP[span_data[:kind]] unless span_data[:kind].nil?
```
fbogsany, line 94:
```suggestion
          zipkin_span['kind'] = KIND_MAP[span_data.kind] unless span_data.kind.nil?
```

fbogsany (commented):
OK, I think I'm done now 😄 - a few small things, otherwise LGTM.

fbogsany:
I think you have some tests that `nil` some things that are never `nil` in practice.

ericmustin:
@fbogsany apologies on the delay on  implementing feedback, I've addressed the cleanup bits and also updated the tests so that i'm not passing `nil` into span data when that is not a default value. Should be good to good now afaik

exporter/zipkin/test/opentelemetry/exporters/zipkin/transformer_test.rb
```diff
@@ -0,0 +1,131 @@
...
+
+    kind_tag = encoded_span['kind']
+    error_tag = encoded_span['tags']['error']
+
+    _(kind_tag).must_equal('SERVER')
+    _(error_tag).must_equal('')
+  end
+
+  # TODO
+
```
fbogsany, line 35:
```suggestion
```

fbogsany (approved):

PR #628, opened 2021-02-27
docs: Add Rake task for updating instrumentations in opentelemetry.io.

### LINKED ISSUE #304: Add instrumentation packages to the registry

fhwang:
Here's a quick draft of a Rake task to make it easy to keep our registry entries up-to-date. This is sort of quick-and-dirty code -- one large Rake task, no tests -- but I'm not sure how much this matters given that this is just a dev script. I can clean that up if we think it's worth it right now.

There's a bit of relying on common patterns in our instrumentation gems, hence the string parsing. Obviously that can be brittle too, but none of these errors would ever affect an end user's system, and the presence of this Rake task does not preclude the current option of falling back to just submitting a PR to opentelemetry.io by hand.

To see what this Rake task produces, you can compare the results here: https://github.com/open-telemetry/opentelemetry.io/compare/main...fhwang:update-ruby-instrumentations?expand=1

Closes #304.

### DIFF
--- a/rakelib/update_registry_instrumentation.rake
+++ b/rakelib/update_registry_instrumentation.rake
@@ -1,0 +1,61 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'erb'
+
+task :update_registry_instrumentations do
+  website_repo_dir = ARGV[1] || "../opentelemetry.io"
+
+  # Load all gemspecs
+  gemspec_files = Dir.glob("instrumentation/*/opentelemetry-*.gemspec")
+  gem_specifications = gemspec_files.map { |gemspec_file|
+    Gem::Specification.load(gemspec_file)
+  }
+
+  # Define template via ERB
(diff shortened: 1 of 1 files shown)


fbogsany:
@dazuma @robertlaurin could one or both of you please review this? It isn't my area of expertise 😕

ericmustin:
I had taken a quick look here....things look reasonable as far as I can tell, the only thing is that this might create unneeded PRs if this gem is already registered? Is there anything we can do with gh api to check if the gem is already added to the registry?

fhwang:
So the code here doesn't do anything with Github, it just creates the `.md` files in your local `opentelemetry.io` repo. You can use `git diff` etc to see which files have changed. It leaves the creation and submission of an `opentelemetry.io` PR as a manual task for now.

robertlaurin (approved):
This looks reasonable helper to me, seems like something we can tweak as needed if it comes up.

ericmustin:
@fhwang ah, woops, i misread (skimmed 😬 ) too fast. lgtm then!

ericmustin (approved):

PR #635, opened 2021-03-03
fix: rescue socket err in otlp exporter to prevent failures when unable to connect

aislinncronin:
We saw errors when the OTel endpoint was invalid or unreachable causing our deployments to fail. This PR simply adds the rescuing of SocketError.


fbogsany:
This is interesting - we've been running the equivalent of this in production for many months (a year maybe?) without seeing anything like this. Do you know where this comes from?

aislinncronin:
Yeah, here's there error:
`/usr/local/lib/ruby/2.5.0/net/http.rb:939:in `rescue in block in connect': Failed to open TCP connection to <removed> (getaddrinfo: Name or service not known) (SocketError)`

 This happens during db:migrate in our deployment when we provide an otel exporter endpoint that does not exist because  it tries to send data before it completes

fbogsany (approved):
Thanks for the context @aislinncronin, and for the fix.

PR #642, opened 2021-03-05
feat: Add peer_service config to lmdb

robertlaurin:
Add a config option to set a `peer.service` attribute to LMDB in the same way we do redis, dalli, mysql, etc..

### DIFF
--- a/instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/instrumentation.rb
+++ b/instrumentation/lmdb/lib/opentelemetry/instrumentation/lmdb/instrumentation.rb
@@ -20,6 +20,8 @@ module OpenTelemetry
           defined?(::LMDB)
         end
 
+        option :peer_service, default: nil, validate: :string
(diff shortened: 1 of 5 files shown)


fbogsany:
I much prefer the flexibility of the `Redis.with_attributes` approach from https://github.com/open-telemetry/opentelemetry-ruby/blob/4d576d5a680736e2798872262aae81c8819d2a78/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb#L59 and https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/redis/lib/opentelemetry/instrumentation/redis.rb but I realize this is more approachable for simple use-cases.

fbogsany (approved):

fbogsany (commented):
This LGTM, but there’s a test failure in 2.6 that isn’t obvious to me.

fbogsany (approved):

PR #644, opened 2021-03-05
fix: Ensure b3 injectors always return a carrier

arielvalentin:
This change ensures the `b3` injectors always return the supplied `carrier` per the method documentation and default API/SDK implementations.

### DIFF
--- a/propagator/b3/lib/opentelemetry/propagator/b3/multi/text_map_injector.rb
+++ b/propagator/b3/lib/opentelemetry/propagator/b3/multi/text_map_injector.rb
@@ -32,7 +32,7 @@ module OpenTelemetry
           # @return [Object] the carrier with context injected
           def inject(carrier, context, setter = nil)
             span_context = Trace.current_span(context).context
-            return unless span_context.valid?
+            return carrier unless span_context.valid?
 
             setter ||= @default_setter
             setter.set(carrier, B3_TRACE_ID_KEY, span_context.hex_trace_id)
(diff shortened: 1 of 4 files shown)


ericmustin (approved):
👍 lgtm, i believe this is the correct behavior

fbogsany:
FYI I'm changing this behaviour in #638. This is because the spec requires that the `carrier` be mutable, and says nothing about it being returned. I believe **NOT** returning the `carrier` is more consistent with the spec. I fully expect this to be contentious 😄.

arielvalentin:
> FYI I'm changing this behaviour in #638. This is because the spec requires that the `carrier` be mutable, and says nothing about it being returned. I believe **NOT** returning the `carrier` is more consistent with the spec. 

I suspect that this may be because some languages have a concept of `void` or no return values, which seems unusual for Ruby programs, but I will update this to be consistent with #638.  

> I fully expect this to be contentious 😄.

mwear:
> FYI I'm changing this behaviour in #638. This is because the spec requires that the `carrier` be mutable, and says nothing about it being returned. I believe **NOT** returning the `carrier` is more consistent with the spec. I fully expect this to be contentious 😄.

I find it a little bit weird that inject both mutates and returns the carrier. I believe the spec was written this way at some point. I would be happy to see us only mutate the carrier and return nothing.

arielvalentin:
> > FYI I'm changing this behaviour in #638. This is because the spec requires that the `carrier` be mutable, and says nothing about it being returned. I believe **NOT** returning the `carrier` is more consistent with the spec. I fully expect this to be contentious 😄.
> 
> I find it a little bit weird that inject both mutates and returns the carrier. I believe the spec was written this way at some point. I would be happy to see us only mutate the carrier and return nothing.

Yes it is indeed strange. I will say that I find it convenient for code structure but this should not be something end users will need to concern themselves with unless they write their own propagators.

fbogsany:
>  I would be happy to see us only mutate the carrier and return nothing.

😌 #638 FTW

PR #653, opened 2021-03-12
feat: instrument active record

robertlaurin:
Todo:
- [x] Get feedback from the SIG group
- [x] Add it to the rails instrumentation gem


robertlaurin:
In it's current state a simple controller create action looks something like this
image

instrumentation/active_record/test/test_helper.rb
```diff
@@ -0,0 +1,51 @@
...
+  host: ENV.fetch('TEST_MYSQL_HOST') { '127.0.0.1' },
+  username: ENV.fetch('TEST_MYSQL_USER') { 'root' },
+  password: ENV.fetch('TEST_MYSQL_PASSWORD') { 'root' },
+  database: ENV.fetch('TEST_MYSQL_DB') { 'mysql' }
+)
+
+# Create User model
+class User < ActiveRecord::Base; end
+
+# Get the current version so we can create a test table
```
robertlaurin, line 34: All the setup below this line is pretty gross, but it works.  I don't think CI really cares if we drop the table after but repeated test runs locally quickly become annoying if we don't clean up after ourselves.

ahayworth (commented):
Overall I think this looks okay, especially for a default implementation. My main concern is that it might end up being noisy, since there are no mechanisms to opt-out of tracing any given query/update method. A given rails request could easily create dozens or hundreds (or thousands!) of spans this way, which would make such a trace quite noisy.

I'm not sure if there's a reasonable way to mitigate that, but it might be worth thinking about.

JordiPolo:
@ahayworth  doesn't the myslq instrumentation have the same level of noise?

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
```diff
@@ -0,0 +1,28 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module ActiveRecord
+      module Patches
+        # Module to prepend to ActiveRecord::Querying for instrumentation
+        module Querying
+          def find_by_sql(sql, binds = [], preparable: nil, &block)
+            tracer.in_span("#{self}.find_by_sql") do
```
jcarres-mdsol, line 14: Maybe not for this first PR but eventually shouldn't this add the sql statement to the span as the [mysql instrumentation does](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb#L57) ? 
Or is the idea to use both instrumentations together? (although in this case this one does not add much)

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence.rb
```diff
@@ -0,0 +1,70 @@
...
+          end
+
+          def update_attribute(name, value)
+            tracer.in_span("#{self.class}#update_attribute") do
+              super
+            end
+          end
+
+          def update!(attributes)
+            tracer.in_span("#{self.class}#update!") do
```
jcarres-mdsol, line 50: Here as well in some extra methods here, can we record the attributes or other information that we have at this stage?

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/querying.rb
robertlaurin: >Or is the idea to use both instrumentations together?

The current direction we're taking with this instrumentation is that it would be used in addition to database instrumentation such as the mysql2 instrumentation.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence.rb
robertlaurin: This is something I would like to separate from this initial PR.  We don't currently have any semantic conventions around ORMs. 

I'm also hesitant adding all attributes to the span in say a call to `create`, it has PII implications that may prevent some early adoption of this instrumentation.

That's not to say we cannot do this eventually, I'd just prefer to do small iterations.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
```diff
@@ -0,0 +1,57 @@
...
+
+        def gem_version
+          Gem.loaded_specs['activerecord'].version
+        end
+
+        def patch
+          ::ActiveRecord::Querying.prepend(Patches::Querying)
+          ::ActiveRecord::Persistence.prepend(Patches::Persistence)
+          ::ActiveRecord::Persistence::ClassMethods.prepend(Patches::PersistenceClassMethods)
+          ::ActiveRecord::Persistence::ClassMethods.prepend(Patches::PersistenceInsertClassMethods) if insert_class_methods_supported?
```
robertlaurin, line 41: Conditionally patching these methods as they were introduced in rails 6.0 https://edgeguides.rubyonrails.org/6_0_release_notes.html#active-record-notable-changes

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/validations.rb
```diff
@@ -0,0 +1,31 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module ActiveRecord
+      module Patches
+        # Module to prepend to ActiveRecord::Validations for instrumentation
+        # We patch these methods because they either raise, or call
+        # the super implementation in persistence.rb
+        # https://github.com/rails/rails/blob/v5.2.4.5/activerecord/lib/active_record/validations.rb#L42-L53
```
robertlaurin, line 14: I moved the patching of `save` and `save!` from the persistence class to the validations class , as we won't trace the call to either if it fails a validation at this level.

robbkidd:
How does this compare to instrumenting ActiveRecord through a subscription to [ActiveSupport::Notifications](https://guides.rubyonrails.org/active_support_instrumentation.html) [`sql.active_record`](https://guides.rubyonrails.org/active_support_instrumentation.html#active-record) events?

ahayworth:
> @ahayworth doesn't the myslq instrumentation have the same level of noise?

@JordiPolo I'd actually presume the mysql instrumentation is even noisier. I would also say we should also have configuration options in that instrumentation as well, to help control the noise. If you wanted to start an issue or discussion about what kinds of configuration options you think a database instrumentation should have, I'd be very interested in discussing that. I put up a draft PR for postgres last night, and I'm actively thinking about what kind of options should be provided (because it is also noisy).

> How does this compare to instrumenting ActiveRecord through a subscription to ActiveSupport::Notifications sql.active_record events?

@robbkidd We discussed this at the working group meeting yesterday - three big takeaways re: ActiveSupport::Notifications from my perspective (others who were on the call - please chime in if you disagree with this summary):

1. There are some reliability concerns - if another subscriber `raise`s in their subscription handling, our own OTel subscribers may never get notified. That can cause problems that range from "annoying" to "maddening", depending on our implementation.

2. To some degree, this functions at a different level than notifications - because we're wrapping the persistence methods (and others), potentially this opens the door to other interesting ways we could annotate spans. We have access to a lot of internals this way, and can get information that just isn't available in notifications.

3. I think we're viewing this as an evolutionary step in our auto-instrumentation. The door is not shut to notification-based instrumentation at all, it's mostly that no one has done the work yet to prove why it's the right way to move forward. And, even if we get some based on notifications, it may or may not complement or supplant what this PR does.

I would say if you're interested, let's chat a bit more on another issue or a discussion - this PR should probably go forward independently of notification discussions 😄

robertlaurin:
> How does this compare to instrumenting ActiveRecord through a subscription to [ActiveSupport::Notifications](https://guides.rubyonrails.org/active_support_instrumentation.html) [`sql.active_record`](https://guides.rubyonrails.org/active_support_instrumentation.html#active-record) events?

@robbkidd Subscribing to `sql.active_record` is more comparable to instrumenting the database adapter ([mysql2 as an example](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/instrumentation/mysql2)).

The `sql.active_record` topic has the nice characteristic that you can subscribe to it and could likely get away without adding any database instrumentation.  But also comes with the disadvantages of using Active Support Notifications, we have been using this approach for tracing in production for quite some time now and it requires managing all the the open spans so that you can close them in the event that one of the notification subscribers raises and prevents your spans from properly closing.

The intention of this PR is to instrument Active Record, and not just a tight wrapper around the db calls.  From this approach we will be able to surface more timing information about persistence actions performed.  

I'll provide an example to better illustrate my point.

Consider a simple User model with a slow callback.
```rb
class User < ApplicationRecord
  after_commit :take_a_nap

  private

  def take_a_nap
    sleep 2
  end
end
```

The corresponding controller.
```rb
class UsersController < ApplicationController
  def index
    @users = User.all
  end

  def new
  end

  def create
    User.create!
    redirect_to users_path
  end
end
```

image

If we know that the `sql.active_record` notification is just a tight wrapper around the db calls(being, insert, commit), you can see in the screenshot that it doesn't really provide much insight into why the request is performing so slowly.  

There are future improvements to this such as instrumenting callbacks.

All this is to say this is just a first step to instrumenting Active Record, I fully expect us to add the option to instrument `sql.active_record` as well as any other topic.

robbkidd:
I finally added the SIG meeting to my calendar, but yesterday I had a conflict. I'll be tuning in to future ones!

@ahayworth @robertlaurin That you both for the explanations! :bow: That clears up the differences for me. 

@robertlaurin ... You had me at "instrumented ActiveRecord callbacks." 😍 

sending love

robertlaurin:
@ahayworth I added a new commit here https://github.com/open-telemetry/opentelemetry-ruby/pull/653/commits/09cffd1458e1d109a82b8e76014995660765907f

If you get some time it would be cool if you try out (even locally) and see if this is still having the same issue as before.  It seems to be patching reliable for me now.

ahayworth:
@robertlaurin We're off work this week, so I'll test this out more thoroughly on Monday. But, at least in my personal, test app, it seems to work:

Screen Shot 2021-07-07 at 9 29 13 AM

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
```diff
@@ -0,0 +1,68 @@
...
+        def insert_class_methods_supported?
+          gem_version >= Gem::Version.new('6.0.0')
+        end
+
+        def gem_version
+          Gem.loaded_specs['activerecord'].version
+        end
+
+        def patch
+          # ::ActiveRecord::Querying.prepend(Patches::Querying)
```
fbogsany, line 38: Curious about the comments in this method - are these documenting what's actually being patched? Or are they just left over from earlier efforts?

robertlaurin: I'll clean this up and leave a proper comment about what's being patched.  Effectively what I was doing before was patching the modules that were being included/extended in ActiveRecord::Base.  Now I'm patching ActiveRecord::Base directly, it appears to consistently have the desired behaviour.

fbogsany (approved):
Minor comment about comments, otherwise this LGTM 👍

fbogsany (approved):

PR #654, opened 2021-03-15
feat!: pass env to url quantization rack config to allow more flexibility

aislinncronin:
This arose from needing a way to more flexibly quantize URLs that have params in the path. This change passes env to the configured `url_quantization` callable, allowing for access to things like `REQUEST_METHOD` that can be used to lower cardinality in these cases.


fbogsany (approved):
Thanks @aislinncronin - I updated the PR title to reflect that this is a breaking change (the signature of the callable changes).

PR #655, opened 2021-03-15
feat: add rack instrumentation config option to accept callable to filter requests to trace

aislinncronin:
This PR adds a configuration option to Rack instrumentation that allows you to configure with a callable that accepts `env` and does not trace any requests in which the callable returns true.  This allows more flexibility in filtering out requests to be untraced than the existing config option of `untraced_endpoints`.

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
@@ -29,7 +29,7 @@ module OpenTelemetry
         option :retain_middleware_names,  default: false, validate: :boolean
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
-        option :filtered_requests,        default: nil,   validate: :callable
+        option :untraced_requests,        default: nil,   validate: :callable
 
         private
 
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
@@ -29,6 +29,7 @@ module OpenTelemetry
         option :retain_middleware_names,  default: false, validate: :boolean
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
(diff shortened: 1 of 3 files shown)


instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb
```diff
@@ -29,6 +29,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         option :retain_middleware_names,  default: false, validate: :boolean
         option :untraced_endpoints,       default: [],    validate: :array
         option :url_quantization,         default: nil,   validate: :callable
+        option :filtered_requests,        default: nil,   validate: :callable
```
fbogsany, line 32: For consistency with `untraced_endpoints`, I think we should call this `untraced_requests`:
```suggestion
        option :untraced_requests,        default: nil,   validate: :callable
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -89,6 +89,11 @@ def untraced_request?(env)
             @untraced_endpoints.include?(env['PATH_INFO'])
           end
 
+          def filtered_request?(env)
+            filter = config[:filtered_requests]
+            filter ? filter.call(env) : false
+          end
```
fbogsany, line 95: I'd combine these two:
```suggestion
            return true if @untraced_endpoints.include?(env['PATH_INFO'])
            return true if config[:filtered_requests]&.call(env)
            
            false
          end
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -53,7 +53,7 @@ def initialize(app)
           end
 
           def call(env) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
-            if untraced_request?(env)
+            if untraced_request?(env) || filtered_request?(env)
```
fbogsany, line 56: I think we can combine the two methods, so this remains:
```suggestion
            if untraced_request?(env)
```

fbogsany (changes requested):

fbogsany (approved):

PR #656, opened 2021-03-16
refactor!: Baggage Overhaul

mwear:
While reviewing some recent context prop PRs it became clear that the Baggage system was in need of some attention. It was still largely in a prototype state from the OTEP-66 era. What started off as a little cleanup turned into a pretty big overhaul. Some of which might be discussion worthy. Below is a summary of the changes:

* We now retain and propagate Baggage properties (aka metadata). While metadata isn't supported by OpenTelemetry (at this time) we should propagate metadata we receive from other W3C compliant implementations.
  * Note, we only save the metadata received on extract and propagate it as-is on inject; there is no validation or further processing. We do not expose metadata to end-users, but currently it's only protected via code comment
* Updated inject to adhere to limits from the W3C baggage spec 
* ~~Renamed API methods from `set_value` `value`, and `values` to `set_entry`, `entry`, and `entries`~~
  * ~~In order to preserve metadata, baggage was changed from being stored as a single value, to being represented as an `Entry` with `value` and `metadata` readers. With the old value-named API methods you'd commonly see `OpenTelemetry.baggage.value('foo').value` which was pretty unfortunate and disappointing to use. Changing the names to be entry-centric fixess that. The spec uses "value" terminology, but it doesn't appear to force us to use it: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/baggage/api.md. The new names are consistent with the JS API: https://github.com/open-telemetry/opentelemetry-js-api/blob/main/src/baggage/Baggage.ts.~~
* Moved the SDK::Baggage::Manager to the API. The Noop Baggage::Manager is still part of the API and is the default if an SDK has not been configured.
  * One reason for this is that the baggage propagation is part of the API and it was incredibly hard to test without having an operational baggage manager. I had to reinvent most of an implementation by way of mocks, and it was pretty terrible.
  * Another insight I had, was that the Baggage::Manager had no dependencies on the SDK, and I could see an alternative SDK implementation wanting to reuse the operational Baggage::Manager from the API.
* Lastly, everything that interacts with baggage does so through the baggage manager. Much of this code was leftover from the OTEP-66 prototype and was reaching into the context and relying on insider knowledge of how baggage works.


fbogsany:
The spec requires the following signatures:
- `set_value(String, String [, Metadata] [, Context])`
- `get_value(String [, Context]) -> String`
- `remove_value(String [, Context])`
- `get_all_values([Context]) -> Enumerable<(String, String)> or Hash<String, String>`

The one thing that isn't specified is a way to access the metadata. The API in this PR exposes the indirection through `Entry`. I think that's less flexible, more verbose, and harder to use than the specified signatures.

The `TextMapPropagator` provided by the API is the only thing currently required to support metadata propagation. It can achieve this with "private" knowledge of the implementation, or perhaps an `# @api private` metadata getter: `get_metadata(String [, Context]) -> Metadata`. In practice, the metadata is likely an opaque string.

mwear:
I updated the method signatures to match the spec. I used the more idiomatic Ruby names of `value` for `get_value` and `values` for `get_all_values`, which is what they were named previously. I added a rogue `raw_entries` method with an `@api private` annotation for the injector to iterate over entries with metadata.

api/lib/opentelemetry/baggage/noop_builder.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Baggage
+    # No op implementation of Baggage::Builder
+    class NoopBuilder
```
fbogsany, line 10: Does the `NoopBuilder` also require an implementation of `entries`?
```suggestion
    class NoopBuilder
      def entries; {}; end

```

fbogsany: 🤔 or should `entries` be `@api private` in `Builder`? I think it is only meant to be called from `Manager`.

api/lib/opentelemetry/baggage/builder.rb
```diff
@@ -6,13 +6,37 @@
...
 module OpenTelemetry
   module Baggage
-    # No op implementation of Baggage::Builder
+    # Operational implementation of Baggage::Builder
     class Builder
-      def set_value(key, value); end
+      attr_reader :entries
 
-      def remove_value(key); end
+      def initialize(entries)
```
fbogsany, line 13: Should this be `@api private`? I think it is only supposed to be called from `Manager`.

api/lib/opentelemetry/baggage/noop_builder.rb
mwear: It shouldn't need one. An operational manager will call Builder#entries in order create a new context containing them, but the NoopManager does nothing. I could add an @api private to the existing `entries` on the operational Manager to clear that up as it's not meant to be exposed to the user.

fbogsany (approved):

PR #659, opened 2021-03-16
feat: Add instrumentation for postgresql (pg gem)

### LINKED ISSUE #523: Support for Postgresql
Postgresql is also a very common engine used together with Ruby on Rails. 
Implement Postgresql instrumentation similar to the one for [Mysql](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb).

BTW, wouldn't it be easier to create instrumentation for ActiveRecord and Sequel than for each database or that would not give enough information?

ahayworth:
This commit adds tracing for postgresql, via the `pg` gem.

The following methods are traced:
- exec / query / sync_exec / async_exec
- exec_params / sync_exec_params / async_exec_params
- prepare / sync_prepare / async_prepare
- exec_prepared / sync_exec_prepared / async_exec_prepared

We instrument all of these methods via meta-programming, since otherwise
it'd be incredibly repetitive. The methods can actually be broken down
into three groups - "exec-ish", "prepare-ish", and "exec prepared-ish"
methods. Every method within a group takes the same arguments, and so
we can trace each method in a group in an identical fashion.

Note that the 'sync' and 'async' variants require no special handling,
because they are not async interfaces _per se_ - rather, they are
non-blocking. See the `pg` docs for more info: https://www.rubydoc.info/gems/pg/PG%2FConnection:exec

We've implemented SQL sanitization the same way that the mysql2 PR did
it - which seems to be done by copying NewRelic. We should consider
extracting that from the NewRelic agent and making it a common shared
library here.

Possibly worth discussion is how we've handled the span name. We've
chosen the format `OPERATION_NAME DATABASE_NAME`, taking the first word
in the SQL string as the operation if it matches a known good list we
pulled from the Postgres docs. This is similar to how the python and
javascript postgres implementations behave, but is slightly different
than how the ruby mysql instrumentation does it. When falling back to
safe value, we also choose to simply use `DATABASE_NAME`, which seems to
be in line with current semantic conventions.

*Definitely* worth discussion is the LRU cache added in d119f8a5 - this tracks
the 50 most recently-prepared statements that we've traced; so that we can
attach it to the span as `db.statement`. Otherwise, if you're using a framework
like Rails that will aggressively prepare statements when the database supports
it, then you just see a lot of 'EXECUTE' calls with no real information attached.

While tests pass and we think this is safe, it should be noted that we
haven't actually put this into production testing yet - there could be
horrible bugs still. :)

Fixes #523


ahayworth:
I'm figuring out how to sign the CLA as part of GitHub - hopefully that should be rectified soon. 😄

ahayworth:
I'm a little mystified as to how I've broken the `mysql` tests, but if anyone has any ideas on how to get another database container running in Actions I'd love to defer to your expertise. 😄

fbogsany:
> We've implemented SQL sanitization the same way that the mysql2 PR did
> it - which seems to be done by copying NewRelic. We should consider
> extracting that from the NewRelic agent and making it a common shared
> library here.

👍 we can do that now or when we hit the next use-case. I asked that we initially tailor it specifically to MySQL to start with, and generalize it later. While we are not required to by the license, it is probably a nice gesture to ask NewRelic if they want to donate that code to OpenTelemetry, if we're going to copy it wholesale.

ahayworth:
> 👍 we can do that now or when we hit the next use-case. I asked that we initially tailor it specifically to MySQL to start with, and generalize it later. While we are not required to by the license, it is probably a nice gesture to ask NewRelic if they want to donate that code to OpenTelemetry, if we're going to copy it wholesale.

Agreed! Do you know anyone there that might be interested in weighing in? Otherwise I can just go open an issue and see how they feel.

.github/workflows/ci.yml
```diff
@@ -86,8 +93,13 @@ jobs:
...
           TEST_KAFKA_PORT: 29092
-          TEST_MYSQL_HOST: sql
-          TEST_MYSQL_PORT: ${{ job.services.mysql.ports['3306'] }}
+          TEST_MYSQL_HOST: mysql
+          TEST_MYSQL_PORT: 3306
+          TEST_POSTGRES_PASSWORD: postgres
+          TEST_POSTGRES_USER: postgres
+          TEST_POSTGRES_HOST: postgres
+          TEST_POSTGRES_PORT: 5432
+          TEST_POSTGRES_DB: postgres
```
ahayworth, line 102: I had a heck of a time getting the tests to run until I set these ports to their default ports. I was unable to convince the mysql tests to find the mysql container when I added the _postgres_ container - suggesting to me that something weird was going on with docker networking and Actions. But this fix worked, and I figured it's ~fine.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/constants.rb
```diff
@@ -0,0 +1,124 @@
...
+          SELECT
+          SET
+          SHOW
+          START
+          TRUNCATE
+          UNLISTEN
+          UPDATE
+          VACUUM
+          VALUES
+        ].freeze
```
ahayworth, line 66: I don't honestly know if all of these are valuable, but I'm erring on the side of including them until we have a better answer for tuning instrumentation noise.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -0,0 +1,155 @@
...
+            # When SQL is being sanitized, we know that this cache will
+            # never be more than 50 entries * 2000 characters (so, presumably
+            # 100k bytes - or 97k). When not sanitizing SQL, then this cache
+            # could grow much larger - but the small cache size should otherwise
+            # help contain memory growth. The intended use here is to cache
+            # prepared SQL statements, so that we can attach a reasonable
+            # `db.sql.statement` value to spans when those prepared statements
+            # are executed later on.
+            @lru_cache ||= ::LruRedux::ThreadSafeCache.new(50)
+          end
```
ahayworth, line 63: I'm not sure if 50 is the right number here - I mostly just made it up, did the math, and thought "gee that seems fine."

instrumentation/pg/lib/opentelemetry/instrumentation/pg/version.rb
```diff
@@ -0,0 +1,14 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module PG
+      VERSION = '0.15.0'
+      MIN_PG_GEM_VERSION = '1.1.0'
+    end
```
ahayworth, line 12: The `exec`/`async_exec`/`sync_exec` methods were added in this version, which is now 5 years old. In theory, if someone really wanted to, we could support earlier versions than this but ¯\\_(ツ)_/¯

instrumentation/pg/opentelemetry-instrumentation-pg.gemspec
```diff
@@ -0,0 +1,47 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'lru_redux', '~> 1.1'
```
ahayworth, line 28: I picked this particular implementation of an LRU cache primarily because it seems to be far and away the most downloaded on rubygems. It's a poor proxy of "is this a good choice," but I feel good that many others have made the same decision. 😆

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -0,0 +1,155 @@
...
+
+          def obfuscate_sql(sql)
+            return sql unless config[:enable_sql_obfuscation]
+
+            # Borrowed from opentelemetry-instrumentation-mysql2
+            return 'SQL query too large to remove sensitive data ...' if sql.size > 2000
+
+            # From:
+            # https://github.com/newrelic/newrelic-ruby-agent/blob/9787095d4b5b2d8fcaf2fdbd964ed07c731a8b6b/lib/new_relic/agent/database/obfuscator.rb
+            # https://github.com/newrelic/newrelic-ruby-agent/blob/9787095d4b5b2d8fcaf2fdbd964ed07c731a8b6b/lib/new_relic/agent/database/obfuscation_helpers.rb
```
ericmustin, line 113: 👋 Hey @mwlang hope all is well. At the SIG today we were chatting on how useful this obfuscation code is for orm/db instrumentation in ruby. It would be wonderful to be able to re-use the Obfuscation Module in opentelemetry-ruby directly. Would you be open to making a contribution to add this if you have bandwidth? And if not, any feedback or guidance on this approach?

ahayworth:
Update: I've removed the dependency on `lru_redux`, and implemented our own minimal LRU cache instead.

johnnyshields:
FYI using this in production, looking very good thanks.

ahayworth:
@johnnyshields Do you have any thoughts on the number of spans its creating? We're not just instrumenting common things like `SELECT`, `UPDATE`, etc - I'm curious if you have any thoughts on if it's perhaps too noisy.

Also, yay! I'm glad it's working well for you!

johnnyshields:
@ahayworth yes, having extra items like `BEGIN`, `COMMIT` etc. are very useful to see. In fact, having these extra spans helped us debug an issue yesterday. In the below example you can see clearly that when using transactions the latency doesn't happen on the `INSERT` step but on the `COMMIT`.

image

ahayworth:
@johnnyshields Thanks - that's good feedback, I appreciate it!

renchap:
Would it make sense to report a transaction as a whole span (starting with a `BEGIN` (or `START TRANSACTION`) and ending with a `COMMIT` or `ROLLBACK`, rather than having two different spans?
`BEGIN` is never supposed to block AFAIK, so we could have a transaction span, and children spans for the individual queries + the final `COMMIT` statement.

ahayworth:
> Would it make sense to report a transaction as a whole span (starting with a BEGIN (or START TRANSACTION) and ending with a COMMIT or ROLLBACK, rather than having two different spans?

The only real downside I see there is that we'd need to do much more parsing of the SQL statements we're getting in the library (they're just strings at this point before they get handed off to the lower-level C extensions in `pg`). That would complicate the PR quite a bit, so I think I'd want to leave that out for now.

ahayworth:
Update: I've been running this in production now for awhile, and have seen no problems from it thus far.

johnnyshields:
Let's merge!

mwear (approved):
LGTM. We have multiple people running this in prod with good results.

mwear:
Any objections to me merging this? If not, I will, but this branch needs to have main merged in.

instrumentation/pg/lib/opentelemetry/instrumentation/pg/version.rb
```diff
@@ -0,0 +1,14 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module PG
+      VERSION = '0.16.0'
+      MIN_PG_GEM_VERSION = '1.1.0'
```
robertlaurin, line 11: Minor nitpick, previously we have put this definition in the base instrumentation class that invokes it.  Whether or not that is the right place for it I'd like to keep that consistence even if we end up moving them all here later.

robertlaurin (approved):
This looks good me.  Thanks a bunch for this contribution :)

ericmustin (approved):

instrumentation/pg/lib/opentelemetry/instrumentation/pg/version.rb
ahayworth: @robertlaurin No worries - I made that change in 409a7d09 . I think I had originally put it in `version.rb` to re-use it in the gemspec, but I agree that the overall consistency is nicer. It's not like we expect it to change very often, so my original location was optimizing for the wrong thing! 😄

ahayworth:
@mwear `main` is merged and the PR is updated to deal with the new `opentelemetry-instrumentation-base` extraction. 😄

mwear:
🚀 Thanks @ahayworth!

johnnyshields:
Great job everyone, really appreciate this. Please cut a release soon.

PR #661, opened 2021-03-17
feat: added http.route in rails instrumentation to match the spec

angeliski:
# Description
The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions) says:

> Given an inbound request for a route (e.g. "/users/:userID?") the name attribute of the span SHOULD be set to this route. If the route does not include the application root, it SHOULD be prepended to the span name.
> 
> If the route cannot be determined, the name attribute MUST be set as defined in the general semantic conventions for HTTP.

Some exporters need more information to provide useful insights to the user,
so add `http.route` could make more easily convert traces to the final destination.

My current case is using Datadog: The [exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/translate_traces.go#L440) uses the `http.route` to define how the resource will be shown, so I think makes sense to add here that information.

I created a really simple implementation, so any feedback would be useful 

The result is something like that:

Screenshot from 2021-03-17 16-20-46

### DIFF
--- a/instrumentation/rails/README.md
+++ b/instrumentation/rails/README.md
@@ -33,7 +33,7 @@ end
 
 ### Configuration options
 
-The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/journey/router.rb#L65)
+The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/v6.1.3/actionpack/lib/action_dispatch/journey/router.rb#L65)
 ```ruby
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Rails', {
--- a/instrumentation/rails/README.md
+++ b/instrumentation/rails/README.md
@@ -30,6 +30,18 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+
+### Configuration options
+
+The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/journey/router.rb#L65)
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Rails', {
+    enable_recognize_route: true
+  }
+end
+```
+
 ## Examples
 
 Example usage can be seen in the `./example/trace_demonstration.rb` file [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rails/example/trace_request_demonstration.ru)
--- a/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
+++ b/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
@@ -22,6 +22,8 @@ module OpenTelemetry
           defined?(::Rails)
         end
 
+        option :enable_recognize_route, default: false, validate: :boolean
+
         private
 
         def require_dependencies
--- a/instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
+++ b/instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
@@ -15,10 +15,20 @@ module OpenTelemetry
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
               rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
 
+              add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
+              super(name, request, response)
+            end
+
+            private
+
+            def add_rails_route(rack_span, request)
               ::Rails.application.routes.router.recognize(request) do |route, _params|
                 rack_span.set_attribute('http.route', route.path.spec.to_s)
               end
-              super(name, request, response)
+            end
+
+            def instrumentation_config
+              Rails::Instrumentation.instance.config
             end
           end
         end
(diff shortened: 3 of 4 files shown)


instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
```diff
@@ -14,6 +14,10 @@ module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
               rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
+
+              ::Rails.application.routes.router.recognize(request) do |route, _params|
```
robertlaurin, line 18: When I originally implemented this instrumentation I had some concerns regarding the overhead of running [.recognize](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/journey/router.rb#L65) a second time for each request to the application.  

I'd like to get a second reviewer to weigh in on it, and if they don't share my same concerns then I have no reason to block on this.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
```diff
@@ -14,6 +14,10 @@ module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
               rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
```
robertlaurin, line 16: The reason we opted for this approach is that we believe using the controller name and action as the span name is more conventional in the context of a rails application.

As you pointed out this is not following the spec, would it make more sense for your use case if we added a configuration that allowed you to set the span name to the path template instead?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
angeliski: Yes, I can understand your concern, I didn't find a better way to get the route here. Maybe we can add a configuration to enable that behavior?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
angeliski: More or less, the spec is a little confused about that.

About the [name](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span)
> The span name concisely identifies the work represented by the Span, for example, an RPC method name, a function name, or the name of a subtask or stage within a larger computation. The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable. That is, "get_user" is a reasonable name, while "get_user/314159", where "314159" is a user ID, is not a good name due to its high cardinality. Generality SHOULD be prioritized over human-readability.

So in that case make sense to me, in a rails app, it behavior you used.
But we have a conflict when we start to talk about the [HTTP Server semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#http-server-semantic-conventions)

> Given an inbound request for a route (e.g. "/users/:userID?") the name attribute of the span SHOULD be set to this route. If the route does not include the application root, it SHOULD be prepended to the span name.
>
> If the route cannot be determined, the name attribute MUST be set as defined in the general semantic conventions for HTTP.

In that case, the span name AND route should be the same? That doesn't make sense to me (could be a problem in specification not here).

The problem (to me) in only change the span name is the current behavior in [Datadog exporter](https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/datadogexporter/translate_traces.go#L441) they use the route to build the resource (and show grouped in UI). I only thought of adding a route here, but maybe we need to change the exporter behavior to look to other attributes.

angeliski: Hey, @ericmustin I see you have some commits in Datadog exporter, could you help us here in that discussion?
What do you think is the better approach?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/patches/action_controller/metal.rb
fbogsany: 👍 I'd be pro-config-option

robertlaurin: Agreed, if we put this behind a config option that defaults to it being off I think that's a good compromise for bringing this functionality in.

angeliski: Great! I will work on it

angeliski: Hey team, I updated the code, please take a look

instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
```diff
@@ -22,6 +22,8 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
           defined?(::Rails)
         end
 
+        option :enable_recognize_route, default: false, validate: :boolean
```
robertlaurin, line 25: I went back and forth on wether or not this was the right name for this setting, typically I wouldn't want to associate the name with the implementation (ie. `router.recognize`).  But in the case I think this makes sense as the reason we are putting this behind a config is due to performance concerns.

instrumentation/rails/README.md
```diff
@@ -30,6 +30,18 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+
+### Configuration options
+
+The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/journey/router.rb#L65)
```
robertlaurin, line 36: Instead of pointing to the main branch which may lose context as it changes, lets just use the current latest tag.
```suggestion
The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/v6.1.3/actionpack/lib/action_dispatch/journey/router.rb#L65)
```

robertlaurin (approved):
I left one suggested change around the readme that I think should come in before merge, otherwise this looks good. 

Thank you.

fbogsany (approved):

PR #680, opened 2021-03-23
fix: Fix Faraday gem dependencies.

fhwang:


instrumentation/faraday/Gemfile
```diff
@@ -7,10 +7,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
-gem 'opentelemetry-api', path: '../../api'
-
-group :test do
-  gem 'opentelemetry-common', path: '../../common'
-  gem 'opentelemetry-sdk', path: '../../sdk'
-end
```
fbogsany, line 16: This stuff is generally required during release builds, when we're bumping versions of all gems at the same time.

fbogsany (approved):

PR #696, opened 2021-03-30
feat: Processors validate exporters on init.

fhwang:
It might reduce configuration headaches if we raised exceptions with helpful messages immediately. I'm assuming here that processors are only ever instantiated on application boot, though I'm not finding much in the spec or elsewhere talking about that specifically.

See #595 for an example of what this could help avoid.


sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -57,6 +57,10 @@ def valid_attributes?(attrs)
           end
         end
       end
+
+      def valid_exporter?(exporter)
+        exporter && %i[export shutdown force_flush].all? { |m| exporter.respond_to?(m) }
+      end
```
fbogsany, line 63: It might be useful to move this to the `opentelemetry-common` gem. We provide a couple of implementations of span processors in the SDK, but clients can implement their own, and they could benefit from calling this too.

fbogsany (approved):

PR #697, opened 2021-03-31
fix!: remove metrics API

fbogsany:
Our Metrics API is old and incomplete. The metrics signal is not yet stable. This PR removes the Metrics API completely. We will start a new implementation in a separate gem, and move it into the API gem when we consider it stable.


ericmustin (approved):
love to see all this 🟥 🟥 🟥 🟥 🟥 🟥

robertlaurin (approved):

PR #698, opened 2021-03-31
refactor!: extract instrumentation base from api

robertlaurin:
Extracts the instrumentation base and registry modules from the API.   Neither is part of the API spec so we are extracting them prior to the 1.0 release.


instrumentation/base/lib/opentelemetry/instrumentation.rb
```diff
@@ -4,12 +4,20 @@
...
 module OpenTelemetry
   # The instrumentation module contains functionality to register and install
   # instrumentation
   module Instrumentation
+    extend self
+
+    # @return [Instrumentation::Registry] registry containing all known
+    #  instrumentation
+    def registry
+      @registry ||= Instrumentation::Registry.new
```
fbogsany, line 20:
```suggestion
    # @return [Registry] registry containing all known
    #  instrumentation
    def registry
      @registry ||= Registry.new
```

instrumentation/base/README.md
```diff
@@ -0,0 +1,30 @@
+# OpenTelemetry Base Instrumentation
+
+Todo:
```
fbogsany, line 3: - [ ] Describe how this is to be used.

.toys/.data/releases.yml
```diff
@@ -31,6 +31,10 @@ commit_lint:
 #  *  changelog_path: Path to CHANGLEOG.md relative to the gem directory.
 #     (Required only if it is not in the expected location.)
 gems:
+  - name: opentelemetry-instrumentation-base
+    directory: instrumentation/base
+    version_constant: [OpenTelemetry, Instrumentation, Base, VERSION]
```
fbogsany, line 36:
```suggestion
    version_constant: [OpenTelemetry, Instrumentation, VERSION]
```

ahayworth (approved):
I really like the idea of moving this out into its own gem. I personally think `Instrumentation::Base` is a nice thing that library authors _should_ use, but it's nice to move it out on its own - since it's not strictly necessary. 😄

instrumentation/base/README.md
```diff
@@ -0,0 +1,30 @@
+# OpenTelemetry Base Instrumentation
```
fbogsany, line 1:
```suggestion
# OpenTelemetry Instrumentation Base
```

ericmustin (approved):
I _think_ this lgtm? it's possible i missed something as this has a lot of files changed but overall i think it's the right approach

instrumentation/base/README.md
```diff
@@ -0,0 +1,154 @@
...
+
+```
+gem install opentelemetry-instrumentation-base
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-base` in your `Gemfile`.
+
+### For SDK Authors
+
+The following is a simplified demonstration of how the `OpenTelemetry::Instrumentation.registry` can be used by an SDK to install auto-instrumentation gems as part of it's configuration.  This should not be used as an example of how to implement an SDK as there are large omissions.  For an example of a complete implementation see the `opentelemetry-sdk` gem.
```
fbogsany, line 17:
```suggestion
The following is a simplified demonstration of how the `OpenTelemetry::Instrumentation.registry` can be used by an SDK to install auto-instrumentation gems as part of its configuration.  This should not be used as an example of how to implement an SDK as there are large omissions.  For an example of a complete implementation see the `opentelemetry-sdk` gem.
```

fbogsany (approved):
Minor typo, otherwise LGTM 👍

PR #702, opened 2021-04-05
build: Redis instrumentation uses real redis in tests.

fhwang:


ahayworth:
@fhwang is there a reason that using docker (like how we do for mysql) isn't sufficient? You can pass along the `redis.conf` to the image (documented [here](https://hub.docker.com/_/redis) under the "Additionally, If you want to use your own redis.conf ..." header). It's not incredibly important, of course.

I do think it'd be nice to have the test suite spin up a redis for you locally, however, rather than assuming that the user has a redis installed and running.

ahayworth:
@fhwang I got redis tests working with a service container, as well as local docker-compose settings for the tests and examples:
- commit that gets it going: https://github.com/github/opentelemetry-ruby/commit/c1a757fb200d73da644f1ae86cbe80608710a95c
- minimal working CI run (just running one of the ruby tests, rather than all of them to shorten the feedback loop): https://github.com/github/opentelemetry-ruby/actions/runs/733376763

There's [another commit](https://github.com/github/opentelemetry-ruby/commit/0c28cd528510658fb3309fec6eda19d4cdfcda03) on that branch that was needed to run CI in a fork, but it isn't needed.

ericmustin:
@fhwang generally changes here lgtm, haven't dug into why tests fail but i see there was a comment above regarding that.

fhwang:
@ahayworth Thanks for finding that bitnami container!

As for whether we should run Redis locally via docker or install, I don't have a super-strong opinion, though I don't think of Redis as something that's a tough install. Is this something where we've set a project-wide standard?

ahayworth:
@fhwang I don't think so - and we can choose to eliminate the docker container here. I mostly added it for parity with the postgres, mysql, kafka containers (etc). As it stands, the commit I linked allows you to run it either way (and defaults to assuming redis is available on localhost). Whatever feels most right to you is what I'd support!

mwear (approved):
LGTM

mwear:
@fhwang Do you mind addressing the merge conflict on this PR and does anyone have concerns if I merge this afterwards?

fhwang:
@mwear Merge conflict has been addressed.

PR #707, opened 2021-04-07
feat: Add Elasticsearch instrumentation

mmontalvo:
Linked to open-telemetry/opentelemetry-ruby-contrib#8


mmontalvo:
to be done:
- update README
- add more tests
- add features not ported yet from dd-trace gem

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,75 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    module Elasticsearch
+      module Patches
+        # Module to prepend to Elasticsearch::Client for instrumentation
+        module Client
+          def self.included(base) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
+            base.class_eval do # rubocop:disable Metrics/BlockLength
+              alias_method :perform_request_without_otel, :perform_request
```
robertlaurin, line 17: I know this is still in draft, but wanted to provide the heads up that we have been consistently sticking to prepending instead of method aliasing for our instrumentation patches.

mmontalvo: Updated the PR and moved away from `class_eval` 👍 

Thanks @robertlaurin!

instrumentation/elasticsearch/Appraisals
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+appraise 'elasticsearch-7.12' do
```
arielvalentin, line 7: @fbogsany Does this [matrix build](https://github.com/open-telemetry/opentelemetry-ruby/blob/1dfea8504716d2a97c3ae1df6be0f523e7ac77e4/.github/workflows/ci.yml#L14) run appraisals without any additional configuration?

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,68 @@
...
+            ) do |span|
+              response = super(*args)
+              span.set_attribute('http.status_code', response.status)
+            end
+            response
+          end
+
+          private
+
+          def tracing_attributes(*args)
```
arielvalentin, line 31: I am not certain we want to add these attributes to the instrumentation since I do not think there aren't any [semantic conventions defined for Elasticsearch](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md).

It looks like the [Java Instrumention](https://github.com/search?q=org%3Aopen-telemetry+otel.instrumentation.elasticsearch.experimental-span-attributes&type=Code) uses a flag to enable experimental tags that are specific to that instrumentation. 

Reference: https://github.com/open-telemetry/opentelemetry-java-instrumentation/pull/2402 

I think that is something we should consider here as well and only limit the tags we capture by default. Thoughts?

cc: @fbogsany @mwear @robertlaurin @ericmustin @ahayworth

arielvalentin:
I also failed to welcome you @mmontalvo ! Thanks for helping get this instrumentation implemented!!!

mmontalvo:
> I also failed to welcome you @mmontalvo ! Thanks for helping get this instrumentation implemented!!!

Thanks @arielvalentin! And also thanks for your review! 👍

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
ericmustin: tbh I'm not sure what the right approach is here regarding non semantic convention attributes when there aren't any specifics for elasticsearch in general. I will review a bit closer this wknd, I think Java's approach is reasonable but I also think it adds a higher config bar for the average user. maybe that's fine

one thing that is in the semantic conventions referenced by @arielvalentin is the attribute `db.system: elasticsearch` which I don't see in this PR. I think we should add that.

Another is that I think we should  make an attempt to quantize the `body`  and `url`, if possible, similar to what we've done for pg and mysql spans, as a config option.  We can leverage the datadog approach https://github.com/DataDog/dd-trace-rb/blob/339c1d2aea1752c43fb34d26a307938f24307ba5/lib/ddtrace/contrib/elasticsearch/patcher.rb#L95 or write our own, I don't have strong opinions here.

Lastly, i'm not sure what the right `span.name` value would be for elasticsearch here. Is a static value the correct approach?

mmontalvo: Thanks for the review @ericmustin 

|> attribute db.system: elasticsearch which I don't see in this PR. I think we should add that
sure, I will add it

Happy to take a look at the Java implementation and will wait for your review this weekend.

|> quantize the `body` and `url`
I was in doubt how to include it: "taking" the dd-trace [approach](https://github.com/DataDog/dd-trace-rb/blob/master/lib/ddtrace/contrib/elasticsearch/quantize.rb) or implement ours. 

Thanks!

ahayworth: From the semantic conventions spec:

> In OpenTelemetry spans can be created freely and it’s up to the implementor to annotate them with attributes specific to the represented operation. 

The "implementor" here is us, I believe - and of course the text goes on to talk about why we should standardize as much as possible. But what this tells me is that we are not restricted to only adding the attributes already defined; but rather that we should adhere to the conventions whenever a convention exists.

Given that `elasticsearch` is mentioned in the database semantic conventions, I think that means we should ensure that we add any of the attributes defined there that are applicable (and of course, we should add any of the other attributes - such as the general attributes, rpc attributes, http attributes, etc - as applicable).

But I also think we should feel free to add any _other_ attributes that we think are necessary, even if they're not in the conventions already. We can use other language implementations as a guide, but they're not necessarily authoritative either. When we're ready to merge this, if it seems like there should be an update to the semantic conventions, we should do that - but I don't think that should block us from adding attributes in any way.

So - bottom line for me is "I think we should add them, but make sure they follow any / all applicable conventions that are already defined, and add definitions upstream to the spec if it makes sense."

FWIW - I don't think we should hide them behind a flag; that seems like a lot of friction to me. I don't think that's how the semantic conventions are supposed to be interpreted, personally.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/instrumentation.rb
```diff
@@ -0,0 +1,33 @@
...
+      # The Instrumentation class contains logic to detect and install the Elasticsearch instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        install do |_config|
+          require_dependencies
+          patch_client
+        end
+
+        present do
+          defined?(::Elasticsearch::Transport)
+        end
```
ahayworth, line 19: @mmontalvo - is there any minimum version that we need to check for? Or does this work for all versions?

(An example of how to do a version check is [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb#L23-L25))

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry'
```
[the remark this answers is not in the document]
ahayworth: I think it is required if you wish to inherit from `OpenTelemetry::Instrumentation::Base`, yeah.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,68 @@
...
+
+            params = JSON.generate(params) if params && !params.is_a?(String)
+            body = JSON.generate(body) if body && !body.is_a?(String)
+
+            {
+              'elasticsearch.url' => url,
+              'elasticsearch.method' => method,
+              'elasticsearch.params' => params,
+              'elasticsearch.body' => body || ''
+            }.merge(connection_attributes)
```
ahayworth, line 46: You might consider using the [http](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md) or [rpc](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md) conventions here, but I'm not sure that either one really applies (what do you think @mmontalvo ?)

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,68 @@
...
+
+          def connection_attributes
+            connection = transport.connections.first
+
+            host = connection.host[:host] if connection
+            port = connection.host[:port] if connection
+
+            {
+              'out.host' => host,
+              'out.port' => port
```
ahayworth, line 57: Unless `out` has a useful meaning for elasticsearch (I am only a casual user), then I think you may wish to instead follow the [network connection attributes](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes) instead.

ahayworth (commented):
I'm excited for this, I think it's a big missing piece in our auto-instrumentation story. I had a few questions/comments/suggestions, but nothing blocking per se. 😄

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
arielvalentin: Thanks for sharing this perspective.

I feel like I want to have a more comprehensive discussion around how to work with experimental conventions that aren't well defined but I also don't want to hold up this PR to resolve it or come to a conclusion. I'll open a discussion and reference this PR so that we could build consensus around this and then provide explicit guidance in a contributors guide. 

That being said, I think something we all agree on is ensuring we follow any existing conventions so at a minimum we should ensure this PR included those; however for those attributes we find useful or not explicitly defined in the spec we should choose naming that resembles the format for existing examples which are [structured](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/database.md#examples) using this format  `db.<system>.<attribute>`  

Thoughts?

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/instrumentation.rb
mmontalvo: From what I found, `perform_request` has been set to current implementation from [this commit](https://github.com/elastic/elasticsearch-ruby/commit/b7b1835638f69c30b9c1fb890a84d4e36806e3d8#diff-815d56877734c3f8654aeb742d159115bb48be9f6f775570a53d3e69a5be8f60R128)
I think we shouldn't need to worry about minor versions.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch.rb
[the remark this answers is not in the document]
mmontalvo: 👍

instrumentation/elasticsearch/.rubocop.yml
```diff
@@ -0,0 +1,27 @@
+AllCops:
```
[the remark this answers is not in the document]
mmontalvo: I was using the `bunny` instrumentation as an example, but after checking others I saw the shared file thing, thanks!

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
mmontalvo: I think we need to consider the http one 👍

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
mmontalvo: |> Unless `out` has a useful meaning for elasticsearch
Not really, it was only part of the dd-trace port.

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/instrumentation.rb
ahayworth: I think they [introduced the method](https://github.com/elastic/elasticsearch-ruby/commit/a2bbb9c5b6b3cb0f062d751eca975843502b571b) on May 17, 2013, which would put the minimum supported version around `0.1.1` - so I agreee, we probably don't need a minimum version check here. 😆

instrumentation/elasticsearch/Appraisals
robertlaurin: You can see here the different configs for the different rubies -> https://github.com/open-telemetry/opentelemetry-ruby/blob/c4949ac8f51822d068c7a8de2a6baddf6970fcc6/.github/workflows/ci.yml#L17-L25

The `--include-appraisal` flag corresponds to these commands -> https://github.com/open-telemetry/opentelemetry-ruby/blob/c4949ac8f51822d068c7a8de2a6baddf6970fcc6/.toys/ci.rb#L63-L64

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
```diff
@@ -0,0 +1,68 @@
...
+            }
+          end
```
robertlaurin, line 59: I think we can simplify this instrumentation patch by moving things inline.  Always a little apprehensive in leaving a block of code as a comment but I think it's the simplest way to illustrate what I mean.
```suggestion
          def perform_request(method, path, params = {}, body = nil, headers = nil)
            attributes = {
              'http.target' => path,
              'http.method' => method,
              'elasticsearch.params' => params,
              'elasticsearch.body' => body || ''
            }

            attributes['elasticsearch.params'] = JSON.generate(params) if params && !params.is_a?(String)
            attributes['elasticsearch.body'] = JSON.generate(body) if body && !body.is_a?(String)

            if (connection = transport.connections.first)
              attributes['http.host'] = connection.host[:host]
              attributes['net.peer.port'] = connection.host[:port]
            end

            tracer.in_span('elasticsearch.query', attributes: attributes, kind: :client) do |span|
              super.tap do |response|
                span.set_attribute('http.status_code', response.status)
              end
            end
          end

          private
```

instrumentation/elasticsearch/test/instrumentation/elasticsearch/patches/client_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+      client.search q: 'test'
+
+      _(span.name).must_equal 'elasticsearch.query'
+      _(span.attributes['out.host']).must_equal host
+      _(span.attributes['out.port']).must_equal port
+      _(span.attributes['elasticsearch.url']).must_equal '_search'
+      _(span.attributes['elasticsearch.method']).must_equal 'GET'
+      _(span.attributes['elasticsearch.params']).must_equal '{"q":"test"}'
+      _(span.attributes['elasticsearch.body']).must_equal ''
```
robertlaurin, line 45: >Attribute values expressing a numerical value of zero, an empty string, or an empty array are considered meaningful and MUST be stored and passed on to processors / exporters.

- [from the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes)

I have about zero time with elasticsearch outside the context of instrumentation 😄 .  Is an empty string body considered meaningful?

instrumentation/elasticsearch/test/instrumentation/elasticsearch/patches/client_test.rb
```diff
@@ -0,0 +1,86 @@
...
+
+  describe 'search query' do
+    let(:path) { '_search?q=test' }
+    it 'produces correct traces' do
+      stub_request(:get, url).to_return(status: 200)
+
+      client.search q: 'test'
+
+      _(span.name).must_equal 'elasticsearch.query'
+      _(span.attributes['out.host']).must_equal host
```
robertlaurin, line 40:
```suggestion
      _(span.attributes['http.host']).must_equal host
```
See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

instrumentation/elasticsearch/test/instrumentation/elasticsearch/patches/client_test.rb
```diff
@@ -0,0 +1,86 @@
...
+      _(span.attributes['out.port']).must_equal port
```
robertlaurin, line 41:
```suggestion
      _(span.attributes['net.peer.port']).must_equal port
```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/span-general.md#general-network-connection-attributes

instrumentation/elasticsearch/test/instrumentation/elasticsearch/patches/client_test.rb
```diff
@@ -0,0 +1,86 @@
...
+      _(span.attributes['elasticsearch.url']).must_equal '_search'
```
robertlaurin, line 42: I think `http.target` makes sense here
```suggestion
      _(span.attributes[http.target']).must_equal '_search'
```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

instrumentation/elasticsearch/test/instrumentation/elasticsearch/patches/client_test.rb
```diff
@@ -0,0 +1,86 @@
...
+      _(span.attributes['elasticsearch.method']).must_equal 'GET'
```
robertlaurin, line 43:
```suggestion
      _(span.attributes['http.method']).must_equal 'GET'
```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

instrumentation/elasticsearch/test/opentelemetry/instrumentation/elasticsearch/instrumentation_test.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
robertlaurin, line 1: I think this file is a repetition of https://github.com/open-telemetry/opentelemetry-ruby/pull/707/files#diff-0f805288429558033dbb57131c19985abaaa93c205e141983d70eb8f9471d475

Probably makes sense to remove this in favour of the other one.

instrumentation/elasticsearch/opentelemetry-instrumentation-elasticsearch.gemspec
```diff
@@ -0,0 +1,48 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'opentelemetry/instrumentation/elasticsearch/version'
+
+Gem::Specification.new do |spec|
+  spec.name        = 'opentelemetry-instrumentation-elasticsearch'
```
robertlaurin, line 12: This instrumentation library in its current state is only instrumenting [elasticsearch-transport](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport), which is a dependency of [elasticsearch](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch/elasticsearch.gemspec#L48-L49).  

I'm a little conflicted on whether or not we should be more explicit in that fact, or because the elasticsearch gem seems to just be a [bundler of the various elasticsearch "sub" gems](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch/lib/elasticsearch.rb) we should just leave this as is.  I'm leaning a bit towards the latter at the moment. 

@open-telemetry/ruby-approvers any stronger opinions here?

instrumentation/elasticsearch/lib/opentelemetry/instrumentation/elasticsearch/patches/client.rb
fbogsany: > Thoughts?

👍

instrumentation/elasticsearch/test/opentelemetry/instrumentation/elasticsearch/instrumentation_test.rb
mmontalvo: Sure! Not sure if it was generated automatically, but happy to remove it 👍

robertlaurin: If you used the instrumentation generator it would have generated a file like this.   But it looks like another one was added in addition to it.  I don't have a strong preference to which one gets kept 😄

instrumentation/elasticsearch/opentelemetry-instrumentation-elasticsearch.gemspec
ahayworth: Well ... at first I agreed, but now I'm not so sure. After reading through both `elasticsearch-transport` and `elasticsearch-api` ... I'm wondering if we should be instrumenting `elasticsearch-api` instead?

The default HTTP implementations supported by `elasticsearch-transport` are actually [provided by Faraday](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-transport/lib/elasticsearch/transport/client.rb#L165-L166), and we already have `opentelemetry-instrumentation-faraday`. I think this PR would actually duplicate many of those spans.

robertlaurin: >Well ... at first I agreed, but now I'm not so sure. After reading through both elasticsearch-transport and elasticsearch-api ... I'm wondering if we should be instrumenting elasticsearch-api instead?

I'm trying to figure out if this is a point in favour of just simply keeping this as a more generic "elasticsearch" instrumentation library to go the more focused route of `opentelemetry-instrumentation-elasticsearch_api` or something to that effect.

>The default HTTP implementations supported by elasticsearch-transport are actually provided by Faraday, and we already have opentelemetry-instrumentation-faraday. I think this PR would actually duplicate many of those spans.

It looks like it will select between these lower level http libs, which we also have instrumentation for (with the exception of Patron).

>For optimal performance, use a HTTP library which supports persistent ("keep-alive") connections, such as patron or Typhoeus. Just require the library (require 'patron') in your code, and it will be automatically used.

>Currently these libraries will be automatically detected and used:

* Patron
* Typhoeus
* HTTPClient
* Net::HTTP::Persistent

- [from the readme](https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport#elasticsearchtransport)

robertlaurin: > I agree, I think it makes sense.
> I will update the PR to instrument `elasticsearch-api` instead 👍
> 
> Thanks!

@mmontalvo mind leaving it for the moment?  Don't want you to change this before we settle on the approach.  Renaming can be a bit tedious.

instrumentation/elasticsearch/README.md
```diff
@@ -0,0 +1,53 @@
...
+
+The `opentelemetry-instrumentation-elasticsearch` gem source is [on github][repo-github], along with related gems including `opentelemetry-api` and `opentelemetry-sdk`.
+
+The OpenTelemetry Ruby gems are maintained by the OpenTelemetry-Ruby special interest group (SIG). You can get involved by joining us in [GitHub Discussions][discussions-url] or attending our weekly meeting. See the [meeting calendar][community-meetings] for dates and times. For more information on this and other language SIGs, see the OpenTelemetry [community page][ruby-sig].
+
+## License
+
+The `opentelemetry-instrumentation-elasticsearch` gem is distributed under the Apache 2.0 license. See [LICENSE][license-github] for more information.
+
+[bundler-home]: https://bundler.io
```
fbogsany, line 48: We need a reference to `[elasticsearch-home]` here too.

instrumentation/elasticsearch/opentelemetry-instrumentation-elasticsearch.gemspec
robertlaurin: Hey @mmontalvo sorry about the delayed response. 

Regarding the name of this instrumentation gem, let's keep it as it is.

Regarding what portion of `elasticsearch` should be instrumented.  I went through the `elasticsearch-api` and the `elasticsearch-transport` gem.  

* All of the elasticsearch-api operations are just thin wrappers around the `perform_request` method that this PR is currently instrumenting.  Here's an [example](https://github.com/elastic/elasticsearch-ruby/blob/master/elasticsearch-api/lib/elasticsearch/api/actions/shutdown/get_node.rb#L18-L54).  Ideally it would be nice if we could add instrumentation points here so we could generate spans with the name of the initiating operation (`elasticsearch.get_node` or something to that effect).  I don't think there's a reasonable way to do this, and is probably not worth pursuing if we have to patch the many many methods like this that exist.  

* The elasticsearch-transport gem on the other hand has a single convenient instrumentation point (as you have done in this PR), the issue from my perspective is that it is not providing a lot of additional information if you take into consideration that we have faraday instrumentation.

If we take a look at the information faraday captures [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/2e25dce78876f45d6774f6dafd0d6e317318ae6b/instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb#L50-L52), we would likely end up with two very similar spans with similar timing.  

I think the approach we want to take here is similar to that of `koala` which is an api client gem for facebook.  https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/koala/lib/opentelemetry/instrumentation/koala/patches/instrumentation.rb

The remaining question is whether or not this gem should depend on *and* add faraday instrumentation as part of its install process.

mmontalvo: Hey @robertlaurin no worries!
All the questions around the PR required some investigation and agreement into which way to go 😁 

I will take a look at the `koala` instrumentation and plan to instrument `elasticsearch-api` with the points you mentioned.

Thanks!

mmontalvo:
Sorry for the delay, definitely will update it this week

johnnyshields:
Looking forward to this!

johnnyshields:
@mmontalvo any time to update this?

johnnyshields:
Anyone else in the community want to take this effort over? It would be great to have this.

plantfansam:

To do that, you can:

kaylareopelle:
👋  Hi, @mmontalvo! Thank you for your PR! 

As [Sam mentioned](https://github.com/open-telemetry/opentelemetry-ruby/pull/707#issuecomment-1159105442), Ruby instrumentation now lives in the [opentelemetry-ruby-contrib repo](https://github.com/open-telemetry/opentelemetry-ruby-contrib). We'd be happy to take another look at your work in that context. 

Since it cannot be merged here, I'm going to close this PR. 

Thank you for your contribution. We hope to work with you again soon!

PR #708, opened 2021-04-09
fix: dynamically upgrade global tracer provider

### LINKED ISSUE #706: API Tracer auto-promotion to SDK Tracer
Currently, if a `Tracer` instance is requested from `OpenTelemetry.tracer_provider` before the SDK is installed, a no-op API `Tracer` is returned. This instance will create no-op API `Span` instances in `#start_span`. Those spans will not be recorded, thereby breaking traces (child and parent spans may be recorded if their `Tracer` instances are "real" SDK tracers).

This is going to be a common situation for first-party instrumentation (e.g. if the `redis` gem is instrumented directly rather than being patched by the `opentelemetry-instrumentation-redis` gem).

We should consider "promoting" API `Tracer` instances to SDK `Tracer` instances when the SDK is installed (this should be true even if an SDK other than `opentelemetry-sdk` is installed).

## Things to consider:
- [x] Survey other language SIGs to see how they handle this situation (proxy tracers?)
- [ ] What's the performance impact of always checking the global `tracer_provider` and forwarding?
- [x] Smalltalk's `#become:` would be lovely here, but this is Ruby 😞 
- [ ] We should only adapt API `Tracer` instances returned by the global `OpenTelemetry.tracer_provider` - perhaps this means we want to introduce special `GlobalTracerProvider` and `ProxyTracer` classes in the API? If someone steps off the beaten path and explicitly creates a `TracerProvider` that returns API `Tracer` instances, those `Tracer` instances should *NOT* be promoted
- [ ] Should we move span creation to the `TracerProvider`? If so, we need to document this behaviour for other SDK implementations to follow
- [ ] Whatever we do here, if it adds API, should be clearly documented for both users and other SDK implementations

fbogsany:
Fixes #706


api/lib/opentelemetry.rb
```diff
@@ -41,10 +42,21 @@ def handle_error(exception: nil, message: nil)
     error_handler.call(exception: exception, message: message)
   end
 
+  # Register the global tracer provider.
+  #
+  # @param [TracerProvider] provider A tracer provider to register as the
+  #   global instance.
+  def tracer_provider=(provider)
+    # TODO: should this use a Mutex?
+    # TODO: should this only allow setting the global once?
```
robertlaurin, line 51: I could see app owners wanting to test custom tracing within the context of a specific test file.  In that situation you might want to reconfigure the sdk with for example the in memory exporter instead of the noop exporter.  If we disallow setting the tracer provider delegate more than once the application owners will be stuck with whatever provider the application initialized with.  

That might not be enough justification to allow setting it multiple times though.  

After typing this out, I think what's probably needed is test helpers that can by-pass these interface restrictions for testing.

robertlaurin: If we did only allow setting it once would we be logging a warning for subsequent attempts at setting the provider delegate?

api/lib/opentelemetry.rb
```diff
@@ -41,10 +42,21 @@ def handle_error(exception: nil, message: nil)
     error_handler.call(exception: exception, message: message)
   end
 
+  # Register the global tracer provider.
+  #
+  # @param [TracerProvider] provider A tracer provider to register as the
+  #   global instance.
+  def tracer_provider=(provider)
+    # TODO: should this use a Mutex?
```
robertlaurin, line 50: What would you be concerned with here that would require the mutex, a race condition with setting the tracer_provider?

api/lib/opentelemetry.rb
mwear: I don't think the spec spells out the thread-safety of globals or whether or not they can be changed. Whatever we decide, it seems like the same policy should apply to all of the globals, and not just the tracer provider. In practice, I don't know that you'd actually ever set these more than once, but I'm not sure how heavy handed we want to be in enforcing that.

mwear (commented):
Overall, LGTM. However, I am suspicious about the test failures as they appear to be related and vary with Ruby version.

api/lib/opentelemetry.rb
robertlaurin: I think barring a good reason to prevent users from allow setting the global more than once, we should allow for it.  As mentioned on the sig call, logging a debug level message is probably a good idea if the tracer is being replaced.

api/lib/opentelemetry.rb
fbogsany: yep. mutex added

robertlaurin (approved):

PR #711, opened 2021-04-13
feat!: Add simple 'recording' attr_accessor to InMemorySpanExporter

ahayworth:
When true, the InMemorySpanExporter will retain the spans passed to it, for later inspection. When false, they will simply be discarded. The default is true, which makes this a no-op.


sdk/lib/opentelemetry/sdk/trace/export/in_memory_span_exporter.rb
```diff
@@ -20,20 +20,28 @@ module Export
...
         #   end
         class InMemorySpanExporter
+          # Controls whether or not the exporter will record spans, or discard them.
+          # @return [Boolean] when true, the exporter is recording. By default, this is false.
+          attr_accessor :recording
+
           # Returns a new instance of the {InMemorySpanExporter}.
           #
           # @return a new instance of the {InMemorySpanExporter}.
           def initialize
```
robertlaurin, line 41: Could we do an optional param for the initializer?  That way we don't have to set the recording value everywhere?  
```suggestion
          def initialize(recording: true)
```

I think defaulting to false might be a little confusing too if you're just playing around with the lib and it's not obvious why you have no spans.

ahayworth: We could, yes. My thought was that defaulting to `false` was nice for users writing tests against the SDK in their own apps; but I suppose we could make that nicer via an example docs of how to flip recording on and off. 😄

ahayworth: I've made that change, and force-pushed the branch so it's a nice commit history. 😄

sdk/test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb
```diff
@@ -12,7 +12,7 @@
   let(:span_data1)   { OpenTelemetry::SDK::Trace::SpanData.new(name: 'name1') }
   let(:span_data2)   { OpenTelemetry::SDK::Trace::SpanData.new(name: 'name2') }
 
-  let(:exporter) { export::InMemorySpanExporter.new }
+  let(:exporter) { sdk::Trace::Export::InMemorySpanExporter.new }
```
robertlaurin, line 15:
```suggestion
  let(:exporter) { OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter.new }
```

robertlaurin (approved):
This is great 🥇 

One suggested change, I think it's what's causing CI to fail.

ahayworth:
> One suggested change, I think it's what's causing CI to fail.

🤦 copy/paste failure on my part when addressing your other feedback 😂

fbogsany (approved):

fbogsany:
Thanks for this. Can you please rebase? This is next in the merge queue 😀

PR #712, opened 2021-04-16
feat: Add configurable truncation of span and event attribute values 

aislinncronin:
We were noticing some exceptionally large spans failing to be exported due to nginx error:
`client intended to send too large body`

It turned out to be due to a case in which large event attributes were being added to spans when exceptions occured.  This PR adds support for configurable truncation of span attribute values (via `OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT ` env variable). The default is to not truncate attribute values and the minimum is 32, following the discussion in [this open spec PR](https://github.com/open-telemetry/opentelemetry-specification/pull/1130).


ericmustin (approved):
that open issue seems stuck in the mud so i think this makes sense to add especially if it unblocks you. nice work

fbogsany (approved):

PR #717, opened 2021-04-20
fix!: Replace Time.now with Process.clock_gettime

### LINKED ISSUE #709: Replace use of Time.now with Process.clock_gettime
`Process.clock_gettime` is more efficient than `Time.now` because it returns a `Numeric` (a float or an integer) rather than a full `Time` object. The clocks of interest are:

1. `Process::CLOCK_REALTIME` - seconds since Epoch (1970-01-01 00:00:00 UTC).
2. `Process::CLOCK_MONOTONIC` - seconds elapsed since some arbitrary point in the past (e.g. system startup).

By default, it returns time in seconds as a float, however we can ask for other time units (float ms, float µs) or different integer resolutions (seconds, ms, µs, ns).

We use `Time.now` in a few different places. Interesting use-cases are:

1. Timestamps in events and start & end times for spans.
2. Timeouts for export, flush and shutdown (specifically `Utilities.maybe_timeout`).

The second use-case is best addressed with the monotonic clock, since all we care about is elapsed time since the start of the operation. For this, we should add `OpenTelemetry::Common::Utilities.timeout_timestamp` and update the description of `maybe_timeout` to expect a `start_time` returned from `timeout_timestamp`.

The first use-case is more challenging. I think we should adhere to the following principles:

1. Strive for accuracy by default. This implies that internally generated timestamps should use monotonic clocks to compute an offset from a timestamp sourced from a local realtime clock, where available.
2. Accept explicit timestamps as `Time` objects rather than expose the complexity of monotonic and realtime clocks
3. Don't create `Time` objects internally - all exported timestamps should be "<time unit> since Epoch" (most efficient is probably integer nanoseconds or microseconds)
4. Explicit timestamps should only have local effect. This means that an explicitly provided timestamp should not be used as a base for any timestamps computed by monotonic offsets.

### Plan

1. Fix the timeout case.
2. Decide on the APIs (essentially just "what should be the format & time unit of exported timestamps?").
3. Implement timestamps internally using `Process.clock_gettime(Process::CLOCK_REALTIME)`.
4. Implement all the hard stuff in the SDK `Span` class.

1-3 will be done here and in #717, and will gate the 1.0 Release Candidate. 4 can be done after the RC.

### Notes

1. Clock resolution is at least milliseconds on Windows, Linux and MacOS.
2. Jaeger and Zipkin require integer microseconds-since-epoch timestamps.
3. OTLP requires nanoseconds-since-epoch timestamps.
4. Possible timestamp formats & units include:
    1. Float seconds
    2. Integer milliseconds
    3. Integer microseconds
    4. Integer nanoseconds

Internally, MRI (the Ruby VM) represents integers as either 63-bit 2s complement immediate or (transparently) as bignums. The Epoch is Jan 1, 1970. 63-bit integers give us 2^62-1 positive nanoseconds, which takes us to 2116. I think we can kick efficiency concerns (overflow to bignum) down the road a bit, and use integer nanoseconds-since-epoch as the exported timestamps.

### Addendum

*WARNING: This is a bunch of text from a rubber-ducking exercise. It may provide context or may just be confusing.*

Two scenarios: we have a local parent or we don't have a local parent ... (either because we're the root or because our parent is remote)... if we don't have a local parent, we need to record both the current "realtime" clock and the current monotonic clock. The "realtime" clock becomes the `start_timestamp` subsequent timestamps on this Span (including for Events) are computed as `start_timestamp + (current_monotonic - start_monotonic)`.

Small wrinkle: callers of `add_event` and `start_span` and `finish` can pass an explicit timestamp. For this, I think we should accept `Time` objects and ignore the timestamp computation, but we want to export timestamps as `Numeric` (unit TBD), so we should `.to_r` or something when we get these explicit timestamps.

What should we do if the caller of `start_span` passed an explicit timestamp and calls to `finish` or `add_event` do **NOT** pass explicit timestamps? it seems like we need to record that fact and then not do the offset computation? :thinking_face:

If we have a local parent, we want compute our `start_timestamp` based on the offset from its `start_timestamp` . Couple of issues there:
1. we need to access its `start_monotonic` too, and handle the case where the parent was created with an explicit timestamp
1. we don't have the parent available in the SDK `Span#initialize`. We have a `parent_context`, which comes to us from `start_span` as a `Context` . We can reach in to get the span, but that's an extra thread-local access that duplicates what we did already in `start_span`

I think there's also some weirdness around explicit start timestamps in the middle of a call chain (e.g. local root has default start time, child has explicit, grandchild has default). Maybe we need to "inherit" the local root start realtime and monotonic clocks? But what about root with explicit and descendants with default start times?

I think this is wrong: https://github.com/open-telemetry/opentelemetry-ruby/blob/22a18c3da0fd3f26a5b064d7[…]telemetry/instrumentation/rack/middlewares/tracer_middleware.rb
It should be passing `start_timestamp`

the one weird case is spans created for someone else (like the frontend span in Rack)
the other place this can happen is when you don't have enough information to create a span initially, so you keep track of the start time manually and do something like `tracer.start_span(start_timestamp: start_time).finish`
example would be creating the `receive` spans for Kafka (we don't do this correctly IMO)

IIRC there's one case where a `receive` span should have the `send` span as its parent, but we don't know the parent span ID and trace ID until the message is received

fbogsany:
Fixes #709


robertlaurin (dismissed):

robertlaurin (approved):
🕐

PR #727, opened 2021-04-24
fix: missing instrumentation classes during configuration

arielvalentin:
When auto-instrumention classes are declared but cannot be loaded, the registry will report an error to logger
and then proceed with the installation.

Before:

```
Could not install OpenTelemetry::Instrumentation::Faraday because it was not found

OpenTelemetry error: unexpected configuration error due to undefined method `name' for nil:NilClass - OpenTelemetry::SDK::ConfigurationError
```

After:
```
Could not install OpenTelemetry::Instrumentation::Faraday because it was not found
```

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/registry.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/registry.rb
@@ -47,9 +47,9 @@ module OpenTelemetry
             instrumentation = find_instrumentation(instrumentation_name)
             if instrumentation.nil?
               OpenTelemetry.logger.warn "Could not install #{instrumentation_name} because it was not found"
-              next
+            else
+              install_instrumentation(instrumentation, instrumentation_config_map[instrumentation.name])
             end
-            install_instrumentation(instrumentation, instrumentation_config_map[instrumentation.name])
           end
         end
       end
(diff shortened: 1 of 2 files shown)


instrumentation/base/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -45,8 +45,10 @@ def install(instrumentation_names, instrumentation_config_map = {})
         @lock.synchronize do
           instrumentation_names.each do |instrumentation_name|
             instrumentation = find_instrumentation(instrumentation_name)
-            OpenTelemetry.logger.warn "Could not install #{instrumentation_name} because it was not found" unless instrumentation
-
+            if instrumentation.nil?
+              OpenTelemetry.logger.warn "Could not install #{instrumentation_name} because it was not found"
+              next
+            end
             install_instrumentation(instrumentation, instrumentation_config_map[instrumentation.name])
```
fbogsany, line 52:
```suggestion
            if instrumentation.nil?
              OpenTelemetry.logger.warn "Could not install #{instrumentation_name} because it was not found"
            else
              install_instrumentation(instrumentation, instrumentation_config_map[instrumentation.name])
            end
```

fbogsany (approved):

robertlaurin (approved):

PR #730, opened 2021-04-28
fix: XRay trace_id parsing length

schanjr:
While testing my  xray traces, I found that my console was outputting incorrect trace id headers that was given. Upon inspection, I found out there was a minor error on the xray propagator of the expected length format of Xray. 

It expects a format like this currently:
`Root=1-608919a-c59d40d74d03e146bb342b875;Parent=e309470015af9a8b;Sampled=1`

It should have a format like this: 
`Root=1-608919ac-59d40d74d03e146bb342b875;Parent=e309470015af9a8b;Sampled=1`

Where the first separator of "-" should be 8 hexadecimal characters, and the next "-" separator should be 24+ hexadecimal characters. 

Multiple other languages repositories is implemented this way. 
https://github.com/aws-observability/aws-otel-js/blob/main/packages/opentelemetry-propagator-aws-xray/src/AWSXRayPropagator.ts#L67

https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/sdk-extension/opentelemetry-sdk-extension-aws/src/opentelemetry/sdk/extension/aws/trace/propagation/aws_xray_format.py#L286

The exact format can be found based on the AWS documentations
https://docs.aws.amazon.com/xray/latest/devguide/xray-concepts.html#xray-concepts-sampling

image

### DIFF
--- a/propagator/xray/CHANGELOG.md
+++ b/propagator/xray/CHANGELOG.md
@@ -1,6 +1,4 @@
 # Release History: opentelemetry-propagator-xray
-### v0.18.0 / 2021-04-28
-* Fix bug on parsing trace id format
 
 ### v0.17.0 / 2021-04-22
 
--- a/propagator/xray/CHANGELOG.md
+++ b/propagator/xray/CHANGELOG.md
@@ -1,4 +1,6 @@
 # Release History: opentelemetry-propagator-xray
+### v0.18.0 / 2021-04-28
+* Fix bug on parsing trace id format
 
 ### v0.17.0 / 2021-04-22
 
--- a/propagator/xray/lib/opentelemetry/propagator/xray/version.rb
+++ b/propagator/xray/lib/opentelemetry/propagator/xray/version.rb
@@ -7,7 +7,7 @@
 module OpenTelemetry
   module Propagator
     module XRay
-      VERSION = '0.17.0'
+      VERSION = '0.18.0'
     end
   end
 end
--- a/propagator/xray/lib/opentelemetry/propagator/xray/version.rb
+++ b/propagator/xray/lib/opentelemetry/propagator/xray/version.rb
@@ -7,7 +7,7 @@
 module OpenTelemetry
   module Propagator
     module XRay
-      VERSION = '0.18.0'
+      VERSION = '0.17.0'
     end
   end
 end
--- a/propagator/xray/opentelemetry-propagator-xray.gemspec
+++ b/propagator/xray/opentelemetry-propagator-xray.gemspec
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 0.17.0'
 
-  spec.add_development_dependency 'bundler', '~> 1.17'
+  spec.add_development_dependency 'bundler', '~> 2.1.4'
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'rake', '~> 12.0'
   spec.add_development_dependency 'rubocop', '~> 0.73.0'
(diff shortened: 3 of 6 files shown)


propagator/xray/CHANGELOG.md
```diff
@@ -1,4 +1,6 @@
 # Release History: opentelemetry-propagator-xray
+### v0.18.0 / 2021-04-28
+* Fix bug on parsing trace id format
```
fbogsany, line 3: Thanks for this, however the release tooling will populate this automatically (and will complain that it is already populated).
```suggestion
```

propagator/xray/lib/opentelemetry/propagator/xray/version.rb
```diff
@@ -7,7 +7,7 @@
 module OpenTelemetry
   module Propagator
     module XRay
-      VERSION = '0.17.0'
+      VERSION = '0.18.0'
```
fbogsany, line 10:
```suggestion
      VERSION = '0.17.0'
```
You don't need to bump this - it'll be bumped automatically during release.

fbogsany:
Thanks for this. I've updated the PR title to conform to the Conventional Commit message format - it helps with the release process. Related: you don't need to update the changelog or version number - the release tooling will handle that (thanks for doing it, though - we are discussing changing the changelog process, at least).

propagator/xray/opentelemetry-propagator-xray.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 0.17.0'
 
-  spec.add_development_dependency 'bundler', '>= 1.17'
+  spec.add_development_dependency 'bundler', '~> 1.17'
```
fbogsany, line 31:
```suggestion
  spec.add_development_dependency 'bundler', '>= 1.17'
```
It looks like this change breaks CI.

schanjr: It might be because mentioned bundler version is too old. I bumped it up a bit, let's see if it gets picked up. Otherwise I'll revert the file.

propagator/xray/lib/opentelemetry/propagator/xray/version.rb
schanjr: unbumped

propagator/xray/CHANGELOG.md
schanjr: unpopulated

propagator/xray/opentelemetry-propagator-xray.gemspec
schanjr: I get this complaint when I build the gem locally 
`jruby -S gem build opentelemetry-propagator-xray.gemspec`

image

robbkidd: I believe this warning can safely be ignored. Libraries should generally be open-ended in their version dependencies until the authors know for certain that use with a later version causes problems.

robbkidd: Another option would be to move all the gems in this repository to require Bundler `~> 2.0` for development and test, which could be a good change but is out-of-scope for this trace header parsing fix.

propagator/xray/opentelemetry-propagator-xray.gemspec
```diff
@@ -28,7 +28,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 0.17.0'
 
-  spec.add_development_dependency 'bundler', '>= 1.17'
+  spec.add_development_dependency 'bundler', '~> 2.1.4'
```
fbogsany, line 31: This seems to be problematic for Ruby 3.0, at least. I think we should remove this change from this PR, and bump the `bundler` version everywhere in the project in a subsequent PR.

schanjr: reverted... sorry for the trouble.

propagator/xray/opentelemetry-propagator-xray.gemspec
schanjr: reverted. Thanks

fbogsany (approved):

PR #732, opened 2021-04-29
fix: rename constant to hide warning message

chrisholmes:
The auto-generated jaeger types are reusing a constant name which is
causing the below warning message at startup:

```
.../exporter/jaeger/thrift/gen-rb/jaeger_types.rb:82: warning: already initialized constant OpenTelemetry::Exporter::Jaeger::Thrift::Log::FIELDS
.../exporter/jaeger/thrift/gen-rb/jaeger_types.rb:80: warning: previous definition of FIELDS was here
```

This is causing  a fair amount of noise in one of my projects that has
 ruby services that print to STDOUT.


ericmustin (approved):
Ah good catch, lgtm

PR #737, opened 2021-04-30
feat: option to obfuscate redis arguments

chrisholmes:
This change introduces an option for the Redis instrumentation that will obfuscate all arguments in the db.statement attribute.

The purpose of this change is to give clients the option to hide potentially sensitive information such as personally identifiable
information. This is similar to changes that have already been made to the mysql and pg instrumentation.

### DIFF
--- a/instrumentation/redis/README.md
+++ b/instrumentation/redis/README.md
@@ -47,8 +47,8 @@ end
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Redis', {
     # The obfuscation of arguments in the db.statement attribute is disabled by default.
-    # To enable, set enable_arg_obfuscation to true.
-    enable_arg_obfuscation: false,
+    # To enable, set enable_statement_obfuscation to true.
+    enable_statement_obfuscation: false,
   }
 end
 ```
--- a/instrumentation/redis/README.md
+++ b/instrumentation/redis/README.md
@@ -41,6 +41,18 @@ OpenTelemetry::Instrumentation::Redis.with_attributes('peer.service' => 'cache')
 end
 ```
 
+### Configuration options
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Redis', {
+    # The obfuscation of arguments in the db.statement attribute is disabled by default.
+    # To enable, set enable_arg_obfuscation to true.
+    enable_arg_obfuscation: false,
+  }
+end
+```
+
 ## Example
 
 An example of usage can be seen in [`example/redis.rb`](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/redis/example/redis.rb).
--- a/instrumentation/redis/README.md
+++ b/instrumentation/redis/README.md
@@ -46,9 +46,9 @@ end
 ```ruby
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::Redis', {
-    # The obfuscation of arguments in the db.statement attribute is disabled by default.
-    # To enable, set enable_statement_obfuscation to true.
-    enable_statement_obfuscation: false,
+    # The obfuscation of arguments in the db.statement attribute is enabled by default.
+    # To disable, set enable_statement_obfuscation to false.
+    enable_statement_obfuscation: true,
   }
 end
 ```
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
@@ -20,7 +20,7 @@ module OpenTelemetry
         end
 
         option :peer_service, default: nil, validate: :string
-        option :enable_arg_obfuscation, default: false, validate: :boolean
+        option :enable_statement_obfuscation, default: false, validate: :boolean
 
         private
 
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
@@ -20,6 +20,7 @@ module OpenTelemetry
         end
 
         option :peer_service, default: nil, validate: :string
+        option :enable_arg_obfuscation, default: false, validate: :boolean
 
         private
 
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
@@ -20,7 +20,7 @@ module OpenTelemetry
         end
 
         option :peer_service, default: nil, validate: :string
-        option :enable_statement_obfuscation, default: false, validate: :boolean
+        option :enable_statement_obfuscation, default: true, validate: :boolean
 
         private
 
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
@@ -50,7 +50,7 @@ module OpenTelemetry
         private
 
         def obfuscate_arg(arg)
-          return PLACEHOLDER if config[:enable_arg_obfuscation]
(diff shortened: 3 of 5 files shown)


chrisholmes:
It looks like this failed due to a problem with Redis being unavailable. Is this a known problem?

instrumentation/redis/README.md
```diff
@@ -41,6 +41,18 @@ OpenTelemetry::Instrumentation::Redis.with_attributes('peer.service' => 'cache')
...
 ```
 
+### Configuration options
+
+```ruby
+  c.use 'OpenTelemetry::Instrumentation::Redis', {
+    # The obfuscation of arguments in the db.statement attribute is disabled by default.
+    # To enable, set enable_arg_obfuscation to true.
+    enable_arg_obfuscation: false,
```
fbogsany, line 51: It's unfortunate that we chose to call the config option for MySQL and Postgres `enable_sql_obfuscation`. I wonder if we'd be better off renaming those and this option to `enable_statement_obfuscation`, both for consistency and because we're obfuscating the `db.statement` attribute.

fhwang:
@chrisholmes the Redis that's started for tests has a non-standard configuration, including the port number. Among other things we're trying to avoid port collisions if the developer already has Redis running for some other reason.

The simplest thing to do is probably to call `redis_with_auth` instead of `Redis.new` in any tests you add.

chrisholmes:
@fhwang thanks. It should be fixed now.

fhwang:
Now you'll need to account for an extra span in your tests -- the default redis setup has a password, so the client will send an `AUTH` command before doing anything else.

chrisholmes:
@fhwang I probably should have run rake before committing 😬

I'm fairly confident it's sorted now.

instrumentation/redis/README.md
johnnyshields: This should be true by default (and should be true for the others as well), following a "secure by default" mindset.

chrisholmes: I agree with you both here. Speaking from my own experience it’s better to not leave this as an opt-in as a security incident will soon follow.

I can change it to `disable_statement_obfuscation` if there’s consensus.

fbogsany: > I can change it to `disable_statement_obfuscation` if there’s consensus.

Please do 🙏

fbogsany (approved):

instrumentation/redis/README.md
fbogsany: Alternatively, just change the default to `true`.

instrumentation/redis/README.md
```diff
@@ -41,6 +41,18 @@ OpenTelemetry::Instrumentation::Redis.with_attributes('peer.service' => 'cache')
...
 ```
 
+### Configuration options
+
+```ruby
+  c.use 'OpenTelemetry::Instrumentation::Redis', {
+    # The obfuscation of arguments in the db.statement attribute is disabled by default.
+    # To enable, set enable_statement_obfuscation to true.
+    enable_statement_obfuscation: false,
```
fbogsany, line 51:
```suggestion
    # The obfuscation of arguments in the db.statement attribute is enabled by default.
    # To disable, set enable_statement_obfuscation to false.
    enable_statement_obfuscation: true,
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb
```diff
@@ -20,6 +20,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         option :peer_service, default: nil, validate: :string
+        option :enable_statement_obfuscation, default: false, validate: :boolean
```
robertlaurin, line 23:
```suggestion
        option :enable_statement_obfuscation, default: true, validate: :boolean
```

chrisholmes: @robertlaurin I've added a commit with this

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
```diff
@@ -17,24 +17,46 @@ module Utils
...
+        private
+
+        def obfuscate_arg(arg)
+          return PLACEHOLDER if config[:enable_statement_obfuscation]
+
+          arg
+        end
+
+        def format_string(input = nil)
+          str = block_given? ? yield(input) : input
```
fbogsany, line 59: This seems over engineered - I don't think this should accept a block. Both call sites can pass an immediate arg.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
```diff
@@ -17,24 +17,46 @@ module Utils
...
+          end
+        end
+
+        private
+
+        def obfuscate_arg(arg)
+          return PLACEHOLDER if config[:enable_statement_obfuscation]
+
+          arg
+        end
```
fbogsany, line 56: This seems over engineered - I think this is sufficient:
```suggestion
          if config[:enable_statement_obfuscation]
            PLACEHOLDER
          else
            format_string(arg.to_s)
          end
        end

        private
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
```diff
@@ -17,24 +17,46 @@ module Utils
...
+          cmd = format_cmd(command_args.first)
+          args = command_args[1..-1].map { |arg| format_arg(arg) }
+          statement = args.unshift(cmd).join(' ')
+          OpenTelemetry::Common::Utilities.truncate(statement, CMD_MAX_LEN)
+        end
+
+        def format_cmd(cmd)
+          format_string do
+            cmd.to_s.upcase
+          end
```
fbogsany, line 37:
```suggestion
          format_string(cmd.to_s.upcase)
```

fbogsany (changes requested):

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
chrisholmes: Part of my original intent with this was to ensure that exceptions in converting the strings above would be caught here. However, I'm now thinking my memory of to_s might be a little shaky and pretty much everything can do it.

chrisholmes: It looks like swapping this caused an internal test to error. You can see in the GH Actions output.

From what I understand `to_s` should be defined on every object except for `BasicObject` so this failure is very unlikely to occur. Are you happy for me to remove the test?

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
```diff
@@ -17,25 +17,44 @@ module Utils
...
+            PLACEHOLDER
+          else
+            arg
+          end
+        end
+
+        def format_string(input = nil)
+          str = OpenTelemetry::Common::Utilities.utf8_encode(input, binary: true)
           OpenTelemetry::Common::Utilities.truncate(str, VALUE_MAX_LEN)
         rescue StandardError => e
```
fbogsany, line 59: This error handling should be moved up to `format_arg` and `format_cmd`. This will resolve the test failure.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
```diff
@@ -17,25 +17,44 @@ module Utils
...
+
+        def obfuscate_arg(arg)
+          if config[:enable_statement_obfuscation]
+            PLACEHOLDER
+          else
+            arg
+          end
+        end
+
+        def format_string(input = nil)
```
fbogsany, line 56:
```suggestion
        def format_string(input)
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/utils.rb
chrisholmes: sure

chrisholmes: Though we're interpolating the input in the debug message so I suppose we'd still run into this problem for `BasicObject`, but let's say that's an edge case.

fbogsany (approved):

PR #738, opened 2021-04-30
fix: Instrument Redis more thoroughly by patching Client#process.

### LINKED ISSUE #197: Improve Redis instrumentation
The Redis instrumentation, which is a port of the dd-trace version, instruments redis by patching the `call` and `call_pipeline` methods. We might be missing some calls by not also patching `call_loop` and `call_pipelined`. We should look into adding coverage for the additional methods.

See: https://github.com/open-telemetry/opentelemetry-ruby/pull/176#pullrequestreview-361604975

fhwang:
Fixes #197.

I've opted to change the span names for pipeline requests to a concatenation of the Redis commands, i.e. `GET SET INCR`. This seems like it's more inline with [what the spec recommends](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#span) than simply `pipeline`:

> The span name SHOULD be the most general string that identifies a (statistically) interesting class of Spans, rather than individual Span instances while still being human-readable.

### DIFF
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
@@ -21,8 +21,8 @@ module OpenTelemetry
             ) do |s|
               reply = super(commands)
               if reply.is_a?(::Redis::CommandError)
-                s&.record_exception(reply)
-                s&.status = Trace::Status.new(
+                s.record_exception(reply)
+                s.status = Trace::Status.new(
                   Trace::Status::ERROR,
                   description: "Unhandled exception of type: #{reply.class}"
                 )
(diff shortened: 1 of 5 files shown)


instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -10,36 +10,25 @@ module Redis
...
+              Utils.format_span_name(commands),
               attributes: attributes,
               kind: :client
-            ) do
-              response = super(*args, &block)
+            ) do |s|
+              reply = super(commands)
+              if reply.is_a?(::Redis::CommandError)
+                s&.record_exception(reply)
+                s&.status = Trace::Status.new(
```
fbogsany, line 25: `s` should definitely be non-`nil` at this point.
```suggestion
                s.record_exception(reply)
                s.status = Trace::Status.new(
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -10,36 +10,25 @@ module Redis
...
+                  Trace::Status::ERROR,
+                  description: "Unhandled exception of type: #{reply.class}"
```
fbogsany, line 27: I don't think we should call this "Unhandled" - any error handling would occur higher in the call chain. Using the `message` is probably more descriptive.
```suggestion
                  description: reply.message
```

Issue #736, opened 2021-04-30
Add option to disable `db.statement` in DB instrumentation


fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1659

We should use the same configuration option in all affected instrumentation.

Affected instrumentation:

- [x] Dalli
- [x] LMDB
- [x] Mongo
- [x] MySQL2
- [x] Postgres
- [x] Redis

robertlaurin:
As mentioned in the description we should use the same config option.

Currently we're a bit inconsistent with the configuration naming here see examples:
https://github.com/open-telemetry/opentelemetry-ruby/blob/9dc42f12f7231bc48e2702d390c54917293d9b65/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb#L28-L29
https://github.com/open-telemetry/opentelemetry-ruby/blob/9dc42f12f7231bc48e2702d390c54917293d9b65/instrumentation/redis/lib/opentelemetry/instrumentation/redis/instrumentation.rb#L24
https://github.com/open-telemetry/opentelemetry-ruby/blob/9dc42f12f7231bc48e2702d390c54917293d9b65/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb#L23

Settling on a consistent name will introduce breaking changes but I think its valuable in the long term.

I have the opinion that the enable/disable prefix is a bit wordy and could be dropped (this is true of other config options, some I introduced myself).

```rb
option :db_statement,             default: true, validate: :boolean
option :db_statement_obfuscation, default: true, validate: :boolean
```

SomalianIvan:
👀

fbogsany:
> I have the opinion that the enable/disable prefix is a bit wordy and could be dropped

I'm not sure I agree 😄. I tend to think boolean config options benefit from being clearly a predicate, e.g. `include_db_statement: false` is clearer than `db_statement: false`. On the other hand, `obfuscate_db_statement: false` is clearer than either `enable_db_statement_obfuscation: false` or `db_statement_obfuscation: false`. So my vote would be for:
```ruby
option :include_db_statement,   default: true, validate: :boolean
option :obfuscate_db_statement, default: true, validate: :boolean
```
Alternatively, these could be unified:
```ruby
option :db_statement, default: :include, validate: ->(opt) { %I[include omit obfuscate].include?(opt) }
```

robertlaurin:
[...]

PR #739, opened 2021-04-30
feat: Add option to postgres instrumentation to disable db.statement

ahayworth:
We leave it on by default, but allow users to disable the inclusion of the attribute entirely if they so wish.

cc #736


ahayworth:
It is unclear to me why the 2.7 tests are failing ... there are a few failures for non-pg related things. 🤔

ericmustin (approved):
generally lgtm, if people have strong opinions on the naming convention of the config option i'm fine with a change also

instrumentation/pg/README.md
```diff
@@ -42,6 +42,12 @@ OpenTelemetry::SDK.configure do |c|
     # You may optionally set a value for 'peer.service', which
     # will be included on all spans from this instrumentation:
     peer_service: 'postgres:readonly',
+
+    # By default, this instrumentation includes the executed SQL as the `db.statement`
+    # semantic attribute (possibly sanitized, according to the value of `enable_sql_obfuscation`).
+    # Optionally, you may disable the inclusion of this attribute entirely by
+    # setting this option to false.
+    enable_db_statement_attribute: true,
```
fbogsany, line 50: Perhaps just:
```suggestion
    enable_statement_attribute: true,
```

In #737 I suggested normalizing the obfuscation option across instrumentation as `enable_statement_obfuscation`. This rename would line up with that. I'm not sure we need the `db_` name spacing on the option.

ahayworth: 👍 That makes sense to me, and it's changed in c996c4e2f2403bd0d3843086059cb52950b57f26.

Re: `enable_statement_obfuscation` specifically, I did not change that in this PR, although I think you have a good point there. Should we rename that in this PR, or do it together with mysql in another (and, do we care about maintaining compatibility for the old name at all)?

fbogsany: Let's do it in a separate PR. I don't think we need to maintain compatibility pre-1.0, but we should call it out prominently in the release notes.

fbogsany (approved):

PR #740, opened 2021-05-01
fix: rescue missed otlp exporter network errors

chrisholmes:
The Net::HTTP client has a few more errors that need to be caught by the OTLP exporter or it will crash if the upstream agent/collector is temporarily unavailable.

You can simulate one of the SystemCallErrors, Errno::ECONNREFUSED by restarting a local OpenTelemetry collector while an app is trying to export spans to it. Without this change the exporter will crash, but with this change the exporter should recover when the collector becomes available again.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -190,6 +190,12 @@ module OpenTelemetry
           rescue SocketError
             retry if backoff?(retry_count: retry_count += 1, reason: 'socket_error')
             return FAILURE
+          rescue SystemCallError
+            retry if backoff?(retry_count: retry_count += 1, reason: 'system_call_error')
+            return FAILURE
+          rescue EOFError
+            retry if backoff?(retry_count: retry_count += 1, reason: 'eof_error')
+            return FAILURE
           end
         ensure
           # Reset timeouts to defaults for the next call.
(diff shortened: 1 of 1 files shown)


chrisholmes:
I'm not sure why the JRuby tests failed. It looks like the failure was in the `TracerProvider` tests that should be unrelated. Any chance that test is flakey?

fbogsany:
> I'm not sure why the JRuby tests failed. It looks like the failure was in the `TracerProvider` tests that should be unrelated. Any chance that test is flakey?

There's definitely a chance JRuby is flaky 😞

robertlaurin (commented):
While you're adding handling for those two exceptions could you add for the following as well?  We recently ran into these two in addition to the `EOFError` you've added handling for.

* `Errno::EHOSTUNREACH` 
* `Errno::ECONNRESET`

chrisholmes:
@robertlaurin: Errno::EHOSTUNREACH and Errno::ECONNRESET are children of `SystemCallError` so they should be covered  by this.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -190,6 +190,12 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
           rescue SocketError
             retry if backoff?(retry_count: retry_count += 1, reason: 'socket_error')
             return FAILURE
+          rescue SystemCallError
+            retry if backoff?(retry_count: retry_count += 1, reason: 'system_call_error')
```
fbogsany, line 194: Is there more detail we can provide here to distinguish different platform-dependent errors? The full `message` is obviously too verbose. Would `.class.name` be reasonable?

chrisholmes: Yes, that's a good idea

fbogsany (approved):
Thanks @chrisholmes !

robertlaurin (approved):

PR #742, opened 2021-05-04
fix: improve attribute error messages

### LINKED ISSUE #741: Improve error message for invalid attributes
At the moment, someone using the MySQL2 (best guess) instrumentation is encountering an error message like:
```
E, [2021-05-04T09:18:39.581149 #31471] ERROR -- : OpenTelemetry error: invalid attribute value type NilClass
```
This doesn't help narrow down the source of the error. We should at least include the span name in the message. This comes from https://github.com/open-telemetry/opentelemetry-ruby/blob/e7da54d8aa5ac89009020c7a7a28afe2e1315c6a/sdk/lib/opentelemetry/sdk/internal.rb#L47-L59

This is called from a few places in the SDK `Span` class:
https://github.com/open-telemetry/opentelemetry-ruby/blob/e7da54d8aa5ac89009020c7a7a28afe2e1315c6a/sdk/lib/opentelemetry/sdk/trace/span.rb#L304-L308

https://github.com/open-telemetry/opentelemetry-ruby/blob/e7da54d8aa5ac89009020c7a7a28afe2e1315c6a/sdk/lib/opentelemetry/sdk/trace/span.rb#L332

https://github.com/open-telemetry/opentelemetry-ruby/blob/e7da54d8aa5ac89009020c7a7a28afe2e1315c6a/sdk/lib/opentelemetry/sdk/trace/span.rb#L349

fbogsany:
Fixes #741

### DIFF
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -50,7 +50,7 @@ module OpenTelemetry
             OpenTelemetry.handle_error(message: "invalid #{kind} attribute key type #{k.class} on span '#{owner}'")
             false
           elsif !valid_value?(v)
-            OpenTelemetry.handle_error(message: "invalid #{kind} attribute value type #{v.class} on span '#{owner}'")
+            OpenTelemetry.handle_error(message: "invalid #{kind} attribute value type #{v.class} for key '#{k}' on span '#{owner}'")
             false
           else
             true
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -44,13 +44,13 @@ module OpenTelemetry
         valid_simple_value?(value) || valid_array_value?(value)
       end
 
-      def valid_attributes?(attrs)
+      def valid_attributes?(owner, kind, attrs)
         attrs.nil? || attrs.all? do |k, v|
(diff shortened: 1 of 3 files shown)


sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -44,13 +44,13 @@ def valid_value?(value)
...
-      def valid_attributes?(attrs)
+      def valid_attributes?(owner, kind, attrs)
         attrs.nil? || attrs.all? do |k, v|
           if !valid_key?(k)
-            OpenTelemetry.handle_error(message: "invalid attribute key type #{k.class}")
+            OpenTelemetry.handle_error(message: "invalid #{kind} attribute key type #{k.class} on span '#{owner}'")
             false
           elsif !valid_value?(v)
-            OpenTelemetry.handle_error(message: "invalid attribute value type #{v.class}")
+            OpenTelemetry.handle_error(message: "invalid #{kind} attribute value type #{v.class} on span '#{owner}'")
```
robertlaurin, line 53: I think it would be valuable to identify which key has a nil value associated with it.

fbogsany: 🤦 of course

fbogsany: done

robertlaurin (approved):

PR #743, opened 2021-05-05
feat: export to jaeger collectors w/ self-signed certs

chrisholmes:
This change covers adding configuration to the Jaeger `CollectorExporter` to support exporting to Collectors that use self-signed certificates or any certificate chain that can't be verified by Ruby's openssl implementation.

It works by passing the configuration parameter onto the `Jaeger::HTTPClientTransport` class.

A similar implementation could be added for the other HTTP based exporters.

I'm a bit unsure about passing the verify mode in as an envvar in its current form, so I'm keen for feedback on this. I believe this is the most correct approach as the flag are meant to be combined together using union operations (e.g. `VERIFY_PEER | VERIFY_FAIL_IF_NO_PEER_CERT`). However, this might not be obvious and I've not been able to find a reliable reference for this.

### DIFF
--- a/exporter/jaeger/README.md
+++ b/exporter/jaeger/README.md
@@ -75,11 +75,14 @@ The agent exporter can be configured explicitly in code, as shown above, or via 
 
 The collector exporter can be configured explicitly in code, as shown above, or via environment variables. The configuration parameters, environment variables, and defaults are shown below.
 
-| Parameter   | Environment variable            | Default                    |
-| ----------- | ------------------------------- | -------------------------- |
-| `endpoint:` | `OTEL_EXPORTER_JAEGER_ENDPOINT` | `"http://localhost:14268"` |
-| `username:` | `OTEL_EXPORTER_JAEGER_USER`     | `nil`                      |
-| `password:` | `OTEL_EXPORTER_JAEGER_PASSWORD` | `nil`                      |
+| Parameter          | Environment variable                   | Default                    |
+| ------------------ | -------------------------------------- | -------------------------- |
+| `endpoint:`        | `OTEL_EXPORTER_JAEGER_ENDPOINT`        | `"http://localhost:14268"` |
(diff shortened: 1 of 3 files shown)


fbogsany:
`OTEL_EXPORTER_JAEGER_SSL_VERIFY_MODE` isn't part of the OpenTelemetry spec for supported environment variables. If this is something that makes sense across languages, it'd be worth opening a PR to add it to the spec. Otherwise, we should name it `OTEL_RUBY_EXPORTER_JAEGER_SSL_VERIFY_MODE `.

chrisholmes:
thanks @fbogsany, I hadn't noticed that these envvars were part of the spec.

This configuration is coupled to the Thrift Ruby implementation and Ruby's OpenSSL implementation so I can't see a generic enough way to add it to the spec. I have found SSL configuration options to be fairly inconsistent when working across languages.

I will rename it then to `OTEL_RUBY_EXPORTER_JAEGER_SSL_VERIFY_MODE`.

fbogsany:
CI failures are related to an upstream Mongo gem change. We'll try to resolve that - no action required in this PR for now.

chrisholmes:
@ahayworth, yeah, I wouldn't say I'm a massive fan of how this is currently configured. 

I had something similar to what you suggested in an earlier version of this, but I swapped it when I saw that the verify_mode values were meant to be used as flags like:
```
flags = OpenSSL::SSL::VERIFY_PEER|OpenSSL::SSL::VERIFY_FAIL_IF_NO_PEER_CERT
```

However, we could use hardcoded envvar values and then use the method parameter if somebody needs the flags.

chrisholmes:
@ahayworth, I've applied your suggestion and I agree this works better.

I also realised that the other flags are just for server side configuration so they can ignored 😅

ahayworth:
@chrisholmes Thanks so much! This looks good to me. 👍

ahayworth (approved):

PR #744, opened 2021-05-06
Fix: Nil value for db.name attribute

robertlaurin:
One of our users reported some log spam during migrations on their application.

It looks like we're expecting a `db.name` to be present when there might not be one available.

```
E, [2021-05-05T15:02:17.437321 #55630] ERROR -- : OpenTelemetry error: invalid span attribute value type NilClass for key 'db.name' on span 'set names'
E, [2021-05-05T15:02:17.440246 #55630] ERROR -- : OpenTelemetry error: invalid span attribute value type NilClass for key 'db.name' on span 'create database'
```

### DIFF
--- a/instrumentation/mysql2/CHANGELOG.md
+++ b/instrumentation/mysql2/CHANGELOG.md
@@ -1,18 +1,22 @@
 # Release History: opentelemetry-instrumentation-mysql2
 
+### Unreleased
+
+* Fix: Nil value for db.name attribute #744
+
 ### v0.17.0 / 2021-04-22
 
 * (No significant changes)
 
 ### v0.16.0 / 2021-03-17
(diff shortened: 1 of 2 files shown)


ericmustin (approved):
could be good to add a test to prevent a regression but generally lgtm

fbogsany (approved):

robertlaurin:
@ericmustin 

> could be good to add a test to prevent a regression but generally lgtm

Hey yeah, sorry responding post merge.  I agree, I find testing for the absence of a log kind of awkward and flakey as a form of test.  It's possible I'm not seeing a more obvious reliable way of testing this change.  

If you have any ideas here I can definitely add the test in a follow up PR.

PR #760, opened 2021-05-20
refactor: redis attribute utils

robertlaurin:
I started this refactor because the redis utils file was getting a little bit difficult to follow, and I suspected there was an unnecessary amount of string manipulation and allocation being done.

I also wanted to have something resembling evidence that this is in fact an improvement beyond "i like this better", so I created a small script to measure the allocations, ran it a few times with the original and the refactor.  The results were consistent and I shared a screenshot of the latest run below along with the script that generated the out.

### Simple script to create profile
```rb
require 'bundler/inline'
require 'benchmark'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'redis'
  # gem 'opentelemetry-instrumentation-redis', path: '/Users/robertlaurin/src/forks/opentelemetry-ruby/instrumentation/redis/'
  gem 'opentelemetry-instrumentation-redis', branch: 'main', git: 'https://github.com/open-telemetry/opentelemetry-ruby', glob: 'instrumentation/redis/opentelemetry-instrumentation-redis.gemspec'
  gem 'heap-profiler'
  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
end

require 'heap-profiler'
require 'opentelemetry-api'
require 'opentelemetry-sdk'
require 'opentelemetry-exporter-otlp'
require 'opentelemetry-instrumentation-redis'

span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
  OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter.new(recording: false)
)

  c.use 'OpenTelemetry::Instrumentation::Redis'
  c.service_name = 'donut matter'
  c.add_span_processor(span_processor)
end

HeapProfiler.report("redis-original") do
  redis = Redis.new

  1..1_000.times do
    redis.queue([:set, 'v1', '0'])
    redis.queue([:incr, 'v1'])
    redis.queue([:get, 'v1'])
    redis.commit
  end
end
```

### New on the left, old on the right
image


instrumentation/redis/test/opentelemetry/instrumentation/redis/utils_test.rb
```diff
@@ -1,113 +0,0 @@
...
-      describe 'an object that can\'t be converted to a string' do
-        let(:arg) { object_class.new }
-        let(:object_class) do
-          Class.new do
-            def to_s
-              raise "can't make a string of me"
-            end
-          end
-        end
-        it { _(subject).must_equal('?') }
```
robertlaurin, line 72: This is an interesting test but it doesn't really fit in the context of redis, maybe if this utils was being generalized outside of the redis instrumentation it would be a better fit.

Redis consistently `to_s` values, so it would raise in redis not the instrumentation patch.

Example -> https://github.com/redis/redis-rb/blob/master/lib/redis.rb#L839

instrumentation/redis/test/opentelemetry/instrumentation/redis/patches/client_test.rb
```diff
@@ -201,6 +201,72 @@ def redis_with_auth(redis_options = {})
...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyyy...
+        SET v1 yyyyyyyyyyyyyyyyyyyyyyyyyy...
+      HEREDOC
+
+      _(last_span.name).must_equal 'SET SET SET SET SET SET SET SET SET'
```
robertlaurin, line 255: I really think we could adhere to the spec and also avoid having span names like this, IMO it makes more sense to call these spans `pipelined` or something very similar to this.

ahayworth: If I saw a span name like this, my first thought would be "gee, there's a bug in the instrumentation"

ahayworth (commented):
The reduction in memory usage is very nice! We should definitely be looking at the amount of string manipulation (especially that which is forced by rubocop...) in our other libraries.

Other than some in-line thoughts, I was also curious what kinds of data structures are passed to the span-name and command-parsing functions? I think it's saying you could get  either `'foo'` or `['foo']`or `[['foo'], ...]` but it wasn't clear to me.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -40,6 +46,41 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
+              command = command[0] if command.is_a?(Array) && command[0].is_a?(Array)
+
+              # If we receive an authentication request command
+              # we want to short circuit parsing the commands
+              # and return the obfuscated command
+              return 'AUTH ?' if command[0] == 'auth'
+
+              cmd = command[0].upcase!
+              if config[:enable_statement_obfuscation]
+                cmd + ' ?' * command[1..-1].size
```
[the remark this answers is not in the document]
robertlaurin: Ha, beat me to this comment :).  I agree and I want to make this change, but I thought I should push up this refactor with parity to the old behaviour first.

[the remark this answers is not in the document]
robertlaurin: I'll do this and the change to `SET SET SET SET SET SET SET SET SET SET SET SET SET SET SET SET SET SET` in a follow up PR.

robertlaurin:
> Other than some in-line thoughts, I was also curious what kinds of data structures are passed to the span-name and command-parsing functions? I think it's saying you could get either `'foo'` or `['foo']`or `[['foo'], ...]` but it wasn't clear to me.

I've added comments to the parsing method that I hope clarifies this a bit.  I think I can simplify the `span_name` method in a follow up PR addressing some of the other comments.

But yeah the method handles when it receives `[["set", "K", "0"]]` or `[[["set", "v1", "0"]], [["incr", "v1"]], [["get", "v1"]]]`

instrumentation/redis/test/opentelemetry/instrumentation/redis/patches/client_test.rb
fbogsany: Agreed - let's use `pipelined`. The `db.statement` will have the full query.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -40,6 +46,49 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
 
           private
 
+          # Ensure all values are properly encoded and truncated before
```
fbogsany, line 49: before ... - please don't leave us hanging!

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -40,6 +46,49 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
+              command = command[0] if command.is_a?(Array) && command[0].is_a?(Array)
+
+              # If we receive an authentication request command
+              # we want to short circuit parsing the commands
+              # and return the obfuscated command
+              return 'AUTH ?' if command[0] == 'auth'
+
+              cmd = command[0].upcase!
+              if config[:enable_statement_obfuscation]
+                cmd + ' ?' * command[1..-1].size
```
fbogsany, line 77: `cmd` will be `nil` here if `upcase!` made no changes.

There's no value in slicing `command` just to compute the size - you can just `' ?' * (command.size - 1)`

robertlaurin: Good catch, I didn't even consider that you could do something like this 👇 
```rb
      redis.queue(['SET', 'v1', '0'])
      redis.commit
```

I was under the impression that the commands always came in as symbols.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -23,9 +27,11 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
             attributes['db.redis.database_index'] = options[:db] unless options[:db].zero?
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
             attributes.merge!(OpenTelemetry::Instrumentation::Redis.attributes)
-            attributes['db.statement'] = Utils.format_statements(commands)
 
-            tracer.in_span(Utils.format_span_name(commands), attributes: attributes, kind: :client) do |s|
+            formatted_commands = format_commands(commands)
+            attributes['db.statement'] = OpenTelemetry::Common::Utilities.truncate(parse_commands(formatted_commands), MAX_STATEMENT_LENGTH)
+
+            tracer.in_span(span_name(formatted_commands), attributes: attributes, kind: :client) do |s|
```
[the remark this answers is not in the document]
robertlaurin: Do we (you) care about doing uniform truncation of command values?  It was part of the reason why I did all the encoding/truncation upfront.  That way if the db.statement did need to be truncated it didn't favour a very long early value, and instead was more likely to even out the values so you could see more of the statement in its entirety.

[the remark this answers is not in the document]
fbogsany: I don't care too much. I'd argue that we should make the statement length configurable and allow users to increase it (or disable the truncation) if they really care about the details. We will likely have obfuscation enabled by default, so most large values will be replaced with `?` anyway.

robertlaurin:
Ended up adding a few changes I planned to do in a separate PR.

- No more `SET SET SET SET SET SET SET SET SET`, now pipelined
- Not truncating each individual command value
- Only encoding and truncating once

Memory allocations is quite a bit lower than before :) 
image

fbogsany (approved):
I like this - a really nice improvement! Thanks @robertlaurin

ahayworth (approved):

PR #761, opened 2021-05-20
Updated rack middleware position to zero

kjg:
PR #699 was opened to place the rack middleware at the very first position. However, it is using `app.middleware.insert_after(0` which has the effect of inserting at the second spot in the chain instead of the first. This updates it to `app.middleware.insert_before(0` to have the originally intended effect


robertlaurin (approved):
Thanks @kjg

arielvalentin:
@open-telemetry/ruby-approvers 👋🏼 looks like I do not have the permissions to merge. Is this restricted to `maintainers`?

robertlaurin:
@arielvalentin merging is restricted to maintainers.  But approvers can create releases.  

I should have got this in with the last release 😞 , so we can merge this and then create a patch release for this instrumentation gem.

@kjg do you mind rebasing on main?

kjg:
@robertlaurin No problem, rebased!

PR #768, opened 2021-05-24
feat: add a SSL verify mode option for the OTLP exporter

chrisholmes:
This change covers adding configuration to the OTLP Exporter
to support exporting to Collectors that have certificates that can't be
 verified by OpenSSL.

This approach is copied from the similar change for the Jaeger
CollectorExporter:
https://github.com/open-telemetry/opentelemetry-ruby/pull/743

### DIFF
--- a/exporter/otlp/README.md
+++ b/exporter/otlp/README.md
@@ -65,6 +65,22 @@ end
 
 For additional examples, see the [examples on github][examples-github].
 
+## How can I configure the Jaeger exporter?
+
+The collector exporter can be configured explicitly in code, as shown above, or via environment variables. The configuration parameters, environment variables, and defaults are shown below.
+
+| Parameter           | Environment variable                         | Default                             |
+| ------------------- | -------------------------------------------- | ----------------------------------- |
+| `endpoint:`         | `OTEL_EXPORTER_OTLP_ENDPOINT`                | `"http://localhost:4317/v1/traces"` |
+| `certificate_file: `| `OTEL_EXPORTER_OTLP_CERTIFICATE`             |                                     |
+| `headers:`          | `OTEL_EXPORTER_OTLP_HEADERS`                 |                                     |
+| `compression:`      | `OTEL_EXPORTER_OTLP_COMPRESSION`             |                                     |
+| `timeout:`          | `OTEL_EXPORTER_OTLP_TIMEOUT`                 | `10`                                |
+| `ssl_verify_mode:`  | `OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_PEER` or | `OpenSSL::SSL:VERIFY_PEER`          |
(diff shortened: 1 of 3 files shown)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -31,8 +31,19 @@ class Exporter # rubocop:disable Metrics/ClassLength
         WRITE_TIMEOUT_SUPPORTED = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
         private_constant(:KEEP_ALIVE_TIMEOUT, :RETRY_COUNT, :WRITE_TIMEOUT_SUPPORTED)
 
-        def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: 'https://localhost:4317/v1/traces'), # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+        def self.ssl_verify_mode
```
arielvalentin, line 34: 💭 _We may have an opportunity here for a [refactoring](https://github.com/open-telemetry/opentelemetry-ruby/pull/743/files#diff-e00633625fe806f1deb53b278b99ddc078438e19f95f3f508e81d1a36f1024ecR18) but it is also small enough that it will likely cause more confusion than clarity._

arielvalentin (commented):
@chrisholmes may I ask you to also add docs to the `README` with something similar to what you added in [the other PR](https://github.com/open-telemetry/opentelemetry-ruby/pull/743/files#diff-26965f6dd30643697c4ad43c4b9043691e8b753969bf42f3e2af964d9282b767R78)?

chrisholmes:
@arielvalentin README added

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
chrisholmes: yeah, I think it's probably worth doing this for the next exporter that needs it. Maybe for the Zipkin exporter?

chrisholmes: I'm there's a few other utilities I've spotted that could be extracted into a base class or mixin for HTTP exporters. There's retry behaviour in the OTLP exporter that could be extract for instance.

exporter/otlp/README.md
```diff
@@ -65,6 +65,22 @@ end
 
 For additional examples, see the [examples on github][examples-github].
 
+## How can I configure the Jaeger exporter?
```
arielvalentin, line 68:
```suggestion
## How can I configure the OTLP exporter?
```

arielvalentin (approved):
Just one suggestion to update in the README. Other than that LGTM

fbogsany (approved):

PR #769, opened 2021-05-24
feat: consistent HTTP Request attributes with utility

chrisholmes:
This change introduces a utility that can be used by instrumentation to add consistent attributes for HTTP requests. Currently, HTTP attributes are set inconsistently across clients, which can be a bit of a headache if the services you're tracing have multiple HTTP client dependencies.

The utility also has an option to hide query parameters so that they don't show up in these attributes. Query parameters often hold sensitive information, which some (hopefully all 😅 ) teams may not want to log. I've made this on by default to align other recent changes to handling sensitive information.

I've applied this utility to the `net/http`, `faraday`, and `http` instrumentation, but I'd also want to apply this to all the other http clients in the repository.

### DIFF
--- a/common/lib/opentelemetry/common/http/request_attributes.rb
+++ b/common/lib/opentelemetry/common/http/request_attributes.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Common
+    module HTTP
+      # RequestAttributes contains common helpers for setting http client attributes in spans
+      module RequestAttributes
+        extend self
+
+        def from_request(request_method, uri, config = {})
+          uri = hide_query_params(uri) if config[:hide_query_params]
+
+          {
+            'http.method' => request_method,
+            'http.scheme' => uri.scheme,
+            'http.target' => uri.request_uri,
+            'http.url' => uri.to_s,
+            'peer.hostname' => uri.host,
+            'peer.port' => uri.port
+          }.compact
+        end
+
+        def hide_query_params(uri)
+          uri.query = '' if uri.query
+
+          uri
+        end
+      end
+    end
+  end
+end
--- a/common/lib/opentelemetry/common/http/request_attributes.rb
+++ b/common/lib/opentelemetry/common/http/request_attributes.rb
@@ -25,6 +25,7 @@ module OpenTelemetry
         end
 
         def hide_query_params(uri)
+          uri = uri.dup
           uri.query = '' if uri.query
 
           uri
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
@@ -4,6 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry/common/http/request_attributes'
+
 module OpenTelemetry
   module Instrumentation
     module Faraday
@@ -42,14 +44,10 @@ module OpenTelemetry
           attr_reader :app
 
           def span_creation_attributes(http_method:, url:)
-            instrumentation_attrs = {
-              'http.method' => http_method, 'http.url' => url.to_s
-            }
             config = Faraday::Instrumentation.instance.config
-            instrumentation_attrs['peer.service'] = config[:peer_service] if config[:peer_service]
-            instrumentation_attrs.merge(
-              OpenTelemetry::Common::HTTP::ClientContext.attributes
-            )
+            attributes = OpenTelemetry::Common::HTTP::RequestAttributes.from_request(http_method, url, config)
+            attributes['peer.service'] = config[:peer_service] if config[:peer_service]
+            attributes.merge(OpenTelemetry::Common::HTTP::ClientContext.attributes)
           end
 
           def tracer
--- a/instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
+++ b/instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
@@ -19,6 +19,7 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
     ::Faraday.new('http://username:password@example.com') do |builder|
       builder.adapter(:test) do |stub|
         stub.get('/success') { |_| [200, {}, 'OK'] }
+        stub.get('/success?foo=bar') { |_| [200, {}, 'OK'] }
         stub.get('/failure') { |_| [500, {}, 'OK'] }
         stub.get('/not_found') { |_| [404, {}, 'OK'] }
       end
@@ -40,19 +41,29 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
 
   describe 'first span' do
     before do
-      instrumentation.install
+      instrumentation.instance_variable_set(:@installed, false)
+      instrumentation.install(hide_query_params: true)
     end
 
     it 'has http 200 attributes' do
-      response = client.get('/success')
+      response = client.get('/success?foo=bar')
 
       _(span.name).must_equal 'HTTP GET'
       _(span.attributes['http.method']).must_equal 'GET'
       _(span.attributes['http.status_code']).must_equal 200
-      _(span.attributes['http.url']).must_equal 'http://example.com/success'
+      _(span.attributes['http.url']).must_equal 'http://example.com/success?'
       _(response.env.request_headers['Traceparent']).must_equal(
         "00-#{span.hex_trace_id}-#{span.hex_span_id}-01"
       )
+    end
+
+    it 'does not hide query params when hide_query_params is false' do
+      instrumentation.instance_variable_set(:@installed, false)
+      instrumentation.install(hide_query_params: false)
+
+      client.get('/success?foo=bar')
+
+      _(span.attributes['http.url']).must_equal 'http://example.com/success?foo=bar'
     end
 
     it 'has http.status_code 404' do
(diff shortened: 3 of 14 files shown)


common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,35 @@
...
+        def from_request(request_method, uri, config = {})
+          uri = hide_query_params(uri) if config[:hide_query_params]
+
+          {
+            'http.method' => request_method,
+            'http.scheme' => uri.scheme,
+            'http.target' => uri.request_uri,
+            'http.url' => uri.to_s,
+            'peer.hostname' => uri.host,
+            'peer.port' => uri.port
```
ericmustin, line 23: I think the spec wants these to be `net.peer.*`,

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
```diff
@@ -42,14 +44,10 @@ def call(env)
...
-            instrumentation_attrs = {
-              'http.method' => http_method, 'http.url' => url.to_s
-            }
             config = Faraday::Instrumentation.instance.config
-            instrumentation_attrs['peer.service'] = config[:peer_service] if config[:peer_service]
-            instrumentation_attrs.merge(
-              OpenTelemetry::Common::HTTP::ClientContext.attributes
-            )
+            attributes = OpenTelemetry::Common::HTTP::RequestAttributes.from_request(http_method, url, config)
+            attributes['peer.service'] = config[:peer_service] if config[:peer_service]
```
ericmustin, line 49: same as https://github.com/open-telemetry/opentelemetry-ruby/pull/769/files#r638374660

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,35 @@
...
+          }.compact
+        end
+
+        def hide_query_params(uri)
+          uri.query = '' if uri.query
```
fbogsany, line 28: Does mutating the `uri` affect the callers of `from_request`? Should this `dup` the uri first?

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,35 @@
...
+
+module OpenTelemetry
+  module Common
+    module HTTP
+      # RequestAttributes contains common helpers for setting http client attributes in spans
+      module RequestAttributes
+        extend self
+
+        def from_request(request_method, uri, config = {})
+          uri = hide_query_params(uri) if config[:hide_query_params]
```
ericmustin, line 15: I think there are a lot of ways to go with this implementation under the hood but overall I like this config option and think it's a reasonable thing to add.  Generally speaking though, for http client helpers, it's an interesting dilemma.

On one hand, you could add more config for more use cases with a middle ground approach here to add some or all query parameters as a span attribute. `hide_query_params` could optionally take an allow-llst maybe, or if changed to `show_query_params` could look something like:

```
require `cgi`
def show_query_params(uri, span)
  if config[:show_query_params] && config[:show_query_params].length > 0

    return if config[:show_query_params][0] == 'all'

    params = CGI::parse(uri.query)

    config[:show_query_params].each do |query_param|
      if params[query_param]
         span.set_attribute("request.#{query_param}", params[query_param])
      end
    end
  else
    uri.query = '' if uri.query
    uri
  end
end
```

But, maybe the ideal end state is to allow users an arbitrary proc on the span. `from_request` could have a similar method like `after_request` and users can do whatever they want to their spans.

```
def after_request(span, request_method, uri, config = {})
  if config[:after_request]
    after_request = config[:after_request]
    after_request.call(span, request_method,uri,config)
  end

  span
end
```

Anyway, just something to think about for future work but this is a great improvement so it'll be great to see this merged and available.

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
ericmustin: Oh, hm this is how it was before. disregard i guess?

common/lib/opentelemetry/common/http/request_attributes.rb
ericmustin: Oh, hm this is how it was before. disregard i guess?

instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
```diff
@@ -40,21 +41,31 @@
 
   describe 'first span' do
     before do
-      instrumentation.install
+      instrumentation.instance_variable_set(:@installed, false)
+      instrumentation.install(hide_query_params: true)
```
ericmustin, line 45:
```suggestion
      instrumentation.install
```

instrumentation/http/test/instrumentation/http/patches/client_test.rb
```diff
@@ -19,7 +19,7 @@
     @orig_propagation = OpenTelemetry.propagation
     propagator = OpenTelemetry::Trace::Propagation::TraceContext.text_map_propagator
     OpenTelemetry.propagation = propagator
-    instrumentation.install({})
+    instrumentation.install(hide_query_params: true)
```
ericmustin, line 22:
```suggestion
    instrumentation.install({})
```

ericmustin: I don't think this should get changed as we want to prevent a regression from the default behavior changing

ericmustin (changes requested):
A couple smaller changes and feedback, but overall I think this is pretty close, i'd like to make sure we reach consensus around the config name and functionality and such if anyone else has thoughts.

Also is there a specific document or comment you were referring to regarding posture around sensitive information? I was wondering this myself. i'd certainly prefer off by default for the query string.

Nice work here!

common/lib/opentelemetry/common/http/request_attributes.rb
chrisholmes: I haven't seen anything like it in tests, but I don't want to leave it to chance so I'll make the change.

instrumentation/http/test/instrumentation/http/patches/client_test.rb
chrisholmes: I believe it has to be like this because it's set as false in another test and as a singleton the state is shared across tests so it could be set to false prior to this. I've run into a similar issue with this before and tests were flakey depending on the order they were run. The `Instrumentation::Base` class needs some enhancements to make it more test friendly, but I'm reluctant to do it in the scope of this PR.

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,35 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Common
+    module HTTP
+      # RequestAttributes contains common helpers for setting http client attributes in spans
+      module RequestAttributes
```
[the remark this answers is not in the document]
chrisholmes: good point!

chrisholmes:
@ericmustin, I was referring to the the comments made here rather than a specific doc: https://github.com/open-telemetry/opentelemetry-ruby/pull/737#discussion_r623876419

fbogsany:
We discussed this on the weekly SIG call yesterday. These helpers don't belong in `opentelemetry-common` - that's already a bit of a grab bag of utilities that needs to be refactored. The specific concern is that the SDK depends on `...-common`, as well as exporters required by the spec.

We would rather build an `opentelemetry-instrumentation-helpers` gem that holds helpers specific to instrumentation. The code in this PR belongs in that gem.

Additionally, we should think about common instrumentation options and how to ensure that related instrumentation uses the same option names, validations and defaults. Factoring those options out into modules that can be included in, e.g., HTTP client instrumentation, would be very useful.

chrisholmes:
> We discussed this on the weekly SIG call yesterday. These helpers don't belong in `opentelemetry-common` - that's already a bit of a grab bag of utilities that needs to be refactored. The specific concern is that the SDK depends on `...-common`, as well as exporters required by the spec.
> 
> We would rather build an `opentelemetry-instrumentation-helpers` gem that holds helpers specific to instrumentation. The code in this PR belongs in that gem.

@fbogsany, this all sounds very sensible. I definitely would not have chosen to add to `-common` if there wasn't already some use. I'm happy to make the change, but would you prefer to see it in a separate PR that we can merge in before this one?

> Additionally, we should think about common instrumentation options and how to ensure that related instrumentation uses the same option names, validations and defaults. Factoring those options out into modules that can be included in, e.g., HTTP client instrumentation, would be very useful.

Yes, there is definitely space for more reuse and abstractions here, but maybe too much for a single PR?

Would it help for me to join SIG to discuss the scope of this next week?

fbogsany:
> Would it help for me to join SIG to discuss the scope of this next week?

Sure, that's a great idea - thanks!

instrumentation/http/test/instrumentation/http/patches/client_test.rb
ericmustin: i see, fair enough.

instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
ericmustin: closing, same as https://github.com/open-telemetry/opentelemetry-ruby/pull/769#discussion_r638955892

plantfansam:

Portions of this PR are related to instrumentation gems, so we'll need you to re-open it against `opentelemetry-ruby-contrib`. Sorry for the inconvenience! 

To do that, you can:

kaylareopelle:
👋  Hi, @chrisholmes! Thank you for your PR! 

As [Sam mentioned](https://github.com/open-telemetry/opentelemetry-ruby/pull/769#issuecomment-1159105049), Ruby instrumentation now lives in the [opentelemetry-ruby-contrib repo](https://github.com/open-telemetry/opentelemetry-ruby-contrib). We'd be happy to take another look at your work in that context. 

Since most of this PR is instrumentation-related, I'm going to close it. 

We appreciate your contribution and hope to work with you again soon.

Issue #250, opened 2020-05-12
Resque instrumentation adapter


mwear:
Convert Datadog instrumentation for Resque to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/resque)

PR #779, opened 2021-05-27
feat: add resque instrumentation

robertlaurin:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/250


instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,44 @@
...
+            attributes = {
+              'messaging.system' => 'resque',
+              'messaging.destination' => queue.to_s,
+              'messaging.destination_kind' => 'queue',
+              'messaging.resque.job_class' => job_class
+            }
+
+            extracted_context = OpenTelemetry.propagation.extract(@payload)
+            OpenTelemetry::Context.with_current(extracted_context) do
+              tracer.in_span("#{queue} process", attributes: attributes, kind: :consumer) do
```
ahayworth, line 25: Similar to the suggestion in #726, it might be nice to support a config option to name the span after the job class as well.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
```diff
@@ -0,0 +1,50 @@
...
+                'messaging.destination' => queue.to_s,
+                'messaging.destination_kind' => 'queue'
+              }
+
+              if (job_class = item[:class])
+                attributes['messaging.resque.job_class'] = job_class
+              end
+
+              tracer.in_span("#{queue} send", attributes: attributes, kind: :producer) do
+                OpenTelemetry.propagation.inject(item)
```
ahayworth, line 33: Also similar to comments in #726 - this forces parent/child propagation semantics for background jobs. Should we use links instead, and also include the option for the propagation kind an end-user wants?

ahayworth (commented):
Yay! 🎉 

I know this is draft still, but I wanted to mention two thoughts about its direction (regarding span names and propagation styles) in line with the way things are going on the ActiveJob PR. I don't think we necessarily need the precise same semantics between them, but it might be nice to harmonize in places.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -0,0 +1,38 @@
...
+          require_dependencies
+          patch
+        end
+
+        present do
+          defined?(::Resque)
+        end
+
+        option :job_class_span_names, default: false, validate: :boolean
+        option :enqueuer_propagation, default: false, validate: :boolean
```
robertlaurin, line 22: Not sure if this config name is too cryptic, open to alternatives.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
```diff
@@ -0,0 +1,38 @@
...
+        install do |_config|
+          require_dependencies
+          patch
+        end
+
+        present do
+          defined?(::Resque)
+        end
+
+        option :job_class_span_names, default: false, validate: :boolean
```
robertlaurin, line 21: sidekiq has this as `enable_job_class_span_names` I think that is less rubyish.  I'll update sidekiq if I don't get any pushback on this 😄

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
robertlaurin: Also need to add this config option to sidekiq.

ahayworth: I called it `context_propagation`, but that might be similarly vague / cryptic. I'll be happy to standardize on whatever we decide, though.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
ahayworth: I picked `enable_job_class_span_names` to mirror sidekiq actually, but as with the span name option I'm really happy to harmonize on whatever. Hopefully its not something people have to actually interact with frequently 😆

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
robertlaurin: What's a succinct way of saying "this config options continues the trace from the enqueuing span to the job process span"?  😄

robertlaurin:
```rb
        option :continue_trace_from_enqueuing_span_to_job_process_span, default: false, validate: :boolean
```

Seems a little wordy

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
robertlaurin: >I picked enable_job_class_span_names to mirror sidekiq actually, but as with the span name option I'm really happy to harmonize on whatever.

I really wanted to shorten the config option name 🙏 

>Hopefully its not something people have to actually interact with frequently

Definitely.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
ahayworth: Yeah .... that's quite the option name there. Perhaps `propagation_style` ?

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
ahayworth: I think `job_class_span_names` is certainly fine then, if one of the goals is shorter config option names!

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
robertlaurin: Apparently config option naming is where I like to bikeshed 😄 

* `enqueuer_propagation` is so so
* `context_propagation` is better
* `propagation_style` leads me to believe I would need to pass in something more than a boolean
* `propagate_context` ... this one I like :)

ahayworth: `propagation_style` <- to be fair, in #726 we accept `:link`, `:child`, or `:none` 😆

fbogsany: I like `propagation_style` with the symbol options - much clearer than the boolean flag.

instrumentation/resque/test/opentelemetry/instrumentation/resque/patches/resque_job_test.rb
```diff
@@ -0,0 +1,91 @@
...
+
+      it 'uses the job class name for the span name' do
+        ::Resque.enqueue(DummyJob)
+        work_off_jobs
+
+        _(job_span.name).must_equal('DummyJob process')
+      end
+    end
+
+    describe 'when job_class_span_names is enabled' do
```
fbogsany, line 67:
```suggestion
    describe 'when propagate_context is enabled' do
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
fbogsany: You still need to inject for links, but otherwise I agree.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,61 @@
...
+            links << OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
+
+            if config[:propagate_context]
+              OpenTelemetry::Context.with_current(extracted_context) do
+                tracer.in_span(span_name, attributes: attributes, links: links, kind: :consumer) do
+                  super
+                end
+              end
+            else
+              tracer.in_span(span_name, attributes: attributes, links: links, kind: :consumer) do
```
fbogsany, line 42: I think we should only have the link if we're starting a new trace. If we're continuing the trace, the link is implicitly provided by the parent context.

```suggestion
            if config[:propagate_context]
              OpenTelemetry::Context.with_current(extracted_context) do
                tracer.in_span(span_name, attributes: attributes, kind: :consumer) do
                  super
                end
              end
            else
              links = []
              links << OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
              tracer.in_span(span_name, attributes: attributes, links: links, kind: :consumer) do
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
robertlaurin: Updated to use `propagation_style` and to match the behaviour from #726

ahayworth (approved):
LGTM! I updated #726 to match the new option names 😄

instrumentation/resque/lib/opentelemetry/instrumentation/resque/instrumentation.rb
fbogsany: 🚲 shed: you could make this an enum like `propagation_style` below. E.g.
```ruby
option :span_naming, default: :queue, validate: ->(opt) { %I[job_class queue].include?(opt) }
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,59 @@
...
+              links << OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
+            end
+
+            parent_context = (config[:propagation_style] == :child ? extracted_context : OpenTelemetry::Context.current)
+
+            OpenTelemetry::Context.with_current(parent_context) do
+              tracer.in_span(span_name, attributes: attributes, links: links, kind: :consumer) do
+                super
+              end
+            end
```
fbogsany, line 43: This is actually more complex to reason about than it seems. I think we need to propagate the `extracted_context` in both cases, because it may contain additional context like `baggage`, but we need to create a root span when `config[:propagation_style] == :link`. I would suggest splitting the two cases and handling them separately rather than the complex logic here that tries to use a single path for the `in_span` and `super`.

```suggestion
            OpenTelemetry::Context.with_current(extracted_context) do
              if config[:propagation_style] == :link
                links = []
                span_context = OpenTelemetry::Trace.current_span(extracted_context).context
                links << OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
                OpenTelemetry::Trace.with_span(tracer.start_root_span(span_name, attributes: attributes, links: links, kind: :consumer)) do
                  super
                end
              else
                tracer.in_span(span_name, attributes: attributes, kind: :consumer) do
                  super
                end
              end
            end
```

fbogsany: Whenever I refactor code in the GitHub web UI it is obviously perfect - I'm sure you simply applied the suggestion incorrectly! :trollface:

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,65 @@
...
+              'messaging.destination' => queue.to_s,
+              'messaging.destination_kind' => 'queue',
+              'messaging.resque.job_class' => job_class
+            }
+
+            span_name = case config[:span_naming]
+                        when :queue then "#{queue} process"
+                        when :job_class then "#{job_class} process"
+                        else "#{queue} process"
+                        end
```
fbogsany, line 27: Just curious: why do you prefer duplicating the `:queue` and default cases rather than:
```suggestion
            span_name = config[:span_naming] == :job_class ? "#{job_class} process" : "#{queue} process"
```
or
```suggestion
            span_name = case config[:span_naming]
                        when :job_class then "#{job_class} process"
                        else "#{queue} process"
                        end
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,65 @@
...
+              else
+                links = []
+                span_context = OpenTelemetry::Trace.current_span(extracted_context).context
+                links << OpenTelemetry::Trace::Link.new(span_context) if config[:propagation_style] == :link && span_context.valid?
+                span = tracer.start_root_span(span_name, attributes: attributes, links: links, kind: :consumer)
+                OpenTelemetry::Trace.with_span(span) do
+                  super
+                rescue Exception => e # rubocop:disable Lint/RescueException
+                  span&.record_exception(e)
+                  span&.status = OpenTelemetry::Trace::Status.new(OpenTelemetry::Trace::Status::ERROR, description: "Unhandled exception of type: #{e.class}")
```
fbogsany, line 43:
```suggestion
                  span.record_exception(e)
                  span.status = OpenTelemetry::Trace::Status.new(OpenTelemetry::Trace::Status::ERROR, description: "Unhandled exception of type: #{e.class}")
```
I don't think `span` can be `nil` here.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -0,0 +1,65 @@
...
+                  raise e
+                ensure
+                  span&.finish
```
fbogsany, line 46: or here
```suggestion
                  span.finish
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
```diff
@@ -0,0 +1,56 @@
...
+
+              if (job_class = item[:class])
+                attributes['messaging.resque.job_class'] = job_class
+              end
+
+              span_name = case config[:span_naming]
+                          when :queue then "#{queue} send"
+                          when :job_class then "#{job_class} send"
+                          else "#{queue} send"
+                          end
```
fbogsany, line 36: Depending on your previous answer:
```suggestion
              span_name = case config[:span_naming]
                          when :job_class then "#{job_class} send"
                          else "#{queue} send"
                          end
```

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
robertlaurin: >Just curious: why do you prefer duplicating the :queue and default cases rather than

Less about duplication and more about the explicitness of it.  The alternative is cleaner though, I'll commit that 👍

fbogsany (approved):

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
ahayworth: I'm updating #726 to match; but it occurs to me that explicitly calling `start_root_span` could have unintended consequences if a trace had already been started by something even lower-level (for instance, `redis` instrumentation).

Is that important to worry about? I'm not sure.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
```diff
@@ -28,17 +28,18 @@ def perform # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
-              tracer.in_span(span_name, attributes: attributes, links: links, kind: :consumer) do
-                super
+            OpenTelemetry::Context.with_current(extracted_context) do
+              if config[:propagation_style] == :link
+                links = []
+                span_context = OpenTelemetry::Trace.current_span(extracted_context).context
+                links << OpenTelemetry::Trace::Link.new(span_context) if span_context.valid?
+                OpenTelemetry::Trace.with_span(tracer.start_root_span(span_name, attributes: attributes, links: links, kind: :consumer)) do
+                  super
+                end
```
ahayworth: Do you need to call `.finish` on the root span that was created? None of my tests pass unless I do that; I'm wondering if this leads to a dangling span here.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
fbogsany: We're explicitly starting a new trace for the job as a unit of work, so I think `start_root_span` makes sense here.

Additional context here: at Shopify we also trace the job worker loop (at a low sample rate) and we don't want the job trace to be associated with that "job infrastructure" trace.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
fbogsany: 🤦 yes, we do. Good catch!

ahayworth: Oh just kidding, I see where that's happening later on in another commit. Whoops.

robertlaurin:
The last thing to do here is testing compatibility with the active job wrapper I realized I haven't tested that manually or otherwise.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_job.rb
ahayworth: > we don't want the job trace to be associated with that "job infrastructure" trace.

I think I could support that as well, it seems reasonable.

ahayworth:
> The last thing to do here is testing compatibility with the active job wrapper I realized I haven't tested that manually or otherwise.

I'd be interested to see what happens - the interaction between the different context-propagation options could be really interesting. And/or confusing. 😆

robertlaurin:
> > The last thing to do here is testing compatibility with the active job wrapper I realized I haven't tested that manually or otherwise.
> 
> I'd be interested to see what happens - the interaction between the different context-propagation options could be really interesting. And/or confusing. 😆

I don't think we've really talked about it much yet, but I always figured you would not use resque/sidekiq instrumentation in addition to active job instrumentation.  They seem to be very closely overlapping.  Or possibly there would be some sort of strange hybrid approach 🤷

robertlaurin:
I added this commit 676bf3417dffcdfce1f77c11cb0ee013124b636e which does a little bit of digging to see if the job is being wrapped with the active job queue adapter.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
```diff
@@ -0,0 +1,60 @@
...
+          # Module to prepend to Resque singleton class
+          module ClassMethods
+            def push(queue, item)
+              # Check if the job is wrapped is being wrapped by ActiveJob
+              # before retrieving the job class name
+              job_class = if item[:args][0]&.is_a?(Hash)
+                            item[:args][0]['job_class']
+                          else
+                            item[:class]
+                          end
```
fbogsany, line 28: Is there any way to do this more reliably or do we have to implicitly depend on AJ internals here?

fbogsany (approved):
One question about AJ-compatibility brittleness, but otherwise this looks great!

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
robertlaurin: I don't think so.  I don't think this is particularly great either but I'm not sure what the better option is to maintain compatibility.  

It's how the resque UI itself does it https://github.com/resque/resque/pull/1525/files#diff-c63425703f7922acdeace3db5e0c9d98fb6ab3ae04fdab863c7161b46bb3d618R2

fbogsany: They do have the extra check there for:
```ruby
job['class'] == 'ActiveJob::QueueAdapters::ResqueAdapter::JobWrapper'
```
Is that worth adding?

ahayworth:
> I don't think we've really talked about it much yet, but I always figured you would not use resque/sidekiq instrumentation in addition to active job instrumentation. They seem to be very closely overlapping. Or possibly there would be some sort of strange hybrid approach 🤷

I think it's fine to say "don't use them both" personally.

instrumentation/resque/lib/opentelemetry/instrumentation/resque/patches/resque_module.rb
ahayworth: > Is that worth adding?

👍 I think so

> Is there any way to do this more reliably or do we have to implicitly depend on AJ internals here?

It is ugly, but I also think this is about the best way to do it. 😞

robertlaurin: Updated https://github.com/open-telemetry/opentelemetry-ruby/pull/779/commits/26e24198f99d8ec739bc9862ade9644997f1f796

PR #781, opened 2021-05-28
fix: Support Case Insensitive Trace and Span IDs

arielvalentin:
This change allows systems that generate uppercase hexadecimal values to properly propagate OTTrace context.

### DIFF
--- a/propagator/ottrace/lib/opentelemetry/propagator/ottrace/text_map_propagator.rb
+++ b/propagator/ottrace/lib/opentelemetry/propagator/ottrace/text_map_propagator.rb
@@ -18,8 +18,8 @@ module OpenTelemetry
       # Propagates context using OTTrace header format
       class TextMapPropagator
         PADDING = '0' * 16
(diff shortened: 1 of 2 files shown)


ahayworth (approved):
We're carrying this monkey-patch internally, because we have an opentracing system that's generating uppercased trace IDs. At least until I can fix it, anyways.

It wasn't really clear to us whether or not there were specifications stating that the hex opentracing IDs needed to be lowercase, but we figured we would upstream this patch and see what people thought. It could improve compatibility for other mixed opentelemetry / opentracing worlds (although this problem may just be one entirely created and experienced by us internally 😆 )

fbogsany:
The [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#retrieving-the-traceid-and-spanid) states that the hex encoded trace ID and span ID returned from `SpanContext` must be lowercase, but there's no restriction on propagators. I don't know anything about the OTTrace format, but this seems fine.

fbogsany (approved):

PR #782, opened 2021-05-29
fix: use monotonic clock where possible

fbogsany:
This is a followup to #717 addressing #709 (which I probably should not have closed). It uses monotonic clock offsets to compute timestamps for `Span` and `Event` unless explicit timestamps are provided.

## TODO:
- [ ] tests
- [x] the approach here does not use monotonic offsets to compute other timestamps if the `start_timestamp` was explicitly provided. Should it? The tradeoff is that we need to store a `realtime_start_timestamp` separately from `start_timestamp`.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -291,13 +291,14 @@ module OpenTelemetry
           @total_recorded_events = 0
           @total_recorded_links = links&.size || 0
           @total_recorded_attributes = attributes&.size || 0
-          @monotonic_start_timestamp = start_timestamp.nil? ? monotonic_now : nil
-          @start_timestamp = if start_timestamp.nil? && parent_span.recording? && parent_span.monotonic_start_timestamp
-                               relative_realtime(parent_span.start_timestamp, parent_span.monotonic_start_timestamp)
-                             elsif start_timestamp.nil?
-                               realtime_now
-                             else
+          @monotonic_start_timestamp = monotonic_now
+          @realtime_start_timestamp = realtime_now
+          @start_timestamp = if start_timestamp
                                time_in_nanoseconds(start_timestamp)
+                             elsif parent_span.recording?
+                               relative_realtime(parent_span.realtime_start_timestamp, parent_span.monotonic_start_timestamp)
+                             else
+                               @realtime_start_timestamp
                              end
           @end_timestamp = nil
           @attributes = attributes.nil? ? nil : Hash[attributes] # We need a mutable copy of attributes.
@@ -311,7 +312,7 @@ module OpenTelemetry
 
         protected
 
-        attr_reader :monotonic_start_timestamp
+        attr_reader :monotonic_start_timestamp, :realtime_start_timestamp
 
         private
 
@@ -389,6 +390,7 @@ module OpenTelemetry
 
         def relative_timestamp(timestamp)
           return time_in_nanoseconds(timestamp) unless timestamp.nil?
+          # TODO francis fix this:
           return relative_realtime(start_timestamp, monotonic_start_timestamp) unless monotonic_start_timestamp.nil?
 
           realtime_now
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -276,7 +276,7 @@ module OpenTelemetry
         end
 
         # @api private
-        def initialize(context, parent_context, parent_span, name, kind, parent_span_id, span_limits, span_processor, attributes, links, start_timestamp, resource, instrumentation_library) # rubocop:disable Metrics/AbcSize
+        def initialize(context, parent_context, parent_span, name, kind, parent_span_id, span_limits, span_processors, attributes, links, start_timestamp, resource, instrumentation_library) # rubocop:disable Metrics/AbcSize,  Metrics/CyclomaticComplexity, Metrics/MethodLength,  Metrics/PerceivedComplexity
           super(span_context: context)
           @mutex = Mutex.new
           @name = name
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -131,7 +131,7 @@ module OpenTelemetry
         #
         # @return [self] returns itself
         def add_event(name, attributes: nil, timestamp: nil)
-          event = Event.new(name, truncate_attribute_values(attributes), wall_clock(timestamp))
+          event = Event.new(name, truncate_attribute_values(attributes), relative_timestamp(timestamp))
 
           @mutex.synchronize do
             if @ended
@@ -233,7 +233,7 @@ module OpenTelemetry
               OpenTelemetry.logger.warn('Calling finish on an ended Span.')
               return self
             end
-            @end_timestamp = wall_clock(end_timestamp)
+            @end_timestamp = relative_timestamp(end_timestamp)
             @attributes = validated_attributes(@attributes).freeze
             @events.freeze
             @ended = true
@@ -276,7 +276,7 @@ module OpenTelemetry
         end
 
         # @api private
-        def initialize(context, parent_context, name, kind, parent_span_id, span_limits, span_processors, attributes, links, start_timestamp, resource, instrumentation_library) # rubocop:disable Metrics/AbcSize
+        def initialize(context, parent_context, parent_span, name, kind, parent_span_id, span_limits, span_processor, attributes, links, start_timestamp, resource, instrumentation_library) # rubocop:disable Metrics/AbcSize
           super(span_context: context)
           @mutex = Mutex.new
           @name = name
@@ -291,7 +291,14 @@ module OpenTelemetry
           @total_recorded_events = 0
           @total_recorded_links = links&.size || 0
           @total_recorded_attributes = attributes&.size || 0
-          @start_timestamp = wall_clock(start_timestamp)
+          @monotonic_start_timestamp = start_timestamp.nil? ? monotonic_now : nil
+          @start_timestamp = if start_timestamp.nil? && parent_span.recording? && parent_span.monotonic_start_timestamp
+                               relative_realtime(parent_span.start_timestamp, parent_span.monotonic_start_timestamp)
+                             elsif start_timestamp.nil?
+                               realtime_now
+                             else
+                               time_in_nanoseconds(start_timestamp)
+                             end
           @end_timestamp = nil
           @attributes = attributes.nil? ? nil : Hash[attributes] # We need a mutable copy of attributes.
           trim_span_attributes(@attributes)
@@ -301,6 +308,10 @@ module OpenTelemetry
         end
 
         # TODO: Java implementation overrides finalize to log if a span isn't finished.
+
+        protected
+
+        attr_reader :monotonic_start_timestamp
 
         private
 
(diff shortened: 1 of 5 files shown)


robertlaurin (approved):
Looks good to me, rubocop isn't happy though.

```
lib/opentelemetry/sdk/trace/span.rb:276:9: C: Metrics/CyclomaticComplexity: Cyclomatic complexity for initialize is too high. [10/6]
        def initialize(context, parent_context, parent_span, name, kind, parent_span_id, trace_config, span_processor, attributes, links, start_timestamp, resource, instrumentation_library) # rubocop:disable Metrics/AbcSize ...

**Review by ericmustin — APPROVED** [review]
rubocop aside

**Comment by fbogsany** [comment]
> * the approach here does not use monotonic offsets to compute other timestamps if the `start_timestamp` was explicitly provided. Should it? The tradeoff is that we need to store a `realtime_start_timestamp` separately from `start_timestamp`.

Any thoughts on this?

Any ideas on how to test this? Just mock the `span.realtime_now` and `span.monotonic_now`?

**Comment by ahayworth** [comment]
> the approach here does not use monotonic offsets to compute other timestamps if the start_timestamp was explicitly provided. Should it? The tradeoff is that we need to store a realtime_start_timestamp separately from start_timestamp.
Any thoughts on this?

It is entirely possible that I don't fully understand the implications here (or anything at all, really). But after thinking about it a bit, I don't think we should use monotonic offsets to compute other timestamps if you provide `start_timestamp`. My line of thinking is "if you provide a `start_timestamp`, then it is likely that other timestamps may not be set in realtime anyways".  

Or put differently - in the use-case I thought of during the SIG meeting (adding spans asynchronously), if you're _starting_ the span after the fact; then I would imagine the events you may add and the actual call to finish the span are also happening asynchronously. And in that world, it seems like it wouldn't make any sense to use the offsets there (and might be less accurate).

Of course: it is quite probable that I am vastly misunderstanding the issue to start with - if that seems like its the case, I'll be happy to retract this comment entirely and let people discuss it who know what's at stake more than me. 😄 😄 😄

**Comment by fbogsany** [comment]
https://github.com/open-telemetry/opentelemetry-ruby/blob/b816345f8f21be7e9a7497abafb0761dfa5d8f0e/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L66-L83 is one example of creating a span representing work done earlier. Here the end time is assumed to line up with the end time of the Rack span, so using a monotonic clock to compute the end time is probably the right thing to do.

The span creation is wrong - it should pass the start time explicitly instead of adding it as an attribute: https://github.com/open-telemetry/opentelemetry-ruby/blob/b816345f8f21be7e9a7497abafb0761dfa5d8f0e/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L97-L106

The other interesting case (which we don't implement yet) is a Kafka `receive` span, which should have a link to the corresponding `send` span - something like:
```ruby
start_time = Time.now
message = Kafka.receive # IDK what the actual API is
parent_context = OpenTelemetry.propagation.extract(message.headers)
span_context = OpenTelemetry::Trace.current_span(parent_context).context
links = [OpenTelemetry::Trace::Link.new(span_context)] if span_context.valid?
tracer.start_span("#{message.topic} receive", links: links, start_time: start_time).finish
```
Here, again, the finish time is local and implicit, although it is less useful in this case to use the monotonic clock.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -30,13 +30,14 @@ def start_root_span(name, attributes: nil, links: nil, start_timestamp: nil, kin
...
-          parent_span_context = OpenTelemetry::Trace.current_span(with_parent).context
+          parent_context = with_parent || Context.current
+          parent_span = OpenTelemetry::Trace.current_span(parent_context)
+          parent_span_context = parent_span.context
           if parent_span_context.valid?
             parent_span_id = parent_span_context.span_id
             trace_id = parent_span_context.trace_id
           end
-          @tracer_provider.internal_create_span(name, kind, trace_id, parent_span_id, attributes, links, start_timestamp, with_parent, @instrumentation_library)
+          @tracer_provider.internal_create_span(name, kind, trace_id, parent_span_id, attributes, links, start_timestamp, parent_context, parent_span, @instrumentation_library)
```
arielvalentin, line 40: Since this now passes in the parent span, does this function need all of the other parent related attributes like `parent_span_id` that could be queried on the `parent_span` object?

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -373,9 +401,26 @@ def append_event(events, event) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
           events << event
         end
 
-        def wall_clock(timestamp)
-          timestamp = (timestamp.to_r * 1_000_000_000).to_i unless timestamp.nil?
-          timestamp || Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
+        def relative_timestamp(timestamp)
```
arielvalentin, line 404: 🤔 I feel like we may be missing a concept here that encapsulates the time management strategy here. 

Perhaps objects like a `RelativeTime` vs a `RealTime` instead of making them functions on the span. 

I don't have an explicit ask here, just a thought experiment.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -291,17 +291,45 @@ def initialize(context, parent_context, name, kind, parent_span_id, span_limits,
...
+                                        relative_realtime(parent_span.realtime_start_timestamp, parent_span.monotonic_start_timestamp)
+                                      else
+                                        realtime_now
+                                      end
+          @start_timestamp = if start_timestamp
+                               time_in_nanoseconds(start_timestamp)
+                             else
+                               @realtime_start_timestamp
+                             end
+          @end_timestamp = nil
```
arielvalentin, line 323: Do you plan on adding unit tests to exercise these flows? I do not think I saw any as part of these changes.

sdk/lib/opentelemetry/sdk/trace/span.rb
fbogsany: They're all local to `SDK::Trace::Span` - we don't need a reusable abstraction, and I'm explicitly trying to avoid object allocation (the values here should all be machine words, except for explicit timestamps).

sdk/lib/opentelemetry/sdk/trace/span.rb
fbogsany: Yep - the "tests" checkbox remains unchecked 😄

sdk/lib/opentelemetry/sdk/trace/tracer.rb
fbogsany: That's fair - I'll move more of this into `internal_create_span` - the current code shape is the result of a merge without much thought.

sdk/lib/opentelemetry/sdk/trace/span.rb
arielvalentin: My apologies. I thought this was ready for review.

fbogsany: Sorry - I can change the title, I guess. I accidentally didn't open it as Draft, and I don't think I can move it back into that state.

robertlaurin: I've rebased and added tests to this PR.  Ready for another round of reviews.

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -200,14 +201,41 @@
       _(events.first.attributes).must_equal(attrs)
     end
 
-    it 'add event with timestamp' do
+    it 'uses the provided timestamp' do
       ts = Time.now
```
fbogsany, line 205: Can we use something completely weird here to ensure we're not getting the realtime or monotonic timestamps instead?

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -200,14 +201,41 @@
...
 
+    it 'sets the end timestamp relative to the span start' do
+      timestamp = Time.new('2021-11-23 12:00:00.000000 -0600')
+      timestamp_in_nano = exportable_timestamp(timestamp)
+      clock_ids_expected = [Process::CLOCK_MONOTONIC, Process::CLOCK_REALTIME]
+      return_values = [500_000_000_000_000, timestamp_in_nano]
+      clock_gettime_mock = lambda do |clock_id, unit|
+        _(clock_id).must_equal(clock_ids_expected.shift)
+        _(unit).must_equal(:nanosecond)
+        return_values.shift
```
fbogsany, line 220: Discussed IRL:
```suggestion
      timestamps = {
        Process::CLOCK_MONOTONIC => 500_000_000_000_000,
        Process::CLOCK_REALTIME => timestamp_in_nano
      }
      clock_gettime_mock = lambda do |clock_id, unit|
        _(timestamps).must_include(clock_id)
        _(unit).must_equal(:nanosecond)
        timestamps[clock_id]
```

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -200,14 +201,41 @@
...
+      end
+
+      Process.stub(:clock_gettime, 500_000_000_000_000 + 100) do
+        test_span.add_event('record something with relative time')
+        event = test_span.events[0]
+        _(event).wont_be_nil
+
+        # The expect timestamp is that of the parent offset by
+        # the drift in the monotonic clock
+        _(event.timestamp).must_equal(timestamp_in_nano + 100)
```
fbogsany, line 235: ✅ 2. Implicit Event timestamp is monotonic offset from realtime Span start timestamp

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -200,14 +201,41 @@
...
+        _(unit).must_equal(:nanosecond)
+        return_values.shift
+      end
+
+      # Create a span with deterministic time values stored on it.
+      test_span = Process.stub(:clock_gettime, clock_gettime_mock) do
+        Span.new(context, Context.empty, OpenTelemetry::Trace::Span::INVALID, 'span', SpanKind::INTERNAL, nil, span_limits, [], nil, nil, timestamp, nil, nil)
+      end
+
+      Process.stub(:clock_gettime, 500_000_000_000_000 + 100) do
```
fbogsany, line 228: Maybe we could do the lambda thing again with:
```ruby
       timestamps = {
        Process::CLOCK_MONOTONIC => 500_000_000_000_000 + 100,
        Process::CLOCK_REALTIME => 0
      }
```

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -200,14 +201,41 @@
...
-    it 'add event with timestamp' do
+    it 'uses the provided timestamp' do
       ts = Time.now
       span.add_event('added', timestamp: ts)
       events = span.events
       _(events.size).must_equal(1)
       _(events.first.timestamp).must_equal(exportable_timestamp(ts))
     end
 
+    it 'sets the end timestamp relative to the span start' do
```
fbogsany, line 212:
```suggestion
    it 'sets the implicit event timestamp relative to the span start' do
```

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -372,14 +400,39 @@
       _(span.finish).must_equal(span)
     end
 
-    it 'sets the end timestamp' do
-      span.finish
-      _(span.to_span_data.end_timestamp).wont_be_nil
+    it 'uses the provided end timestamp' do
+      ts = Time.now
```
fbogsany, line 404: Can we use some weird non-sensical timestamp here?

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -372,14 +400,39 @@
...
+      test_span = Process.stub(:clock_gettime, clock_gettime_mock) do
+        Span.new(context, Context.empty, OpenTelemetry::Trace::Span::INVALID, 'span', SpanKind::INTERNAL, nil, span_limits, [], nil, nil, timestamp, nil, nil)
+      end
+
+      Process.stub(:clock_gettime, 500_000_000_000_000 + 100) do
+        test_span.finish
+        # The expect timestamp is that of the parent offset by
+        # the drift in the monotonic clock
+        _(test_span.end_timestamp).must_equal(timestamp_in_nano + 100)
+      end
```
fbogsany, line 430: Same comments as above.

✅ 3. Implicit End timestamp is monotonic offset from realtime Span start timestamp

sdk/test/opentelemetry/sdk/trace/span_test.rb
fbogsany: ✅ Explicit end timestamp is honoured

sdk/test/opentelemetry/sdk/trace/span_test.rb
fbogsany: ✅ Explicit event timestamp is honoured

sdk/test/opentelemetry/sdk/trace/span_test.rb
robertlaurin: >Can we use some weird non-sensical timestamp here?

Not sure what you mean.  As in have a separate test that receives junk instead of a valid timestamp, or just a weird fixed timestamp?

fbogsany: Weird fixed timestamp - really anything other than `Time.now`, so we know that we're honouring the provided timestamp.

robertlaurin: Updated

robertlaurin (approved):

PR #784, opened 2021-06-02
fix: handle OTLP exporter encoding exceptions

robertlaurin:
```
gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:338:in `method_missing': "\xC3" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:338:in `as_otlp_any_value'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:331:in `as_otlp_key_value'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:294:in `block (2 levels) in as_otlp_span'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:294:in `each'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:294:in `map'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:294:in `block in as_otlp_span'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:290:in `map'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:290:in `as_otlp_span'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:269:in `block (3 levels) in encode'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:269:in `map'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:269:in `block (2 levels) in encode'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:263:in `each'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:263:in `map'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:263:in `block in encode'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:256:in `each'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:256:in `map'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:256:in `encode'
    from gems/opentelemetry-exporter-otlp-0.18.0/lib/opentelemetry/exporter/otlp/exporter.rb:76:in `export'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:187:in `block in export_batch'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:187:in `synchronize'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:187:in `export_batch'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:170:in `block in work'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:159:in `loop'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:159:in `work'
    from gems/opentelemetry-sdk-1.0.0.rc1/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:180:in `block in reset_on_fork'

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -341,6 +341,9 @@ module OpenTelemetry
 
         def as_otlp_key_value(key, value)
           Opentelemetry::Proto::Common::V1::KeyValue.new(key: key, value: as_otlp_any_value(value))
+        rescue Encoding::UndefinedConversionError => e
+          OpenTelemetry.handle_error(exception: e, message: "encoding error for key #{key} and value #{value}")
+          Opentelemetry::Proto::Common::V1::KeyValue.new(key: key, value: as_otlp_any_value('Encoding Error'))
         end
 
         def as_otlp_any_value(value)
--- a/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
+++ b/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
@@ -189,6 +189,27 @@ describe OpenTelemetry::Exporter::OTLP::Exporter do
       _(result).must_equal(SUCCESS)
     end
 
+    it 'handles encoding errors with poise and grace' do
+      log_stream = StringIO.new
+      logger = OpenTelemetry.logger
+      OpenTelemetry.logger = ::Logger.new(log_stream)
+
+      stub_request(:post, 'https://localhost:4317/v1/traces').to_return(status: 200)
+      span_data = create_span_data(total_recorded_attributes: 1, attributes: { 'a' => "\xC2".dup.force_encoding(::Encoding::ASCII_8BIT) })
+
+      result = exporter.export([span_data])
+
+      # Needed so it doesn't crash the test
+      encoded_logger_output = log_stream.string.encode('UTF-8', invalid: :replace, undef: :replace, replace: '?')
+
+      _(encoded_logger_output).must_match(
+        /ERROR -- : OpenTelemetry error: encoding error for key a and value ?/
+      )
+      _(result).must_equal(SUCCESS)
+    ensure
+      OpenTelemetry.logger = logger
+    end
+
     it 'exports a span from a tracer' do
       stub_post = stub_request(:post, 'https://localhost:4317/v1/traces').to_return(status: 200)
       processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(exporter, max_queue_size: 1, max_export_batch_size: 1)

### CODE REVIEW

File: exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb [hunk 1]
```diff
@@ -341,6 +341,9 @@ def as_otlp_span_kind(kind)
 
         def as_otlp_key_value(key, value)
           Opentelemetry::Proto::Common::V1::KeyValue.new(key: key, value: as_otlp_any_value(value))
+        rescue Encoding::UndefinedConversionError => e
+          OpenTelemetry.handle_error(exception: e, message: "encoding error for key #{key} and value #{value}")
+          Opentelemetry::Proto::Common::V1::KeyValue.new(key: key, value: as_otlp_any_value('Encoding Error'))
```
ericmustin, line 346: Can the key itself be malformed and cause a `Encoding::UndefinedConversionError`?

Also, is there a specification around behavior here that says `encode as value "Encoding Error"`

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -341,6 +341,9 @@ def as_otlp_span_kind(kind)
 
         def as_otlp_key_value(key, value)
           Opentelemetry::Proto::Common::V1::KeyValue.new(key: key, value: as_otlp_any_value(value))
+        rescue Encoding::UndefinedConversionError => e
+          OpenTelemetry.handle_error(exception: e, message: "encoding error for key #{key} and value #{value}")
```
robertlaurin, line 345: I think this should be enough information to track down the source of these errors in a production environment.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: >Can the key itself be malformed and cause a Encoding::UndefinedConversionError?

I'd expect it is vulnerable to the same issue.  Though we're not encoding the key with `as_otlp_any_value` as we are the value, which is why I have it this way.  

Maybe we should be? @fbogsany 

>Also, is there a specification around behavior here that says encode as value "Encoding Error"

I didn't find anything, and I'm open to alternatives/using spec'd conventions I missed.  I found some mention of discarding attributes in certain situations, but I think discarding in this situation might be more frustrating than what is present in the PR currently.  

Chasing down why attributes sometimes are not present sounds frustrating.

fbogsany: The [error handling guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md) suggest that this is the correct thing to do: call the configurable error handler and insert a default value (there's no guidance on what that value should be in the case of invalid encoding).

While it is technically possible for the `key` to be malformed, it is unlikely to be a problem in practice because attribute keys are typically explicitly provided rather than extracted or computed from other data. I think we should deal with the known `value` problem first, so we can figure out which key & instrumentation needs to be fixed (this is likely to be an auto-instrumentation bug).

fbogsany (approved):

PR #796, opened 2021-06-07
feat: Add ActionView instrumentation via ActiveSupport::Notifications

ahayworth:
This commit adds a safe (I hope) subscription framework for
ActiveSupport::Notifications and then uses that framework to instrument
ActionView.

Key to this PR is the fact that Rails allows you to swap out the queue for
ActiveSupport::Notifications; although documentation on that is sparse.

We implement a thin wrapper around the existing (and default) Fanout
queue and a "SpanSubscriber" that handles the actual event instrumentation.
Together, they do three special things:

- Sort the subscribers for a notification so that our "SpanSubscriber"
  class always runs first.
- Start a span when the event starts
- Finish the span when the event finishes, merges in the payload as
  span attributes, and then toggles the error bit if there was an
  uncaught exception.

There are two main problems with ActiveSupport::Notifications as they
relate to telemetry, and we attempt to solve them both here:

1. Any *other* subscriber to the notification in question could throw,
   and that could mean we are left with dangling spans. Our solution to
   this lies in how we sort the subscribers for a notification when
   handling the events in the queue wrapper. By ensuring that our
   handlers run first, we ensure that anyone *else* who might throw
   doesn't affect our span handling.

2. Span stack handling can be a nightmare. Here, we rely on a
   undocumented (?) feature (??) of notifications: if the callable
   passed in with the subscription responds to `start`/`finish` methods,
   then Rails will call `start` and `finish` on your queue object. And
   the return value of `start` from your queue is saved around the
   execution of the instrumented block, and passed back in to the
   `finish` method on your queue object. We can use that and pass around
   the span that was started *and* the previous context to restore, and
   that greatly simplifies the state management.

There are, of course, tradeoffs:

1. In this implementation, we don't allow for any kind of real
   specialization in the "SpanSubscriber" object. We could in the
   future; and individual instrumentations could always subclass it.
2. You always lose some detail by instrumenting in this way - it's never
   going to be as rich as patching the classes directly; but it is
   something of a handy-dandy shortcut.
3. We add a dependency on the `rails` instrumentation. I don't know if
   this is the right thing to do here or not, and I don't know if
   potentially installing the instrumentation many times will have a
   huge problem.

It should be noted that the Rails instrumentation is required at the moment, if you choose to use this instrumentation.


ahayworth:
cc @open-telemetry/ruby-approvers - if you'd like to give some early feedback on this, I'd appreciate it. And, for kicks and giggles, here's a screenshot of it all working:

Screen Shot 2021-06-07 at 3 08 46 PM

robbkidd:
I'm in.

instrumentation/action_view/.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.2
```
fbogsany, line 1: Mistaken commit?

instrumentation/action_view/README.md
```diff
@@ -0,0 +1,52 @@
+# OpenTelemetry ActionView Instrumentation
+
+Todo: Add a description.
```
fbogsany, line 3: TODO

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
```diff
@@ -0,0 +1,51 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module Rails
+      class SpanSubscriber
+        def start(name, id, payload)
+          prev_ctx = OpenTelemetry::Context.current
+
+          span_name = name.split('.')[0..1].reverse.join(' ')
+          span = otel_tracer.start_span(span_name, kind: :internal)
+          OpenTelemetry::Context.current = OpenTelemetry::Trace.context_with_span(span)
```
fbogsany, line 16: 😬 so... technically, `OpenTelemetry::Context.current=` should not be exposed as part of the public API (it should be `@api private`). The `attach` and `detach` methods should really be class methods (see ["Optional Global operations](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/context.md#optional-global-operations)) and they should be used here instead (they should *not* be `@api private`). The main reason to use `attach` and `detach` is that `attach` will return the previous context and `detach` will check that the correct context is being restored.

We need to refactor `Context` to more closely match the API spec.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
```diff
@@ -0,0 +1,51 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Rails
+      class SpanSubscriber
+        def start(name, id, payload)
+          prev_ctx = OpenTelemetry::Context.current
+
+          span_name = name.split('.')[0..1].reverse.join(' ')
```
fbogsany, line 14: It would be lovely to avoid these repeated allocations & string manipulation. I assume `name` does not vary much? Is it worth caching this mapping?

fbogsany:
This is really nice!

instrumentation/action_view/.ruby-version
ahayworth: 🤦 

Again. I'm going to open a PR to add these to .gitignore 😂

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
ahayworth: ...oh! That would actually make things a bit easier, I think.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
ahayworth: I think it could be cached; and I also think there are some other small wins too (could cache the sorted listeners, as well).

I'll make those changes!

instrumentation/action_view/.ruby-version
ahayworth: Removed in 97e67bf5 and banished forever in #801

instrumentation/all/opentelemetry-instrumentation-all.gemspec
```diff
@@ -26,6 +26,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 2.5.0'
 
+  spec.add_dependency 'opentelemetry-instrumentation-action_view', '~> 1.0.0.rc1'
```
robbkidd, line 29: This version constraint from `inst-all` to `inst-action_view` uses a version number later than [`inst-action_view`'s current version](https://github.com/open-telemetry/opentelemetry-ruby/blob/97e67bf59a7c379232f6d8d827b542ec5c2ecaba/instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/version.rb#L10) (0.19.0).

Found attempting to use this branch for smoke testing.

<details><summary>Unhappy Bundler</summary>

```
Resolving dependencies...
Bundler could not find compatible versions for gem "opentelemetry-instrumentation-action_view":
  In Gemfile:
    opentelemetry-instrumentation-all was resolved to 0.18.0, which depends on
      opentelemetry-instrumentation-action_view (~> 1.0.0.rc1)

Could not find gem 'opentelemetry-instrumentation-action_view (~> 1.0.0.rc1)',
  which is required by gem 'opentelemetry-instrumentation-all', in any of the sources.
```

</details>

ahayworth: I think this is because the instrumentation generator uses the api/sdk version as a default, when that might not actually be what anyone wants now that they are not kept in lockstep. cc @robertlaurin

ahayworth: @robbkidd I [updated](https://github.com/open-telemetry/opentelemetry-ruby/pull/796/commits/23b8c55e408a8f359db35c3ad455de84f8b18544) to the correct version, but since the gem is unreleased I don't know how bundler will actually behave with the `-all` gem. When I was testing locally I had just installed the gem from a `file` path in Bundler.

ahayworth: I am very interested in your smoke-testing though; this "works on my machine" ™️ but I would not be at all surprised if it breaks horribly in ways I haven't foreseen.

robertlaurin: I don't think we actually talked about what the new default should be but lets just start by setting it to something that makes more sense than the api version https://github.com/open-telemetry/opentelemetry-ruby/pull/803

robbkidd: I rubbed a lot of `gem: _, git: _, branch: _, glob: _` into my Gemfile which got Bundler to resolve them at install time OK, but I'm still having issues with library load at runtime. So I'm cloning your fork and trying it locally with `file: _` now.

instrumentation/action_view/test/opentelemetry/instrumentation/action_view/instrumentation_test.rb
```diff
@@ -0,0 +1,29 @@
...
+
+require 'test_helper'
+
+require_relative '../../../../lib/opentelemetry/instrumentation/action_view'
+
+describe OpenTelemetry::Instrumentation::ActionView do
+  let(:instrumentation) { OpenTelemetry::Instrumentation::ActionView::Instrumentation.instance }
+
+  it 'has #name' do
+    _(instrumentation.name).must_equal 'OpenTelemetry::Instrumentation::ActionView'
```
robbkidd, line 15: I am seeing spans emitted by the ActionView instrumentation appear with: `library.name: OpenTelemetry::Instrumentation::Rails`. Is that a result of [which instrumentation gets used in the railtie](https://github.com/open-telemetry/opentelemetry-ruby/pull/796/files#diff-6a1bb501f356942c4ec7170c3e818300c0509c1e0a6ac890dcfc64f54a6a714aR18)?

ahayworth: It is. I think I have an idea on how to fix that though.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
robertlaurin: The context refactor is here I kept the existing tests intact and added some new ones for the attach/detach class methods https://github.com/open-telemetry/opentelemetry-ruby/pull/807

ahayworth:
@open-telemetry/ruby-approvers I've updated to the latest `context` and `status` methods, and I think this is ready for legitimate review.

instrumentation/all/opentelemetry-instrumentation-all.gemspec
ahayworth: @robbkidd did this ever work for you?

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
```diff
@@ -0,0 +1,49 @@
...
+
+        def finish(_name, _id, payload)
+          span = payload.delete(:__opentelemetry_span)
+          token = payload.delete(:__opentelemetry_ctx_token)
+          return unless span && token
+
+          attrs = payload.reject do |k, v|
+            %i[exception exception_object].include?(k) || v.nil?
+          end
+          span.add_attributes(attrs.transform_keys(&:to_s))
```
fbogsany, line 36: Is there any risk here of the payload values (other than `exception` and `exception_object`, which you removed) having types other than String, Numeric or boolean?

fbogsany (approved):
This looks awesome. I have one question about payload value types, but otherwise this LGTM 🎉

instrumentation/all/opentelemetry-instrumentation-all.gemspec
robbkidd: I ... think yes? I'll refresh mental state on it and report back tomorrow!

robbkidd: Succeeded in using the new gem, yes.

```ruby
gem 'opentelemetry-api'
gem 'opentelemetry-sdk'

github 'ahayworth/opentelemetry-ruby', branch: 'ahayworth-activesupport-notifications', glob: 'instrumentation/*/*.gemspec' do
  gem 'opentelemetry-instrumentation-action_view'
  gem 'opentelemetry-instrumentation-rails'
end
```

Currently seeing errors and no spans produced for ActionView renderings:

```
E, [2021-06-26T00:11:38.603174 #1] ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
E, [2021-06-26T00:11:38.603274 #1] ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
E, [2021-06-26T00:11:38.603454 #1] ERROR -- : OpenTelemetry error: calls to detach should match corresponding calls to attach.
```

Of course, I could be doing something wrong.

robbkidd: @ahayworth With some judicious `puts` statements sprinkled in `SpanSubscriber`'s `start()` and `finish()`, I'm seeing that the `payload` does not have the [span or token](https://github.com/open-telemetry/opentelemetry-ruby/pull/796/files#diff-e29d49095a6b1901328545a9e5571958045335dfd677018d2c1c4e99c53ee886R29-R30) tucked into it. So, I'm thinking [the OTel replacement for Fanout](https://github.com/open-telemetry/opentelemetry-ruby/pull/796/files#diff-f637f95bc8fb1aad5c059ae044b926502075c6797033e81ff7b18c713ad4c0b4R15) isn't taking effect, maybe?

ahayworth: > So, I'm thinking the OTel replacement for Fanout isn't taking effect, maybe?

@robbkidd Quite probably, yes. Can you share how you're initializing the OpenTelemetry SDK in your tests? (And/or if you have any kind of public test app that reproduces it, that'd be fantastic too)

robbkidd: @ahayworth My test lab is very messy which could very well be the origin of my problem. I'll work on getting it into a state that is shareable.

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
fbogsany: I don't have any strong feelings other than it shouldn't blow up 😄

fbogsany: Actually, I do have one feeling I'd forgotten about...

One of the things we found very useful in our (much less impressive) ASN instrumentation was a "payload transform" capability. We allow subscribers to provide a denylist of payload keys to not include as attributes, and a transform lambda that takes the payload and returns a hash of attributes to add. We've used this to set the `peer.service` attribute, for example, and to remove things like the `:connection_id` and `:connection` object from `sql.active_record` spans.

ahayworth: Makes sense. Check out 596239ead915f10be23fa4d2f20934fd8b5f6e45, 83e5fa06c62ec60afe11639ca66a62702ea3c1bc, and 3bedfda56112ac055bd04124da6b9dfd8db1a458 for the implementation!

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
```diff
@@ -0,0 +1,89 @@
...
+      class SpanSubscriber
+        ALWAYS_VALID_PAYLOAD_TYPES = [TrueClass, FalseClass, String, Numeric, Symbol].freeze
+
+        def initialize(name:, tracer:)
+          @span_name = name.split('.')[0..1].reverse.join(' ').freeze
+          @tracer = tracer
+        end
+
+        def start(_name, _id, payload)
+          span = @tracer.start_span(@span_name, kind: :internal)
```
fbogsany, line 22: Is there a situation where we'd use a span subscriber to produce DB client spans (for example)? We've done that previously at Shopify, but this project has lower level instrumentation for a bunch of DBs, so we're not doing that anymore.

fbogsany (approved):

robertlaurin (approved):

instrumentation/rails/lib/opentelemetry/instrumentation/rails/span_subscriber.rb
ahayworth: I think you certainly could. I don't know if we _would_ or not but I don't see anything prohibiting it.

fbogsany: Should we allow passing in the span type as a config option? I'm also totally OK with punting that to a subsequent PR.

ahayworth: I think we should punt. Not that it would be terribly difficult to add or anything, but we also don't have a real need for it yet (unless you know of something coming up that I don't).

ahayworth:
For any passers-by: I'm going to test this internally at GitHub a bit more on Monday, before we merge; just to be safe. Unless someone has been lurking and testing it in their app all along, which would be fantastic. 😆

ahayworth:
@robbkidd and I found a way to initialize the queue-swapping more reliably in 967bbdc2544e46214d59c04a49b8e8f5ea6a0619, and I'm testing this now in production.

robbkidd (approved):
With the Rails app I was using previously, the change in 967bbdc results in no more `OpenTelemetry error: calls to detach should match corresponding calls to attach.` errors and a successful child span for `action_view render_template` being emitted within a trace.

ahayworth:
I ran this in production for a few hours, and noted nothing obviously or dramatically wrong. I think we're good to merge @open-telemetry/ruby-maintainers 👍

PR #797, opened 2021-06-07
fix: Rescue error in Faraday instrumentaion

aislinncronin:
We have added some customizations to OTel ruby in our projects and noticed errors in the faraday instrumentation causing our deployments to fail.  The errors being thrown did need to be fixed, but ideally OTel would fail gracefully and log the error to allow deployments to go ahead.
The specific error we experienced was
`#<NameError: uninitialized constant OpenTelemetry::Common>`

### DIFF
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
@@ -50,6 +50,9 @@ module OpenTelemetry
             instrumentation_attrs.merge(
               OpenTelemetry::Common::HTTP::ClientContext.attributes
             )
+          rescue StandardError => e
+            OpenTelemetry.logger.debug("[Faraday] Error creating span attrs: #{e}")
+            {}
           end
 
           def tracer


robertlaurin:
Hey @aislinncronin I think this might be the fix we want here https://github.com/open-telemetry/opentelemetry-ruby/pull/798

I'm curious this scenario was reached though, the SDK requires `common` during configuration.  https://github.com/open-telemetry/opentelemetry-ruby/blob/302ae8a95cdc865da42325e2f5a8dc270f202b7c/sdk/lib/opentelemetry/sdk.rb#L8

Are you perchance using the instrumentation libraries without the `opentelemetry-sdk` ?

aislinncronin:
@robertlaurin  we have a use-case where we need to add the faraday middleware to a connection that is instantiated before OTel is loaded/configured.  Seeing your fix in #798 however makes me think that requiring the sdk when we do this could have been the proper fix for us 🤔  thank you!

robertlaurin:
>@robertlaurin we have a use-case where we need to add the faraday middleware to a connection that is instantiated before OTel is loaded/configured. Seeing your fix in #798 however makes me think that requiring the sdk when we do this could have been the proper fix for us 🤔 thank you!

@aislinncronin this is just a plain old bug on our part 😞, there should not have been any hard dependency on the SDK being present. The instrumentation is not supposed to depend on the SDK at all.

I went ahead and released `0.18.1` for the faraday instrumentation which contains the missing require.

aislinncronin:
@robertlaurin ah got it! thank you - I'll close this PR then 👍

PR #799, opened 2021-06-08
feat: Add Tracer.non_recording_span to API

### LINKED ISSUE #793: Explicitly define a NonPropagatingSpan function
Correct me if I'm wrong: I believe you are supporting wrapping the `NonRecordingSpan` notion through the no-op API `Tracer`, i.e. in API `Tracer.start_span()` you do:

```ruby
Span.new(span_context: span_context)
```

Effectively wrapping a `SpanContext`. `Span` itself doesn't _seem_ to publicly allow this though:

```ruby
      # @api private
      def initialize(span_context: nil)
```

But we need to offer this publicly. Maybe consider exposing a `Span.wrap(SpanContext)` function?

fbogsany:
Fixes #793

There's a lot going on here. The TL;DR is that
1. we had some crufty code that was long overdue for cleanup; and
2. in order to comply with the spec, we need to make `non_recording_span` an *instance* method on `Tracer` (because `Span`s can only be created by a `Tracer`), and we had a bunch of places that abused `Span.new` (which we marked `@api private`) and didn't have a `Tracer` instance handy.


sdk/test/opentelemetry/sdk/trace/export/console_span_exporter_test.rb
```diff
@@ -22,11 +22,11 @@
     $stdout = @original_stdout
   end
 
-  it 'accepts an Array of Spans as arg to #export and succeeds' do
+  it 'accepts an Array of SpanData as arg to #export and succeeds' do
     _(exporter.export(spans)).must_equal export::SUCCESS
   end
 
-  it 'accepts an Enumerable of Spans as arg to #export and succeeds' do
+  it 'accepts an Enumerable of SpanData as arg to #export and succeeds' do
```
ahayworth, line 29: With the change at L13 (going from an array of `Span` to an array of `Object`) we're not really testing that the `console_span_exporter` accepts "an Enumerable of SpanData" (and same for the test at L25).

...although, I suppose one could make the case that we were _never_ testing for an array/enumerable of `SpanData` specifically, since we were actually sending an array of `Span`. 😆 

Do we have any desire to ensure that this exporter accepts specifically `SpanData`? If we do, we should probably update the test to actually assert that specifically (and maybe have the exporter fail otherwise). Otherwise we could rename the test to something like `it accepts an Enumerable as arg to #export and succeeds` since the exporter as-implemented really doesn't care?

FWIW part of the reason I noticed this is that we have an in-house `LogSpanExporter` for debugging that's less verbose (and I've been meaning to contribute it back, honestly). We inherit from the `ConsoleSpanExporter` and do actually assume that we're getting an array of `SpanData`, and so if that's an unsafe assumption I can/will update our exporter to double-check what it's getting first.

ahayworth (approved):
On a technical level, I think this change is 👍 

On a spec level, it's a tad confusing to me so I feel like someone a little more knowledgeable about the spec should also sign-off. 😄

sdk/test/opentelemetry/sdk/trace/export/console_span_exporter_test.rb
fbogsany: It's a safe assumption (that you're getting an `Enumerable` of `SpanData`), but `ConsoleSpanExporter` doesn't actually care what it receives. If you do contribute your `LogSpanExporter` (ideally as improvements to the `ConsoleSpanExporter`), then a more strict test would be warranted.

Honestly, I was just in the file to fix some stuff actually related to the PR, and I noticed that this was obviously wrong. I'll fix the test to do what it says.

fbogsany:
In light of https://github.com/open-telemetry/opentelemetry-ruby/issues/793#issuecomment-856776966 (TL;DR this method doesn't have to be an instance method on `Tracer`), do we prefer to:
1. Keep things the way they are in this PR (instance method on `Tracer`)
2. Make the method static on `Tracer` (leads to calls like `OpenTelemetry::Trace::Tracer.non_recording_span(...)`
3. Move the method to `Trace` (leads to calls like `OpenTelemetry::Trace.non_recording_span(...)`)

My 2c: static method on `Tracer` doesn't help much; method on `Trace` aligns with helpers like `with_span` and means we don't need to sprinkle `Tracer` instances around.

robertlaurin:
>Move the method to Trace (leads to calls like OpenTelemetry::Trace.non_recording_span(...))

Seems like the most convenient/consistent choice.

sdk/test/opentelemetry/sdk/trace/export/console_span_exporter_test.rb
fbogsany: Done

instrumentation/rack/test/opentelemetry/instrumentation/rack_test.rb
```diff
@@ -10,6 +10,7 @@
 
 describe OpenTelemetry::Instrumentation::Rack do
   let(:instrumentation) { OpenTelemetry::Instrumentation::Rack::Instrumentation.instance }
+  let(:tracer) { OpenTelemetry.tracer_provider.tracer(__FILE__) }
```
robertlaurin, line 13:
```suggestion
  let(:new_span) { OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new) }
```

instrumentation/rack/test/opentelemetry/instrumentation/rack_test.rb
```diff
@@ -53,4 +54,8 @@
       end
     end
   end
+
+  def new_span
+    OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new)
+  end
```
robertlaurin, line 60:
```suggestion
```

robertlaurin (approved):

Issue #790, opened 2021-06-05
Consider removing the no-op Baggage `Manager`


carlosalberto:
Likewise for the `Builder` counterpart. Other SIGs have no no-op classes for this, and we could add these back in the future if/as needed.

PR #800, opened 2021-06-08
fix!: refactor Baggage to remove Noop*

### LINKED ISSUE #789: Consider exposing Baggage `Manager` as functions
As `BaggageManager` is not required in other SIGs API (as there won't be specialized implementations, as opposed to traces/metrics), consider removing this class and expose the functionality directly, e.g.

```ruby
module Baggage
  def build(context:Context) # etc
```

This is also how these functions are exposed in Python, btw. Not a strong feeling though.

fbogsany:
Fixes #789 and #790


api/lib/opentelemetry/baggage.rb
```diff
@@ -7,13 +7,103 @@
...
+    extend self
+
+    BAGGAGE_KEY = OpenTelemetry::Baggage::Propagation::ContextKeys.baggage_key
+    EMPTY_BAGGAGE = {}.freeze
+    private_constant(:BAGGAGE_KEY, :EMPTY_BAGGAGE)
+
+    # Used to chain modifications to baggage. The result is a
+    # context with an updated baggage. If only a single
+    # modification is being made to baggage, use the other
+    # methods on +Manager+, if multiple modifications are being made, use
```
ahayworth, line 24:
```suggestion
    # methods on +Baggage+, if multiple modifications are being made, use
```

ahayworth (approved):
LGTM, one suspected typo but otherwise 👍

PR #805, opened 2021-06-08
fix!: total order constraint on span.status=

fbogsany:
See https://github.com/open-telemetry/opentelemetry-specification/pull/1685

## BREAKING CHANGES

- The `OpenTelemetry::Trace::Util::HttpToStatus` module has been removed. It previously contributed the `http_to_status` class method to the `Status` class. That method incorrectly set the span status to `OK` for response codes in the range `100..399`. Calls to that method should be changed as follows:
```ruby
# change:
span.status = OpenTelemetry::Trace::Status.http_to_status(response_code)
# to:
span.status = OpenTelemetry::Trace::Status.error unless (100..399).include?(response_code.to_i)
```
- The `Status.new(code, description:)` initializer has been hidden in favour of simpler constructors for each status code: `Status.ok`, `Status.error` and `Status.unset`. Each constructor takes an optional `description`.


sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -167,20 +167,23 @@ def record_exception(exception, attributes: nil)
 
         # Sets the Status to the Span
         #
-        # If used, this will override the default Span status. Default is OK.
+        # If used, this will override the default Span status. Default has code = Status::UNSET.
```
robertlaurin, line 170:
```suggestion
        # If used, this will override the default Span status. Default is UNSET.
```

fbogsany: 🤷 I was being pedantic - technically, `UNSET` is the `status.code`, not the `status`.

robertlaurin: Pedantic is permitted :)

robertlaurin (dismissed):

robertlaurin:
Excon is doing something fishy with span status here, probably needs to be updated.
https://github.com/open-telemetry/opentelemetry-ruby/blob/353c381262cb085f6cd911677ca6a3495c30ce51/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb#L82-L84

robertlaurin:
If instrumentation is not supposed to set an OK status anymore we probably need to look for uses of this too
https://github.com/open-telemetry/opentelemetry-ruby/blob/353c381262cb085f6cd911677ca6a3495c30ce51/api/lib/opentelemetry/trace/util/http_to_status.rb#L11-L24

fbogsany:
I think we need to 👀 the semantic conventions for HTTP instrumentation - I *think* there was a requirement to set the span status based on the response code.

robertlaurin:
>Span Status MUST be left unset if HTTP status code was in the 1xx, 2xx or 3xx ranges, unless there was another error (e.g., network error receiving the response body; or 3xx codes with max redirects exceeded), in which case status MUST be set to Error.

>For HTTP status codes in the 4xx and 5xx ranges, as well as any other code the client failed to interpret, status MUST be set to Error.

>Don't set the span status description if the reason can be inferred from http.status_code.


https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status

ahayworth (approved):
LGTM!

PR #807, opened 2021-06-09
fix: update context to match spec

robertlaurin:
Moves the `attach` and `detach` methods from private instance methods to public global methods
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/context/context.md

Related: https://github.com/open-telemetry/opentelemetry-ruby/pull/796/files#r647055849

### DIFF
--- a/api/lib/opentelemetry/context.rb
+++ b/api/lib/opentelemetry/context.rb
@@ -30,7 +30,7 @@ module OpenTelemetry
       #
       # @return [Context]
       def current
-        Thread.current[KEY] ||= ROOT
+        stack.last || ROOT
       end
 
       # @api private
@@ -46,10 +46,9 @@ module OpenTelemetry
       # @param [Context] context The new context
       # @return [Object] A token to be used when detaching
       def attach(context)
-        prev = current
-        self.current = context
-        stack.push(prev)
-        stack.size
+        s = stack
+        s.push(context)
+        s.size
       end
 
       # Restores the previous Context associated with the current Fiber.
@@ -60,10 +59,11 @@ module OpenTelemetry
       # @param [Object] token The token provided by the matching call to attach
       # @return [Boolean] True if the calls matched, false otherwise
       def detach(token)
-        calls_matched = (token == stack.size)
+        s = stack
+        calls_matched = (token == s.size)
         OpenTelemetry.handle_error(exception: DetachError.new('calls to detach should match corresponding calls to attach.')) unless calls_matched
 
-        self.current = stack.pop
+        s.pop
         calls_matched
       end
 
--- a/api/lib/opentelemetry/context.rb
+++ b/api/lib/opentelemetry/context.rb
@@ -10,10 +10,9 @@ require 'opentelemetry/context/propagation'
 module OpenTelemetry
   # Manages context on a per-fiber basis
   class Context
-    KEY = :__opentelemetry_context__
     EMPTY_ENTRIES = {}.freeze
     STACK_KEY = :__opentelemetry_context_storage__
-    private_constant :KEY, :EMPTY_ENTRIES, :STACK_KEY
+    private_constant :EMPTY_ENTRIES, :STACK_KEY
 
     DetachError = Class.new(OpenTelemetry::Error)
 
@@ -33,10 +32,6 @@ module OpenTelemetry
         stack.last || ROOT
       end
 
-      # @api private
-      def current=(ctx)
-        Thread.current[KEY] = ctx
-      end
 
       # Associates a Context with the caller's current Fiber. Every call to
       # this operation should be paired with a corresponding call to detach.
--- a/api/lib/opentelemetry/context.rb
+++ b/api/lib/opentelemetry/context.rb
@@ -29,11 +29,32 @@ module OpenTelemetry
         Thread.current[KEY] ||= ROOT
       end
 
-      # Sets the current context
-      #
-      # @param [Context] ctx The context to be made active
+      # @api private
       def current=(ctx)
         Thread.current[KEY] = ctx
+      end
+
+      # Returns the previous context so that it can be restored
+      #
+      # @param [Context] context The new context
+      # @return [Context] prev The previous context
+      def attach(context)
+        return current.parent if context == current
+
+        prev = current
+        self.current = context
+        context.parent = prev
+      end
+
+      # Restores the current context to the context supplied or the parent context
+      # if no context is provided
+      #
+      # @param [Context] previous_context The previous context to restore
+      def detach(previous_context = nil)
+        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if current.parent != previous_context
+
+        previous_context ||= current.parent || ROOT
+        self.current = previous_context
       end
 
       # Executes a block with ctx as the current context. It restores
@@ -42,10 +63,10 @@ module OpenTelemetry
       # @param [Context] ctx The context to be made active
       # @yield [context] Yields context to the block
       def with_current(ctx)
-        prev = ctx.attach
+        prev = attach(ctx)
         yield ctx
       ensure
-        ctx.detach(prev)
+        detach(prev)
       end
 
       # Execute a block in a new context with key set to value. Restores the
@@ -58,10 +79,10 @@ module OpenTelemetry
       #   the block
       def with_value(key, value)
         ctx = current.set_value(key, value)
-        prev = ctx.attach
+        prev = attach(ctx)
         yield ctx, value
       ensure
-        ctx.detach(prev)
+        detach(prev)
       end
 
       # Execute a block in a new context where its values are merged with the
@@ -75,17 +96,10 @@ module OpenTelemetry
       #   to the block
       def with_values(values)
         ctx = current.set_values(values)
-        prev = ctx.attach
+        prev = attach(ctx)
         yield ctx, values
       ensure
-        ctx.detach(prev)
-      end
-
-      # Returns the value associated with key in the current context
-      #
-      # @param [String] key The lookup key
-      def value(key)
-        current.value(key)
+        detach(prev)
       end
 
       def clear
@@ -93,13 +107,15 @@ module OpenTelemetry
       end
 
       def empty
-        new(nil, EMPTY_ENTRIES)
+        new(EMPTY_ENTRIES)
       end
     end
 
-    def initialize(parent, entries)
-      @parent = parent
+    attr_accessor :parent
+
+    def initialize(entries)
       @entries = entries.freeze
+      @parent = parent
     end
 
     # Returns the corresponding value (or nil) for key
@@ -120,7 +136,7 @@ module OpenTelemetry
     def set_value(key, value)
       new_entries = @entries.dup
       new_entries[key] = value
-      Context.new(self, new_entries)
+      Context.new(new_entries)
     end
 
     # Returns a new Context with the current context's entries merged with the
@@ -131,22 +147,7 @@ module OpenTelemetry
     # @param [Object] value Object to be stored under key
     # @return [Context]
     def set_values(values) # rubocop:disable Naming/AccessorMethodName:
-      Context.new(self, @entries.merge(values))
-    end
-
-    # @api private
-    def attach
-      prev = self.class.current
-      self.class.current = self
-      prev
-    end
-
-    # @api private
-    def detach(ctx_to_attach = nil)
-      OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if self.class.current != self
-
-      ctx_to_attach ||= @parent || ROOT
-      ctx_to_attach.attach
+      Context.new(@entries.merge(values))
     end
 
     ROOT = empty.freeze
--- a/api/lib/opentelemetry/context.rb
+++ b/api/lib/opentelemetry/context.rb
@@ -35,6 +35,9 @@ module OpenTelemetry
         Thread.current[KEY] = ctx
       end
 
+      # Associates a Context with the caller's current Fiber. Every call to
+      # this operation should be paired with a corresponding call to detach.
+      #
       # Returns a token to be used with the matching call to detach
       #
       # @param [Context] context The new context
@@ -46,17 +49,19 @@ module OpenTelemetry
         stack.size
       end
 
-      # Restores the previous context, if a token is supplied it will
-      # be used to check if the call to detach is balanced with
-      # the original attach call. A warning is logged if the
+      # Restores the previous Context associated with the current Fiber.
+      # The supplied token is used to check if the call to detach is balanced
+      # with a corresponding attach call. A warning is logged if the
       # calls are unbalanced.
       #
       # @param [Integer] token The token provided by the matching call to attach
-      def detach(token = nil)
-        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if token && token != stack.size
+      # @return [Boolean] True if the calls matched, false otherwise
+      def detach(token)
+        calls_matched = (token == stack.size)
+        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' unless calls_matched
 
-        previous_context = stack.pop || ROOT
-        self.current = previous_context
+        self.current = stack.pop || ROOT
+        calls_matched
       end
 
       # Executes a block with ctx as the current context. It restores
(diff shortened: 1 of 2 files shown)


api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +29,44 @@ def current
...
-      # @param [Context] ctx The context to be made active
+      # @api private
       def current=(ctx)
         Thread.current[KEY] = ctx
       end
 
+      # Returns the previous context so that it can be restored
+      #
+      # @param [Context] context The new context
+      # @return [Context] prev The previous context
```
fbogsany, line 40: The spec refers to this as a "Token" rather than necessarily being the previous Context. Is there value in giving ourselves that flexibility or do we not need it?
```suggestion
      # @return [Context] The previous context
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +29,44 @@ def current
...
+      def attach(context)
+        return current.parent if context == current
```
fbogsany, line 42: This is an interesting corner case 🤔 - it is weirdly open to abuse, in that you can use it to retrieve the parent of the current context. Should we document this?

Actually, this makes me mildly concerned about our current implementation with the stack of contexts represented as an internally linked list. With other implementations (e.g. I think Java maintains the context stack as an array), it is possible for a context to be pushed (attached) twice. Our implementation won't allow that, which could result in some legitimate code patterns 💥 .

api/lib/opentelemetry/context.rb
```diff
@@ -75,31 +96,26 @@ def with_value(key, value)
...
       end
     end
 
-    def initialize(parent, entries)
-      @parent = parent
+    attr_accessor :parent
+
+    def initialize(entries)
       @entries = entries.freeze
+      @parent = parent
```
fbogsany, line 118: 🤔 I am very confused - won't `parent` always be `nil` here?

api/lib/opentelemetry/context.rb
```diff
@@ -75,31 +96,26 @@ def with_value(key, value)
...
+        prev = attach(ctx)
         yield ctx, values
       ensure
-        ctx.detach(prev)
-      end
-
-      # Returns the value associated with key in the current context
-      #
-      # @param [String] key The lookup key
-      def value(key)
```
fbogsany, line 87: Why lose this? This is just a convenience, right? I.e. instead of:
```ruby
OpenTelemetry::Context.current.value('foo')
```
we can
```ruby
OpenTelemetry::Context.value('foo')
```

api/lib/opentelemetry/context.rb
```diff
@@ -75,31 +96,26 @@ def with_value(key, value)
...
 
       def empty
-        new(nil, EMPTY_ENTRIES)
+        new(EMPTY_ENTRIES)
       end
     end
 
-    def initialize(parent, entries)
-      @parent = parent
+    attr_accessor :parent
```
fbogsany, line 114: This should be `protected` or otherwise clearly not part of the public API.

api/lib/opentelemetry/context.rb
robertlaurin: Unintentional, I'll restore it and add some test.

api/lib/opentelemetry/context.rb
robertlaurin: This corner case was to prevent repeatedly attaching the same context and (due to the way I implemented this) causing a circular dependency that would prevent users from properly detaching the context.

I agree that this is concerning, I was admittedly a little too focused on keeping parity with the previous implementation but with making attach/detach public it certainly changes considerations.

I suppose we want to support attaching the same context multiple times?  Otherwise it's too easy to have your attach/detach calls fall out of sync.

api/lib/opentelemetry/context.rb
robertlaurin: I'll admit I find the returning of a Token a bit unclear in the context of the detach documentation.

```
This operation is intended to help making sure the correct Context is associated with the caller's current execution unit. Users can rely on it to identify a wrong call order, i.e. trying to detach a Context that is not the current instance. In this case the operation can emit a signal to warn users of the wrong call order, such as logging an error or returning an error value.

The API MUST accept the following parameters:

A Token that was returned by a previous call to attach a Context.
```

What purpose does this token serve if we want to keep attaching and detaching balanced.  Is there a situation where you want to detach and provide a token that doesn't match the current context?

api/lib/opentelemetry/context.rb
robertlaurin: Much more likely a bug though.  I expect this to change based on the comments above.

api/lib/opentelemetry/context.rb
ahayworth: > I suppose we want to support attaching the same context multiple times? Otherwise it's too easy to have your attach/detach calls fall out of sync.

Unless the spec forbids it, I think that'd be a nice thing to support.

api/lib/opentelemetry/context.rb
fbogsany: > Is there a situation where you want to detach and provide a token that doesn't match the current context?

It helps double-check that `attach` and `detach` are balanced. These functions are meant to be low-level building blocks for higher level block-structured mechanisms. Passing an opaque Token around provides implementation flexibility while also being able to check for correct use of the functions, and report meaningful errors (oh yeah, we should do that 😄 ).

fbogsany: So, why a Token instead of the Context? If we were to use an array as the Context stack instead of chaining contexts together internally, we'd likely use the array index as the Token instead of the Context itself.

robertlaurin: Ok so that makes it a lot clearer, I was under the impression that the token was to be used to actually retrieve a specific context instead of a means to assert that the attach/detach calls were lined up.  

I think with the fact that should allow for attaching the same context multiple times it makes sense to use an array to keep track of the context stack, and to provide the array index as token to check against.

robertlaurin: Changes based on discussion here 0c57bd649ff7b86319151dac48abbe332ec4a64e

api/lib/opentelemetry/context.rb
robertlaurin: Removed.

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
 
-      # Sets the current context
-      #
-      # @param [Context] ctx The context to be made active
+      # @api private
       def current=(ctx)
         Thread.current[KEY] = ctx
       end
 
+      # Returns a token to be used with the matching call to detach
```
fbogsany, line 38:
```suggestion
      # Associates a Context with the caller's current Fiber. Every call to
      # this operation should be paired with a corresponding call to detach.
      #
      # Returns a token to be used with the matching call to detach
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+        prev = current
+        self.current = context
+        stack.push(prev)
+        stack.size
+      end
+
+      # Restores the previous context, if a token is supplied it will
+      # be used to check if the call to detach is balanced with
+      # the original attach call. A warning is logged if the
+      # calls are unbalanced.
```
fbogsany, line 52:
```suggestion
      # Restores the previous Context associated with the current Fiber.
      # The supplied token is used to check if the call to detach is balanced
      # with a corresponding attach call. A warning is logged if the
      # calls are unbalanced.
```
The `token` arg should not be optional.

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+      #
+      # @param [Integer] token The token provided by the matching call to attach
+      def detach(token = nil)
+        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if token && token != stack.size
```
fbogsany, line 56:
```suggestion
      def detach(token)
        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if token != stack.size
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+
+        previous_context = stack.pop || ROOT
+        self.current = previous_context
+      end
```
fbogsany, line 60:
```suggestion
        self.current = stack.pop || ROOT
        calls_matched
      end
```

api/lib/opentelemetry/context.rb
fbogsany: Also, do you think there's value in logging the stack at this point, or at least the caller?

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+        stack.push(prev)
+        stack.size
+      end
+
+      # Restores the previous context, if a token is supplied it will
+      # be used to check if the call to detach is balanced with
+      # the original attach call. A warning is logged if the
+      # calls are unbalanced.
+      #
+      # @param [Integer] token The token provided by the matching call to attach
```
fbogsany, line 54:
```suggestion
      # @param [Integer] token The token provided by the matching call to attach
      # @return [Boolean] True if the calls matched, false otherwise
```

fbogsany: I think it's useful to return this indication, even if it isn't used. The spec suggests:
> The API MAY return a value used to check whether the operation was successful or not.

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+      def detach(token = nil)
+        OpenTelemetry.logger.warn 'Calls to detach should match corresponding calls to attach' if token && token != stack.size
+
+        previous_context = stack.pop || ROOT
```
fbogsany, line 58: I think the `|| ROOT` implies the calls were mismatched - we should probably incorporate that into the `calls_matched` check (if the stack is empty, the calls must not have been matched).

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
+      end
+
       # Executes a block with ctx as the current context. It restores
       # the previous context upon exiting.
       #
       # @param [Context] ctx The context to be made active
       # @yield [context] Yields context to the block
       def with_current(ctx)
-        prev = ctx.attach
+        prev = attach(ctx)
```
fbogsany, line 68:
```suggestion
        token = attach(ctx)
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +30,45 @@ def current
...
         yield ctx
       ensure
-        ctx.detach(prev)
+        detach(prev)
```
fbogsany, line 71:
```suggestion
        detach(token)
```

api/lib/opentelemetry/context.rb
```diff
@@ -58,10 +81,10 @@ def with_current(ctx)
       #   the block
       def with_value(key, value)
         ctx = current.set_value(key, value)
-        prev = ctx.attach
+        prev = attach(ctx)
```
fbogsany, line 84:
```suggestion
        token = attach(ctx)
```

api/lib/opentelemetry/context.rb
```diff
@@ -58,10 +81,10 @@ def with_current(ctx)
...
         yield ctx, value
       ensure
-        ctx.detach(prev)
+        detach(prev)
```
fbogsany, line 87:
```suggestion
        detach(token)
```

api/lib/opentelemetry/context.rb
```diff
@@ -75,10 +98,10 @@ def with_value(key, value)
       #   to the block
       def with_values(values)
         ctx = current.set_values(values)
-        prev = ctx.attach
+        prev = attach(ctx)
```
fbogsany, line 101:
```suggestion
        token = attach(ctx)
```

api/lib/opentelemetry/context.rb
```diff
@@ -75,10 +98,10 @@ def with_value(key, value)
...
         yield ctx, values
       ensure
-        ctx.detach(prev)
+        detach(prev)
```
fbogsany, line 104:
```suggestion
        detach(token)
```

api/lib/opentelemetry/context.rb
```diff
@@ -89,16 +112,22 @@ def value(key)
...
 
       def empty
-        new(nil, EMPTY_ENTRIES)
+        new(EMPTY_ENTRIES)
+      end
+
+      private
+
+      def stack
+        @stack ||= []
```
fbogsany, line 126: The `stack` needs to be thread-local (well, Fiber-local).

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +33,50 @@ def current
...
         Thread.current[KEY] = ctx
       end
 
+      # Associates a Context with the caller's current Fiber. Every call to
+      # this operation should be paired with a corresponding call to detach.
+      #
+      # Returns a token to be used with the matching call to detach
+      #
+      # @param [Context] context The new context
+      # @return [Integer] A token to be used when detaching
```
fbogsany, line 47: I think we should be vague here.
```suggestion
      # @return [Object] A token to be used when detaching
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +33,50 @@ def current
...
+        stack.push(prev)
+        stack.size
+      end
+
+      # Restores the previous Context associated with the current Fiber.
+      # The supplied token is used to check if the call to detach is balanced
+      # with a corresponding attach call. A warning is logged if the
+      # calls are unbalanced.
+      #
+      # @param [Integer] token The token provided by the matching call to attach
```
fbogsany, line 60:
```suggestion
      # @param [Object] token The token provided by the matching call to attach
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +33,50 @@ def current
...
+      #
+      # Returns a token to be used with the matching call to detach
+      #
+      # @param [Context] context The new context
+      # @return [Object] A token to be used when detaching
+      def attach(context)
+        prev = current
+        self.current = context
+        stack.push(prev)
+        stack.size
```
fbogsany, line 52: There's an efficiency concern here because this requires 4 thread local lookups (reading `current`, writing `current`, and 2 `stack` accesses). It would be more efficient to only store the `stack` as a thread-local and have `current` implicitly stored at the top of the stack. (`detach` has a similar issue below)
```suggestion
        s = stack
        s.push(context)
        s.size
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +33,50 @@ def current
...
+      # calls are unbalanced.
+      #
+      # @param [Object] token The token provided by the matching call to attach
+      # @return [Boolean] True if the calls matched, false otherwise
+      def detach(token)
+        calls_matched = (token == stack.size)
+        OpenTelemetry.handle_error(exception: DetachError.new('calls to detach should match corresponding calls to attach.')) unless calls_matched
+
+        self.current = stack.pop
+        calls_matched
```
fbogsany, line 67:
```suggestion
        s = stack
        calls_matched = (token == s.size)
        OpenTelemetry.handle_error(exception: DetachError.new('calls to detach should match corresponding calls to attach.')) unless calls_matched

        s.pop
        calls_matched
```

api/lib/opentelemetry/context.rb
```diff
@@ -29,23 +33,50 @@ def current
         Thread.current[KEY] ||= ROOT
```
fbogsany, line 33:
```suggestion
        stack.last || ROOT
```

api/lib/opentelemetry/context.rb
```diff
@@ -26,26 +30,53 @@ def create_key(name)
...
+        stack.last || ROOT
       end
 
-      # Sets the current context
-      #
-      # @param [Context] ctx The context to be made active
+      # @api private
       def current=(ctx)
         Thread.current[KEY] = ctx
       end
```
fbogsany, line 39: I think this is not needed anymore?
```suggestion
```

api/lib/opentelemetry/context.rb
```diff
@@ -12,6 +12,10 @@ module OpenTelemetry
   class Context
     KEY = :__opentelemetry_context__
     EMPTY_ENTRIES = {}.freeze
+    STACK_KEY = :__opentelemetry_context_storage__
+    private_constant :KEY, :EMPTY_ENTRIES, :STACK_KEY
```
fbogsany, line 16: I don't think `KEY` is used anymore?
```suggestion
    EMPTY_ENTRIES = {}.freeze
    STACK_KEY = :__opentelemetry_context_storage__
    private_constant :EMPTY_ENTRIES, :STACK_KEY
```

fbogsany (approved):

PR #808, opened 2021-06-09
Use 55681 as the default OLTP exporter port

ahayworth:
In the interest of making this library easier to use by default, this PR
proposes a simple change: allow our OTLP exporter to default to using
port 55681, rather than 4317. Port 4317 is the collector's GRPC port,
and our exporter doesn't support GRPC. The default HTTP port is 55681,
and so we use it throughout.

However, this does seem to be against the specification in some ways.
The specification doesn't *require* that you support GRPC, however it
defines the default value for the environment variable as if you do.

I think this is acceptable; and we can default to port 4317 and GRPC
once we have support for it and are comfortable that it performs well.

https://github.com/open-telemetry/opentelemetry-collector/blob/main/receiver/otlpreceiver/README.md#getting-started

### DIFF
--- a/exporter/otlp/README.md
+++ b/exporter/otlp/README.md
@@ -2,6 +2,8 @@
 
 The `opentelemetry-exporter-otlp` gem provides an [OTLP](https://github.com/open-telemetry/opentelemetry-proto) exporter for OpenTelemetry for Ruby. Using `opentelemetry-exporter-otlp`, an application can configure OpenTelemetry to export collected tracing data to [the OpenTelemetry Collector][opentelemetry-collector-home].
 
+This OTLP exporter only supports OTLP/HTTP at this time.
+
 ## What is OpenTelemetry?
 
 [OpenTelemetry][opentelemetry-home] is an open source observability framework, providing a general-purpose API, SDK, and related tools required for the instrumentation of cloud-native software, frameworks, and libraries.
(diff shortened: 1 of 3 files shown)


fbogsany:
Unfortunately, there's a conflict between the spec and the collector because the [collector has had long-standing challenges implementing a mux that works for both gRPC and HTTP endpoints](https://github.com/open-telemetry/opentelemetry-collector/issues/1256).

[The spec states](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-default-port):
> The default network port for OTLP/HTTP is 4317.

The [OTLP Exporter configuration options](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options) also makes it clear that the default port for OTLP/HTTP is 4317.

ahayworth:
That'd be lovely if they could get back to a single port. I wish them luck.

It does stink that our default configuration doesn't work, however. Any other thoughts on ways to resolve that?

(Other than implementing GRPC, I suppose, which we still could/should do)

ahayworth:
(Closing this out in light of comments from @fbogsany but I still think it's valuable to discuss a better "out of box" experience given the conflicts between the standards and the collector, and I'm fine to do that here still or wherever)

fbogsany:
The environment variable is the simplest way to handle this - i.e. just document this case prominently. An alternative is to provide a "distribution" that defaults to the thing that works (AWS have done this - I'm not sure if they've done it for Ruby). The thing we talked about to unify common elements of Shopify's and GitHub's configuration gems as a public convenience gem could be that.

ahayworth:
Both of those are reasonable approaches, thanks. 😄

PR #822, opened 2021-06-11
refactor!: TraceConfig -> SpanLimits

### LINKED ISSUE #813: Define TraceConfig values as SpanLimits
As per the [Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#span-limits) these values should be under a `SpanLimits` object. It's definitely not a biggie, but it would keep uniformity with other SIGs.

### LINKED ISSUE #815: Consider hiding `TracerProvider.stopped`
We don't offer this in other SIGs, and unless it's really required by some components, we could hide it (at least for now).

### LINKED ISSUE #816: Consider hiding `TracerProvider.active_span_processor`
Similarly to `TracerProvider.stopped`, we don't offer/expose this and I think it's rather dangerous to allow users to consume it.

### LINKED ISSUE #817: Consider moving `Sampler` out of `TraceConfig`
Specially if `TraceConfig` is renamed to `SpanLimits`.  This is also how other SIGs are doing it, by putting it into the SDK `TracerProvider` directly.

fbogsany:
Fixes #817
Fixes #813
Fixes #816
Fixes #815

This is a significant refactor of the tracing configuration mechanism in the SDK. The most significant pieces involve renaming `Config::TraceConfig` to `SpanLimits` and moving the `sampler` configuration out of the old `TraceConfig` and into `TracerProvider`. It also moves span creation responsibility to `TracerProvider` and hides the `active_span_processor` and `stopped` fields. This matches the SDKs for other languages, and updates to the spec that I missed earlier. Needless to say, these are breaking changes.

## BREAKING CHANGES

- `OpenTelemetry::SDK::Trace::Config::TraceConfig` is renamed to `OpenTelemetry::SDK::Trace::SpanLimits`
- The `sampler` configuration is moved from `TraceConfig` to `TracerProvider`
- The configuration attributes in `SpanLimits` are renamed to align with the environment variable names
- The `active_span_processor` and `stopped`/`stopped?` attributes are now private in `TracerProvider`


sdk/lib/opentelemetry/sdk/trace/span_limits.rb
```diff
@@ -0,0 +1,60 @@
...
+                       event_count_limit: Integer(ENV.fetch('OTEL_SPAN_EVENT_COUNT_LIMIT', 128)),
+                       link_count_limit: Integer(ENV.fetch('OTEL_SPAN_LINK_COUNT_LIMIT', 128)),
+                       attribute_per_event_count_limit: attribute_count_limit,
+                       attribute_per_link_count_limit: attribute_count_limit)
+          raise ArgumentError, 'attribute_count_limit must be positive' unless attribute_count_limit.positive?
+          raise ArgumentError, 'attribute_length_limit must not be less than 32' unless attribute_length_limit.nil? || Integer(attribute_length_limit) >= 32
+          raise ArgumentError, 'event_count_limit must be positive' unless event_count_limit.positive?
+          raise ArgumentError, 'link_count_limit must be positive' unless link_count_limit.positive?
+          raise ArgumentError, 'attribute_per_event_count_limit must be positive' unless attribute_per_event_count_limit.positive?
+          raise ArgumentError, 'attribute_per_link_count_limit must be positive' unless attribute_per_link_count_limit.positive?
```
ahayworth, line 45: You know, I didn't actually know about the `positive?` predicate. Learn something new every day! 😄

sdk/test/opentelemetry/sdk/configurator_test.rb
```diff
@@ -167,7 +167,7 @@
       it 'defaults to NoopSpanProcessor if no valid exporter is available' do
         configurator.configure
 
-        _(OpenTelemetry.tracer_provider.active_span_processor).must_be_instance_of(
+        _(OpenTelemetry.tracer_provider.instance_variable_get(:@active_span_processor)).must_be_instance_of(
```
ahayworth, line 170: This is the thing that I was not looking forward to with this change. Debugging / testing gets harder here, but I suppose it's still a good idea.

ahayworth (approved):
The move to creating spans via the `TracerProvider` feels odd to me, semantically. Is it more important because other SDKs do it that way (not wrong to match, necessarily) or because it helps us with the requirements that configuration changes on the `TracerProvider` need to automatically be picked up by the `Tracer`s it creates? (...or, something else altogether?)

Not saying we shouldn't do it at all, I trust others here. Just trying to understand what the rationale is so I can get behind it. 😄 

Overall though, this change seems sound, and I'm excited about moving closer to 1.0! 👍

sdk/test/opentelemetry/sdk/configurator_test.rb
fbogsany: Agreed. In general, though, nothing outside of the SDK should know or care about the specific TracerProvider that's installed. I don't have the energy to do it now, but I think we could clean up these tests to stub the appropriate class's `on_span_start` and start a span instead.

fbogsany:
> The move to creating spans via the `TracerProvider` feels odd to me, semantically. Is it more important because other SDKs do it that way (not wrong to match, necessarily) or because it helps us with the requirements that configuration changes on the `TracerProvider` need to automatically be picked up by the `Tracer`s it creates? (...or, something else altogether?)

This part of the change is because the `Tracer` was mostly grabbing state from the `TracerProvider`, that should have been internal, before creating the `Span`. The `Tracer` really only contributes the `InstrumentationLibrary`, which was the sole reason for introducing the `TracerProvider` abstraction early on in the project. So we're really moving the responsibility to the right place, which allows us to hide internal state.

sdk/test/opentelemetry/sdk/configurator_test.rb
ahayworth: >  I don't have the energy to do it now

No need, either. That sounds like good rainy-day refactoring for someone who feels up to it, definitely not necessary at all! 😄

ahayworth:
> The Tracer really only contributes the InstrumentationLibrary, which was the sole reason for introducing the TracerProvider abstraction early on in the project.

@fbogsany That makes sense - I really appreciate the context that you give me on these threads. 😄

PR #826, opened 2021-06-14
refactor!: remove MultiSpan*

### LINKED ISSUE #825: Remove, move or hide MultiSpanExporter and MultiSpanProcessor
`MultiSpanProcessor` and `MultiSpanExporter` are not required by the specification. `MultiSpanExporter` is not referenced within the SDK, and has dubious value (other than perhaps debugging in production by tee-ing the spans to the console). `MultiSpanProcessor` is an implementation detail in `TracerProvider` that tries to remove a level of indirection for the common case of a single span processing and export pipeline (at the cost of an extra level of indirection for multiple pipelines).

Should we remove `MultiSpanExporter`, or move it to a convenience gem?

Should we remove `MultiSpanProcessor` or hide it as an implementation detail in `TracerProvider`? It is highly unlikely that the extra level of indirection (`TracerProvider -> array -> SpanProcessor` vs `TracerProvider -> SpanProcessor`) is measurable compared with the encoding overhead, for example. Removing `MultiSpanProcessor` will likely simplify the implementation of `TracerProvider` and reduce the surface area of the SDK.

fbogsany:
Fixes #825

For now, I've chosen to remove `MultiSpanExporter`. If we think it's a good candidate for a convenience gem, we can grab it back out of `git` history.


sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
```diff
@@ -64,14 +63,24 @@ def tracer(name = nil, version = nil)
...
+            start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
+            results = @span_processors.map do |processor|
+              remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+              break [Export::TIMEOUT] if remaining_timeout&.zero?
+
+              processor.shutdown(timeout: remaining_timeout)
             end
-            @active_span_processor.shutdown(timeout: timeout)
             @stopped = true
+            results.uniq.max || Export::SUCCESS
```
ahayworth, line 83: Do you need to `uniq` before `max`?

ahayworth (approved):
Seems alright to me!

robertlaurin (approved):

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
fbogsany: No. Good catch, thanks! This was C&P from a couple other places, so I'll check & fix them too.

fbogsany: Oh, well, I deleted the places that this was C&P from 🤷

PR #828, opened 2021-06-17
refactor: Remove redundant methods in instrumentation/rack

ahayworth:
While I do not know the precise provenance of these methods (nor do I
care to futz around with `git` any further to find out), the current API
contains similar methods that we can rely upon for this functionality.

Specifically, we remove:

- `OpenTelemetry::Instrumentation::Rack.context_with_span`
- `OpenTelemetry::Instrumentation::Rack.with_span`

`with_span` was being used, but could be replaced by the API method.
`current_span` is not being used _here_ but is being used by the `Rails`
instrumentation, so it stays.

### DIFF
--- a/instrumentation/rack/gemfiles/rack_2.0.gemfile
+++ b/instrumentation/rack/gemfiles/rack_2.0.gemfile
@@ -3,10 +3,14 @@
 source "https://rubygems.org"
 
 gem "opentelemetry-api", path: "../../../api"
+gem "opentelemetry-instrumentation-base", path: "../../base"
 gem "rack", "2.0.8"
 
 group :test do
+  gem "byebug"
+  gem "opentelemetry-common", path: "../../../common"
   gem "opentelemetry-sdk", path: "../../../sdk"
+  gem "pry-byebug"
 end
 
 gemspec path: "../"
--- a/instrumentation/rack/gemfiles/rack_2.1.gemfile
+++ b/instrumentation/rack/gemfiles/rack_2.1.gemfile
@@ -3,10 +3,14 @@
 source "https://rubygems.org"
 
 gem "opentelemetry-api", path: "../../../api"
+gem "opentelemetry-instrumentation-base", path: "../../base"
 gem "rack", "~> 2.1.2"
 
 group :test do
+  gem "byebug"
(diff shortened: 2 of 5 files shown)


arielvalentin (approved):

ahayworth:
I knew that was too easy; the rails tests are broken now. I'll fix.

ahayworth:
Ah, dang. Okay - so the `rails` instrumentation will modify the name of the top-level `rack` span if it can, and that would account for why the duplication is still needed. Creating our own `Context.key` in the `rack` instrumentation allows us to fairly easily get at that specific span.

That said - I'd really like to remove the duplication still. It feels like the kind of thing where bugs will creep in if we're not careful, and it feels like it's violating the way instrumentation is supposed to only rely on publicly exposed bits of the API.

Do we have any strong feelings on altering our approach here? I feel like one way would be storing a reference to the context object explicitly in the rack env, and then using that in the rails instrumentation to modify the span name.

fbogsany:
We chose the current approach because it is somewhat similar to what we do with HTTP and Redis instrumentation - using Context to provide additional information. I don't think this is a violation of the way instrumentation is supposed to work - Rails is pretty dependent on Rack, and can provide more details than Rack can discern, so it makes sense to me for Rails to be able to pickup the Rack span from the Context in this way and augment it.

ahayworth:
> it makes sense to me for Rails to be able to pickup the Rack span from the Context in this way and augment it.

I agree with that - I think what I was concerned about was duplicating methods from the API into the `rack` instrumentation to achieve it. I wonder if there's another way we could obtain the "rack span" from within the thread-local context? If the context key wasn't opaque, it would seem straightforward...

We could expose the `CURRENT_SPAN_KEY` constant from the Rack instrumentation, and perhaps that would allow us a more straightforward way to accomplish it?

(Also: this is definitely not some huge, pressing issue - I just saw an opportunity to refactor a bit and wanted to see what others thought!)

ahayworth:
After today's discussion, I'm just going to close this. I started to refactor the PR to only remove the "redundant" methods that weren't necessary for the `rails` instrumentation and started to end up with a PR that did very little, if anything. 😆

PR #829, opened 2021-06-17
[WIP] Ignore links with invalid span context

### LINKED ISSUE #735: Ignore Links with invalid span context
See https://github.com/open-telemetry/opentelemetry-specification/pull/1492.

In the following code we should check that the links each have a valid span context, and remove any with an invalid span context. https://github.com/open-telemetry/opentelemetry-ruby/blob/f6d156353a838f47daf167156c9b861b7eac2c30/sdk/lib/opentelemetry/sdk/trace/span.rb#L327-L344

suryart:
Check links with only the valid span context 

Fixes #735

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -331,7 +331,7 @@ module OpenTelemetry
           # Fast path (likely) common cases.
           return nil if links.nil?
 
-          links = links.map{ |link| link.span_context.valid? }
+          links = links.map { |link| link.span_context.valid? }
 
           if links.size <= link_count_limit &&
              links.all? { |link| link.attributes.size <= attribute_per_link_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -331,6 +331,8 @@ def trim_links(links, link_count_limit, attribute_per_link_count_limit) # ruboco
           # Fast path (likely) common cases.
           return nil if links.nil?
 
+          links = links.map { |link| link.span_context.valid? }
```
fbogsany, line 334: This turns all the links into booleans, which isn't what we want.
```suggestion
          links = links.select { |link| link.span_context.valid? }
```
Is correct, but inefficient. We have a fast path below that should also check for validity.
```suggestion
          links = links.map { |link| link.span_context.valid? }
```

suryart: Did you mean this?

    link.attributes.size <= attribute_per_link_count_limit && Internal.valid_attributes?(name, 'link', link.attributes)

Issue #814, opened 2021-06-10
Log an error upon empty string passed to `TracerProvider.tracer()`


carlosalberto:
Not technically part of 1.0, but useful nevertheless, i.e. it will signal users that an unexpected value was specified.

PR #830, opened 2021-06-21
feat: add unnamed tracer warning message

robertlaurin:
Fixes https://github.com/open-telemetry/opentelemetry-ruby/issues/814

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
@@ -8,7 +8,7 @@ module OpenTelemetry
   module SDK
     module Trace
       # {TracerProvider} is the SDK implementation of {OpenTelemetry::Trace::TracerProvider}.
-      class TracerProvider < OpenTelemetry::Trace::TracerProvider
+      class TracerProvider < OpenTelemetry::Trace::TracerProvider # rubocop:disable Metrics/ClassLength
         Key = Struct.new(:name, :version)
         private_constant(:Key)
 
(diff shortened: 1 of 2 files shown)


ericmustin (approved):

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
```diff
@@ -50,6 +50,7 @@ def initialize(sampler: sampler_from_environment(Samplers.parent_based(root: Sam
         def tracer(name = nil, version = nil)
           name ||= ''
           version ||= ''
+          OpenTelemetry.logger.warn 'calling TracerProvider#tracer without providing a tracer name.' if name.empty?
```
arielvalentin, line 53: @robertlaurin If I am reading this correctly, the message should be generated as part of calls made to the `API` and not the `SDK`. 

Is there a way this check could be inherited from the`API` `TracingProvider` or if we should duplicate this check and message there?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#get-a-tracer

fbogsany: > In case an invalid name (null or empty string) is specified, a working Tracer implementation MUST be returned as a fallback rather than returning null or throwing an exception, its name property SHOULD be set to an empty string, and a message reporting that the specified value is invalid SHOULD be logged. 

Emphasis here on "a working Tracer implementation". The API Tracer is considered a (mostly) no-op implementation rather than a working one, so only doing this in the SDK is correct.

fbogsany (approved):

PR #831, opened 2021-06-22
feat: allow disabling of install messages

robertlaurin:
Allows controlling the severity of the opentelemetry log messages emitted independently of the logger supplied.

This still allows for a user to override the `OpenTelemetry.logger` entirely via the accessor `OpenTelemetry.logger =`, if you wish to supply a logger and still respect the `OTEL_LOG_LEVEL` supplied environment variable this is done through the configuration.

```rb
OpenTelemetry::SDK.configure { |c| c.logger = Logger.new(....) }
```

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/registry.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/registry.rb
@@ -75,11 +75,10 @@ module OpenTelemetry
       end
 
       def install_instrumentation(instrumentation, config)
-        log_install_messages = ENV['OTEL_RUBY_INSTRUMENTATION_INSTALL_MESSAGES'] != 'false'
         if instrumentation.install(config)
-          OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was successfully installed" if log_install_messages
+          OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was successfully installed"
         else
-          OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install" if log_install_messages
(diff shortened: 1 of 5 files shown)


ahayworth:
`OTEL_RUBY_INSTRUMENTATION_INSTALL_MESSAGES`

Is there a general log-level argument we could use instead? `OTEL_DEBUG=1` or `OTEL_LOG_LEVEL=debug` or something - I would expect this would be off by default (as an end user) but be something I could opt into when needed.

(Unless this is required by the spec, of course).

robertlaurin:
> `OTEL_RUBY_INSTRUMENTATION_INSTALL_MESSAGES`
> 
> Is there a general log-level argument we could use instead? `OTEL_DEBUG=1` or `OTEL_LOG_LEVEL=debug` or something - I would expect this would be off by default (as an end user) but be something I could opt into when needed.
> 
> (Unless this is required by the spec, of course).

Hmm, we do have `OTEL_LOG_LEVEL` https://github.com/open-telemetry/opentelemetry-ruby/blob/943e44071121402ae9c72c77d8a7d7884c351dda/api/lib/opentelemetry.rb#L31 so that could be a better choice here.  

The context for this change is around users getting tired of the install message log spam, previously they would use `OTEL_LOG_LEVEL` but we're now setting the OpenTelemetry logger to use the Rails.logger as part of our railtie so that env var isn't working the same.  

I was already a little apprehensive to add a one off config for this.

ahayworth (approved):

fbogsany (approved):

ahayworth (approved):
This looks great!

amomchilov (commented):
This is neat, thanks!

Issue #832, opened 2021-06-22
XRay not working


scar-lovevery:
Not sure if it's related to:
https://github.com/open-telemetry/opentelemetry-collector/issues/3405
or
https://github.com/aws-observability/aws-otel-collector/issues/537
Seems like both of these indicate that the trace id wasnt correctly generated, which i'm also guessing is the issue i'm facing.

I'm getting a similar behavior to the above issues.

When sending data to the aws-otel-collector I'm getting the "Permanent error: SerializationException" much like they are.  I have tried many different configurations.  

I know that the aws-otel-collector is configured correctly if I run a java example, I get traces forwarded to XRay:
```
# create namespace
apiVersion: v1
kind: Namespace
metadata:
  name: aws-otel-eks
  labels:
    name: aws-otel-eks
---
# create deployment
apiVersion: apps/v1
kind: Deployment
metadata:
  name: aws-otel-eks
  namespace: aws-otel-eks
  labels:
    name: aws-otel-eks
spec:
  replicas: 1
  selector:
    matchLabels:
      name: aws-otel-eks
  template:
    metadata:
      labels:
        name: aws-otel-eks
      annotations:
        linkerd.io/inject: enabled
    spec:
      containers:
        - name: aws-otel-emitter
          image: "aottestbed/aws-otel-collector-java-sample-app:0.9.0"
          env:
            - name: OTEL_EXPORTER_OTLP_ENDPOINT
              value: "http://aws-otel-collector.example.svc.cluster.local:4317"
            - name: OTEL_RESOURCE
              value: ClusterName=example
            - name: OTEL_RESOURCE_ATTRIBUTES
              value: "service.namespace=AWSObservability,service.name=CloudWatchEKSService"
            - name: S3_REGION
              value: eu-central-1
            - name: LISTEN_ADDRESS
              value: "0.0.0.0:4567"
          imagePullPolicy: Always
        - name: curl
          imagePullPolicy: Always
          image: ellerbrock/alpine-bash-curl-ssl:latest
          command: [ "/bin/bash", "-c", "sleep 10; while :; do curl localhost:4567/outgoing-http-call > /dev/null 1>&1; sleep 2; curl localhost:4567/aws-sdk-call > /dev/null 2>&1; sleep 5; done" ]
```

I'm attempting to use the XRay propagator (0.18.0)
Tried(environment var) OTEL_PROPAGATORS:
          value: xray,tracecontext,baggage
&
[...]

PR #840, opened 2021-06-24
feat: Xray compliant ids

scar-lovevery:
Need some feedback here.  @fbogsany @ahayworth @robertlaurin  This general implementation works on my local test environment.  

Questions:
Seems like this is in the wrong place? Where should I put this "extension" for AWS XRay?  Perhaps I should include it in the xray propagator gem?  Seems like people should use that gem in conjunction with this.

Remaining Tasks:
1. Decide where this should live
2. Write test(s)
3. Update some documentation

This is all in reference to:
https://github.com/open-telemetry/opentelemetry-ruby/issues/832


ahayworth:
> Seems like this is in the wrong place? Where should I put this "extension" for AWS XRay? Perhaps I should include it in the xray propagator gem? Seems like people should use that gem in conjunction with this.

Yes - I would agree that it shouldn't go here. I think your suggestion of placing it alongside the xray propagator somehow would make the most sense. Then as part of this PR we could also update the xray propagator docs to reference the extra installation step. 😄

fbogsany:
IIRC AWS has included this sort of thing in their OpenTelemetry distributions. AFAIK they don't yet have a distribution for OpenTelemetry Ruby. IDK whether they regard XRay propagation and XRay trace ID generation as going hand-in-hand - technically, you could use a different trace context propagator with XRay, and potentially only use the XRay propagator when talking to AWS services.

All that aside, unless someone from AWS has a strong opinion (and I don't know who to ping about this), sticking it in the XRay propagator gem seems reasonable to start with.

scar-lovevery:
Looking at where Java put it:
https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/aws/src/main/java/io/opentelemetry/sdk/extension/aws/trace/AwsXrayIdGenerator.java

If we follow their example it seems it should go into some type of SDK extension, I dont see a similar pattern for something like this in the current repo?  It's not a big deal to me one way or another.  🤷   

I'm good with moving it into the propagator.  Seems like a good place for now.  The propagator doesn't automatically instrument so you can still use a different propagator and the xray id generation functionality independently.

I'll do this now.

scar-lovevery:
Alright, updated.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # This module is entended to only be used as an override for how to generate IDs to be compliant with XRay
```
ahayworth, line 8:
```suggestion
  # This module is intended to only be used as an override for how to generate IDs to be compliant with XRay
```

ahayworth (approved):

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+  module AWSXRayTrace
```
fbogsany, line 9: The module hierarchy here does not match the directory structure. I would expect:
```ruby
module OpenTelemetry
  module Propagator
    module AWSXRayTrace
```
However, I think this should be nested under the `XRay` module as `IDGenerator`, i.e.
```ruby
module OpenTelemetry
  module Propagator
    module XRay
      module IDGenerator
```

fbogsany (changes requested):

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+    # non-zero byte.
+    # AWS Docs: https://docs.aws.amazon.com/xray/latest/api/API_PutTraceSegments.html
+    # hi - 4 bytes timestamp, 4 bytes random (Mid)
+    # low - 8 bytes random.
+    # Since we include timestamp, impossible to be invalid.
+    # @return [bytes] a valid trace ID that is compliant with AWS XRay.
+    def generate_trace_id
+      time_hi = generate_time_byte
+      mid = generate_variable_byte_id(4)
+      low = generate_variable_byte_id(8)
```
fbogsany, line 22: Could you replace this with:
```suggestion
      rest = generate_variable_byte_id(12)
```
Better yet would be:
```suggestion
      rest = RANDOM.bytes(12)
```
Since you don't need to avoid generating 0.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+    # respond to `#bytes` and uses the system PRNG. The current logic is
+    # compatible with Ruby 2.5 (which does not implement the `Random.bytes`
+    # class method) and with Ruby 3.0+ (which deprecates `Random::DEFAULT`).
+    # When we drop support for Ruby 2.5, this can simply be replaced with
+    # the class `Random`.
+    #
+    # @return [#bytes]
+    RANDOM = Random.respond_to?(:bytes) ? Random : Random::DEFAULT
+
+    def generate_time_byte
```
fbogsany, line 46: I'd prefer:
```suggestion
    def generate_time_bytes
```

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+      # get the seconds from epoch, convert to hex, then convert to bytes
+      [Time.now.to_i.to_s(16)].pack('H*')
```
fbogsany, line 48: It seems inefficient to convert to a string first. You should be able to do:
```suggestion
      [Time.now.to_i].pack('V')
```
(You might need to use 'N' instead of 'V' - I'm not sure offhand whether the time stamp bytes should be little-endian or big-endian.)

fbogsany (changes requested):

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+
+    def generate_variable_byte_id(bytes = 8)
+      return nil unless bytes.is_a? Numeric
+      return nil unless bytes.between?(1, 256)
+
```
fbogsany, line 54: You don't need these checks since this is a private method and you control all callers (in fact, with my suggestion above, there's only one caller, which suggests it may be worth inlining this into the caller - it has little value on its own).
```suggestion
```

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,62 @@
...
+      invalid_id = ("\0" * bytes).b
```
fbogsany, line 55: This will allocate a string and convert it to a byte array on every call. Our implementation in `OpenTelemetry::Trace` stores this in a constant to avoid this overhead.

fbogsany (changes requested):

scar-lovevery:
@fbogsany Thanks for the feedback, total noob here with ruby.  I'll respond to each comment, but they should all be resolved.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: Removed all of this.  Makes sense.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: Removed.  I guess I was trying to protect myself from the future.  🤷

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: Looked it up:  'N' is correct.  Docs: "N             | Integer | 32-bit unsigned, network (big-endian) byte order".
Updated.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: Makes sense.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: Since I removed this function, i'm directly using RANDOM like suggested.

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
scar-lovevery: 🤦

propagator/xray/lib/opentelemetry/propagator/aws_xray_trace.rb
```diff
@@ -0,0 +1,55 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Propagator
+    module XRay
+      # This module is intended to only be used as an override for how to generate IDs to be compliant with XRay
+      module IDGenerator
```
fbogsany, line 11: The file needs to be renamed for this too.

scar-lovevery: done.

propagator/xray/lib/opentelemetry/propagator/xray.rb
```diff
@@ -46,3 +46,4 @@ def text_map_propagator
 end
 
 require_relative './xray/version'
+require_relative './id_generator'
```
fbogsany, line 49:
```suggestion
require_relative './xray/id_generator'
```
and rename the file accordingly

propagator/xray/lib/opentelemetry/propagator/id_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+        # class method) and with Ruby 3.0+ (which deprecates `Random::DEFAULT`).
+        # When we drop support for Ruby 2.5, this can simply be replaced with
+        # the class `Random`.
+        #
+        # @return [#bytes]
+        RANDOM = Random.respond_to?(:bytes) ? Random : Random::DEFAULT
+
+        def generate_time_bytes
+          # get the seconds from epoch, convert to hex, then convert to bytes
+          # big-endian
```
fbogsany, line 49:
```suggestion
        # Seconds since epoch converted to 4 bytes in big-endian order.
        def generate_time_bytes
```

propagator/xray/lib/opentelemetry/propagator/id_generator.rb
```diff
@@ -0,0 +1,55 @@
...
+        # non-zero byte.
+        # AWS Docs: https://docs.aws.amazon.com/xray/latest/api/API_PutTraceSegments.html
+        # hi - 4 bytes timestamp, 4 bytes random (Mid)
+        # low - 8 bytes random.
+        # Since we include timestamp, impossible to be invalid.
+        # @return [bytes] a valid trace ID that is compliant with AWS XRay.
+        def generate_trace_id
+          time_hi = generate_time_bytes
+          low_and_low = RANDOM.bytes(12)
+          time_hi << low_and_low
```
fbogsany, line 24: For consistency with the method doc:
```suggestion
          mid_and_low = RANDOM.bytes(12)
          time_hi << mid_and_low
```

scar-lovevery:
Updated.

fbogsany (approved):
LGTM - thanks @scar-lovevery !

PR #841, opened 2021-06-24
feat: Make API's NoopTextMapPropagator private

### LINKED ISSUE #812: Consider making NoopTextMapPropagator private
https://github.com/open-telemetry/opentelemetry-ruby/blob/main/api/lib/opentelemetry/context/propagation/noop_text_map_propagator.rb can probably be defined as private, and be enabled public later on if/as needed. This is done to reduce API surface ;)

ahayworth:
To do so, we wholesale duplicate it into the SDK.

fixes #812


api/lib/opentelemetry/context/propagation/noop_text_map_propagator.rb
```diff
@@ -7,6 +7,8 @@
 module OpenTelemetry
   class Context
     module Propagation
+      # @api private
```
fbogsany, line 10: I feel like, if we're making this `@api private`, we should just delete all the method documentation. The documentation on `TextMapPropagator` is mostly pretty similar, and folks looking to implement their own will probably discover that implementation before this one 🤷

sdk/lib/opentelemetry/sdk/context/propagation/noop_text_map_propagator.rb
```diff
@@ -0,0 +1,53 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    class Context
+      module Propagation
+        # A no-op text map propagator implementation
```
fbogsany, line 11: This should also be `@api private` and mostly undocumented.

api/lib/opentelemetry/context/propagation/noop_text_map_propagator.rb
ahayworth: 👍 I removed all documentation for the `NoopTextMapPropagator` (API and SDK) in f0688bff .

sdk/lib/opentelemetry/sdk/context.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    class Context < OpenTelemetry::Context
```
fbogsany, line 9: This feels weird. Why do we need this subclass? The no-op propagator should be able to stand alone without it. 🤔 

OIC! You're just trying to mirror the structure from the API. I don't think we need to do that. We could really just make it a private module inside the `Configurator` class.

ahayworth: Yeah, I was going for mirroring the structure rather than just dumping it somewhere. But if you're fine with putting it directly into the Configurator, that's fine with me - see e2a2219e .

fbogsany (approved):

PR #851, opened 2021-06-30
Update CODEOWNERS to use OpenTelemetry team

MikeGoldsmith:
Updates the codeowners file to use the Github team @open-telemetry/ruby-approvers instead of discrete list of users. This will make managing the group easier as it doesn't require code changes when the approvers list is updated.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -12,4 +12,4 @@
 #  https://help.github.com/en/articles/about-code-owners
 #
 
-* @fbogsany @mwear @robertlaurin @dazuma @ericmustin @arielvalentin @ahayworth
+* @open-telemetry/ruby-approvers


ahayworth (dismissed):

fbogsany:
While I appreciate the convenience of this from a maintenance perspective, is there a way to be notified of changes to the membership of @open-telemetry/ruby-approvers? Updating this file manually currently notifies anyone watching this repo.

Side note: your PR description refers to rust-approvers rather than ruby-approvers.

MikeGoldsmith:
I'm not certain if Github teams have a notification feature on edit. Is it important to update all repository watchers of approver changes?

(I've updated the PR description to reference the correct type - must have been auto-complete error).

fbogsany:
> Is it important to update all repository watchers of approver changes?

It's a nice side effect. Most discussion of approver changes happen in SIG meetings, which likely have a more limited audience than PRs to this repo. I know I would like to know if the approvers list was changed 😄 - this PR made me wonder how I would be informed of that.

arielvalentin:
> It's a nice side effect. Most discussion of approver changes happen in SIG meetings, which likely have a more limited audience than PRs to this repo. I know I would like to know if the approvers list was changed 😄 - this PR made me wonder how I would be informed of that.

@fbogsany if I understand your need here you want to subscribe to team changed events? That way you could be notified is someone is added or removed from a team. Is that correct?

robertlaurin:
This may not be the best justification for keeping it the way it is, but having an explicit code owners list creates a fun little ceremony when adding members to the group.  I see it as a positive thing, and given the final say I would keep it as is because it promotes the type of community I seek to be part of.

robertlaurin:
Going to close this as I don't see us making this change in the near future.

ericmustin:
long live _the ceremony_

PR #852, opened 2021-06-30
Support Active Jobs with keyword args across ruby versions 

kjg:
Ruby 2.7 / 3.0 changes the way that keyword args work: see [Separation of positional and keyword arguments in Ruby 3.0](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) for full details.

In order to support passing keyword args in ruby versions both prior to and after this change, we need to make use of the `ruby2_keywords` method in ruby versions where it is defined. [ActiveJob base](https://github.com/rails/rails/blob/6-0-stable/activejob/lib/active_job/core.rb#L90) (Starting in Rails 6) does this as well, so we're also following their lead

Running the new test without the code change produces the following error on ruby 3

```ArgumentError: wrong number of arguments (given 1, expected 0; required keyword: keyword)```

### DIFF
--- a/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb
+++ b/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb
@@ -20,6 +20,7 @@ module OpenTelemetry
             @metadata = {}
             super
           end
+          ruby2_keywords(:initialize) if respond_to?(:ruby2_keywords, true)
 
           def serialize
             super.merge('metadata' => serialize_arguments(metadata))
--- a/instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
+++ b/instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
@@ -46,6 +46,12 @@ describe OpenTelemetry::Instrumentation::ActiveJob::Patches::ActiveJobCallbacks 
     end
   end
 
+  describe 'compatibility' do
+    it 'works with keyword args' do
+      _(KeywordArgsJob.perform_now(keyword: :keyword)).must_be_nil # Make sure this runs without raising an error
+    end
+  end
+
   describe 'exception handling' do
     it 'sets span status to error' do
       _ { ExceptionJob.perform_now }.must_raise StandardError, 'This job raises an exception'
--- a/instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
+++ b/instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
@@ -47,8 +47,16 @@ describe OpenTelemetry::Instrumentation::ActiveJob::Patches::ActiveJobCallbacks 
   end
 
   describe 'compatibility' do
+    it 'works with positional args' do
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb [hunk 1]
```diff
@@ -20,6 +20,7 @@ def initialize(*args)
             @metadata = {}
             super
           end
+          ruby2_keywords(:initialize) if respond_to?(:ruby2_keywords, true)
```
arielvalentin, line 23: I was curious if using `double splat` would be enough, i.e. `def initialize(**kwargs)`, but it looks like this is how it is defined in `rails/active_job/core`:

https://github.com/rails/rails/blob/53d54694e9a423642ba9c984207097e9522db26f/activejob/lib/active_job/core.rb#L100

It looks like once we have minimal support for 2.7 we could remove the `respond_to` check.
Is there a min version of `ActiveJob`, and by extension Ruby, that we are looking to support for this instrumentation?

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
```diff
@@ -46,6 +46,12 @@
     end
   end
 
+  describe 'compatibility' do
```
arielvalentin, line 49: Would it be too much trouble to add test cases using non-keyword args and positional attributes?

arielvalentin (approved):

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb
ahayworth: @arielvalentin 5.2 is the lowest supported version: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb#L12

ahayworth (approved):
Ah, good catch. None of our internal testing jobs use kwargs, and I had forgotten about the behavior change.

This LGTM, although I agree @arielvalentin's comment about adding a few additional tests for other arg types, if it's not too much trouble.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
kjg: I added some additional tests, are there any other cases that you can think of that should be added?

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb
kjg: I *think* double splat will only capture keyword args and not positional args, but I could be wrong about that. I do think it makes sense to mirror what canonical activejob is doing as closely as possible.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
arielvalentin: I think that covers it thanks!

fbogsany (approved):

PR #854, opened 2021-06-30
feat: GRPC instrumentation

### LINKED ISSUE #96: Code examples: gRPC instrumentation example

### LINKED ISSUE #248: gRPC instrumentation adapter
Convert Datadog instrumentation for gRPC to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/grpc)

ahayworth:
This PR adds GRPC auto-instrumentation. The following features are implemented:

- [x] Client / Server "unary" RPCs (single request/response)
- [x] Client / Server "server streaming" RPCs (single resquest / stream of responses)
- [x] Client / Server "client streaming" RPCs (stream of requests / single response)
- [x] Client / Server "bi-directional streaming" RPCs (stream of requests / stream of responses)
- [x] Log events for individual sent / received messages during a stream's lifecycle

The following features are not yet implemented, but will be finished in this PR in short order:

- [x] Tests, of any kind
- [ ] `net.peer.*` and `net.transport` semantic convention attributes
- [ ] Span error handling (this might actually be working, but I haven't tested it yet)

I might not implement:

- [ ] `message.compressed_size` and `message.uncompressed_size` semantic convention attributes. I think I might be able to get to those sizes, but I'm not completely certain. They are not listed as a `MUST` in the spec, so I think it might be okay to leave out.

Things of note to reviewers:

- As stated before, I'm not completely done here. But if you have feedback or any thoughts, they are more than welcome at this point.
- I initially tried to use client / server interceptors, but they only handle the beginning of the calls and are thus unsuited for the task of instrumenting a long-running call.
- Clients using the ruby GRPC libraries can invoke a service passing the `return_op: true` flag ... and then they get back a version of an `ActiveCall` that can actually be invoked _later_. So we don't necessarily know that the RPC call will take place when `ClientStub#request_response` is called, for instance. That leads to some annoying duplication throughout, and passing spans around as instance attributes on `ActiveCall` / `BidiCall` objects.
- If a client does not pass a block to a streaming call, they'll receive an enumerable that can be called later to lazily read in the actual responses. This causes additional annoyances because we don't really know what spans may be active in the context when we actually start to iterate through the responses - another reason we pass the spans around as instance variables.

I'm not sure that this approach is a good one, but it feels okay to me. I'm worried primarily that we won't clean up spans correctly (what if a client never enumerates their results?) but I'm not sure I see another way to do it while still adhering to the [GRPC event spec requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#events).

<details>
<summary>Expand me for a bunch of screenshots</summary>
Screen Shot 2021-06-30 at 4 54 30 PM
Screen Shot 2021-06-30 at 4 55 56 PM
Screen Shot 2021-06-30 at 4 55 41 PM
Screen Shot 2021-06-30 at 4 55 23 PM
Screen Shot 2021-06-30 at 4 55 07 PM
Screen Shot 2021-06-30 at 4 54 45 PM
</details>

Fixes #248
Fixes #96


instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
```diff
@@ -0,0 +1,72 @@
...
+          # from set_{input,output}_stream_done. We use it to finish
+          # off the span we've been tracking throughout the
+          # lifecycle of this ActiveCall.
+          def maybe_finish_and_close_call_locked
+            super
+
+            if @call_finished && @span.recording?
+              unless status.nil?
+                @span.set_attribute("rpc.grpc.status_code", status.code)
+                @span.status = OpenTelemetry::Trace::Status.error unless status.code == ::GRPC::Core::StatusCodes::OK
```
arielvalentin, line 63: IIRC we recently discussed that auto-instrumentation should not set the `span.status`. Is that correct?

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
```diff
@@ -0,0 +1,72 @@
...
+          end
+
+          # This method is locked with a mutex and only called
+          # from set_{input,output}_stream_done. We use it to finish
+          # off the span we've been tracking throughout the
+          # lifecycle of this ActiveCall.
+          def maybe_finish_and_close_call_locked
+            super
+
+            if @call_finished && @span.recording?
```
arielvalentin, line 60: Since we are finishing the span, should this or at least the `finish` be in an `ensure` block?

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
ahayworth: IIRC the outcome was that yes, generally speaking instrumentation shouldn't set the status (and certainly should not set to "OK", which is now a "final" state that can't be changed). But, I think errors _can_ still be set.

I was working from the [relevant part of the RPC semantic conventions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/rpc.md#grpc-status), which state:

> The Span Status MUST be left unset for an OK gRPC status code, and set to Error for all others.

The current [relevant portion of the API docs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status) seem to say that this is still allowable:

> The status code SHOULD remain unset, except for the following circumstances:
> When the status is set to Error by Instrumentation Libraries, the status codes SHOULD be documented and predictable. The status code should only be set to Error according to the rules defined within the semantic conventions. For operations not covered by the semantic conventions, Instrumentation Libraries SHOULD publish their own conventions, including status codes.
> Generally, Instrumentation Libraries SHOULD NOT set the status code to Ok, unless explicitly configured to do so. Instrumention libraries SHOULD leave the status code as Unset unless there is an error, as described above.

So - I think this is okay still, based on the specs. But I'm not 100% certain actually.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
ahayworth: Good point. Done in a3a9fa83 .

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
fbogsany: The spec is verbose, but the essence is (for instrumentation):
1. Don't ever set `span.status = OpenTelemetry::Trace::Status.ok`
2. If there's an error, and the semantic conventions say you should set an error, set `span.status = OpenTelemetry::Trace::Status.error`
3. Otherwise leave `span.status` alone! What did it ever do to you anyway?!

ahayworth:
Behold, I am alive!

I think this PR is ready for testing and review. Sadly I don't have much of a real application to test this with. If anyone does, I'd greatly appreciate it!

instrumentation/grpc/Gemfile
```diff
@@ -0,0 +1,21 @@
...
+
+gem 'opentelemetry-api', path: '../../api'
+gem 'opentelemetry-instrumentation-base', path: '../base'
+
+group :test do
+  gem 'opentelemetry-common', path: '../../common'
+  gem 'opentelemetry-sdk', path: '../../sdk'
+  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
+
+  gem 'pry'
```
arielvalentin, line 19: I think we are moving these to the Gemspec as development dependencies. E.g.
https://github.com/open-telemetry/opentelemetry-ruby/pull/896/files#diff-0e18c3a8e005f4da1b58da58cdd9c29ef449fb3526504bcb81d70893de6e078aR36

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
```diff
@@ -0,0 +1,73 @@
...
+          def initialize(call, marshal, unmarshal, deadline, started: true, metadata_received: false, metadata_to_send: nil)
+            super
+            @span = OpenTelemetry::Trace.current_span
+          end
+
+          # We need a setter because while we can easily capture a correct trace during initialization
+          # on the client side, it's a little harder in the accept loop on the server. So, we provide
+          # a way for server-side ActiveCalls to set the span after the ActiveCall is created.
+          def span=(span)
+            @span = span
```
arielvalentin, line 22: I am unfamiliar with the execution model here. Does this require a threadsafe writer?

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/core_call.rb
```diff
@@ -0,0 +1,66 @@
...
+    module GRPC
+      module Patches
+        # Patches to prepend to the ::GRPC::CoreCall, which is a C extension.
+        module CoreCall
+          # This is the heart of sending / receiving GRPC messages on the wire.
+          # Even though it's a C extension, we can still intercept it here to
+          # handle the log message requirements of the OpenTelemetry semantic conventions.
+          def run_batch(ops)
+            if ops.keys.include?(::GRPC::Core::CallOps::SEND_MESSAGE)
+              OpenTelemetry::Trace.current_span.add_event(
```
arielvalentin, line 18: I wish I had a better grasp of the execution model here. I keep thinking back to the way that the Rack instrumentation keeps track of its own Span instead of using the "current" span in case users add custom interceptors that generate their own spans and I want to make sure the gRPC span is the one that is augmented with the event.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/rpc_desc.rb
```diff
@@ -0,0 +1,44 @@
...
+            span = tracer.start_span(
+              "#{mth.owner.service_name}/#{mth.name}",
+              with_parent: context,
+              kind: :server,
+              attributes: attrs
+            )
+            # The server's ActiveCall is created during the accept loop, and so the
+            # current_span saved off when created will likely be a non-recording API
+            # span. So we replace the saved-off span with whatever we just created.
+            active_call.span = span
```
arielvalentin, line 30: Something I do not quite grok is why this does not rely on `current_span` since it is passed in to `with_span` on L32?

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/rpc_desc.rb
```diff
@@ -0,0 +1,44 @@
...
+    module GRPC
+      module Patches
+        # Patches to prepend to ::GRPC::RpcDesc.
+        module RpcDesc
+          # This is the entry point to all server-side calls.
+          def run_server_method(active_call, mth, inter_ctx = ::GRPC::InterceptionContext.new)
+            attrs = {
+              'rpc.system' => 'grpc',
+              'rpc.service' => mth.owner.service_name,
+              'rpc.method' => mth.name.to_s
```
ericmustin, line 18: small nit (was working on something related and figured i'd see how this was implemented, havent done a full review)

I think semantic conventions require either `net.peer.ip` _or_ `net.peer.name`: https://github.com/open-telemetry/opentelemetry-specification/blob/87a5ed7f0d4c403e2b336f275ce3e7fd66a8041b/specification/metrics/semantic_conventions/rpc.md#attributes

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/client_stub.rb
```diff
@@ -0,0 +1,62 @@
...
+          private
+
+          def start_span(method, metadata)
+            method_parts = method.to_s.sub(%r{^\/}, '').split('/')
+            service_name = method_parts.shift
+            method_name = method_parts.join('/')
+            attrs = {
+              'rpc.system' => 'grpc',
+              'rpc.service' => service_name.to_s,
+              'rpc.method' => method_name.to_s
```
ericmustin, line 42: same as https://github.com/open-telemetry/opentelemetry-ruby/pull/854/files#r683679304

instrumentation/grpc/Gemfile
ahayworth: Oh, good point. Looks like it was already in the gemspec with a matching line in the `test_helper.rb`, so I just removed this in 8aaea741 .

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/active_call.rb
ahayworth: I don't _think_ so. I will admit that fundamentally I am not 100% sure. This ActiveCall setter is only called in two places:
- when the ActiveCall is initialized (which sets the "correct" span for client side things using the ActiveCall class)
- when we start to handle a call on the server-side - just after it is dispatched for processing, but technically _after_ the ActiveCall was already created.

The main reason I added it was because on the server-side, I couldn't find a great way to cleanly patch the accept loop that handles and schedules all of the server methods without re-implementing a lot of that code myself. That felt like a great way to introduce bugs to me 😆.

The "accept loop" (so to speak) is at [`RpcServer#loop_handle_server_calls`](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/rpc_server.rb#L448) - and the bit I did choose to patch was [`RpcDesc#run_server_method`](https://github.com/grpc/grpc/blob/master/src/ruby/lib/grpc/generic/rpc_desc.rb#L117), which is called from within `RpcServer#loop_handle_server_calls`. If you take a peek at the handling code, you can probably see why I chose to do it this way. However, a fresh set of eyes would be appreciated sometime, because I may well have overlooked the easy way to do this 🤣 

Anyways - back to the thread safety question - the reason I think this is okay is because we only ever change this variable in one place, and that specifically is when we dispatch the server-side ActiveCall to a thread for processing. A different ActiveCall is created for each request coming in, and my read of the work-processing system within the GRPC code is that requests are explicitly scheduled _with_ their context to a single thread, rather than multiple threads potentially pulling the same work item or something. I don't immediately see a way in which there could be concurrent access to this setter given the scope and the overall way that the GRPC library is doing things.

That said ... the GRPC library is rather confusing, so I very much could be wrong there. I'm going to say I'm about 95% certain we are fine here, in my opinion.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/core_call.rb
ahayworth: > I wish I had a better grasp of the execution model here.

Me too, honestly. 🤣 This one is rough, for sure.

> I keep thinking back to the way that the Rack instrumentation keeps track of its own Span instead of using the "current" span in case users add custom interceptors that generate their own spans and I want to make sure the gRPC span is the one that is augmented with the event.

I think this is an excellent concern. I'm going to try and refactor it along those lines. I _think_ the scenario you described could happen if someone wrote their own interceptor, and that would definitely be a really upsetting / embarrassing bug for us. Not the first one I'd have written, either. 🤦 

I think what really makes this hard is that the execution model of the official GRPC library is just all over the place - and the bit we're discussing here is actually a C extension ... and a lot of work happens in that C extension, which just makes it all harder.

That said, I think I see a way to get to the state we want. 👍

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/rpc_desc.rb
ahayworth: You know - at this point it's been so long since I wrote it that I don't precisely recall why. I suspect it's something like what you were talking about in another area - I wanted to make sure we're augmenting the correct span, so the stuff inside the `ActiveCall` patches all operate on the `@span` variable instead of `current_span`. But; there are places (the c extension, etc) where getting at the current `ActiveCall` isn't possible so I then need to use `current_span` ... and it gets to be a mess.

Like I said elsewhere, I'm going to try harder to encapsulate "our spans" more cleanly, and then as part of that refactor I'll try to make all of this stuff a lot more clear somehow.

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/rpc_desc.rb
ahayworth: Ah, you're right. It's even on the TODO list for the PR and I neglected to do it. 😆

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/client_stub.rb
ahayworth: Yup - I need to do that still. I think I punted on it because it looked rather hard to figure out, but I suspect we can figure it out. 😄

ahayworth:
Putting this back into draft; the comments from @arielvalentin and @ericmustin were pretty helpful - and I need to do a major revision to try to get this ready to merge. 😄  Thanks!

instrumentation/grpc/lib/opentelemetry/instrumentation/grpc/patches/rpc_desc.rb
arielvalentin: We should also look at other SDK implementations as well and see if the generated code looks similar to see how they are handling it. That may help us puzzle through some options.

jimeh:
What's progress on this looking like? How much is left to get it across the line?

I've got a quick and dirty GRPC interceptor implementation working within a project, but coming up against some limitations with how the GRPC interceptors are implemented in Ruby. Which is why I assume you're not using them here.

I'd be happy to spare a couple of evenings to try and help out if it's of any use :)

And if it helps, regarding `message.uncompressed_size`, you should be able to get that by simply calling `.to_proto.size` on GRPC request and response objects.

ahayworth:
@jimeh it's definitely stalled, my day job no longer requires ruby and I don't have as much free and clear time to devote to otel-ruby :(

The general status was, however:
- Needs to be cleaned up overall. The flow is confusing and I think it's likely incorrect in places.
- Needs testing. Even when I wrote this, I didn't have a GRPC ruby app that I could do much testing with.

I'd be happy to work on it with you, or for you to take over the branch if you have the motivation to do so!

jimeh:
@ahayworth Thanks for the info :)

I've actually ended up taking a quite different approach in an attempt to quickly hack something together for my day job that just about works. Turned out to not be that quick, but I think I have it all working now with message events, error handling, etc. It lacks automated tests at the moment though, so no there's no guarantees around those promises... lol

I'll try and get things cleaned up and organized so I can put together a draft PR here sometime soon. Any help you're able and willing to offer at that point @ahayworth would be greatly appreciated :)

As for the details of my implementation; It's all operating through gRPC client and server interceptors, which means a smaller footprint of monkey-patched / prepended classes, so it's a bit easier to follow how it works (I think/hope). However, due to the limitations of interceptors, the bulk of the tracing is done by monkey-patching methods on individual instances of objects exposed to interceptors. Both long-lived and per-request objects at that. It's kind of horrible, but yet somewhat elegant, and reasonably easy to follow the logical flow of... I think :P

ahayworth:
@jimeh I look forward to the PR, that's awesome!

robertlaurin:
Hey @jimeh, did you still have any time/bandwidth to share a draft with us?  We're looking get some GRPC instrumentation out, and I'd be interested to see what your approach looks like.

ahayworth:
I'm going to go ahead and close this out - it's unlikely that I'll be able to work on it now that my day job does not involve ruby. However, hopefully it's instructive for a future attempt!

jimeh:
@robertlaurin Apologies for the silence, it's been a busy few months, and my focus has actually ended up being on the Golang side of our project that uses gRPC and otel. But I'm planning to put aside a couple of evenings this week or next to extract the Ruby implementation I wrote to get a draft PR submitted. After that, next steps are tests, and fixing a bug where message events are sometimes assigned to the wrong span during high volumes of parallel gRPC requests.

@ahayworth Thanks for your efforts on this PR. It was a really valuable resource for me end of last year when I went about my "quick and dirty hack" implementation. I hope you won't mind if I poke you when I open the above mentioned PR within the next week or two :)

PR #858, opened 2021-07-01
feat: Add auto-generated semantic conventions gem

### LINKED ISSUE #733: Semantic convention constants should be autogenerated from YAML
See https://github.com/open-telemetry/opentelemetry-specification/pull/1515. Specifically, we should provide Ruby-specific support to the [code generator](https://github.com/open-telemetry/build-tools/tree/main/semantic-conventions#code-generator).

ahayworth:
Generated from the YAML, complete with Rake task to build it.

Fixes #733

### DIFF
--- a/.toys/.data/releases.yml
+++ b/.toys/.data/releases.yml
@@ -191,3 +191,8 @@ gems:
     directory: resource_detectors
     version_rb_path: lib/opentelemetry/resource/detectors/version.rb
     version_constant: [OpenTelemetry, Resource, Detectors, VERSION]
+
+  - name: opentelemetry-semantic_conventions
+    directory: semantic_conventions
+    version_rb_path: lib/opentelemetry/semantic_conventions/version.rb
+    version_constant: [OpenTelemetry, SemanticConventions, VERSION]
--- a/semantic_conventions/.rubocop.yml
+++ b/semantic_conventions/.rubocop.yml
@@ -1,0 +1,37 @@
+AllCops:
+  TargetRubyVersion: '2.5.0'
+
+Bundler/OrderedGems:
+  Exclude:
+    - gemfiles/**/*
+Lint/UnusedMethodArgument:
+  Enabled: false
+Metrics/AbcSize:
+  Max: 18
+Metrics/LineLength:
+  Enabled: false
+Metrics/MethodLength:
+  Max: 20
+Metrics/ModuleLength:
+  Enabled: false
+Metrics/ParameterLists:
+  Enabled: false
+Style/FrozenStringLiteralComment:
+  Exclude:
+    - gemfiles/**/*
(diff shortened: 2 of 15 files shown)


ahayworth:
Note that:
- I didn't include tests (wasn't sure what to test)
- I disabled a few additional rubocop rules for this, rather than fighting Jinja to produce perfect ruby style (it's close enough for an autogenerated thing, I think).

Example:
```ruby
~/projects/opentelemetry-ruby/semantic_conventions ahayworth-yaml-intensifies 19s
❯ pry
[1] pry(main)> require_relative './lib/opentelemetry-semantic_conventions'
=> true
[2] pry(main)> OpenTelemetry::SemanticConventions::Trace::FAAS_DOCUMENT_NAME
=> "faas.document.name"
[3] pry(main)> OpenTelemetry::SemanticConventions::Resource::PROCESS_PID
=> "process.pid"
[4] pry(main)>
```

We could alternately put these under the `OpenTelemetry::Trace` and `OpenTelemetry::Resource` namespaces as well, if we think that's a good idea - cc @open-telemetry/ruby-approvers for any thoughts there.

fbogsany (approved):
This is awesome! Thanks @ahayworth!

fbogsany:
> We could alternately put these under the `OpenTelemetry::Trace` and `OpenTelemetry::Resource` namespaces as well, if we think that's a good idea

I think it is fine under `OpenTelemetry::SemanticConventions`. I 🤔 about whether we want additional nesting to produce names like `OpenTelemetry::SemanticConventions::Trace::DB::Cassandra::KEYSPACE` but I think that's getting too verbose and it'd arguably slow down lookup.

robertlaurin (approved):
🥇

robertlaurin:
>I think it is fine under OpenTelemetry::SemanticConventions

Agreed

> I 🤔 about whether we want additional nesting to produce names like OpenTelemetry::SemanticConventions::Trace::DB::Cassandra::KEYSPACE but I think that's getting too verbose and it'd arguably slow down lookup.

I appreciate the desire for it, but that's a really long namespace.

fbogsany:
I think the follow-on to this should be to remove the `Resources` constants from https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/resources/constants.rb and replace their use with the constants from this gem instead. This will mean the SDK depends on the `semantic_conventions` gem, and I think that is fine (and likely intended by the spec). We should probably be careful with the dependency so that the `semantic_conventions` gem can be bumped without requiring a new version of the SDK gem.

ahayworth:
Agreed about the namespacing.

@fbogsany I opened https://github.com/open-telemetry/opentelemetry-ruby/issues/860 for SDK follow-on work.

PR #859, opened 2021-07-02
feat: Initial logging SDK implementation

ahayworth:
This implements most of the logging SDK otep [1], aside from the OTLP exporters. It uses a few since-removed concepts in the SDK, such a logging equivalent of the 'MultiSpanExporter', but that's okay for now.

I wanted to put this up for people to look at - not because I think it's really *ready* for anything, but I wanted any feedback that people care to offer on the general approach.

Notably:
- There exists, perhaps, an opportunity to generalize the batch span / log processors/exporters into something that can be re-used, somehow. And maybe the MetricsReporter, too!
- This gem is built as something that plugs into the sdk after it's been required, and the configuration is not precisely how I'd want to actually ship this.
- The gem isn't in the release data file, not yet anyways.
- The additional legwork in the example to associate trace/log context from a background thread is an interesting example as to what might be useful from a 'diagnostics api', or something like it.
- I have not yet written tests, so I can only assert that it works correctly because I can see it dump logs to my terminal. But it's probably broken in some ways.
- I copied and pasted a *lot* from the trace SDK, and I think that's obvious once you start reading through it.
- I included a Semantic Logger "appender", which is not required in the OTEP. However, it's not really easy to use the logging SDK without plugging it into another framework: so that shows how it could be done. We'd probably want to ship any "appenders" as their own gems or contribute them directly to the logging libraries in question.

We include a runnable example for people to try.

This is super, super experimental. It'll probably set your computer on fire.

[1] https://github.com/open-telemetry/oteps/blob/main/text/logs/0150-logging-library-sdk.md

### DIFF
--- a/log/CHANGELOG.md
+++ b/log/CHANGELOG.md
@@ -1,0 +1,3 @@
+# Release History: opentelemetry-sdk-log
+
+### Unreleased
--- a/log/Gemfile
+++ b/log/Gemfile
@@ -1,0 +1,11 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+gemspec
+
+gem 'opentelemetry-sdk', path: '../sdk'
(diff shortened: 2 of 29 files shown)


ahayworth:
I'm going to go ahead and close this out, as it's very old. But it may be useful for others to review in the future. ❤️

PR #865, opened 2021-07-07
feat: Use auto-generated resource constants in sdk and resource_detectors

### LINKED ISSUE #860: Use auto-generated Resource constants in SDK
I think the follow-on to this should be to remove the `Resources` constants from https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/resources/constants.rb and replace their use with the constants from this gem instead. This will mean the SDK depends on the `semantic_conventions` gem, and I think that is fine (and likely intended by the spec). We should probably be careful with the dependency so that the `semantic_conventions` gem can be bumped without requiring a new version of the SDK gem.

__Originally posted by @fbogsany in https://github.com/open-telemetry/opentelemetry-ruby/pull/858#issuecomment-874291155__

ahayworth:
This PR uses the auto-generated constants in the new `opentelemetry-semantic_conventions` gem.

NB: This will fail CI until the `opentelemetry-semantic_conventions` gem is released; and we certainly should not merge until that release happens.

Fixes #860


robertlaurin (approved):
CI is having a fit, but the code looks good.

We have to make sure when we merge this we use the `!` on the commit message so it identifies this as a breaking change (removing the constants file from the SDK).

ahayworth:
> We have to make sure when we merge this we use the ! on the commit message so it identifies this as a breaking change (removing the constants file from the SDK).

Good call; I force-pushed a fix to the commit history.

fbogsany (approved):

PR #871, opened 2021-07-12
feat: add explicit BSP export error

robertlaurin:


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -194,7 +194,7 @@ def report_result(result_code, batch)
               @metrics_reporter.add_to_counter('otel.bsp.export.success')
               @metrics_reporter.add_to_counter('otel.bsp.exported_spans', increment: batch.size)
             else
-              OpenTelemetry.handle_error(message: "Unable to export #{batch.size} spans")
+              OpenTelemetry.handle_error(exception: ExportError.new("Unable to export #{batch.size} spans"))
```
arielvalentin, line 197: I think this will end up passing in an exception without the backtrace being set. Is that OK?

```
irb(main):001:0> e = StandardError.new("some bug")
=> #<StandardError: some bug>
irb(main):002:0> e.backtrace
=> nil
```

robertlaurin: The backtrace getting set is a function of raising the exception.  So you're right that the exception backtrace is not set.  I think this fine because the error handler by default does not even look at the backtrace https://github.com/open-telemetry/opentelemetry-ruby/blob/d6ca70f6b47c7aad71985a64a5e28b7b7bd6d314/api/lib/opentelemetry.rb#L37

```
=> #<StandardError: some bug>
[4] pry(main)> e.backtrace
=> nil
[5] pry(main)> e = begin
  raise e
rescue => error
  error
[5] pry(main)> e = begin
  raise e
rescue => error
  error
end
=> #<StandardError: some bug>
[6] pry(main)> e.backtrace
=> ["(pry):20:in `__pry__'",
 "/Users/robertlaurin/.gem/ruby/2.5.3/gems/pry-0.14.1/lib/pry/pry_instance.rb:290:in `eval'",
 "/Users/robertlaurin/.gem/ruby/2.5.3/gems/pry-0.14.1/lib/pry/pry_instance.rb:290:in `evaluate_ruby'",
 "/Users/robertlaurin/.gem/ruby/2.5.3/gems/pry-0.14.1/lib/pry/pry_instance.rb:659:in `handle_line'",
 "/Users/robertlaurin/.gem/ruby/2.5.3/gems/pry-0.14.1/lib/pry/pry_instance.rb:261:in `block (2 levels) in eval'
 ```
 
 However if you decide to make the `error_handler` dangerous, the above would give you the stacktrace and the customer error.
 ```rb
   OpenTelemetry.error_handler = ->(exception: nil, message: nil) do
    raise exception || OpenTelemetry::Error, "OpenTelemetry error: #{[message, exception&.message].compact.join(" - ")}"
  end
```

arielvalentin: This is something that we should explicitly document at least for folks who want to write custom error handlers. I personally would always anticipate when given an exception it was generated as a results of `Kernel.raise` and have a backtrace explicitly set.

robbkidd:
As a followup to today's conversation about custom exception classes, while a bunch of little exception classes within a particular Rails app might be discouraged, [core Ruby suggests it for libraries](https://ruby-doc.org/core-2.7.0/Exception.html#class-Exception-label-Custom+Exceptions).

> A good practice is for a library to create a single “generic” exception class (typically a subclass of StandardError or RuntimeError) and have its other exception classes derive from that class. This allows the user to rescue the generic exception, thus catching all exceptions the library may raise even if future versions of the library add new exception subclasses.

So, 👍🏻 👍🏻

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
robertlaurin: I think that's fair, I'm open to alternative approaches here too.  I wonder how reasonable it is to just set the backtrace manually.  I want to avoid raising just to rescue though, seems wasteful (which was my original approach for the context class).

fbogsany: I think we should clearly document the `OpenTelemetry.error_handler` to explicitly state that the error **may** not have been generated as a result of `Kernel.raise` and **may** not have a backtrace set. One prominent use-case for a custom `error_handler` is to be able to ignore specific errors and minimize their performance impact. Raising and catching just to set a backtrace (or even explicitly setting the backtrace) is counterproductive for this use-case.

fbogsany (approved):
Let's add an issue to improve documentation of `OpenTelemetry.handle_error`. I think this PR should be merged as-is.

PR #874, opened 2021-07-13
Update docs to rely more on environment variable configuration

### LINKED ISSUE #872: Ensure documentation and examples reference configuration via ENV
We've closed #438 with the decision that we are largely comfortable with the Configurator API at the moment; and rather than spending more energy smoothing out the remaining rough edges we will update our documentation.

Specifically we want to ensure that people use ENV variables whenever possible; they are the happy path and well-documented upstream in the API.

ahayworth:
This PR updates our documentation in a few ways:
- Replaces some outdated configuration examples referencing a tracer factory
- Cleans up a reference to a non-existent LMDB example
- Sets the console span exporter for all current instrumentation examples
- Updates docs to reflect how to configure things via the environment whenever possible or reasonable.

Note that this means our examples may not flush to the console since the
BatchSpanProcessor is selected by default. I think this is okay,
honestly, but if it comes up we could figure out a way to select the
SimpleSpanProcessor (as some examples were doing) via the environment.

Fixes #872

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -29,11 +29,11 @@ Then, use the OpenTelemetry interfaces to produces traces and other telemetry da
 ```ruby
 require 'opentelemetry'
 
-# Obtain the current default tracer factory
-factory = OpenTelemetry.tracer_factory
+# Obtain the current default tracer provider
+provider = OpenTelemetry.tracer_provider
 
 # Create a trace
-tracer = factory.tracer('my_app', '1.0')
+tracer = provider.tracer('my_app', '1.0')
 
 # Record spans
 tracer.in_span('my_task') do |task_span|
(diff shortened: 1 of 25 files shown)


fbogsany:
> Note that this means our examples may not flush to the console since the
BatchSpanProcessor is selected by default. I think this is okay,
honestly, but if it comes up we could figure out a way to select the
SimpleSpanProcessor (as some examples were doing) via the environment.

The console exporter (when set via the environment variable) is wrapped in a SimpleSpanProcessor rather than a BSP: https://github.com/open-telemetry/opentelemetry-ruby/blob/f7cdc410365f69791ec6222a07d1b902274c6186/sdk/lib/opentelemetry/sdk/configurator.rb#L180

fbogsany (approved):
Beautiful cleanup, thank you!

Issue #530, opened 2020-12-23
Add GraphQL error handling.


robertlaurin:
GraphQL instrumentation currently provides poor visibility into failed GraphQL requests as the typical configuration is done in the form of a post which returns a 200 even if the query is malformed or fails for other reasons.

Some [exploration](https://github.com/open-telemetry/opentelemetry-ruby/pull/527/commits/caac3293a8ac5a34a7ccf8eff1788098d919d82f) was while [creating the instrumentation gem](https://github.com/open-telemetry/opentelemetry-ruby/pull/527), however it was removed to be addressed in a separate PR.

PR #876, opened 2021-07-14
proposal: add support for graphql errors

SomalianIvan:
https://github.com/open-telemetry/opentelemetry-ruby/issues/530

Implementing span error status based on [graphql spec](https://spec.graphql.org/June2018/#sec-Errors). I looked at other tracing implementations in https://graphql-ruby.org/api-doc/1.12.13/GraphQL/Tracing.html  and didn't see anything similar, but I think this would be helpful.

I also think logging `locations` and `path` if available should be added as attributes as well.

copy pastad @robertlaurin from https://github.com/open-telemetry/opentelemetry-ruby/pull/527/commits/6eb983d9be2d057ef27def65007507b20065a4d4 

graphql spec: 
```
Every error must contain an entry with the key message with a string description of the error intended for the developer as a guide to understand and correct the error.

If an error can be associated to a particular point in the requested GraphQL document, it should contain an entry with the key locations with a list of locations, where each location is a map with the keys line and column, both positive numbers starting from 1 which describe the beginning of an associated syntax element.

If an error can be associated to a particular field in the GraphQL result, it must contain an entry with the key path that details the path of the response field which experienced the error. This allows clients to identify whether a null result is intentional or caused by a runtime error.
```

example:
image
image


instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,20 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
...
-              yield
+            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
+               yield.tap do |response|
+                if key == 'validate' && !response[:errors]&.empty?
+                  span.status = OpenTelemetry::Trace::Status.error
+                  response[:errors].each do |error|
+                    span.add_event(
+                      "graphql error",
+                      attributes: {
+                        "error" => error.to_h.to_json
```
SomalianIvan, line 38: error.to_h might be a nested hash, https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes. To make sure we include everything, maybe report it as json string?

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,18 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
...
-            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |_span|
-              yield
+            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
+              yield.tap do |response|
+                if key == 'validate' && !response[:errors]&.empty?
+                  span.status = OpenTelemetry::Trace::Status.error
+                  response[:errors].each do |error|
+                    span.add_event(
+                      'graphql error',
+                      attributes: error.respond_to?(:to_h) ? { 'error' => error.to_h.to_json } : nil
```
rmosolgo, line 37: If you want greater granularity here, you could use `message`, `locations`, or `path`, which are described in the spec: https://spec.graphql.org/draft/#sec-Errors.Error-result-format

SomalianIvan:
https://github.com/open-telemetry/opentelemetry-js-contrib/blob/aec35d8a5e3d06c39db58c1ac7d20555163dc709/plugins/node/opentelemetry-instrumentation-graphql/src/enums/AttributeNames.ts#L24

https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L381

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
SomalianIvan: https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/error.rb

I think its ok to rely on `https://github.com/rmosolgo/graphql-ruby/blob/master/lib/graphql/static_validation/error.rb` `to_h`. It will include `message, locations, path`

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,18 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
           def platform_trace(platform_key, key, data)
             return yield if platform_key.nil?
 
-            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |_span|
-              yield
+            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
+              yield.tap do |response|
+                if key == 'validate' && !response[:errors]&.empty?
+                  span.status = OpenTelemetry::Trace::Status.error
```
robertlaurin, line 33: Remove setting error status as per sig.
```suggestion
```

SomalianIvan:
image

robertlaurin (approved):

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,17 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
           def platform_trace(platform_key, key, data)
             return yield if platform_key.nil?
 
-            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |_span|
-              yield
+            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
+              yield.tap do |response|
+                if key == 'validate' && !response[:errors]&.empty?
+                  response[:errors].each do |error|
```
fbogsany, line 33: If `response[:errors]` is `nil`, the `if` condition will be `true` and the `response[:errors].each` call will 💥
```
irb(main):001:0> x = {}
irb(main):002:0> !x[:foo]&.empty?
=> true
irb(main):003:0> x[:foo].each {|y| p y}
Traceback (most recent call last):
        4: from /opt/rubies/2.7.1/bin/irb:23:in `<main>'
        3: from /opt/rubies/2.7.1/bin/irb:23:in `load'
        2: from /opt/rubies/2.7.1/lib/ruby/gems/2.7.0/gems/irb-1.2.3/exe/irb:11:in `<top (required)>'
        1: from (irb):4
NoMethodError (undefined method `each' for nil:NilClass)
```

One option is to simply skip the test for `response[:errors]&.empty?` and use the safe navigation operator on the `response[:errors]&.each`.
```suggestion
                if key == 'validate'
                  response[:errors]&.each do |error|
```

However, see my next comment 😄

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,17 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
...
+                    span.add_event(
+                      'graphql.validation.error',
+                      attributes: error.respond_to?(:to_h) ? { 'message' => error.to_h.to_json } : nil
+                    )
```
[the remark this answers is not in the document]
SomalianIvan: The later approach is what [javascript folks have chosen to do](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L381).

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,24 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
           def platform_trace(platform_key, key, data)
             return yield if platform_key.nil?
 
-            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |_span|
-              yield
+            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
+              yield.tap do |response|
+                if key == 'validate' &&
+                   response.include?(:errors) &&
```
SomalianIvan, line 33: not sure how to make this less verbose ;(

instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
```diff
@@ -27,8 +27,23 @@ class GraphQLTracer < ::GraphQL::Tracing::PlatformTracing
...
+                    error&.to_h&.to_json
+                  end
+                  errors.compact!
+                  span.add_event(
+                    'graphql.validation.error',
+                    attributes: {
+                      'message' => errors
+                    }
+                  )
+                end
```
fbogsany, line 45:
```suggestion
                errors = response[:errors]&.compact&.map { |e| e.to_h.to_json } if key == 'validate'
                unless errors.nil?
                  span.add_event(
                    'graphql.validation.error',
                    attributes: {
                      'message' => errors
                    }
                  )
                end
```

instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb
```diff
@@ -154,6 +154,21 @@
...
+        <<-GRAPHQL
+          {
+            nonExistentField
+          }
+        GRAPHQL
+      )
+      span = spans.find { |s| s.name == 'graphql.validate' }
+      event = span.events.find { |e| e.name == 'graphql.validation.error' }
+      _(event.attributes['message']).must_equal(
+        ["{\"message\":\"Field 'nonExistentField' doesn't exist on type 'Query'\",\"locations\":[{\"line\":2,\"column\":13}],\"path\":[\"query\",\"nonExistentField\"],\"extensions\":{\"code\":\"undefinedField\",\"typeName\":\"Query\",\"fieldName\":\"nonExistentField\"}}"]
```
arielvalentin, line 169: I am interested in hearing from spec SIG, vendors, and other SDK implementors about using escaped JSON text in event attribute values. It seems like there will be a bit of post processing required here to represent the event attributes.

I see e.g. in the spec when using arrays or dictionaries they recommend breaking out key-value pairs into jQuery like `application/x-www-form-urlencoded`:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attributes

So maybe the attributes would be something like:

```
errors[0].message =  "Field 'nonExistentField' doesn't exist on type 'Query'"
errors[0].locations[0][line] = 2
# ...
``` 

Thoughts?

SomalianIvan: The [JS instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/e8c207b40faa4e2de1bac95d4b9c83f0833ee455/plugins/node/opentelemetry-instrumentation-graphql/src/instrumentation.ts#L381) uses JSON string. If we supply a string with json list, jaeger at the very least pretty prints it.
image

SomalianIvan:
```diff
commit 388d208c67e0deb6c2063934ec02450a9992ec63 (HEAD -> add-errors-events-to-graphql-instrumentation, origin/add-errors-events-to-graphql-instrumentation)
Author: Timur Borkhodoev <tim.borkhodoev@shopify.com>
Date:   Mon Aug 9 13:24:25 2021 -0400

    wip: better

────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
modified: instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
@ instrumentation/graphql/lib/opentelemetry/instrumentation/graphql/tracers/graphql_tracer.rb:32 @ module OpenTelemetry

            tracer.in_span(platform_key, attributes: attributes_for(key, data)) do |span|
              yield.tap do |response|
                errors = response[:errors]&.compact&.map { |e| e.to_h.to_json } if key == 'validate'
                errors = response[:errors]&.compact&.map { |e| e.to_h }&.to_json if key == 'validate'
                unless errors.nil?
                  span.add_event(
                    'graphql.validation.error',
```

fbogsany (approved):

robertlaurin (approved):

PR #878, opened 2021-07-15
refactor: split out actionpack from rails gem

robertlaurin:
Todo:
- [x] Separate Action Pack from the main rails instrumentation gem 
- [x] Manual tests (run this branch on couple rails apps)
- [x] Add some example files for Action Pack and Active Record

Followup improvements (not for this PR):
- Add config options for opting out of any sub gem instrumentation.
- Have the tests in the rails gem be more comprehensive (test Action View, Active Record).
- Split Active Support Notifications out of the Action View instrumentation and into its own proper gem.


instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
```diff
@@ -13,34 +13,29 @@ module Rails
       # instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         install do |_config|
+          OpenTelemetry::Instrumentation::ActiveRecord::Instrumentation.instance.install({})
+          OpenTelemetry::Instrumentation::ActionPack::Instrumentation.instance.install({})
```
ahayworth, line 17: What kind of final, end-state did you have in mind?

Personally I think we should go one of two ways:
- Install all rails-related gems by default (and provide an option that allows you to selectively turn them off)
- Install no rails-related gems by default (and provide an option that allows you to choose which to enable)

Of those two, I would actually prefer the first one. My rationale, over all, is that we probably don't want to be in the business of telling people precisely which parts of rails they want instrumented, because we can't really make a good judgement call there. For example, why ActiveRecord and not ActionView? (Not that this is a particularly bad choice you've made, just using it as an example).

robertlaurin: > Of those two, I would actually prefer the first one.

I'm glad, that's the direction I had planned for this :) 

> My rationale, over all, is that we probably don't want to be in the business of telling people precisely which parts of rails they want instrumented, because we can't really make a good judgement call there.

Yup, let people turn off the bits they are not interested in.

robertlaurin:
I need to add some missing examples, and do some manual testing against rails apps to validate that the initializers and boot order have not become problematic with this shuffle.  

The code itself has primarily been shuffled around, ready for review.

robertlaurin:
Using a typical `config/initializers/otel.rb` to configure the sdk is working as well.

```rb
require 'opentelemetry/exporter/jaeger'

instrumentation_config = {
  'OpenTelemetry::Instrumentation::ActionPack' => { enable_recognize_route: true },
}

  c.use_all(instrumentation_config)
  c.service_name = 'opentelemetry_ruby_example'
  c.add_span_processor(OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(
    OpenTelemetry::Exporter::Jaeger::AgentExporter.new(
      max_packet_size: 9 * 1024
    )
  ))
end
```

Also confirmed it it respecting config options passed in.

image

robertlaurin:
Just tested again on a real application with everything working as expected.

ahayworth (approved):

PR #880, opened 2021-07-15
refactor: split out active support from rails gem

robertlaurin:
Splitting apart the rails component instrumentation dependencies from the rails instrumentation gem itself.

Similar to https://github.com/open-telemetry/opentelemetry-ruby/pull/878


fbogsany (approved):

ahayworth (approved):
:shipit: :shipit: :shipit:

fbogsany:
I marked this as a breaking change because we've moved config options from one gem to another.

robertlaurin:
The gem was never released but, works for me :)

PR #887, opened 2021-07-23
Update CODEOWNERS

theletterf:
Updating CODEOWNERS to set docs team as reviewers for docs files.

This change was approved and merged to GDI specs in https://github.com/signalfx/gdi-specification/pull/115/files


ericmustin:
@theletterf It's possible i missed some comms but I think this got pointed at the wrong repo, I assume there's a splunk distro this should be for?

theletterf:
Ha! You're totally right. Let's close this one. :)

PR #888, opened 2021-07-26
prune invalid links

SomalianIvan:
closes https://github.com/open-telemetry/opentelemetry-ruby/issues/735

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -331,15 +331,13 @@ module OpenTelemetry
           # Fast path (likely) common cases.
           return nil if links.nil?
 
-          links.filter! { |link| link.span_context.valid? }
-
           if links.size <= link_count_limit &&
              links.all? { |link| link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
             return links.frozen? ? links : links.clone.freeze
           end
 
           # Slow path: trim attributes for each Link.
-          links.last(link_count_limit).map! do |link|
+          links.select { |link| link.span_context.valid? }.last(link_count_limit).map! do |link|
             attrs = Hash[link.attributes] # link.attributes is frozen, so we need an unfrozen copy to adjust.
             attrs.keep_if { |key, value| Internal.valid_key?(key) && Internal.valid_value?(value) }
             excess = attrs.size - link_attribute_count_limit
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -332,7 +332,7 @@ module OpenTelemetry
           return nil if links.nil?
 
           if links.size <= link_count_limit &&
-             links.all? { |link| link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
+             links.all? { |link| link.span_context.valid? && link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
             return links.frozen? ? links : links.clone.freeze
           end
 
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -338,7 +338,8 @@ module OpenTelemetry
 
           # Slow path: trim attributes for each Link.
           valid_links = links.select { |link| link.span_context.valid? }
-          valid_links.shift(links.size - link_count_limit) if valid_links.size > link_count_limit
+          excess_link_count = valid_links.size - link_count_limit
+          valid_links.pop(excess_link_count) if excess_link_count.positive?
           valid_links.map! do |link|
             attrs = Hash[link.attributes] # link.attributes is frozen, so we need an unfrozen copy to adjust.
             attrs.keep_if { |key, value| Internal.valid_key?(key) && Internal.valid_value?(value) }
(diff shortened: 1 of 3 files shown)


sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -331,6 +331,8 @@ def trim_links(links, link_count_limit, link_attribute_count_limit) # rubocop:di
           # Fast path (likely) common cases.
           return nil if links.nil?
 
+          links.filter! { |link| link.span_context.valid? }
```
fbogsany, line 334: Technically, we shouldn't modify the `links` array. It is handed to us by the caller (well, a few levels up the stack) and they could modify it subsequently. On the fast(er) path, we check the content and either use it as is (if already frozen) or clone & freeze the array. On the slow path, we'll create a new array before modifying its contents.

All of which to say, I think we should not `filter!` here - instead we should validate on the fast(er) path below (I'll show what I mean) and then `filter` before `last` on the slow path.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -331,6 +331,8 @@ def trim_links(links, link_count_limit, link_attribute_count_limit) # rubocop:di
...
+
           if links.size <= link_count_limit &&
              links.all? { |link| link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
```
SomalianIvan, line 337:
```suggestion
             links.all? { |link| link.span_context.valid? && link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }

```

sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -446,5 +446,13 @@
                       [], nil, links, Time.now, nil, nil)
       _(span.links.size).must_equal(1)
     end
+
+    it 'prunes invalid links' do
+      invalid_context = OpenTelemetry::Trace::SpanContext.new(trace_id: OpenTelemetry::Trace::INVALID_TRACE_ID)
+      links = [OpenTelemetry::Trace::Link.new(invalid_context)]
+      span = Span.new(context, Context.empty, 'name', SpanKind::INTERNAL, nil, span_limits,
+                      [], nil, links, Time.now, nil, nil)
+      _(span.links.size).must_equal(0)
```
arielvalentin, line 455: Consider using predicate assertions since it will print more details about the contents of the `span.links` array as part of the error message:

```suggestion
      _(span.links).must_be_empty
```

It may also be helpful, but not required, to include some details in the test to ensure the `links` parameter is not mutated per @fbogsany comment above. E.g. check that the links array is unchanged or use a frozen array as an argument.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -332,12 +332,12 @@ def trim_links(links, link_count_limit, link_attribute_count_limit) # rubocop:di
...
 
           if links.size <= link_count_limit &&
-             links.all? { |link| link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
+             links.all? { |link| link.span_context.valid? && link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
             return links.frozen? ? links : links.clone.freeze
           end
 
           # Slow path: trim attributes for each Link.
-          links.last(link_count_limit).map! do |link|
+          links.select { |link| link.span_context.valid? }.last(link_count_limit).map! do |link|
```
fbogsany, line 340: I can't see a clean way to only allocate one array here, unfortunately. 🤔 I guess we could do the equivalent of `excess.times { links.shift }` after the `links.select`.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -332,12 +332,14 @@ def trim_links(links, link_count_limit, link_attribute_count_limit) # rubocop:di
...
           if links.size <= link_count_limit &&
-             links.all? { |link| link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
+             links.all? { |link| link.span_context.valid? && link.attributes.size <= link_attribute_count_limit && Internal.valid_attributes?(name, 'link', link.attributes) }
             return links.frozen? ? links : links.clone.freeze
           end
 
           # Slow path: trim attributes for each Link.
-          links.last(link_count_limit).map! do |link|
+          valid_links = links.select { |link| link.span_context.valid? }
+          valid_links.shift(links.size - link_count_limit) if valid_links.size > link_count_limit
```
fbogsany, line 341: Technically, this still allocates (and throws away) an array 😞

fbogsany, line 341:
```suggestion
          valid_links.shift(valid_links.size - link_count_limit) if valid_links.size > link_count_limit
```

fbogsany, line 341: There's no expectation of *which* links are dropped, if too many are provided. It is more efficient to drop the end of the array (because we don't need to move anything down).
```suggestion
          valid_links = links.select { |link| link.span_context.valid? }
          excess_link_count = valid_links.size - link_count_limit
          valid_links.pop(excess_link_count) if excess_link_count.positive?
```

fbogsany: It's on a (very) slow path, though, so let's not worry about the extra allocation.

fbogsany (approved):

PR #891, opened 2021-08-03
feat: add configuration options for dalli

SomalianIvan:
part of #736

### DIFF
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
@@ -14,10 +14,10 @@ module OpenTelemetry
             operation = Utils.opname(op, multi?)
             attributes = {
               'db.system' => 'memcached',
-              'db.statement' => Utils.format_command(operation, args),
               'net.peer.name' => hostname,
               'net.peer.port' => port
             }
+            attributes['db.statement'] = Utils.format_command(operation, args) if config[:db_statement] == :include
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
             tracer.in_span(operation, attributes: attributes, kind: :client) do
               super
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
@@ -20,6 +20,7 @@ module OpenTelemetry
         end
 
         option :peer_service, default: nil, validate: :string
+        option :db_statement, default: :include, validate: ->(opt) { %I[omit include].include?(opt) }
 
         private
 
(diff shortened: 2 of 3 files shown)


instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```diff
@@ -14,9 +14,9 @@ def request(op, *args)
             operation = Utils.opname(op, multi?)
             attributes = {
               'db.system' => 'memcached',
-              'db.statement' => Utils.format_command(operation, args),
               'net.peer.name' => hostname,
-              'net.peer.port' => port
+              'net.peer.port' => port,
+              **db_statement(operation, args),
             }
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
```
robertlaurin, line 21: The double splat there doesn't idiomatic, I totally believe it works but it's awkward looking :)
```suggestion
            }
            attributes['db.statement'] = db_statement if config[:db_statement] == :include || config[:db_statement] == :obfuscate
            attributes['peer.service'] = config[:peer_service] if config[:peer_service]
```

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```diff
@@ -26,6 +26,25 @@ def request(op, *args)
...
+
+          def db_statement(operation, args)
+            placeholder = "#{operation} BLOB (OMITTED)"
+            case config[:db_statement]
+            when :omit
+              command = placeholder
+            when :obfuscate
+              unsanitized_command = [operation, *args].join(' ').strip
+              command = OpenTelemetry::Common::Utilities.utf8_encode(unsanitized_command, binary: true, placeholder: placeholder)
+            end
```
robertlaurin, line 39: Sanitizing the statement is not the same as obfuscation.  I'd encourage you to test this by running a test and see if any of the values you set show up in the resulting span.  

Instead of taking on adding obfuscation here, I'd encourage you to simply add the `omit` and `include` options for now.

SomalianIvan: so this logic should be under `include` flag then. Note I mostly  copy pastad from https://github.com/SomalianIvan/opentelemetry-ruby/blob/7667c3c8e969b74627c5c1afc21847e156c4d6ab/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/utils.rb#L47-L55

robertlaurin: Right if it is included, then run the format code.  Otherwise do no work and do not add the `db.statement` attribute to the span.

robertlaurin: We can punt obfuscation to a follow up PR/issue.

instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
```diff
@@ -14,10 +14,10 @@ def request(op, *args)
             operation = Utils.opname(op, multi?)
             attributes = {
               'db.system' => 'memcached',
-              'db.statement' => Utils.format_command(operation, args),
               'net.peer.name' => hostname,
-              'net.peer.port' => port
+              'net.peer.port' => port,
             }
+            attributes['db.statement'] = db_statement if config[:db_statement] == :include
```
robertlaurin, line 20: I think if you do the following, you can remove the additional code below and then you're just left with adding a test case.  
```suggestion
            attributes['db.statement'] = Utils.format_command(operation, args) if config[:db_statement] == :include
```

robertlaurin (approved):

PR #894, opened 2021-08-04
feat: add option for db.statement

SomalianIvan:
part of #736


instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
```diff
@@ -20,7 +20,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         option :peer_service, default: nil, validate: :string
-        option :enable_sql_obfuscation, default: false, validate: :boolean
+        option :db_statement, default: :include, validate: ->(opt) { %I[omit include obfuscate].include?(opt) }
```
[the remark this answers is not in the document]
SomalianIvan: > I know we're not obligated to do so as we're not 1.0, but having someone bump this gem without watching the changelogs could put them in a position where all of a sudden they are emitting PII and secrets.

👍  Ill try to add something inside install clause

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -53,9 +53,7 @@ module Client
           def query(sql, options = {})
             tracer.in_span(
               database_span_name(sql),
-              attributes: client_attributes.merge(
-                'db.statement' => obfuscate_sql(sql)
-              ),
+              attributes: %I[include obfuscate].include?(config[:db_statement]) ? client_attributes.merge('db.statement' => obfuscate_sql(sql)) : client_attributes,
```
arielvalentin, line 56: 💭 It looks like we are evaluating this configuration every time we execute this query. 

How would we feel about memoizing this value or evaluating this at configuration time?

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -65,7 +63,7 @@ def query(sql, options = {})
           private
 
           def obfuscate_sql(sql)
-            return sql unless config[:enable_sql_obfuscation]
+            return sql if config[:db_statement] == :include
```
arielvalentin, line 66: Now that there is a check on L56, is this check redundant?

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
```diff
@@ -133,10 +133,47 @@
       assert(!span.events.first.attributes['exception.stacktrace'].nil?)
     end
 
+    describe 'when db_statement set as obfuscate' do
```
arielvalentin, line 136: I think we may also want at least one more check here where for some reason someone sets both the legacy attribute and the new one. _Just in case_

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
SomalianIvan: 🤔  I think the check on line 56, if its include or obfuscate it would call `obfuscate_sql(sql)` which will check the flag again. In this state its a necessary check. However I see how checking the same flag twice is a code smell.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
SomalianIvan: I will be an exercise for me to memoize and run a simple benchmark on both approaches 👍

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
SomalianIvan: Ill try to refactor, see if I can do better

instrumentation/mysql2/test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb
SomalianIvan: Would need to be able to tell then when something is a default, vs user chose default value. I might have to remove https://github.com/open-telemetry/opentelemetry-ruby/pull/894/files#diff-335a880214cc43314c9840b69feddb39868a51db2dda1ffb126094486f6ec39bL23 the default setting and handle it inside the `install`, to be able to differentiate between the two.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
```diff
@@ -20,7 +27,8 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         option :peer_service, default: nil, validate: :string
-        option :enable_sql_obfuscation, default: false, validate: :boolean
+        option :enable_sql_obfuscation, default: nil, validate: :boolean
```
robertlaurin, line 30: Does the default need to change here for the deprecation warning to work?

SomalianIvan: that's an oopsie, I wanted to warn when enable_obfuscation is supplied. Let me fix it up.

SomalianIvan: https://github.com/open-telemetry/opentelemetry-ruby/pull/894/commits/f02423a7799f7c18dd8ca9d2c2b6981ad4dc7bca

fbogsany: Would it be unreasonable to add a `deprecated_option` helper to make this clearer to consumers (and, perhaps, to simplify option deprecation for instrumentation authors)? E.g.:
```suggestion
        deprecated_option :enable_sql_obfuscation, default: nil, validate: :boolean, alternative: :db_statement
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
```diff
@@ -10,7 +10,15 @@ module Mysql2
...
       class Instrumentation < OpenTelemetry::Instrumentation::Base
-        install do |_config|
+        install do |config|
+          if config[:enable_sql_obfuscation]
+            config[:db_statement] = :obfuscate
+            OpenTelemetry.logger.warn(
+              'Instrumentation mysql2 configuration option enable_sql_obfuscation has been deprecated,' \
+              'use db_statement option instead'
+            )
+          end
```
robertlaurin, line 20: Tim and I discussed this a bit on a call, I encouraged taking this approach where we would let the legacy option override the `db_statement` option if it was set.

The reason for this is that it is defensive towards someone who bumps the gem without checking the changelog.  If someone sets both, we'll assume they probably wanted to keep the obfuscation enabled and will get the warning.

We'll leave it in for a version or two and then remove the setting entirely.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
robertlaurin: >Would it be unreasonable to add a deprecated_option helper to make this clearer to consumers (and, perhaps, to simplify option deprecation for instrumentation authors)? E.g.:

🤔 , I think adding something like this makes sense.  I'd want to see it come in as a follow up PR on its own where we can bikeshed the details of it.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
fbogsany: 🤔 if they have the new option set, it most likely means that they added it explicitly and forgot to remove the old one. Anyway, this approach is the safest, so 👍

fbogsany: Also, I guess we can't distinguish between "default value" and "explicitly set to default value"?

fbogsany (approved):

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/instrumentation.rb
robertlaurin: Definitely focusing on safety here.

We should be able to distinguish between default and explicitly set, those changes would require pushing the deprecation support into the instrumentation base as you suggested above.

https://github.com/open-telemetry/opentelemetry-ruby/blob/9e86944a0347a63b393246cf94d585e04e17b539/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L267-L277

robertlaurin (approved):

PR #895, opened 2021-08-04
fix: rails instrumentation to not explicitly install sub gems

robertlaurin:
> Add config options for opting out of any sub gem instrumentation.

Follow up from https://github.com/open-telemetry/opentelemetry-ruby/pull/878

### DIFF
--- a/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
+++ b/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
@@ -14,19 +14,12 @@ module OpenTelemetry
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('5.2.0')
 
-        install do |_config|
-          OpenTelemetry::Instrumentation::ActionPack::Instrumentation.instance.install({})
-          OpenTelemetry::Instrumentation::ActionView::Instrumentation.instance.install({})
-          OpenTelemetry::Instrumentation::ActiveRecord::Instrumentation.instance.install({})
-        end
-
-        present do
-          defined?(::Rails)
-        end
-
-        compatible do
-          gem_version >= MINIMUM_VERSION
-        end
+        # This gem requires the instrumentantion gems for the different
+        # components of Rails, as a result it does not have any explicit
+        # work to do in the install step.
+        install { true }
+        present { defined?(::Rails) }
+        compatible { gem_version >= MINIMUM_VERSION }
 
         private
 
--- a/instrumentation/action_pack/README.md
+++ b/instrumentation/action_pack/README.md
@@ -30,6 +30,17 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
+### Configuration options
+
+The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/v6.1.3/actionpack/lib/action_dispatch/journey/router.rb#L65)
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::ActionPack', {
+    enable_recognize_route: true
+  }
+end
+```
+
 ## Examples
 
 Example usage can be seen in the `./example/trace_demonstration.rb` file [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/action_pack/example/trace_demonstration.ru)
--- a/instrumentation/rails/README.md
+++ b/instrumentation/rails/README.md
@@ -14,15 +14,7 @@ Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-r
 
 ## Usage
 
-To use the instrumentation, call `use` with the name of the instrumentation:
-
-```ruby
-OpenTelemetry::SDK.configure do |c|
-  c.use 'OpenTelemetry::Instrumentation::Rails'
-end
-```
-
-Alternatively, you can also call `use_all` to install all the available instrumentation.
+To use the Rails instrumentation, call `use_all` so it installs all the instrumentation gems.
 
 ```ruby
 OpenTelemetry::SDK.configure do |c|
@@ -30,15 +22,14 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
-
 ### Configuration options
 
-The `http.route` attribute is disabled by default because we use [.recognize](https://github.com/rails/rails/blob/v6.1.3/actionpack/lib/action_dispatch/journey/router.rb#L65)
+The Rails instrumentation attempts to mirror the structure of the Ruby on Rails.  It is a collection of instrumentation gems for components of Rails such as Action View, Active Record, Action Pack, etc...
+
+You may want to include all of the Rails instrumentation but disable a single instrumentation gem that it includes.  Here is an example of how you can disable Active Record when using this instrumentation gem.
 ```ruby
 OpenTelemetry::SDK.configure do |c|
-  c.use 'OpenTelemetry::Instrumentation::Rails', {
-    enable_recognize_route: true
-  }
+  c.use_all({ 'OpenTelemetry::Instrumentation::ActiveRecord' => { enabled: false } })
 end
 ```
 
--- a/instrumentation/rails/test/opentelemetry/instrumentation/rails/instrumentation_test.rb
+++ b/instrumentation/rails/test/opentelemetry/instrumentation/rails/instrumentation_test.rb
@@ -7,7 +7,7 @@
 require 'test_helper'
 
 describe OpenTelemetry::Instrumentation::Rails::Instrumentation do
-  it 'adds the tracing middleware' do
+  it 'adds the rack tracing middleware' do
     _(DEFAULT_RAILS_APP.config.middleware).must_include OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware
   end
 end
--- a/instrumentation/rails/test/test_helper.rb
+++ b/instrumentation/rails/test/test_helper.rb
@@ -20,7 +20,7 @@ EXPORTER = OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter.new
 span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(EXPORTER)
 
 OpenTelemetry::SDK.configure do |c|
-  c.use 'OpenTelemetry::Instrumentation::Rails'
+  c.use_all
   c.add_span_processor span_processor
 end
 


instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
```diff
@@ -14,19 +14,12 @@ module Rails
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('5.2.0')
 
-        install do |_config|
-          OpenTelemetry::Instrumentation::ActionPack::Instrumentation.instance.install({})
-          OpenTelemetry::Instrumentation::ActionView::Instrumentation.instance.install({})
-          OpenTelemetry::Instrumentation::ActiveRecord::Instrumentation.instance.install({})
-        end
```
robertlaurin, line 21: This change is required as with it being present you're unable to disable the via config being passed in.  It's more consistent with the `all` instrumentation gem than any of the other ones so I think that's ok.  Happy to bikeshed on this.

arielvalentin: If this is no longer auto-installing sub-gems, should it also remove requires?

https://github.com/open-telemetry/opentelemetry-ruby/blob/fix-remove-explicit-installs/instrumentation/rails/lib/opentelemetry/instrumentation/rails.rb#L18

ahayworth (commented):
I have some undirected thoughts that this has provoked, @robertlaurin - curious what you think on all of it.

---

I think I understand where we're going with this - correct me if I'm wrong:
- `-instrumentation-rails` is now more like `-instrumentation-all` (but for Rails)
- We expect (based on this PR) that users of Rails will install `-instrumentation-rails` and then call `c.use_all` in their SDK configuration block.
- If folks only want a specific subset, they still call `c.use_all` but pass `enabled: false` as an option for the specific Rails instrumentation they do *not* want enabled

I think that as an outsider coming to the instrumentation for the first time, I would expect that I could do:

```ruby
  c.use 'OpenTelemetry::Instrumentation::Rails'
end
```

...and for that to load all parts of the Rails instrumentation. I don't think this approach is necessarily a bad idea at all, but it's a little different than how I had imagined. I think I had imagined something like being able to do this if you wanted to tweak the rails instrumentation components:

```ruby
  c.use 'OpenTelemetry::Instrumentation::Rails', {
    action_pack: { enable: false },
    active_job: { job_class_span_names_or_whatever_we_called_it: :sure_i_guess },
  }
end
```

---

I think what I'm running into, personally, is that there seems to be a relatively unstated preference in the instrumentation parts of the SDK that you will be using all instrumentation loaded, almost all of the time. I think that's actually probably okay? But we should probably make this preference explicit somehow. It feels like an assumption in the repo that we should make plain, so that people know "if you choose _not_ to use `c.use_all`, you're on your own for things".

I think there are some docs that just need to be written - I feel like they're taking shape in my head. Almost like an "end-users guide to otel-ruby" or something...

---

I wonder if this second layer of abstraction will make it more difficult for wrapper gems to insert "required options" that they might desire? For example, GitHub's railtie around all of this stuff requires that you pass in a hash of instrumentations that you want and the options you want; and if we see you've passed in one that needs an option set (for whatever reason we deem necessary) we merge that in over whatever options you passed ... and then configure the sdk. Something like:

```ruby
# in the railtie
REQUIRED_DEFAULTS = {
  'OpenTelemetry::Instrumentation::Foo' => { some_important_opt: true },
}

# I don't remember how we get the opts from configuration precisely
# My last day was yesterday so I can't look anymore, but pretend that
# `opts` is an ActiveSupport ordered options thing with user-provided stuff
instrumentation_with_opts = opts.instrumentation.merge(REQUIRED_DEFAULTS)

# In the railtie, we ran all the configuration for the user
  instrumentation_with_opts.each do |name, opts|
    c.use name, opts
  end
end
```

That's obviously not the only way to write such a helper - and if memory serves me correct (it may not, honestly) I think yours works differently?

I wonder if maybe that just highlights the fact that we should expose something in the instrumentation base class - perhaps a way to globally override options, or something... :hmm:

---

robertlaurin:
> I think I understand where we're going with this - correct me if I'm wrong:
> 
> * `-instrumentation-rails` is now more like `-instrumentation-all` (but for Rails)

That's what I'm suggesting, I think it makes sense but it's worth going through the implications of it.  Instrumentation all requires the use of `c.use_all`, and that's probably the biggest consideration to make here.

> * We expect (based on this PR) that users of Rails will install `-instrumentation-rails` and then call `c.use_all` in their SDK configuration block.
> * If folks only want a specific subset, they still call `c.use_all` but pass `enabled: false` as an option for the specific Rails instrumentation they do _not_ want enabled

Yes to both points.

> I think that as an outsider coming to the instrumentation for the first time, I would expect that I could do:
> 
> ```ruby
> OpenTelemetry::SDK.configure do |c|
>   c.use 'OpenTelemetry::Instrumentation::Rails'
> end
> ```
> 
> ...and for that to load all parts of the Rails instrumentation. I don't think this approach is necessarily a bad idea at all, but it's a little different than how I had imagined. I think I had imagined something like being able to do this if you wanted to tweak the rails instrumentation components:
> 
> ```ruby
> OpenTelemetry::SDK.configure do |c|
>   c.use 'OpenTelemetry::Instrumentation::Rails', {
>     action_pack: { enable: false },
>     active_job: { job_class_span_names_or_whatever_we_called_it: :sure_i_guess },
>   }
> end
> ```

I did originally consider that approach, but I questioned if it was actually simpler/clearer.  It adds a (small) layer of indirection, and requires the rails instrumentation to handle mapping and passing a long additional configs, though maybe that's really not a big deal.  

I honestly have not yet found a strong inclination towards either approach.  If we do the above I believe it prevents users from passing in their configs via `c.use_all`, so we will have to settle on one direction.  

> I think what I'm running into, personally, is that there seems to be a relatively unstated preference in the instrumentation parts of the SDK that you will be using all instrumentation loaded, almost all of the time. I think that's actually probably okay? But we should probably make this preference explicit somehow. It feels like an assumption in the repo that we should make plain, so that people know "if you choose _not_ to use `c.use_all`, you're on your own for things".
> 
> I think there are some docs that just need to be written - I feel like they're taking shape in my head. Almost like an "end-users guide to otel-ruby" or something...

Agreed, this feels bit more like a documentation communication challenge.


> I wonder if this second layer of abstraction will make it more difficult for wrapper gems to insert "required options" that they might desire? For example, GitHub's railtie around all of this stuff requires that you pass in a hash of instrumentations that you want and the options you want; and if we see you've passed in one that needs an option set (for whatever reason we deem necessary) we merge that in over whatever options you passed ... and then configure the sdk. Something like:
> 
> ```ruby
> # in the railtie
> REQUIRED_DEFAULTS = {
>   'OpenTelemetry::Instrumentation::Foo' => { some_important_opt: true },
> }
> 
> # I don't remember how we get the opts from configuration precisely
> # My last day was yesterday so I can't look anymore, but pretend that
> # `opts` is an ActiveSupport ordered options thing with user-provided stuff
> instrumentation_with_opts = opts.instrumentation.merge(REQUIRED_DEFAULTS)
> 
> # In the railtie, we ran all the configuration for the user
> OpenTelemetry::SDK.configure do |c|
>   instrumentation_with_opts.each do |name, opts|
>     c.use name, opts
>   end
> end
> ```
> 
> That's obviously not the only way to write such a helper - and if memory serves me correct (it may not, honestly) I think yours works differently?

Ours is a bit different which is what probably influence this decision.

It looks something like this.
```rb
      instrumentation_config = {
        'OpenTelemetry::Instrumentation::Redis' => { peer_service: "#{service_name}:redis", trace_root_spans: false },
        'OpenTelemetry::Instrumentation::Mysql2' => { peer_service: "#{service_name}:sql" },
        'OpenTelemetry::Instrumentation::Sidekiq' => { span_naming: :job_class },
        'OpenTelemetry::Instrumentation::Rack' => { untraced_endpoints: ['/services/ping'] },
      }

        c.use_all(instrumentation_config)
      end
 ```

> 
> I wonder if maybe that just highlights the fact that we should expose something in the instrumentation base class - perhaps a way to globally override options, or something... :hmm:

🤔

File: instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb (hunk 1 above)
Comment (robertlaurin) [reply instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb:21]: I think we want it to still require the instrumentation sub components to make sure that the instrumentation hooks get called.  We do something similar in [`instrumentation-all`](https://github.com/open-telemetry/opentelemetry-ruby/blob/9e86944a0347a63b393246cf94d585e04e17b539/instrumentation/all/lib/opentelemetry/instrumentation/all.rb#L7-L35) which is where I pulled my inspiration from.

**Comment by robertlaurin** [comment]
So if we took an approach like this, which is what I think you're suggesting.
```diff
diff --git a/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb b/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
index 3ac7030c..5ba22c01 100644
--- a/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
+++ b/instrumentation/rails/lib/opentelemetry/instrumentation/rails/instrumentation.rb
@@ -14,17 +14,23 @@ module OpenTelemetry
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('5.2.0')
 
-        # This gem requires the instrumentantion gems for the different
-        # components of Rails, as a result it does not have any explicit
-        # work to do in the install step.
-        install { true }
+        install do |config|
+          OpenTelemetry::Instrumentation::ActionPack::Instrumentation.instance.install(config[:action_pack])
+          OpenTelemetry::Instrumentation::ActionView::Instrumentation.instance.install(config[:action_view])
+          OpenTelemetry::Instrumentation::ActiveRecord::Instrumentation.instance.install(config[:active_record])
+        end
+
         present { defined?(::Rails) }
         compatible { gem_version >= MINIMUM_VERSION }
 
+        option :action_pack,   default: {}, validate: :hash
+        option :action_view,   default: {}, validate: :hash
+        option :active_record, default: {}, validate: :hash
+
         private
 
         def gem_version
-          Gem.loaded_specs['actionpack'].version
+          Gem.loaded_specs['rails'].version
         end
       end
     end
```


### Assume rails instrumentation is present
#### Configuring it looks like this
```rb
instrumentation_all_config = {
  'OpenTelemetry::Instrumentation::Rails' => {
    'action_pack' => { enabled: false },
    'action_view' => { disallowed_notification_payload_keys: ['abc'] }
  },
}

  c.use_all(instrumentation_all_config)
end
```
#### or this
```rb
  c.use 'OpenTelemetry::Instrumentation::Rails', {
    action_pack: { enable: false },
    active_job: { job_class_span_names_or_whatever_we_called_it: :sure_i_guess },
  }
end
```

### Assume rails instrumentation is not present but the sub gems are
#### Configuring it looks like this
```rb
instrumentation_all_config = {
  'OpenTelemetry::Instrumentation::ActiveRecord' => {},
  'OpenTelemetry::Instrumentation::ActionPack' => { enabled: false},
  'OpenTelemetry::Instrumentation::ActiveView' => { disallowed_notification_payload_keys: ['abc'] },
}

  c.use_all(instrumentation_all_config)
end
```

or this
```rb

  c.use 'OpenTelemetry::Instrumentation::ActiveRecord', {}
  c.use 'OpenTelemetry::Instrumentation::ActiveView', { disallowed_notification_payload_keys: ['abc'] }
end
```

This is probably clearer than what is currently committed PR.  If you use rails instrumentation you have to pass config opts through rails, if you're not using it you pass the config opts directly.

Seems good to me.  @ahayworth wdyt?

ahayworth (approved):
As discussed in spec SIG, I think this is totally fine to go forward with as-is; I opened #898 and #899 for future evolution of how we handle configuration. 😄

robertlaurin:
Waiting for CI before bringing this in image

PR #905, opened 2021-08-11
feat: add db_statement toggle for postgres

SomalianIvan:
part of #736

### DIFF
--- a/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb
+++ b/instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb
@@ -11,7 +11,23 @@ module OpenTelemetry
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('1.1.0')
 
-        install do |_config|
+        install do |config|
+          if config[:enable_sql_obfuscation]
+            config[:db_statement] = :obfuscate
+            OpenTelemetry.logger.warn(
+              'Instrumentation pg configuration option enable_sql_obfuscation has been deprecated,' \
+              'use db_statement option instead'
+            )
+          end
+
+          unless config[:enable_statement_attribute]
+            config[:db_statement] = :omit
+            OpenTelemetry.logger.warn(
+              'Instrumentation pg configuration option enable_sql_obfuscation has been deprecated,' \
+              'use db_statement option instead'
+            )
+          end
+
           require_dependencies
           patch_client
         end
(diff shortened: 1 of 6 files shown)


instrumentation/pg/lib/opentelemetry/instrumentation/pg/instrumentation.rb
```diff
@@ -11,7 +11,23 @@ module PG
...
+            OpenTelemetry.logger.warn(
+              'Instrumentation pg configuration option enable_sql_obfuscation has been deprecated,' \
+              'use db_statement option instead'
+            )
+          end
+
+          unless config[:enable_statement_attribute]
+            config[:db_statement] = :omit
+            OpenTelemetry.logger.warn(
+              'Instrumentation pg configuration option enable_sql_obfuscation has been deprecated,' \
```
fbogsany, line 26:
```suggestion
              'Instrumentation pg configuration option enable_statement_attribute has been deprecated,' \
```

fbogsany (approved):
Small change, otherwise LGTM

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -66,15 +66,15 @@ def lru_cache
...
               operation = extract_operation(args[0])
-              sql = obfuscate_sql(args[0])
+              sql = args[0]
             else
               statement_name = args[0]
 
               if kind == :prepare
-                sql = obfuscate_sql(args[1])
+                sql = args[1]
                 lru_cache[statement_name] = sql
```
robertlaurin, line 78: Trying to think this one through, is the performance better or worse with this change? 

Before we would cache the obfuscated code, but we would obfuscate before we knew that whether or not we would record the attribute for it, so that could have been wasted work.

Now we only obfuscate if the setting is enabled (which means we are recording the attribute), but now we are not caching the obfuscated code so we may obfuscate the same SQL multiple times.  

@ahayworth @fbogsany wdyt?  Is this being a bit too nit picky?

SomalianIvan: I was too zoned in on my changes, I think it makes sense to use LRU cache for obfuscated call

fbogsany: I agree - I think the simplest change here is to just replace the original conditionals `if config[:enable_statement_attribute]` and `unless config[:enable_sql_obfuscation]` with (respectively) `unless config[:db_statement] == :omit` and `unless config[:db_statement] == :obfuscate`.

robertlaurin (approved):

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ahayworth: @robertlaurin you're not being too nitpicky - we definitely want to cache the results of obfuscation if we're going to obfuscate it later on. @arielvalentin and I discovered that the string-munging obfuscation we do in the mysql instrumentation has a performance penalty (and the pg implementation is quite similar). So if we only have to pay that once for obfuscation, that's much better.

It's a moot point now that the change was made differently, but I don't think you were being overly cautious at all!

PR #915, opened 2021-08-17
fix: rack middleware assuming script_name presence

robertlaurin:
```
Minitest::UnexpectedError: NoMethodError: undefined method `+' for nil:NilClass
--
  | /app/vendor/bundle/ruby/2.7.0/gems/opentelemetry-instrumentation-rack-0.19.0/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb:133:in `fullpath'
```

It seems it is not safe to assume that `SCRIPT_NAME` will always be present.

>SCRIPT_NAME
>The initial portion of the request URL's “path” that corresponds to the application object, so that the application knows its >virtual “location”. This may be an empty string, if the application corresponds to the “root” of the server.
https://github.com/rack/rack/blob/master/SPEC.rdoc

### DIFF
--- a/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
+++ b/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
@@ -121,18 +121,11 @@ module OpenTelemetry
               'http.method' => env['REQUEST_METHOD'],
               'http.host' => env['HTTP_HOST'] || 'unknown',
               'http.scheme' => env['rack.url_scheme'],
-              'http.target' => fullpath(env)
+              'http.target' => env['QUERY_STRING'].empty? ? env['PATH_INFO'] : "#{env['PATH_INFO']}?#{env['QUERY_STRING']}"
             }
+
             attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
-            attributes.merge(allowed_request_headers(env))
-          end
-
-          # e.g., "/webshop/articles/4?s=1":
-          def fullpath(env)
-            query_string = env['QUERY_STRING']
-            path = env['SCRIPT_NAME'] + env['PATH_INFO']
-
-            query_string.empty? ? path : "#{path}?#{query_string}"
+            attributes.merge!(allowed_request_headers(env))
           end
 
           # https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#name
--- a/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
+++ b/instrumentation/rack/test/opentelemetry/instrumentation/rack/middlewares/tracer_middleware_test.rb
@@ -29,6 +29,7 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware do
   let(:default_config) { {} }
   let(:config) { default_config }
   let(:env) { {} }
+  let(:uri) { '/' }
 
   before do
     # clear captured spans:
@@ -54,7 +55,7 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware do
 
   describe '#call' do
     before do
-      Rack::MockRequest.new(rack_builder).get('/', env)
+      Rack::MockRequest.new(rack_builder).get(uri, env)
     end
 
     it 'records attributes' do
@@ -72,6 +73,15 @@ describe OpenTelemetry::Instrumentation::Rack::Middlewares::TracerMiddleware do
 
     it 'has no parent' do
       _(first_span.parent_span_id).must_equal OpenTelemetry::Trace::INVALID_SPAN_ID
+    end
+
+    describe 'when a query is passed in' do
+      let(:uri) { '/endpoint?query=true' }
+
+      it 'records the query path' do
+        _(first_span.attributes['http.target']).must_equal '/endpoint?query=true'
+        _(first_span.name).must_equal '/endpoint'
+      end
     end
 
     describe 'config[:untraced_endpoints]' do


instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -121,18 +121,11 @@ def request_span_attributes(env:)
...
-          end
 
-          # e.g., "/webshop/articles/4?s=1":
-          def fullpath(env)
-            query_string = env['QUERY_STRING']
-            path = env['SCRIPT_NAME'] + env['PATH_INFO']
-
-            query_string.empty? ? path : "#{path}?#{query_string}"
+            attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
+            attributes.merge!(allowed_request_headers(env))
```
ahayworth, line 128: Honestly I thought `merge!` returned `nil`, but that must be some other bang-type method I was thinking of.

ahayworth (approved):
LGTM

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -121,18 +121,11 @@ def request_span_attributes(env:)
               'http.method' => env['REQUEST_METHOD'],
               'http.host' => env['HTTP_HOST'] || 'unknown',
               'http.scheme' => env['rack.url_scheme'],
-              'http.target' => fullpath(env)
+              'http.target' => env['QUERY_STRING'].empty? ? env['PATH_INFO'] : "#{env['PATH_INFO']}?#{env['QUERY_STRING']}"
```
arielvalentin, line 124: I have a suggestion to use functional style here but not sure if this is generating too many objects for our performance needs:
```suggestion
 [env['PATH_INFO'], env['QUERY_STRING']].join("?")
```
Is there a unit test that covers this functionality? 

Will `env['QUERY_STRING']` always return a `String` object? Will it ever return a `nil`?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
```diff
@@ -121,18 +121,11 @@ def request_span_attributes(env:)
...
-            attributes.merge(allowed_request_headers(env))
-          end
 
-          # e.g., "/webshop/articles/4?s=1":
-          def fullpath(env)
-            query_string = env['QUERY_STRING']
-            path = env['SCRIPT_NAME'] + env['PATH_INFO']
-
-            query_string.empty? ? path : "#{path}?#{query_string}"
+            attributes['http.user_agent'] = env['HTTP_USER_AGENT'] if env['HTTP_USER_AGENT']
```
arielvalentin, line 127: Have we decided to start using [Semantic Conventions Constants](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb#L162) everywhere? If so should we update this as part of this change?

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: Yeah I usually double check when I use anything with a bang.  Probably thinking of `compact!`

```
compact!
Removes nil elements from the array.

Returns nil if no changes were made, otherwise returns the array.
```

https://ruby-doc.org/core-2.4.1/Array.html#method-i-compact-21

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: >I have a suggestion to use functional style here but not sure if this is generating too many objects for our performance needs

🤔 , definitely looks nicer.

> Is there a unit test that covers this functionality?

Hmm, I think it is missing a case for when a query string is present.  I'll add that in.

> Will env['QUERY_STRING'] always return a String object? Will it ever return a nil?

According to the spec it can be "empty" but is always required.  I think that means an empty string?
```
QUERY_STRING
The portion of the request URL that follows the ?, if any. May be empty, but is always required!
```

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: >Have we decided to start using Semantic Conventions Constants everywhere? 

I'm not sure actually, I'll bring it up at the next sig to see what we want to do there.

>If so should we update this as part of this change?

I think this change would happen as a "big sweeping change" across the gems.  So I'd prefer to keep it out of this PR.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
robertlaurin: Doesn't look like that suggestion works the way we want it to. 
irb(main):001:0> ["/", ""].join("?")
=> "/?"
irb(main):002:0> ["/", nil].join("?")
=> "/?"

We expect it to be `/`

arielvalentin: Ah fair because `QUERY_STRING` could be a blank string. Makes sense.

instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb
fbogsany: > > Have we decided to start using Semantic Conventions Constants everywhere?
> 
> I'm not sure actually, I'll bring it up at the next sig to see what we want to do there.

Yes, we should use the Semantic Conventions Constants everywhere.

> 
> > If so should we update this as part of this change?
> 
> I think this change would happen as a "big sweeping change" across the gems. So I'd prefer to keep it out of this PR.

Agreed

PR #917, opened 2021-08-18
feat: trace update_all and delete_all calls in ActiveRecord

indrekj:
We often use code like:
```
  Product
    .where(...)
    .where(...)
    .update_all(attributes)
```
This allows doing atomic updates and avoid N+1 queries. `update_all` and
`delete_all` however were not traced.

I initially thought about putting these to `PersistenceClassMethods`
module but that doesn't work when scopes are used. These calls create
`ActiveRecord::Relation` instance which seemed a better place to patch.
I also added tests to ensure these calls are also traced when scopes are
not used.


robertlaurin (commented):
This looks great, I'm going to pull down the changes and run them on a locally against a real app before bringing it.  

The patching against Active Record has been brittle in my experience so far, and green in CI doesn't always mean it applies correctly for a real app.

indrekj:
@robertlaurin have you had a chance to take a look?

robertlaurin:
Looks good to me, tested on a very simple app here
image

image

As well as tested on a real app locally.

robertlaurin (approved):

PR #918, opened 2021-08-18
feat: add Que instrumentation

indrekj:
From the home page:
> [que][1] is a high-performance job queue that improves the reliability of your application by protecting your jobs with the same ACID guarantees as the rest of your data

It's similar to delayed_job and sidekiq.

There are two parts here:
* Patching `Que::Job.enqueue` which is used when enqueuing and a new job.
* Adding a `job_middleware` which is used when processing the job.

Que doesn't have a good place to store metadata. There are
basically two options: the job payload and the job tags.

Using the job payload is very brittle. We'd have to modify
existing Hash arguments or add a new argument when there are
no arguments we can modify. If the server side is not using
this instrumentation yet (e.g. old jobs before the
instrumentation was added or when instrumentation is being
added to client side first) then the server can error out due
to unexpected payload.

The second option (which we are using here) is to use tags.
They also are not meant for tracing information but they are
much safer to use than modifying the payload.

This is my first instrumentation pull request so it's likely I missed something. I mostly used the code from already existing instrumentations to build this. (e.g. I'm not entirely sure how versioning should work for new libraries)

cc @siegy22 (que maintainer, in case you have any comments)

[1]: https://github.com/que-rb/que


instrumentation/que/lib/opentelemetry/instrumentation/que/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Que
+      VERSION = '0.19.1'
```
fbogsany, line 10: Since we started the 1.0 release candidates, we've decoupled instrumentation versions, so this can start at 0.1.0.

instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+              "#{job_class} process",
+              attributes: attributes_before_job_complation(job, job_class),
+              kind: :consumer
+            ) do |span|
+              block.call
+
+              span.set_attribute('messaging.que.attempts', job.que_attrs[:error_count])
+              if (error = job.que_error)
+                span.record_exception(error)
+                span.status = OpenTelemetry::Trace::Status.error("Unhandled exception of type: #{error.class}")
```
fbogsany, line 26: The general guidelines suggest that instrumentation should not set `span.status` to `error` - only application code should do that. That said, I *think* this is OK, since we're pretty sure at this point that the job's error is "real".

instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+    module Que
+      module Middlewares
+        # Server middleware to trace Que jobs
+        class ServerMiddleware
+          def self.call(job, &block)
+            job_class = job.que_attrs[:job_class] || job.class.name
+
+            tracer.in_span(
+              "#{job_class} process",
+              attributes: attributes_before_job_complation(job, job_class),
```
fbogsany, line 18:
```suggestion
              attributes: attributes_before_job_completion(job, job_class),
```

instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb
```diff
@@ -0,0 +1,56 @@
...
+            # return value is not important
+            nil
+          end
+
+          def self.tracer
+            OpenTelemetry::Instrumentation::Que::Instrumentation.instance.tracer
+          end
+          private_class_method :tracer
+
+          def self.attributes_before_job_complation(job, job_class)
```
fbogsany, line 39:
```suggestion
          def self.attributes_before_job_completion(job, job_class)
```

fbogsany:
Thanks for this contribution @indrekj! It'd be really great if we could find a suitable mechanism for context propagation, but if there's no extensible metadata facility we can leverage in `que`, I think it's fine to at least get this level of instrumentation in.

instrumentation/que/lib/opentelemetry/instrumentation/que/version.rb
indrekj: https://github.com/open-telemetry/opentelemetry-ruby/compare/4e646730ac5f7f3fde5ee479127a449d1e30d6ed..b8fff233a2309740f43a163a6a7330b78602e348

instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb
indrekj: https://github.com/open-telemetry/opentelemetry-ruby/compare/4e646730ac5f7f3fde5ee479127a449d1e30d6ed..b8fff233a2309740f43a163a6a7330b78602e348

instrumentation/que/lib/opentelemetry/instrumentation/que/middlewares/server_middleware.rb
indrekj: https://github.com/open-telemetry/opentelemetry-ruby/compare/4e646730ac5f7f3fde5ee479127a449d1e30d6ed..b8fff233a2309740f43a163a6a7330b78602e348

indrekj:
> It'd be really great if we could find a suitable mechanism for context propagation, but if there's no extensible metadata facility we can leverage in que, I think it's fine to at least get this level of instrumentation in.

One thought I had: to have a configuration field to allow propagating context through the payload. So the user can opt-in if they want, but we don't accidentally break any existing users. But yes, to my knowledge there's no good metadata field in the que right now.

ahayworth:
> One thought I had: to have a configuration field to allow propagating context through the payload. So the user can opt-in if they want, but we don't accidentally break any existing users. But yes, to my knowledge there's no good metadata field in the que right now.

@indrekj We have an option for that in the  [ActiveJob instrumentation](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb#L37-L46) as well as the [Sidekiq instrumentation](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/instrumentation.rb#L31). If you implement such an option, you should allow it to be set in the same way as those two for consistency's sake (if possible, of course).

Regarding the lack of metadata for propagation - both the sidekiq instrumentation and the ActiveJob instrumentation end up actually propagating context through the payload (the Sidekiq instrumentation does it explicitly; the ActiveJob instrumentation does it implicitly via some Rails magic). That approach can work well, if you take care not to mangle the job's payload! I'd be excited to see if it can work for Que as well! 😄

indrekj:
@ahayworth I added the same option.

I was also able to use `tags` to propagate the tracing information.

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/que_job.rb
```diff
@@ -0,0 +1,74 @@
...
+                #
+                # The second option (which we are using here) is to use tags.
+                # They also are not meant for tracing information but they are
+                # much safer to use than modifying the payload.
+                if otel_config[:propagation_style] != :none
+                  tracing_info = {}
+                  OpenTelemetry.propagation.inject(tracing_info)
+
+                  tags ||= []
+                  tags += tracing_info.map { |k, v| "#{k}:#{v}" }
```
fbogsany, line 45: You can define your own `setter` that appends to `tags` directly, rather than create a `Hash` and map it. See https://github.com/open-telemetry/opentelemetry-ruby/blob/7e314bde62bb0fb46357b459e726ff93b8737795/api/lib/opentelemetry/context/propagation/text_map_setter.rb#L13-L17
```suggestion
                  tags ||= []
                  OpenTelemetry.propagation.inject(tags, setter: tag_setter)
```
where:
```ruby
class TagSetter
  def set(carrier, key, value)
    carrier << "#{key}:#{value}"
  end
end

tag_setter = TagSetter.new
```

indrekj: https://github.com/open-telemetry/opentelemetry-ruby/compare/21b528af11288591298b67f656aacb29bf90444f..b828d9146b67d623f84b6b963c7b8bd8abb302ee

instrumentation/que/lib/opentelemetry/instrumentation/que/tag_getter.rb
```diff
@@ -0,0 +1,37 @@
...
+          tags_array_to_hash(carrier)[key]
+        end
+
+        def keys(carrier)
+          tags_array_to_hash(carrier).keys
+        end
+
+        private
+
+        def tags_array_to_hash(tags)
```
fbogsany, line 26: This is rough 😞 - the API really isn't designed well for this sort of data format. I think you'd be better off with the original code (converting to a hash and using the default getter) for the `extract` operation. Doing it this way is just adding abstraction for no value - we're still allocating the hash, we're just hiding it.

indrekj: yes, I agree. It felt somewhat odd. Reverted it: https://github.com/open-telemetry/opentelemetry-ruby/compare/b828d9146b67d623f84b6b963c7b8bd8abb302ee..1723d99922224afec6a1f1b9370f02feada1a0e3

indrekj:
I think I've addressed all comments so far. If there's anything else I need to do or improve here please let me know.

indrekj:
@fbogsany @ahayworth pinging you as you've been active before in this PR. I would like to get this merged/released. Is there anything else I am missing?

fbogsany (approved):

fbogsany:
> @fbogsany @ahayworth pinging you as you've been active before in this PR. I would like to get this merged/released. Is there anything else I am missing?

Thanks @indrekj - sorry this took so long to review 😞.

indrekj:
@fbogsany Thank you :bow:. I'd appreciate it if someone could also publish it to rubygems. This would allow us to start migrating all our ruby projects to opentelemetry.

fbogsany:
🤦 will do, thanks for the polite nudge

PR #925, opened 2021-08-18
fix: OTLP exporter missing failure metrics

robertlaurin:
We were not capturing failures for a common failure path, we were also under
reporting failures on the final retry.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -187,6 +187,7 @@ module OpenTelemetry
             when Net::HTTPBadRequest, Net::HTTPClientError, Net::HTTPServerError
               # TODO: decode the body as a google.rpc.Status Protobuf-encoded message when https://github.com/open-telemetry/opentelemetry-collector/issues/1357 is fixed.
               response.body # Read and discard body
+              @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response.code })
               FAILURE
             when Net::HTTPRedirection
               @http.finish
@@ -237,9 +238,8 @@ module OpenTelemetry
         end
 
         def backoff?(retry_after: nil, retry_count:, reason:)
-          return false if retry_count > RETRY_COUNT
-
           @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => reason })
+          return false if retry_count > RETRY_COUNT
 
           sleep_interval = nil
           unless retry_after.nil?


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -187,6 +187,7 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
             when Net::HTTPBadRequest, Net::HTTPClientError, Net::HTTPServerError
               # TODO: decode the body as a google.rpc.Status Protobuf-encoded message when https://github.com/open-telemetry/opentelemetry-collector/issues/1357 is fixed.
               response.body # Read and discard body
+              @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response.code })
```
robertlaurin, line 190: Currently this failure was not being reported so comparing `otel.bsp.export.failure` against `otel.otlp_exporter.failure` was not providing proper insight into export failures.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -237,9 +238,8 @@ def measure_request_duration
         end
 
         def backoff?(retry_after: nil, retry_count:, reason:)
-          return false if retry_count > RETRY_COUNT
-
           @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => reason })
```
robertlaurin, line 241: We were also not reporting the final failed attempt, thanks to @SomalianIvan for catching this.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
arielvalentin: ✨

arielvalentin: @robertlaurin Is `reason` free text other instances? Would it be sensible to use `http.status_code` as a label instead?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
arielvalentin: Any chance to add a test for this 🙏🏼 ?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: I'd consider it free text at the moment
https://github.com/open-telemetry/opentelemetry-ruby/blob/469aa89ba180b1716a6a33c12647b32f1df4fed9/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L200-L214

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -187,6 +187,7 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/AbcSize, Metrics/Cyclo
             when Net::HTTPBadRequest, Net::HTTPClientError, Net::HTTPServerError
               # TODO: decode the body as a google.rpc.Status Protobuf-encoded message when https://github.com/open-telemetry/opentelemetry-collector/issues/1357 is fixed.
```
fbogsany, line 188: Note that the issue referenced was fixed almost a year ago. We should be able to decode the body and potentially use information it contains in the `reason` label below.

fbogsany (approved):
This LGTM. It’d be nice if we could decode the google.rpc.Status body, but don’t let that hold up this fix - maybe open an issue to track that?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: I remember why I didn't have a test for this originally. 

The backoff behaviour basically makes running tests really slow unless you stub backoff, which is where the metric reporting behaviour is.  
https://github.com/open-telemetry/opentelemetry-ruby/blob/469aa89ba180b1716a6a33c12647b32f1df4fed9/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L239-L264

It's pretty awkward to test, if you have some ideas to simplify testing this I can circle back and add it.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robertlaurin: Tracking here https://github.com/open-telemetry/opentelemetry-ruby/issues/927

PR #932, opened 2021-08-24
Use 4318 as the default OLTP exporter port

bcharna:
The collector has changed to use port 4318 by default in https://github.com/open-telemetry/opentelemetry-collector/pull/3743.

### DIFF
--- a/exporter/otlp/README.md
+++ b/exporter/otlp/README.md
@@ -70,7 +70,7 @@ The collector exporter can be configured explicitly in code, or via environment 
 
 | Parameter           | Environment variable                         | Default                             |
 | ------------------- | -------------------------------------------- | ----------------------------------- |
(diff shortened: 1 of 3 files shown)


arielvalentin (approved):

fbogsany (changes requested):
Blocking this for now. The port change in the spec was reverted in https://github.com/open-telemetry/opentelemetry-specification/pull/1847 until a prototype PR in the collector determines whether it *is* in fact possible for gRPC and HTTP to share a port. We can merge or close this once that’s resolved one way or another.

indrekj:
https://github.com/open-telemetry/opentelemetry-specification/pull/1970 it has been changed to 4318

fbogsany (approved):

PR #933, opened 2021-08-25
feat: move activesupport notification subsciber out of action_view gem

SomalianIvan:
this will allow to subsribe to arbitrary active support notifications.
## TODO
  * [x] Move the subscriber, fanout stuff from `action_view` to new `active_support_instrumentation` gem
  * [x] pass the subscriptions as a config option
  * [x] add `active_support_instrumentation` dependency to `action_view`


SomalianIvan:
> This seems like a reasonable idea overall, and I had originally added it to `opentelemetry-instrumentation-rails` in earlier versions of the ActionView PRs. I'm curious how we'll deal with initialization of the `Fanout` subscriber from potentially multiple instrumentations, though?

I wonder if there is a need for `action_view_instrumentation` at all then, we can let the `active_support_instrumentation` handle subscribing to active support notifications and also provide default subscription set that users can override? Wdyt @ahayworth @robertlaurin.

```ruby
option action_view_enabled?
... 
option arbitrary_notifications
```

robertlaurin:
So…
- ActiveSupport would be it’s own instrumentation
- Rails would depend on ActiveSupport instrumentation
- Rails would configured a default set of notifications (including action view?)
- ActionView wouldn't exist as it's own instrumentation gem?
- Someone could pull ActiveSupport without rails and configure it all manually themselves?

SomalianIvan:
> is there another part of Rails we want to instrument right now using ActiveSupport notifications?

I was migrating one of the repos to open telemetry and encountered this bit

```
# This is all the action view stuff
Shopify::Tracing::Subscriber.subscribe('render_collection.action_view')
Shopify::Tracing::Subscriber.subscribe('render_partial.action_view')
Shopify::Tracing::Subscriber.subscribe('render_template.action_view')
Shopify::Tracing::Subscriber.subscribe('instantiation.active_record')
# Action Mailer
Shopify::Tracing::Subscriber.subscribe('deliver.action_mailer')
Shopify::Tracing::Subscriber.subscribe('process.action_mailer')
# Active Job
Shopify::Tracing::Subscriber.subscribe('enqueue_at.active_job', transform: job_transform)
Shopify::Tracing::Subscriber.subscribe('enqueue.active_job', transform: job_transform)
Shopify::Tracing::Subscriber.subscribe('enqueue_retry.active_job', transform: job_transform)
Shopify::Tracing::Subscriber.subscribe('perform.active_job', transform: job_transform)
Shopify::Tracing::Subscriber.subscribe('retry_stopped.active_job', transform: job_transform)
Shopify::Tracing::Subscriber.subscribe('discard.active_job', transform: job_transform)
```

robertlaurin:
>I guess to some extent, that depends on what the motivation for this PR is in the first place. Is there another part of Rails we want to instrument right now using ActiveSupport notifications? I think that might change how we approach it...

I would say the motivation right now is to be able to support generating spans from any Active Support Notification topic.  In the case of the action view instrumentation we preconfigure which topics.  

I believe there is an appetite for allowing an app owner to supply their own list of topics, similar to what @SomalianIvan shared above.

> I think that could work, although that's maybe a little bit at odds with the overall direction we've been taking with Rails instrumentation, which has been to break everything out into sub-gems.

Yeah I think this is interesting because some sub gems would instrumented purely with Notifications, while some will get patched.  I imagine a situation where both may be true.  

I'm trying to balance what makes the most sense here.  Is it worth having gems that only add subscriptions to notification topics?

robertlaurin:
>I'm curious how we'll deal with initialization of the Fanout subscriber from potentially multiple instrumentations, though?

I think if we can come up with a good answer to this question we can continue with the approach taken on this PR, and I would feel good about that.  Even if it means some gems just add topic subscriptions.

ahayworth:
> Seems unnecessary to have split packages, wdyt?

¯\\_(ツ)_/¯ - we have been kinda going back and forth on this for awhile: is it better to have split packages for rails things, or a mega package for all things rails (with lots of knobs to adjust). Honestly, there are pros and cons to both approaches. 

> I would say the motivation right now is to be able to support generating spans from any Active Support Notification topic. In the case of the action view instrumentation we preconfigure which topics.
> I believe there is an appetite for allowing an app owner to supply their own list of topics, similar to what @SomalianIvan shared above.

Cool - that's kinda what I had presumed, but wanted to be sure.

> Yeah I think this is interesting because some sub gems would instrumented purely with Notifications, while some will get patched. I imagine a situation where both may be true.
> I'm trying to balance what makes the most sense here. Is it worth having gems that only add subscriptions to notification topics?

It honestly feels like overkill having gems that just subscribe to notifications, just a little bit. _But_, I think it makes more sense to be consistent - either we keep splitting everything, or we smush it all up in a mega-package. I feel like the in-between case is really worse. We've already started down the route of splitting things up, and I honestly think it's fine to continue with splitting things up.

> I think if we can come up with a good answer to this question we can continue with the approach taken on this PR, and I would feel good about that. Even if it means some gems just add topic subscriptions.

So two things have to happen for the subscription logic to work right:
- We _must_ replace the default `ActiveSupport::Notifications::Fanout` queue with our own `OpenTelemetry::Instrumentation::ActiveSupport::Fanout` queue.
    - This _must_ happen before any subscriptions are made, because subscriptions are essentially "stored" on the queue instance. If we don't do it first thing, then other initializers could add subscriptions that are never honored - which would be very unfortunate. 😆 
- We _may_ subscribe to notifications some point after the swap, but each subscription _must_ be made only once (or we'd have lots of nested, duplicate spans).

With that in mind I think the most straightforward thing to do is:
1. Have each instrumentation that needs subscriptions depend on the `-active_support` instrumentation, and install it _if it's not already installed_.
    1. Can we reliably detect that condition today?
    2. The main reason to not blindly re-install is because we have options to preserve that are likely valid for all subscription things - they're only in `-action_view` today because that's the only place we used subscriptions right now, but I think people would want them for all subscription things? I could be really wrong about that though, and if so then it's a lot less important to preserve options.
2. Each instrumentation that needs subscriptions should install their own railtie, essentially identical to what we do today. We should change _all_ of the railties to basically say:

```ruby
config.before_initialize do
  unless ::ActiveSupport::Notifications.notifier.is_a?(::OpenTelemetry::Instrumentation::ActiveSupport::Fanout)
    ::ActiveSupport::Notifications.notifier = ::OpenTelemetry::Instrumentation::ActiveSupport::Fanout.new 
  end
end 
```

Ugly, I know.

3. The `after_initialize` block of the railties should just contain their subscription logic.


I don't really know if that's the most elegant or the _best_ way, but I think something like it might work today? There could be (probably are) better ways to do it, though...

robertlaurin:
>It honestly feels like overkill having gems that just subscribe to notifications, just a little bit. But, I think it makes more sense to be consistent - either we keep splitting everything, or we smush it all up in a mega-package. I feel like the in-between case is really worse. We've already started down the route of splitting things up, and I honestly think it's fine to continue with splitting things up.

Yeah I've been bouncing around a bit too much because I want to make sure we get to a good place with this before it gets too hard to change.  That said I completely agree with you, let's just stick with splitting it even if some gems are really thin and feel a bit over kill.  Like you said, we get consistency and that has a lot of value here.

robertlaurin (approved):

PR #934, opened 2021-08-27
Exporter should use gzip compression by default

johnnyshields:
Gzip should be the standard for all tracing. 99.9% of practical application use cases would want it; it should only be disabled in very specific scenarios (I'm not even sure what they would be).


arielvalentin (dismissed):
Thanks for this contribution.

Based on my understanding of the specification there should not be a default value for compression, which is why  we have left it unset by default. 

Am I misinterpreting this?

Supported values for OTEL_EXPORTER_OTLP_*COMPRESSION options:

If the value is missing, then compression is disabled.
gzip is the only specified compression method for now. Other options MAY be supported by language SDKs and should be documented for each particular language.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options

mjobin-mdsol (approved):

fbogsany (changes requested):
It looks like we're going with https://github.com/open-telemetry/opentelemetry-specification/pull/1923 which allows SDKs to decide the default (valid choices are `none` and `gzip`). We can go ahead with this PR when the spec PR is approved & merged.

johnnyshields:
@fbogsany it looks like the gzip error was fixed in #983. Are we good to merge now?

fbogsany:
There's a CI failure - this expects a default of no compression:
https://github.com/open-telemetry/opentelemetry-ruby/blob/61b84aec36130799603e1fb6f0b22155a42edfcd/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb#L20

fbogsany (approved):

PR #935, opened 2021-08-31
feat: Faraday instrumentation installs OTel middleware first in handler list

nicksieger:
This allows propagation headers to be added before other middleware that may depend on them being installed, such as
[faraday_middleware-aws-sigv4](/winebarrel/faraday_middleware-aws-sigv4), which does a signature calculation that hashes all headers. If the headers are installed last, the signature calculation will not match the servers's hash value.

### DIFF
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/patches/rack_builder.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/patches/rack_builder.rb
@@ -11,12 +11,11 @@ module OpenTelemetry
         # Module to be prepended to force Faraday to use the middleware by
         # default so the user doesn't have to call `use` for every connection.
         module RackBuilder
-          def adapter(*args)
-            use(:open_telemetry) unless @handlers.any? do |handler|
-              handler.klass == Faraday::Middlewares::TracerMiddleware
+          def initialize(*args, &block)
+            super(*args) do |*bargs|
+              use :open_telemetry
+              block&.call(*bargs)
             end
-
-            super
           end
         end
       end
--- a/instrumentation/faraday/opentelemetry-instrumentation-faraday.gemspec
+++ b/instrumentation/faraday/opentelemetry-instrumentation-faraday.gemspec
@@ -34,7 +34,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'faraday', '~> 0.17.0'
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-sdk', '~> 1.0.0.rc1'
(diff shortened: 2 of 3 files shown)


nicksieger:
Upon further inspection, while I think this might still be worth considering, instead the AWS signer should be configured to ignore tracing headers, like it already does for authorization. Let me know if you think the contribution is worthwhile, otherwise feel free to close.

fbogsany:
> instead the AWS signer should be configured to ignore tracing headers, like it already does for authorization

I think this is the right approach.

It's also worth considering how to allow users to adjust the position of the middleware in the stack. IMO the right default is to place it as "close to the wire" as possible. I think the "escape hatch" should be an option to not install the middleware, in which case users would install it explicitly in their Faraday middleware stacks.

I think we have part of this already: if the user explicitly inserts the middleware, the `RackBuilder` module will skip adding it explicitly: https://github.com/open-telemetry/opentelemetry-ruby/blob/ec213f0b752f499e15c3c0e695c3173c85d52086/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/patches/rack_builder.rb#L15-L17

nicksieger:
Thanks for the feedback. I agree with your analysis. Closing this.

Would appropriately instrumenting the AWS signing library to ignore trace headers be worth a new library? If so I can throw up a PR with a stab at that.

fbogsany:
> Would appropriately instrumenting the AWS signing library to ignore trace headers be worth a new library? If so I can throw up a PR with a stab at that

I don't know what's involved in that. Is that a configuration option for the AWS signing library, or does it require monkey-patching? It doesn't quite feel like "instrumentation".

nicksieger:
Yeah, it's not really instrumentation, just configuration. It gets [configured via an initializer](https://github.com/aws/aws-sdk-ruby/blob/version-3/gems/aws-sigv4/lib/aws-sigv4/signer.rb#L107-L109), so I have to monkey-patch it like so. Maybe just having this info in the PR is enough documentation for people to fix this for themselves.

```ruby
  # Patch Aws::Sigv4::Signer to always ignore tracing headers
  begin
    require "aws-sigv4"
    module IgnoreTracingHeaders
      def initialize(options = {})
        (options[:unsigned_headers] ||= []).concat %w(traceparent tracestate baggage)
        super(options)
      end
    end

    class Aws::Sigv4::Signer
      prepend IgnoreTracingHeaders
    end
  rescue LoadError
    # Skip since aws_sigv4 isn't present
  end
```

leemeichin:
As an alternative to monkeypatching, you can pass the unsigned headers in directly: 

```ruby
f.request(:aws_sigv4, service: 'es', region: 'eu-west-1', unsigned_headers: %w[traceparent tracestate baggage])
```

PR #938, opened 2021-09-03
fix: keep Active Support subscriptions intact when patching

SomalianIvan:
Recently encountered an issue where subscriptions that happen earlier get dropped when we override `notifier`. Here is my attempt at keeping the subscriptions by using `DelegateClass`

I confirmed by running the added test on main branch

image

### DIFF
--- a/instrumentation/action_view/test/opentelemetry/instrumentation/action_view/fanout_test.rb
+++ b/instrumentation/action_view/test/opentelemetry/instrumentation/action_view/fanout_test.rb
@@ -74,10 +74,13 @@ describe OpenTelemetry::Instrumentation::ActionView::Fanout do
       end
 
       ::ActiveSupport::Notifications.notifier = OpenTelemetry::Instrumentation::ActionView::Fanout.new(::ActiveSupport::Notifications.notifier)
+      ::ActiveSupport::Notifications.subscribe('render', span_subscriber)
 
       ActiveSupport::Notifications.instrument('render', extra: :information)
 
-      assert notification_fired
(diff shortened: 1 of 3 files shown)


instrumentation/action_view/test/opentelemetry/instrumentation/action_view/fanout_test.rb
```diff
@@ -62,4 +62,22 @@
...
+      notification_fired = false
+      ActiveSupport::Notifications.subscribe('render') do |*_args|
+        notification_fired = true
+      end
+
+      ::ActiveSupport::Notifications.notifier = OpenTelemetry::Instrumentation::ActionView::Fanout.new(::ActiveSupport::Notifications.notifier)
+
+      ActiveSupport::Notifications.instrument('render', extra: :information)
+
+      assert notification_fired
```
robertlaurin, line 80:
```suggestion
      ::ActiveSupport::Notifications.notifier = OpenTelemetry::Instrumentation::ActionView::Fanout.new(::ActiveSupport::Notifications.notifier)
      ::ActiveSupport::Notifications.subscribe('render', span_subscriber)

      ActiveSupport::Notifications.instrument('render', extra: :information)

      _(last_span).wont_be_nil
      _(last_span.name).must_equal('foo bar')
      _(notification_fired).must_equal(true)
```

ahayworth (approved):
I'm fairly certain this will work. 👍

robertlaurin (approved):
This makes sense to me.

PR #946, opened 2021-09-15
Add instrumentation helpers gem and hide_query_param option

ericmustin:
### summary

building off of open-telemetry/opentelemetry-ruby#769 this pr introduces an `instrumentation_helpers` gem as discussed in open-telemetry/opentelemetry-ruby-contrib#27.

Most of this is just code reshuffling from the great work in open-telemetry/opentelemetry-ruby#769 , but there's a few new bits here.

And, for context, what open-telemetry/opentelemetry-ruby#769 provides is the option `hide_query_params` to a number of http client instrumentations, that removes the query_string on `http.target` and `http.url` attributes https://github.com/open-telemetry/opentelemetry-ruby/pull/946/files#diff-4ff80b55e04dbd9500019e54eb35f43d50c71a881bf318e3ac74e0738d19e9adR17-R18

### details

There's basically one mixin introduced here, `HTTP::RequestAttributes`. 
 
`HTTP::RequestAttributes` is pretty straightforward, it provides helper methods that can get included into instrumentation and be used to ensure consistent http span attributes are extracted out of the http.client gem's metadata. It's almost entirely a reshuffling of open-telemetry/opentelemetry-ruby#769. 

### Questions

- would people prefer if i just broke out the `hide_query_params` stuff into a small quick pr to get this feature out
- should `hide_query_params` instead be a `callable` option that allows ppl to define a proc that takes the `query_string` as an argument and allows them to obfuscate it in whatever way they prefer. This would be more useful if, say, 1 specific query param had PII but the rest didn't, and the user doesn't want to drop the entire query string?


instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/instrumentation.rb
```diff
@@ -11,6 +11,8 @@ module Instrumentation
     module HttpClient
       # The Instrumentation class contains logic to detect and install the HttpClient instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
+        include InstrumentationHelpers::HTTP::RequestAttributes
```
fbogsany, line 14:
```suggestion
        include InstrumentationHelpers::HTTP::InstrumentationOptions
```

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
```diff
@@ -42,14 +46,10 @@ def call(env)
...
-              'http.method' => http_method, 'http.url' => url.to_s
-            }
             config = Faraday::Instrumentation.instance.config
-            instrumentation_attrs['peer.service'] = config[:peer_service] if config[:peer_service]
-            instrumentation_attrs.merge(
-              OpenTelemetry::Common::HTTP::ClientContext.attributes
-            )
+            attributes = from_request(http_method, url, config)
+            attributes['peer.service'] = config[:peer_service] if config[:peer_service]
+            attributes.merge(OpenTelemetry::Common::HTTP::ClientContext.attributes)
```
fbogsany, line 52: Should the HTTP helpers implicitly merge the HTTP client context attributes?

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
```diff
@@ -42,14 +46,10 @@ def call(env)
...
-            instrumentation_attrs = {
-              'http.method' => http_method, 'http.url' => url.to_s
-            }
             config = Faraday::Instrumentation.instance.config
-            instrumentation_attrs['peer.service'] = config[:peer_service] if config[:peer_service]
-            instrumentation_attrs.merge(
-              OpenTelemetry::Common::HTTP::ClientContext.attributes
-            )
+            attributes = from_request(http_method, url, config)
+            attributes['peer.service'] = config[:peer_service] if config[:peer_service]
```
fbogsany, line 51: `peer.service` feels like something common we should extract as well. It might be more general than just HTTP, though? Maybe we need inheritance of options?

instrumentation_helpers/lib/opentelemetry/instrumentation_helpers/http/instrumentation_options.rb
```diff
@@ -0,0 +1,18 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module InstrumentationHelpers
+    module HTTP
+      # Instrumentation options contains instrumentation helpers for http instrumentation client requests
+      module InstrumentationOptions
+        def self.included(klass)
+          klass.option(:hide_query_params, default: true, validate: :boolean) if klass.respond_to?(:option)
```
fbogsany, line 13: This feels overly defensive to me. If we want to do this, though, we should `raise` if the class doesn't respond to `option`.

We need a convention for documenting these options, and referencing them from the class that includes them. I don't think we want to copy option documentation - that seems very brittle.

I wonder if there's value in "self documenting" options in a way that'd let you ask (programmatically) both "what options does this instrumentation support?" and "what are the configuration values for this 'instance' of the instrumentation, and which values differ from the default?"

fbogsany: The boolean option weirds me out a bit, probably because we had something similar for DB instrumentation that we ended up deprecating in favour of a tristate option supporting obfuscation. I wonder if we're going to end up in the same place with this.

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -70,6 +69,14 @@ def annotate_span_with_response!(span, response)
             def tracer
               Net::HTTP::Instrumentation.instance.tracer
             end
+
+            def config
+              Net::HTTP::Instrumentation.instance.config
+            end
+
+            def uri_from_req(req)
+              req.uri || URI.join("#{USE_SSL_TO_SCHEME[use_ssl?]}://#{@address}:#{@port}", req.path)
```
fbogsany, line 78: This is awkward - we're piecing together the URL string so the helper can pull it apart, which is somewhat perversely inefficient when we already have the fragments we need in hand. IIRC we did a bunch of work to reduce allocations & string manipulation as much as possible in this instrumentation.

chrisholmes: I think this is from one of my commits on the earlier PR. I'm a bit rusty on this since it's been a while, but I think my reasoning can be summarised as:
- req.uri is not guaranteed to be present as it also accepts a path only parameter which is a string
- I wanted to set the `http.url` so that it's consistent across clients. The services teams run have a few different http clients as dependencies and I find it really annoying not knowing which attributes to use when searching for a particular endpoint's client spans.
- The original PR added a hide query params option to emit query params as they often include sensitive information.  This was done using the `URI` class' `#query=` method.

I think that I probably would be happy about not setting `http.url` if we are consistently setting the `http.target` using `uri.path, `uri.request_uri`, or `path` as that should always be possible.

If we choose to not construct a URI then we can probably do something by splitting the path String at `&` to omit query params

instrumentation_helpers/lib/opentelemetry/instrumentation_helpers/http/instrumentation_options.rb
ericmustin: spent a bit of time thinking through some ideas this morning. I'm starting to think this is going to be a separate issue/PR from the request attribute http helpers, and we should scope down this pr to only address the request attribute helper stuff.

tl;dr is I am reviewing how dd-trace-rb instrumentation handles organizing [settings](https://github.com/DataDog/dd-trace-rb/blob/e391d2eb64d3c6151a4bdd2710c6a8c7c1d57457/lib/ddtrace/contrib/active_record/configuration/settings.rb#L11), as we were able to programmatically expose setting info via [diagnostic logs](https://github.com/DataDog/dd-trace-rb/blob/e391d2eb64d3c6151a4bdd2710c6a8c7c1d57457/lib/ddtrace/diagnostics/environment_logger.rb#L262) in a way that I think is similar to the needs for `self.documenting` instrumentation, and I think the approach taken there could be applied here.

also Apologies if my terminology or pseudo code is a bit off.

Anyway, I think allowing programmatic control over instrumentation options involves refactoring (or at least further review as it's been awhile) of `Instrumentation::Base` class,  and likely an introduction of something like a `Instrumentation::Base::Settings` or `Instrumentation::Base::Configuration` class, which specific instrumentation can inherit from, and augment. This class would basically encapsulate all the various `Instrumentation::Base.option(option_details)` invocations sprinkled in the Instrumentation class, and be a single source of truth for what keys are available for an instrumentation's options. Ideally it would expose some helper methods that could, for example, build a hash of all the instrumentations `option_name / default / validate` configuration settings, and then instrumentations like say `OpenTelemetry::Instrumentation::Bunny` could simply expose method like `default_configuration`, that could gather information about the settings class (or instantiate a new settings class instance, etc) without us necessarily having to create an instance of the entire instrumentation.

I haven't really fully spec'd this out but, i think for this PR, i'd rather keep the old/somewhat hacky pattern of just manually adding `.option(:hide_query_params,...` method calls to all the relevant instrumentation, and scoping this PR to only be a helper for the http request attributes.

As this has been sitting on my plate for awhile, and the other comments are all pretty reasonable and straightforward, it probably doesn't make sense to tie up releasing that work with a chunky refactor of instrumentation.

wdyt?

ericmustin: been working through this a bit more...i think i have something that is maybe managable to include in this pr, gonna try to push it up by eod

ericmustin:
so, i added an approach to better manage default settings, by using an instrumentation::settings class. And i've had a chance to look at this with fresh eyes and I don't like the implementation at all upon further reflection. I think it would make sense to throw out this approach and do something much lighter/less controversial such as exposing another block method (`configurable`, maybe?) on the instrumentation::base class similar to `compatible`, and in that block `option` can be invoked lazily against a set of `name/default/validate` combinations that are defined/stored in constants on the instrumentation class. will try to update this pr shortly to reflect new approach

ericmustin:
i've signed the cla multiple times at this point 😢

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/instrumentation.rb
ericmustin: i'm removing the `InstrumentationOptions` portion of this for now,  i think there's a few related issues and the scope (add better config option declaration and documentation syntax, add deprecation support, add env var for config option support, etc) has enough scope to warrant a larger PR, and i don't want that to block the request attribute helpers work

instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
ericmustin: I agree and think this can be accomplished via inheritance but as per https://github.com/open-telemetry/opentelemetry-ruby/pull/946#discussion_r723467509 i'm pulling out all the instrumentationoptions stuff into a seperate pr

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
ericmustin: i tried to clean the request attributes helper up to both merge in the http client context attribs automagically, and also expose some more optional arguments so that it's flexible enough to not have to construct a URI in the case where it doesn't already exist.

chrisholmes:
@ericmustin is there anything that I can help with to get this PR ready to merge?

ericmustin:
@chrisholmes hey, sorry, iirc i wasn't able to get reviews or feedback here and eventually forgot about it and moved onto other things. I can rebase, clean it up, resolve conflicts otw and i'll ask nudge the right folks to get this merged and approved. sorry that this fell through the cracks.

ericmustin:
@chrisholmes see https://github.com/open-telemetry/opentelemetry-ruby/pull/1133

plantfansam:

To do that, you can:

ericmustin:
no clue where this stands, was nit'd to death and the repo structure and instrumentations have changed pretty significantly, just going to close it.

Issue #857, opened 2021-07-01
OTEL_EXPORTER_OTLP_TRACES_HEADERS / OTEL_EXPORTER_OTLP_HEADERS should allow values with multiple equals signs


arielvalentin:
I have run into an issue where exporter headers cannot be parsed if the values contain equals signs, e.g. `token` with an encoded value that contains an equals sign:

```
irb(main):001:0> require 'csv'
=> true
irb(main):002:0> headers = "token=abccafas="
=> "token=abccafas="
irb(main):003:0> CSV.parse(headers, col_sep: '=', row_sep: ',').to_h
Traceback (most recent call last):
        5: from /usr/bin/irb:23:in `<main>'
        4: from /usr/bin/irb:23:in `load'
        3: from /Library/Ruby/Gems/2.6.0/gems/irb-1.0.0/exe/irb:11:in `<top (required)>'
        2: from (irb):3
        1: from (irb):3:in `to_h'
ArgumentError (wrong array length at 0 (expected 2, was 3))
```

This should be allowed per the [specification](https://github.com/w3c/baggage/blob/main/baggage/HTTP_HEADER_FORMAT.md#value). 

> Note, value MAY contain any number of the equal sign (=) characters. Parsers MUST NOT assume that the equal sign is only used to separate key and value.

We ran in to a different problem when working with the Golang SDK were it expects the keys and values to be CGI escaped so I started a discussion in their repo to clarify why they do this in their implementation: 

https://github.com/open-telemetry/opentelemetry-go/discussions/2050  

There isn't any consistency in other SDKs so I wonder if this ambiguity should be resolved in the specification:

| language | strategy | code |
| --- | --- | --- |
| golang | `QueryUnescape` | https://github.com/open-telemetry/opentelemetry-go/blob/main/exporters/otlp/otlptrace/internal/otlpconfig/envconfig.go#L180 |
| js | [`decodeURIComponent`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/decodeURIComponent) | https://github.com/open-telemetry/opentelemetry-js/blob/f2a958dfeead4f6ae20901ab3fac04e4fc424a3f/packages/opentelemetry-core/src/baggage/utils.ts#L42 |
[...]

PR #948, opened 2021-09-19
fix: OTLP Export Header Format

arielvalentin:
This updates OTLP header parser to conform to the [specification](https://github.com/w3c/baggage/blob/main/baggage/HTTP_HEADER_FORMAT.md#value):

> **key**
> ASCII string according to the token format, defined in RFC2616, Section 2.2.
> Leading and trailing whitespaces (OWS) are allowed but MUST be trimmed when converting the header into a data structure.

> **value**
> A value contains a URL encoded UTF-8 string.
> Leading and trailing whitespaces (OWS) are allowed but MUST be trimmed when converting the header into a data structure.
> Note, value MAY contain any number of the equal sign (=) characters.
> Parsers MUST NOT assume that the equal sign is only used to separate key and value.

Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/857


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -50,7 +50,6 @@ def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_
                        metrics_reporter: nil)
           raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" if invalid_url?(endpoint)
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || compression == 'gzip'
-          raise ArgumentError, 'headers must be comma-separated k=v pairs or a Hash' unless valid_headers?(headers)
```
fbogsany, line 53: Do we want to lose this validation? Does `parse_headers` have any failure cases that could `raise`? If so, will the error be obvious?

arielvalentin: Looking at the other SDK implementations, there wasn't any validation there as far as I could tell, so I had mixed feelings about keeping the validation. E.g. It looks like blank or invalid values are omitted but do not raise errors. 

Do you feel strongly that we should fail-fast here?

Also, I may have missed some test cases that could reveal some issues with this PR, e.g. Invalid Byte Sequence. I will add some for those I could think of.

fbogsany: The spec says:

> The API or SDK may _fail fast_ and cause the application to fail on initialization, e.g. because of a bad user config or environment

We've generally tried to be helpfully verbose with initialization errors rather than silently ignoring config problems. I'd like to retain that, if possible. We don't *have* to, though.

arielvalentin: I have mixed feelings about failing-fast in production environments because that seems to run contrary to the experience with other components like instrumentation, which intentionally ignores errors to not interfere with normal application use. 

Part of me wants to guide our users to write a custom error handler that would "fail fast", giving them some control of what to ignore and what to address, however that could be daunting for them in production environments when sifting through errors. 

That lead me down the path of reporting errors using `OpenTelemetry.handle_error`. If it is indeed our preference to Fail-Fast when it comes to Exporter configurations then I will go back to the fail-fast approach and raise `ArgumentError` instead.

fbogsany: > I have mixed feelings about failing-fast in production environments because that seems to run contrary to the experience with other components like instrumentation, which intentionally ignores errors to not interfere with normal application use.

The choice to ignore errors in instrumentation [aligns with the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/error-handling.md#basic-error-handling-principles):

> ..., but MUST NOT cause the application to fail later at run time

Initialization is a different matter.

> Part of me wants to guide our users to write a custom error handler that would "fail fast", giving them some control of what to ignore and what to address, however that could be daunting for them in production environments when sifting through errors.

I think that's good guidance.

> That lead me down the path of reporting errors using `OpenTelemetry.handle_error`. If it is indeed our preference to Fail-Fast when it comes to Exporter configurations then I will go back to the fail-fast approach and raise `ArgumentError` instead.

Any errors raised during configuration are passed to the error handler anyway, so we should simply `raise` and let `configure` handle the error: https://github.com/open-telemetry/opentelemetry-ruby/blob/666b81204fc5d0a34744d9e88788127897becd01/sdk/lib/opentelemetry/sdk.rb#L71

fbogsany: :+1: `raise hell`

arielvalentin: I drafted something but gotta say. I don't love it. Need to sleep on this.

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -148,31 +148,68 @@
...
+                     end
+                   }.must_raise(ArgumentError)
+        _(error.message).must_match(/headers/i)
 
-        exp = with_env('OTEL_EXPORTER_OTLP_HEADERS' => 'a=,c=hi%F3,,token=abc455==afs') do
-          OpenTelemetry::Exporter::OTLP::Exporter.new
-        end
-        _(exp.instance_variable_get(:@headers)).must_equal('token' => 'abc455==afs')
+        error = _(){
+                     with_env('OTEL_EXPORTER_OTLP_TRACES_HEADERS' => 'a = ') do
```
fbogsany: I'd like to clean up all these tests at some point, to remove the duplication between `OTEL_EXPORTER_OTLP_` and `OTEL_EXPORTER_OTLP_TRACES_` prefixes. It's pretty pervasive in this file, so let's do that in a subsequent PR rather than polluting this one.

fbogsany (approved):

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
arielvalentin: Sounds good. I don't often use minitest spec format and teams usually have very specific style rules about how they prefer test layout. It occurs now to me that I don't know what the maintainers preferred style is here. 

E.g. do y'all prefer using a `describe` block per variation and an `it` block per example?

Issue #947, opened 2021-09-18
Add fallback for applications that bypass RubyGem to load dependencies


arielvalentin:
The GitHub monolith bypasses RubyGems to load dependencies and as a result is unable to install any auto-instrumentation gems that use [`Gem.loaded_specs`](https://github.com/open-telemetry/opentelemetry-ruby/search?q=Gem.loaded_specs) to determine compatibility. 

@ericmustin implemented [several fixes in `ddtrace`](https://github.com/DataDog/dd-trace-rb/pull/1510) to mitigate this issue. We discussed introducing a helper method but then also pointed out that there are some cases where gem authors do not include a `VERSION` constant or follow a consistent pattern to namespace the constant. 

We may want to consider adding a helper similar to what @ericmustin [described in one of the PRs](https://github.com/DataDog/dd-trace-rb/pull/1510#issue-895859173):

```ruby
# This allows instrumentation authors to provide a default value
Datadog::Utils::Gem.loaded_specs_version('dogstatsd-ruby') { Datadog::Statsd::VERSION }
```

That would make it so instrumentation authors could do something like this:

```ruby
module OpenTelemetry
  module Instrumentation
    module MyGem
      class Instrumentation < OpenTelemetry::Instrumentation::Base
        compatible do
          loaded_specs_version("my-gem") { MyGem::VERSION } <= "2.0.1"
        end
      end
    end
  end
end
```

Longer term I propose that the `Instrumentation::Base` DSL could possibly include a declaration that uses [RubyGems pessimistic version constraint](https://guides.rubygems.org/patterns/#pessimistic-version-constraint) e.g.:

```ruby

module OpenTelemetry
  module Instrumentation
    module MyGem
      class Instrumentation < OpenTelemetry::Instrumentation::Base
        supported_versions '>= 2.2.0', '< 2.3.0'
      end
    end
  end
end

# When the version is located in an submodule:  
module OpenTelemetry
  module Instrumentation
    module MyGem
      class Instrumentation < OpenTelemetry::Instrumentation::Base
        supported_versions '>= 2.2.0', '< 2.3.0', "MyGem::ChildModule::GrandChild::VERSION"
      end
    end
  end
end
```

**Related**
- [Slack Conversation](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1631894529019900)
- [Fix in ddtrace](https://github.com/DataDog/dd-trace-rb/issues/1505)
[...]

PR #949, opened 2021-09-19
fix: Use Kafka::VERSION fallback for compatibility

arielvalentin:
Some applications bypass Ruby Gems when loading dependencies,
which results in a `nil` variable being returned by `Gem.loaded_specs`.

This change adds a fallback to use `Kafka::VERSION` constant to check
the minimum version when one is not available in Ruby Gems.

See https://github.com/open-telemetry/opentelemetry-ruby/issues/947

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -180,6 +180,29 @@ services:
     image: memcached:alpine
     command: memcached -m 64
 
+  zookeeper:
+    image: confluentinc/cp-zookeeper:latest
+    ports:
+      - 2181:2181
+    environment:
+      ZOOKEEPER_CLIENT_PORT: 2181
+      ZOOKEEPER_TICK_TIME: 2000
+
+  kafka:
+    image: confluentinc/cp-kafka:latest
+    ports:
+      - "9092:9092"
+      - "29092:29092"
+    environment:
+      KAFKA_BROKER_ID: 1
+      KAFKA_ZOOKEEPER_CONNECT: zookeeper:2181
+      KAFKA_ADVERTISED_LISTENERS: PLAINTEXT://localhost:29092,PLAINTEXT_HOST://kafka:9092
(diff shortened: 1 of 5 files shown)


ahayworth (approved):

ericmustin (approved):
lgtm (some minor rubocops), especially if it's an issue for users of the ruby kafka instrumentation gem, which it seems to be in this case. Long term, can add incrementally to other instrumentation imo. We also may want to make more clear the different stability guarantees around instrumentation (i'd like to actually have this understood better myself 😅 )

shorter term, aside from instrumentation gems, should we also be concerned about:
```
opentelemetry-ruby % grep -r "spec.add_dependency" ./ | grep -v "'opentelemetry-"

.//exporter/jaeger/opentelemetry-exporter-jaeger.gemspec:  spec.add_dependency 'thrift'
.//exporter/otlp/opentelemetry-exporter-otlp.gemspec:  spec.add_dependency 'google-protobuf', '~> 3.7'
.//resource_detectors/opentelemetry-resource_detectors.gemspec:  spec.add_dependency 'google-cloud-env'
```

As mentioned on the attached issue, some libs don't like to declare their `VERSION` constant  via a `version.rb` file, such as `thrift` as far as i can tell [here](https://github.com/apache/thrift/blob/e3be32112aa8029577b88a9ee431ef67f8a357db/lib/rb/thrift.gemspec#L6). So, definitely open to other ideas on if there's a more generalized way to address this rather than going case by case.

arielvalentin:
Re: only being concerned with instrumentation I _think_ the answer is yes. In our case bundler will check and install any runtime dependencies declared in gemspecs but not development dependencies. Unlike the exporters, the instrumentation gems use development dependencies on the libraries they are trying to patch so the problems should only exist thereZ 

I plan on opening a PRs for other gems but wanted to narrow my focus on those I'm going to start with those I want to use in the monolith.

I'll propose a refactoring PR once a pattern emerges.

fbogsany (approved):

PR #956, opened 2021-09-24
fix: do not replace fanout

SomalianIvan:


instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/span_subscriber.rb
```diff
@@ -10,6 +10,16 @@ module ActionView
       # The SpanSubscriber is a special ActiveSupport::Notification subscription
       # handler which turns notifications into generic spans, taking care to handle
       # context appropriately.
+
+      def self.subscribe(pattern = nil, callable = nil)
```
robertlaurin, line 14: This is a hacky version of this call path
https://github.com/rails/rails/blob/v6.1.4.1/activesupport/lib/active_support/notifications.rb#L234-L236
https://github.com/rails/rails/blob/v6.1.4.1/activesupport/lib/active_support/notifications/fanout.rb#L24-L36

The idea is if we push ourselves to the front of the list we will avoid the issue of a caller before us raising and preventing us from starting/ending a span.  This is a similar approach to what exists currently without pushing the ordering responsibility onto the notifier.

ahayworth:
This looks like it might be alright, if we're okay with pushing the span into the notification payload explicitly (which is fine, I suppose). The tests aren't passing so it's hard for me to give a firm thumbs-up. 😆 

Also, FWIW - this isn't nearly as kludgy as was described in the SIG meeting? I feel like you all are far too modest. 😄

SomalianIvan:
> > ch is fine, I suppose). The tests aren't passing so it's hard for me to give a firm thumbs-up. 😆
> 
> On it 😁

Oh dang, `Subscribers::subscribe` method has 3 different signatures 

```
## 6.2
def subscribe(pattern = nil, callable = nil, monotonic: false, &block)
## 6.0
def subscribe(pattern = nil, callable = nil, &block)
## 5.2
def subscribe(pattern = nil, block = proc.new)
```

instrumentation/action_view/lib/opentelemetry/instrumentation/action_view/span_subscriber.rb
```diff
@@ -6,10 +6,27 @@
...
+    # rubocop:disable Style/Documentation
     module ActionView
       # The SpanSubscriber is a special ActiveSupport::Notification subscription
       # handler which turns notifications into generic spans, taking care to handle
       # context appropriately.
+
+      # A very hacky way to make sure that OpenTelemetry::Instrumentation::ActionView::SpanSubscriber
+      # gets invoked first
+      def self.subscribe(pattern = nil, callable = nil)
+        ActiveSupport::Notifications.subscribe(pattern, callable)
```
SomalianIvan, line 18: 1. vanilla subscribe
2. reach inside and push the span_subsciber to the front

robertlaurin: Is it safe to assume that between subscribing and reaching in to pop the subscriber out that it will definitely be the last item appended to the array?

SomalianIvan: yeah I skimmed the sources

ahayworth: If we really wanted to be defensive, we could check to see that the thing we `pop` off is what we expect. But I don't know that we need to.

instrumentation/action_view/test/opentelemetry/instrumentation/action_view/span_subscriber_test.rb
```diff
@@ -38,6 +38,19 @@
...
+
+    expect do
+      ::ActiveSupport::Notifications.instrument('bar.foo', extra: 'context')
+    end.must_raise RuntimeError
+
+    _(last_span).wont_be_nil
+    _(last_span.name).must_equal('foo bar')
+    _(last_span.attributes['extra']).must_equal('context')
+  end
+
```
ahayworth, line 53: Did you mean to add this test twice?

ahayworth (approved):
Yay! I had one comment about a possibly duplicated test, but otherwise I find no fault with this! 😄 

I think this is a nice simplification and will make things less brittle + more re-useable in the long run.

Awesome work, thank you!

instrumentation/action_view/test/opentelemetry/instrumentation/action_view/span_subscriber_test.rb
SomalianIvan: Nope

instrumentation/action_view/test/opentelemetry/instrumentation/action_view/span_subscriber_test.rb
```diff
@@ -157,4 +157,40 @@
...
+      ::ActiveSupport::Notifications.instrument('bar.foo', extra: 'context')
+
+      _(last_span).wont_be_nil
+      _(last_span.name).must_equal('foo bar')
+      _(last_span.attributes['extra']).must_equal('context')
+    end
+
+    it 'finishes spans even when other subscribers blow up' do
+      ::OpenTelemetry::Instrumentation::ActionView.subscribe('bar.foo', subscriber)
+      ::ActiveSupport::Notifications.subscribe('bar.foo') { raise 'boom' }
```
robertlaurin, line 185: I think we should change the ordering to this, or have this as a separate test case as well.  It asserts that our forced ordering logic works.
```suggestion
      ::ActiveSupport::Notifications.subscribe('bar.foo') { raise 'boom' }
      ::OpenTelemetry::Instrumentation::ActionView.subscribe('bar.foo', subscriber)
```

In our span subscriber we define the `start` and `finish` methods so we can start and finish our span.

It would be nice to have a subscriber that raises on the start hook, and one that raises on the finish hook in addition to the existing tests I think it would a lot of confidence that this is the right solution.

robertlaurin (commented):
This looks good, looking for adding two more test cases here.

Issue #959, opened 2021-09-27
BatchSpanProcessor#shutdown does not return proper results


ykitamura-mdsol:
**Description of the bug**

The [`BatchSpanProcessor#shutdown` method](https://github.com/open-telemetry/opentelemetry-ruby/blob/11c97228f35ce9338175e63de774a521d593c55a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb#L133-L152) is supposed to return `SUCCESS`, `FAILURE` or `TIMEOUT`:

https://github.com/open-telemetry/opentelemetry-ruby/blob/11c97228f35ce9338175e63de774a521d593c55a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb#L137-L138

But it always returns `nil`.

So if you set multiple span processors, it will raise an exception when the `TracerProvider#shutdown` method is called:

https://github.com/open-telemetry/opentelemetry-ruby/blob/11c97228f35ce9338175e63de774a521d593c55a/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L85

> `max': comparison of Integer with nil failed (ArgumentError)

**Share details about your runtime**

Operating system details: macOS, Darwin 19.6.0
RUBY_ENGINE: "ruby"
RUBY_VERSION: "2.7.3"
RUBY_DESCRIPTION: "ruby 2.7.3p183 (2021-04-05 revision 6847ee089d) [x86_64-darwin19]"

**Share a simplified reproduction if possible**

```rb
gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-exporter-otlp'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'
require "opentelemetry/exporter/otlp"

simple_span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
  OpenTelemetry::SDK::Trace::Export::ConsoleSpanExporter.new
)

batch_span_processor = OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor.new(
  OpenTelemetry::Exporter::OTLP::Exporter.new
)

OpenTelemetry::SDK.configure do |c|
  c.add_span_processor(simple_span_processor)
  c.add_span_processor(batch_span_processor)
end

OpenTelemetry.tracer_provider.shutdown
[...]

PR #960, opened 2021-09-27
Update bsp .shutdown return value

ericmustin:
Addresses regression as mentioned in https://github.com/open-telemetry/opentelemetry-ruby/issues/959

Tests are pretty basic but should be enough to help prevent a similar regression.


fbogsany (approved):
Thanks @ericmustin and @ykitamura-mdsol

fbogsany (approved):

PR #961, opened 2021-09-28
Add example and tests for Concurrent::Async

chrisholmes:
The recent changes to the concurrent-ruby instrumentation also added context propagation support to classes that mixin `Concurrent::Async` so I've added an example and test to demonstrate this behaviour.

It's possible that other features of `concurrent-ruby` will also behave similarly but I haven't explored this.


fbogsany (approved):
Thanks @chrisholmes !

Issue #775, opened 2021-05-26
Documentation Tracking Issue


arielvalentin:
# Summary

_This issue is an action item outcome of the SIG meeting on 2021-05-25_

There are several requests for documentation to this project by both end users and contributors. As we are ready to work on tasks they may merit further exploration and discussion in sub-issues.

# Issues and Tasks

- [ ] [Documentation: Add design rationale to contributing guide](https://github.com/open-telemetry/opentelemetry-ruby/issues/105)
- [x] [Documentation: Add links or information on style to contributing](https://github.com/open-telemetry/opentelemetry-ruby/issues/106)
- [x] [Documentation: Add how to for bug reports](https://github.com/open-telemetry/opentelemetry-ruby/issues/107)
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/872
- [x] https://github.com/open-telemetry/opentelemetry-ruby/pull/691
- [ ] [Span Processors are basically undocumented](https://github.com/open-telemetry/opentelemetry-ruby/issues/397)
- [ ] Create contribution guides for API, SDK, Instrumentation, and Exporters
- [ ] Create a sample Rails application that demonstrates steps required to configure the SDK
- [x] Stretch Goal: Contribute a microservice to [Hipster Shop](https://github.com/GoogleCloudPlatform/microservices-demo)

fbogsany:
- [ ] [Add opentelemetry.io docs](https://github.com/open-telemetry/opentelemetry-ruby/pull/691)

plantfansam:
Bug report template has [been added](https://github.com/open-telemetry/opentelemetry-ruby/pull/890), so I checked that off

plantfansam:
Updated otel website docs here https://github.com/open-telemetry/opentelemetry.io/pull/1452. Probably would be good to automate this process in the `opentelemetry.io` repo 🤔

ahayworth:
Our configuration docs could be better - see https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/88#issuecomment-1193145469

PR #968, opened 2021-09-30
docs: Stable Release Updates

arielvalentin:
- Moved "who's using" section to the website docs to increase visibility
- Added table of contents on the repo homepage

See https://github.com/open-telemetry/opentelemetry-ruby/issues/775

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,13 @@
 
 The Ruby [OpenTelemetry](https://opentelemetry.io/) client.
 
+- [Getting Started][getting-started]
+- [Contributing](#contributing)
+- [Instrumentation Libraries](#instrumentation-libraries)
+- [Versioning](#versioning)
+- [Useful links](#useful-links)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -35,70 +42,6 @@ Maintainers ([@open-telemetry/ruby-maintainers](https://github.com/orgs/open-tel
 
 *Find more about the maintainer role in [community repository](https://github.com/open-telemetry/community/blob/master/community-membership.md#maintainer).*
 
-## Installation
```
robertlaurin, line 38: Do we want to move this info outside of the repo readme?  I know it's not great have docs in two places as they are almost guaranteed to diverge but I feel strongly towards to keeping something like this in the readme.  It's really handy when the simple getting started bit is inside the codebase.  We could push anything beyond the basic "getting started" into the microsite, as it has better support for structuring documentation.

WDYT?

PR #970, opened 2021-10-04
chore: Pin Collector Version

arielvalentin:
Some folks ran into issues running throught the example docs.

This update pins the otel-collector version to a production release to
mitigate issues with breaking changes introduced in the development version of the collector.

https://github.com/open-telemetry/opentelemetry-ruby/discussions/966


ericmustin (approved):
seems fine to me we may want to make add an accompanying issue to remind ourselves to update back to latest once there's a 1.0 release

PR #973, opened 2021-10-06
fix: Prevent high cardinality rack span name as a default

robertlaurin:
It is possible to maintain the high cardinality behaviour by passing in a
url quantization function that simply forwards the URI path.

> Instrumentation MUST NOT default to using URI path as span name, but MAY
> provide hooks to allow custom logic to override the default span name.

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.0.0/specification/trace/semantic_conventions/http.md#name


ericmustin:
I _think_ the failing action controller tests just need to get hardcoded updates?

+1 to @arielvalentin's comment. Though we probably need a broader update to the `automatic instrumentation` section of the opentelemetry.io docs, in the meantime maybe a quick section/note in just the opentelemetry-instrumentation-rack readme would suffice?

Issue #904, opened 2021-08-11
instrumentation-base: Support config option overrides via environment variables


robertlaurin:
We currently support [disabling a specific instrumentation gem via environment variables](https://github.com/open-telemetry/opentelemetry-ruby/blob/f7108f2089794c1a407e395ef6b6fa8b47db142a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L297-L305).  We should look to add support for modifying any of the configuration options for a specific library via environment.  

There will be some expected limitations around the options that expect [callables](https://github.com/open-telemetry/opentelemetry-ruby/blob/f7108f2089794c1a407e395ef6b6fa8b47db142a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L68).

Issue #903, opened 2021-08-11
instrumentation-base: Add config option deprecation support


robertlaurin:
Provide a simple mechanism for instrumentation library authors to declare a configuration option is deprecated, and point to the alternative.
https://github.com/open-telemetry/opentelemetry-ruby/pull/894#discussion_r685437541

Issue #899, opened 2021-08-10
instrumentation-base / configurator: support global configuration for all instrumentation


ahayworth:
We want to allow the ability to provide configuration for all instrumentation in one pass; and to also support one instrumentation reading the configuration of another instrumentation. There are a few use-cases:
- This allows us to more easily support various combinations of `c.use / c.use_all` with either the `-instrumentation-all` gem or individually installed gems.
- We can more flexibly express dependencies between instrumentation (eg: between `-rails` and its sub-gems, or between something that uses Faraday and the `-faraday` instrumentation). Currently, we override user options when configuring inter-instrumentation dependencies.

Issue #898, opened 2021-08-10
instrumentation-base: Support enums as an option 


ahayworth:
Currently we typically validate "enum-like" options as a callable, like so:

```ruby
option :foo, default: :bar, validate: ->(opt) { %i[bar baz].include?(opt) }
```

Instead, it'd be nice to support an enum option. Perhaps something like:

```ruby
option :foo, default: :bar, validate: =[:bar, :baz]
```

ericmustin:
addressed by https://github.com/open-telemetry/opentelemetry-ruby/pull/994

PR #977, opened 2021-10-08
[wip]: Add config improvements

ericmustin:
### Summary 

WIP, but meant to address:
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/904
- [ ] https://github.com/open-telemetry/opentelemetry-ruby/issues/903
and possibly while i'm in here
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/898
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/899

Originally some of this work was in https://github.com/open-telemetry/opentelemetry-ruby/pull/946 but i made a new branch based on top of it that includes only the config specific code.

### Notes

This PR is an attempt to clean up a bit the way we handle configuration options in instrumentation. It does this by
- Establishing a constant on the instrumentation-base and -`<subclasses>` that is opt-in for storing all configuration option entries (`name`, `default`, `validator`), as well as storing all deprecated configuration options. This will allow us to have a canonical home for the valid configuration options of an instrumentation without needing to instantiate an instance of it. This ought to make it easier to track config options over time, document them, deprecate them, and so on.
- Add convenience methods to `-instrumentation-base` class for both generating a hash of all default options for the subclass and also applying that hash as options. The idea would be to add these helpers and then migrate individual instrmentations to this approach in a followup PR. For now, only `http.rb` instrumentation is using it as an example.
- Add environment variable support for all instrumentation configuration options (excluding callables). This uses the syntax `OTEL_RUBY_INSTRUMENTATION_<INSTRUMENTATION_NAME>_<INSTRUMENTATION_OPTION_NAME>`. For ex: `OTEL_RUBY_INSTRUMENTATION_HTTP_HIDE_QUERY_PARAMS=false`. These environment variables take precedence _at runtime_ over local configuration, this PR also include heuristics for coercing them to the appropriate `validate` types (`boolean`, `integer`, `array`, `string`).
- An easy storage and checking mechanism for determining whether a config option that's being used is `deprecated`. This is largely just a skeleton still but i'm open to ideas of what we want to expose, in my mental model right now there would be perhaps a list of deprecated options, as well as the ability to either _warn_ (via a log line, this would be default behavior)  or _raise_ (via an exception, in cases where a config option was potentially doing something _very bad_ like causing a memory leak or what have you).

Anyway, PR was getting kinda chunky so wanted to push it up, it's only being used by the `http` instrumentation but after we get agreement on a form/syntax/etc it would be easy (and can be done in followup prs) to migrate the other instrumentation over, and then once everything is migrated it would be pretty straightforward to add some rake tasks that generate documentation of config options per instrumentation auto-magically.

### DIFF
--- a/instrumentation/http/lib/opentelemetry/instrumentation/http/instrumentation.rb
+++ b/instrumentation/http/lib/opentelemetry/instrumentation/http/instrumentation.rb
@@ -25,16 +25,6 @@ module OpenTelemetry
           !(defined?(::HTTP) && Gem.loaded_specs['http']).nil?
         end
 
-        # option :hide_query_params, default: true, validate: :boolean
-
-        # def self.default_options
-        #   if defined?(super)
-        #     super.merge(DEFAULT_OPTIONS)
-        #   else
-        #     DEFAULT_OPTIONS
-        #   end
-        # end
-
         initialize_default_options
 
         private


robbkidd:
Here's a convenience [diff of only the Environment changes](https://github.com/ericmustin/opentelemetry-ruby/compare/add_instrumentation_helpers_gem...add_config_improvements) with the changes from #946 excluded.

PR #981, opened 2021-10-13
feat: add instrumentation for rspec

chrisholmes:
This PR covers adding instrumentation for RSpec using a RSpec formatter. It could then be used as part of a CI/CD pipeline that is configured to emit traces. At this point, parent context would need to passed manually from other processes, but I understand this being discussed as [an addition to the spec](https://github.com/open-telemetry/opentelemetry-specification/issues/740).

There are two ways to enable instrumentation either by configured Otel to use the instrumentation or by adding the formatter to RSpec. You can see examples of this in `examples/`.

Some things :
- The formatter is using `::attach` and `::detach` and a stack to add and remove span context. I think this is appropriate considering how example and example groups are currently executed, but I'm assuming this might break if a later RSpec version offers multi-threaded execution.
- Should test framework instrumentation be auto-instrumented? 
- I've had to add a monkey patch at the `test/rspec_patches.rb` so that RSpec doesn't clobber minitest's `#describe` method. This might not be needed if specs were written in RSpec, but I don't think this is a good idea as there would then be multiple test frameworks used in this repo.

Also, the README is still a WIP

### DIFF
--- a/instrumentation/rspec/Appraisals
+++ b/instrumentation/rspec/Appraisals
@@ -1,0 +1,16 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+## TODO: Include the supported version to be tested here.
+## Example:
+# appraise 'rack-2.1' do
+#   gem 'rack', '~> 2.1.2'
+# end
+
+# appraise 'rack-2.0' do
+#   gem 'rack', '2.0.8'
+# end
+
--- a/instrumentation/rspec/Appraisals
+++ b/instrumentation/rspec/Appraisals
@@ -6,11 +6,6 @@
 
 ## TODO: Include the supported version to be tested here.
 ## Example:
-# appraise 'rack-2.1' do
-#   gem 'rack', '~> 2.1.2'
+# appraise 'rspec-3.10' do
+#   gem 'rspec', '~> 3.10.0'
 # end
-
-# appraise 'rack-2.0' do
-#   gem 'rack', '2.0.8'
-# end
-
--- a/instrumentation/rspec/README.md
+++ b/instrumentation/rspec/README.md
@@ -49,6 +49,10 @@ end
 
 If you need to test trace behaviour in your specs then you should be able to use a custom tracer provider and the instrumentation's output should not interfere with your specs.
 
+### Sampling
+
+To avoid spans from being dropped, which will mean you lose insight into your specs, you may want to set sampling to 'ALWAYS_ON'. The easiest way to do this is by setting the `OTEL_TRACES_SAMPLER` environment variable to `always_on`.
+
 ## Examples
 
 Example usage can be seen in the `/example` directory [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rspec/example)
--- a/instrumentation/rspec/README.md
+++ b/instrumentation/rspec/README.md
@@ -1,6 +1,6 @@
-# OpenTelemetry Rspec Instrumentation
+# OpenTelemetry RSpec Instrumentation
 
-Todo: Add a description.
+The RSpec instrumentation is a community-maintained instrumentation for the [RSpec][rspec-home] BDD testing framework.
 
 ## How do I get started?
 
@@ -22,17 +22,36 @@ OpenTelemetry::SDK.configure do |c|
 end
 ```
 
-Alternatively, you can also call `use_all` to install all the available instrumentation.
+Alternatively, you can add it directly to RSpec as a customer formatter:
 
 ```ruby
-OpenTelemetry::SDK.configure do |c|
-  c.use_all
+RSpec.configure do |config|
+  config.formatter = OpenTelemetry::Instrumentation::RSpec::Formatter
 end
 ```
 
+### Using multiple tracer providers
+
+By default the instrumentation will use the global tracer provider.
+
+If you want the instrumentation to use something other than the global tracer provider you can configure the instrumentation with a custom tracer provider using the `OpenTelemetry::Instrumentation::RSpec::Formatter` constructor:
+
+```ruby
+exporter = OpenTelemetry::SDK::Trace::Export::InMemorySpanExporter.new
+span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(exporter)
+tracer_provider = OpenTelemetry::SDK::Trace::TracerProvider.new
+tracer_provider.add_span_processor(span_processor)
+
+RSpec.configure do |config|
+  config.formatter = OpenTelemetry::Instrumentation::RSpec::Formatter.new(config.output_stream, tracer_provider)
+end
+```
+
+If you need to test trace behaviour in your specs then you should be able to use a custom tracer provider and the instrumentation's output should not interfere with your specs.
+
 ## Examples
 
-Example usage can be seen in the `./example/trace_demonstration.rb` file [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rspec/example/trace_demonstration.rb)
+Example usage can be seen in the `/example` directory [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rspec/example)
 
 ## How can I get involved?
 
@@ -50,3 +69,4 @@ The `opentelemetry-instrumentation-rspec` gem is distributed under the Apache 2.
 [ruby-sig]: https://github.com/open-telemetry/community#ruby-sig
 [community-meetings]: https://github.com/open-telemetry/community#community-meetings
 [discussions-url]: https://github.com/open-telemetry/opentelemetry-ruby/discussions
+[rspec-home]: https://rspec.info
--- a/instrumentation/rspec/README.md
+++ b/instrumentation/rspec/README.md
@@ -1,0 +1,52 @@
+# OpenTelemetry Rspec Instrumentation
+
+Todo: Add a description.
+
+## How do I get started?
+
+Install the gem using:
+
+```
+gem install opentelemetry-instrumentation-rspec
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-rspec` in your `Gemfile`.
+
+## Usage
+
+To use the instrumentation, call `use` with the name of the instrumentation:
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Rspec'
+end
+```
+
+Alternatively, you can also call `use_all` to install all the available instrumentation.
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use_all
+end
+```
+
+## Examples
+
+Example usage can be seen in the `./example/trace_demonstration.rb` file [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rspec/example/trace_demonstration.rb)
+
+## How can I get involved?
+
+The `opentelemetry-instrumentation-rspec` gem source is [on github][repo-github], along with related gems including `opentelemetry-api` and `opentelemetry-sdk`.
+
+The OpenTelemetry Ruby gems are maintained by the OpenTelemetry-Ruby special interest group (SIG). You can get involved by joining us in [GitHub Discussions][discussions-url] or attending our weekly meeting. See the [meeting calendar][community-meetings] for dates and times. For more information on this and other language SIGs, see the OpenTelemetry [community page][ruby-sig].
+
+## License
+
+The `opentelemetry-instrumentation-rspec` gem is distributed under the Apache 2.0 license. See [LICENSE][license-github] for more information.
+
+[bundler-home]: https://bundler.io
+[repo-github]: https://github.com/open-telemetry/opentelemetry-ruby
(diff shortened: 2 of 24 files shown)


instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/version.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module RSpec
+      VERSION = '0.0.0'
```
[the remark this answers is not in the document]
robertlaurin: >that makes sense to me, would then be 0.19.4-experimental, 0.0.1, or 0.0.1-experimental

I would go with `0.0.1`, or `0.0.1-experimental`. I don't actually have a really strong opinion on this it was just a loose suggestion, I was just thinking it might be nice to signal the uncertainty as this is likely to evolve.  

All that said, I think `0.0.1` is reasonable as technically none of the instrumentation gems have any stability guarantee until there's a formal spec for it and we 1.0.0 them.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,101 @@
...
+        end
+
+        def initialize(output = StringIO.new, tracer_provider = OpenTelemetry.tracer_provider)
+          @spans_and_tokens = []
+          @output = output
+          @tracer_provider = tracer_provider
+        end
+
+        def tracer
+          @tracer ||= @tracer_provider.tracer('RSpec')
```
robertlaurin, line 31: This would keep it consistent with the other auto instrumentation packages.
```suggestion
          @tracer ||= @tracer_provider.tracer('OpenTelemetry::Instrumentation::RSpec', OpenTelemetry::Instrumentation::RSpec::Version)
```

chrisholmes: sorted

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/version.rb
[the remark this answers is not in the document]
chrisholmes: done

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,102 @@
...
+          span = tracer.start_span(example.description, attributes: attributes, start_timestamp: current_timestamp)
+          track_span(span)
+        end
+
+        def example_finished(notification)
+          pop_and_finalize_span do |span|
+            result = notification.example.execution_result
+            notification.example.metadata
+
+            span.set_attribute('result', result.status.to_s)
```
ericmustin, line 74: I think one of the key value-adds of test instrumentation is the ability to track failing/flaky tests. In OpenTelemetry I _think_ (we're kinda invented spec as we go here) span.status is the appropriate place to record that information. I'd like to see span.status get set here if possible if we _know_ it's a failed test, `passed` / `failed` / `skipped` appear to be the possible values (I actually am rustier with rspec than i'd like to admit so entirely possible i'm wrong), so i think in the case of `result.status == :failed` we should set the span status to error. This is the approach some OSS vendor approaches take, like our good friends at [ddog](https://github.com/DataDog/dd-trace-rb/blob/988431adf07bb09e41bb845d4536816b0e61579a/lib/datadog/ci/test.rb#L58-L71), and also an approach taken by other OpenTelemetry Test instrumentations floating around in the community, like @rakyll's work in go-test-trace: https://github.com/rakyll/go-test-trace/blob/732baad7c542b6afa16c38eca52c34172a619b5e/main.go#L149-L151

An example of how to use the `Trace::Status` helpers can be found here: https://github.com/open-telemetry/opentelemetry-ruby/blob/87e0ccb0e30bf179ee826d67020ba600216afd29/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb#L49. Wdyt? Does this make sense for your use case specifically?

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,102 @@
...
+        def example_group_finished(notification)
+          pop_and_finalize_span
+        end
+
+        def example_started(notification)
+          example = notification.example
+          attributes = {
+            'location' => example.location.to_s,
+            'full_description' => example.full_description.to_s,
+            'described_class' => example.metadata[:described_class].to_s
```
ericmustin, line 63: I'd like to see `test.` prefix on the attribute names, or just some prefix if there's a more appropriate convention, not picky, just feels like a nice housekeeping thing to do while it's still greenfield

instrumentation/rspec/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+Alternatively, you can add it directly to RSpec as a customer formatter:
+
+```ruby
+RSpec.configure do |config|
+  config.formatter = OpenTelemetry::Instrumentation::RSpec::Formatter
+end
+```
+
+### Using multiple tracer providers
```
ericmustin, line 33: This is great, i appreciate the detail here. My only nit is I think it should be encouraged to `Always` sample, since a failing test that gets dropped due to random sampling sort've defeats the whole point of rspec instrumentation imo. I guess that's the implicit default behavior in the code example anyway, but I wouldn't mind seeing this be made explicit.

ericmustin (changes requested):
Sorry for the delay in reviewing here, apologies.

I think, this is super close imo. Great work, appreciate the awesome contribution.

Since we're treating this as experimental i'm in favor of avoiding as much bikeshedding as possible over stuff like span attribute naming conventions or whatever, and just nailing down the concrete bits and getting this shipped. Most of my feedback falls into the category of nits and i'm open to what you think are the right choices here.

Setting span status if the `result.status` is a failing test looks like the only missing bit here to me. Once that's done and the other comments have either been addressed or resolved according to your best judgement, I'd like to approve this and get it out the door.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
chrisholmes: Yes, I was considering `example.` or `rspec.` in order to make it specific to this test framework

I was thinking we should avoid `test.` in case it become part of the standard later and this library conflicts with it. Lot’s of speculation there anyway.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
chrisholmes: Yes, that works for me.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,102 @@
...
+          }
+          span = tracer.start_span(example.description, attributes: attributes, start_timestamp: current_timestamp)
+          track_span(span)
+        end
+
+        def example_finished(notification)
+          pop_and_finalize_span do |span|
+            result = notification.example.execution_result
+            notification.example.metadata
```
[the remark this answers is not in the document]
chrisholmes: Just something I left in accidentally probably.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,102 @@
...
+
+            span.set_attribute('result', result.status.to_s)
+
+            if (exception = result.exception)
+              span.record_exception(exception)
+              span.set_attribute('message', exception.message) if exception.is_a? ::RSpec::Expectations::ExpectationNotMetError
```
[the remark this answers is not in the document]
chrisholmes: Initially, it was to try and show a distinction between examples that error and examples that fail, but in practice it just adds duplication so I’m happy to remove.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
ericmustin: yea, no strong opinions, `rspec` seems hard to argue with and future-proof enough for now?

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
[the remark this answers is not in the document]
chrisholmes: I had a look at this. My current preference is to keep the failure message attached to the span, as `rspec.example.failure_message` rather than on an attached span event so that it can be found easily next to the `result` attribute. A lot of this is dependent on how tracing tools visualize the trace, span, span events.

It would also be convenient to be able to do queries of the span data like:
```
select description, failure_message from spans where result = false
```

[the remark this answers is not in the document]
chrisholmes: Actually, would adding this as the description to the `span.status` be sensible?

instrumentation/rspec/README.md
chrisholmes: I’ve added something for this

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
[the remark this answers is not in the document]
ericmustin: hm, yes I think this would fit nicely into the status description if possible.

Thinking a bit more, I am fine with leaving it attached in a few places tbh, not the end of the world, we'd just probably want to clean it up before moving it out of experimental.  

I recognize in practice that basically all observability UIs out there do a poor job of showing span events, but imo it's a bit of their responsibility to support OTEL better, less than it is our responsibility to ensure compatibility. That being said i don't want having this live in -contrib degrade your experience using it as a user, so fine with this for now.

ericmustin (approved):
Lgtm! ty for implementing that feedback. As discussed, ideally we'd set the status description as well on error, but not blocking imo. Looks like a small conflict with main but rebasing ought to clean that up, otherwise keen to see this merged and released.

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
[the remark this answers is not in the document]
chrisholmes: OK, I've added it as the description now.

chrisholmes:
@ericmustin, I think I'm finished with making changes. Please merge when convenient.

ericmustin:
@chrisholmes i'm not a maintainer so i can't merge, i've asked for another review here and then we'll get this merged. Great work again, thanks!

instrumentation/rspec/Appraisals
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+## TODO: Include the supported version to be tested here.
+## Example:
+# appraise 'rspec-3.10' do
```
SomalianIvan, line 9: are you planning to uncomment this?

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -0,0 +1,102 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module RSpec
+      # An RSpec Formatter that outputs Otel spans
+      class Formatter
+        attr_reader :output
+
+        ::RSpec::Core::Formatters.register self, :example_started, :example_finished, :example_group_started, :example_group_finished, :start, :stop
+
+        @clock = Time.method(:now)
```
SomalianIvan, line 19: wondering what's the difference between `Time.method(:now).call` and `Time.now`?

instrumentation/rspec/opentelemetry-instrumentation-rspec.gemspec
```diff
@@ -0,0 +1,48 @@
...
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 2.5.0'
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.0.0.rc3'
```
SomalianIvan, line 28:
```suggestion
  spec.add_dependency 'opentelemetry-api', '~> 1.0'
```

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
SomalianIvan: ahh I see, its to save the reference to the original methods if tests alter Time

SomalianIvan (approved):
This is 🥇 !! Sending my fake approval

robertlaurin (approved):
🥇

PR #987, opened 2021-10-18
Docs: add aliases

chalin:
- Contributes to https://github.com/open-telemetry/opentelemetry.io/issues/665.
- An alias (in this context) is an OTel website path that will be redirected to this page containing the alias definition. These aliases existed before, they're just being moved.

/cc @austinlparker


chalin:
Anything more to be done before this is merged? Thanks

PR #994, opened 2021-10-26
feat: add default options config helper + env var config option support

ericmustin:
## Summary

This PR follows up/isolates the work in https://github.com/open-telemetry/opentelemetry-ruby/pull/977 with some minor modifications

- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/904
- [x] https://github.com/open-telemetry/opentelemetry-ruby/issues/898 (enum support).

## Details

~This PR introduces convenience methods on the base class so that instrumentation can have a single uniform constant (`DEFAULT_OPTIONS`) and single convenience method (`initialize_default_options`) to declare+initialize their configuration options, that is backward compatible and works alongside current `option` declarations in code.~ _removed_

This PR also introduces environment variable support with the syntax `OTEL_RUBY_INSTRUMENTATION_<INSTRUMENTATION_NAME>_OPTS=<option_name_1>=<option_value_1>;<option_name_2>=<option_value_2>;`

```
OTEL_RUBY_INSTRUMENTATION_HTTP_CONFIG_OPTS' => "query_params=false;some_other_option=true"
```

It uses the `mysql2` instrumentation as the guinea pig instrumentation here. If this format is amendable to folks, can followup with a 2nd PR to pass over and update+test the remaining instrumentation

Additionally, its worth mentioned that, as hinted in #904, env var support has some limitations around `:callable` validation types. `:callable` represents two states 
1. an option that is itself a proc, 
2. an option that uses a custom proc to validate it's input. 
 
~The 1st obviously can't be managed via environment variable without some difficulty, but the the 2nd, in practice is used exclusively in our instrumentation to represent an `:enum` option type where the value is one among a selection of `Symbol` options (ie: `[:low :medium :high]`). Because of this, i've chosen to limit the functionality of Env var config option support for `callables` to always assume the input is a string that should be coerced to a `symbol`. It's a limitation but is clearly defined in the tests, and in practice makes this feature far more valuable than if we included `:callable` validations entirely.~ _callable option types are excluded from environment variable configuration._

_this pr has been updated to include an `:enum` type that is possible to define by passing an Array of Symbols as the `validate` value in the `option` method. This allows us to safely handle environment variable configuration for the many `callable` option types that use a custom proc to validate their input, as all of those custom procs are merely wrappers around an enum._


robertlaurin (commented):
While I was reviewing this I found I was jumping around the code a lot trying to follow what it was doing.  It seems like we could potentially simplify this by splitting out the option names from the env var like you were and just inject that hash into the existing config_options method.  

I'm not sure I understand the benefit of introducing a new default config options format hash in place of the existing option declaration. 

This approach leaves the original app level config unmodified so if there's a problem with the env var passed in and it doesn't validate it falls back to the app level config and then finally to the default.
 
```diff
diff --git a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
index 77ccbeec..14fa71d1 100644
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -146,11 +146,11 @@ module OpenTelemetry
         # a key in the VALIDATORS hash.  The supported keys are, :array, :boolean,
         # :callable, :integer, :string.
         def option(name, default:, validate:)
-          validate = VALIDATORS[validate] || validate
-          raise ArgumentError, "validate must be #{VALIDATORS.keys.join(', ')}, or a callable" unless validate.respond_to?(:call)
+          validator = VALIDATORS[validate] || validate
+          raise ArgumentError, "validate must be #{VALIDATORS.keys.join(', ')}, or a callable" unless validator.respond_to?(:call)
 
           @options ||= []
-          @options << { name: name, default: default, validate: validate }
+          @options << { name: name, default: default, validator: validator, validation_type: validate }
         end
 
         def instance
@@ -260,13 +260,17 @@ module OpenTelemetry
       def config_options(user_config) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
         @options ||= {}
         user_config ||= {}
+        config_overrides = config_overrides_from_env
         validated_config = @options.each_with_object({}) do |option, h|
           option_name = option[:name]
           config_value = user_config[option_name]
+          config_override = coerce_env_var(config_overrides[option_name], option[:validation_type]) if config_overrides[option_name]
 
-          value = if config_value.nil?
+          value = if config_value.nil? && config_override.nil?
                     option[:default]
-                  elsif option[:validate].call(config_value)
+                  elsif option[:validator].call(config_override)
+                    config_override
+                  elsif option[:validator].call(config_value)
                     config_value
                   else
                     OpenTelemetry.logger.warn(
@@ -288,6 +292,40 @@ module OpenTelemetry
         validated_config
       end
 
+      def config_overrides_from_env
+        var_name = name.dup.tap do |n|
+          n.upcase!
+          n.gsub!('::', '_')
+          n.gsub!('OPENTELEMETRY_', 'OTEL_RUBY_')
+          n << '_OPTS'
+        end
+
+        env_config_options = ENV[var_name].split(';')
+        environment_config_overrides = {}
+        env_config_options.each_with_object(environment_config_overrides) do |env_config_option, eco|
+          parts = env_config_option.split('=')
+          option_name = parts[0].to_sym
+          eco[option_name] = parts[1]
+        end
+        environment_config_overrides
+      end
+
+      def coerce_env_var(env_var, validation_type)
+        case validation_type
+        when :array
+          env_var.split(',').map(&:strip)
+        when :boolean
+          env_var.to_s.strip.downcase == 'true'
+        when :integer
+          env_var.to_i
+        when :string
+          env_var.to_s.strip
+        when :callable
+          # Should we try to support this?
+          nil
+        end
+      end
+
       # Checks to see if this instrumentation is enabled by env var. By convention, the
       # environment variable will be the instrumentation name upper cased, with '::'
       # replaced by underscores, OPENTELEMETRY shortened to OTEL_{LANG} and _ENABLED appended.
@@ -295,6 +333,7 @@ module OpenTelemetry
       # will be OTEL_RUBY_INSTRUMENTATION_SINATRA_ENABLED. A value of 'false' will disable
       # the instrumentation, all other values will enable it.
       def enabled_by_env_var?
+        # TODO: Add deprecation notice
         var_name = name.dup.tap do |n|
           n.upcase!
           n.gsub!('::', '_')
diff --git a/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb b/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb
index 8bef8190..da24f8c1 100644
--- a/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb
+++ b/instrumentation/graphql/test/instrumentation/graphql/tracers/graphql_tracer_test.rb
@@ -126,6 +126,18 @@ describe OpenTelemetry::Instrumentation::GraphQL::Tracers::GraphQLTracer do
         span = spans.find { |s| s.name == 'Query.resolvedField' }
         _(span).wont_be_nil
       end
+
+      it 'can be overriden by the env var' do
+        with_env('OTEL_RUBY_INSTRUMENTATION_GRAPHQL_OPTS' => 'enable_platform_field=false') do
+          instrumentation.instance_variable_set(:@installed, false)
+          instrumentation.install(config)
+        end
+
+        SomeGraphQLAppSchema.execute(query_string, variables: { 'id': 1 })
+
+        span = spans.find { |s| s.name == 'Query.resolvedField' }
+        _(span).must_be_nil
+      end
     end
 
     describe 'when platform_authorized is enabled' do
diff --git a/instrumentation/graphql/test/test_helper.rb b/instrumentation/graphql/test/test_helper.rb
index 15369d8a..b07feda1 100644
--- a/instrumentation/graphql/test/test_helper.rb
+++ b/instrumentation/graphql/test/test_helper.rb
@@ -19,3 +19,12 @@ span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(EXPO
 OpenTelemetry::SDK.configure do |c|
   c.add_span_processor span_processor
 end
+
+def with_env(new_env)
+  env_to_reset = ENV.select { |k, _| new_env.key?(k) }
+  keys_to_delete = new_env.keys - ENV.keys
+  new_env.each_pair { |k, v| ENV[k] = v }
+  yield
+  env_to_reset.each_pair { |k, v| ENV[k] = v }
+  keys_to_delete.each { |k| ENV.delete(k) }
+end
```

ericmustin:
> I'm not sure I understand the benefit of introducing a new default config options format hash in place of the existing option declaration.

Just found it a bit cleaner, exposing it as a constant on the class would makes it easy to generate instrumentation documentation via a rake task for example. Don't have strong opinions here

ericmustin:
@robertlaurin 

> This approach leaves the original app level config unmodified so if there's a problem with the env var passed in and it doesn't validate it falls back to the app level config and then finally to the default.

yea, i wasn't sure what the "correct" behavior should be. Happy to take that approach, with the exception in the above diff that I think we need to treat env vars for callables as symbols and can't ignore them, or we need to do separate work here to define an `enum` validation type.

robertlaurin:
> with the exception in the above diff that I think we need to treat env vars for callables as symbols and can't ignore them, or we need to do separate work here to define an `enum` validation type.

Yeah agreed we should add explicit support for enums. 

I just don't know if we should try support config options like [this](https://github.com/open-telemetry/opentelemetry-ruby/blob/c9836ac791d789a96329ae8fbd4b89a48eb5a3f4/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb#L31-L32) where we accept some arbitrary lambda/proc.  That seems a bit scary and error prone. 

```rb
irb(main):006:0> callable = eval "->() { puts \"this is scary\" }"
=> #<Proc:0x00007fe49b07b4d0 (eval):1 (lambda)>
irb(main):007:0> callable.call
this is scary
```

ericmustin:
@robertlaurin right, yea no agreed, that's why in this pr i'm explicitly coercing to a symbol. anyway its just messy either way so i guess an `enum` option is better here

ericmustin:
simplified the base instrumentation class helpers along the lines of the `.diff` above, threw out the `DEFAULT_OPTIONS`  constant approach, and added an `:enum` option validation type that is a bit ugly imo but seems to work. Using `mysql2` as the guinea pig here. If this is agreeable, can followup with a separate PR migrating all the other instrumentation _callable option validation's that are really enums_.

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -303,6 +320,46 @@ def enabled_by_env_var?
...
+
+        environment_config_overrides
+      end
+
+      def coerce_env_var(env_var, validation_type) # rubocop:disable Metrics/CyclomaticComplexity
+        case validation_type
+        when :array
+          env_var.split(',').map(&:strip)
+        when :boolean
+          env_var.to_s.strip.downcase == 'true'
```
robertlaurin, line 351: What happens is the user provides a random word here?  Will it evaluate to false?  

Could we add a test here for that?

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -303,6 +320,46 @@ def enabled_by_env_var?
...
+        when :integer
+          env_var.to_i
+        when :string
+          env_var.to_s.strip
+        when :enum
+          env_var.to_s.strip.to_sym
+        when :callable
+          # Should we try to support this?
```
robertlaurin, line 359: We should probably log warning/error message here to indicate this field is unsupported.

robertlaurin (commented):
Two small comments to address and then I think it's good to go.  Will want to follow up with a smaller PR to modify the option api for better support enums.

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
ericmustin: yup the intention is to (evaluate to false). will add test to confirm

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
ericmustin: agreed, will do

robertlaurin (approved):

PR #999, opened 2021-11-01
fix: change net attribute names to match the semantic conventions spec for http

### LINKED ISSUE #998: net_http instrumentation not following semantic conventions for net.peer.name
**Description of the bug**
Spans created from `net_http` instrumentations are [started with attribute `peer.hostname` and `peer.port`](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb#L28) which are not in the specification. They should be changed to `net.peer.name` and `net.peer.port` [according to the specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes)

If it's not intentionally, I can submit a PR with a fix

blumamir:
fixes #998

### DIFF
--- a/common/README.md
+++ b/common/README.md
@@ -40,7 +40,7 @@ tracer.in_span(
     'http.method' => req.method,
     'http.scheme' => USE_SSL_TO_SCHEME[use_ssl?],
     'http.target' => req.path,
-    'peer.hostname' => @address,
+    'net.peer.name' => @address,
     'peer.port' => @port
   ),
   kind: :client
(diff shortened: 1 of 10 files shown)


arielvalentin (approved):

blumamir:
Thanks for the quick review @arielvalentin .
After you reviewed it, I noticed that the wrong attribute names are also used in other places in the codebase. Update them as well.

fbogsany (approved):

PR #1004, opened 2021-11-02
feat: Initial outline of Metrics API

fbogsany:
This is starting point for the Metrics API. It provides a no-op implementation and the proxy implementation to allow upgrading `Meter`, `MeterProvider` and instruments to the SDK implementations once they're available.

The only functionality of significance is the proxy mechanism and the validation of instrument creation arguments.


metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -0,0 +1,62 @@
...
+        create_instrument(:up_down_counter, name, unit, description, nil) { UP_DOWN_COUNTER }
+      end
+
+      def create_observable_up_down_counter(name, unit: nil, description: nil, callback:)
+        create_instrument(:observable_up_down_counter, name, unit, description, callback) { OBSERVABLE_UP_DOWN_COUNTER }
+      end
+
+      private
+
+      def create_instrument(kind, name, unit, description, callback)
```
fbogsany, line 53: This should enforce [instrument naming requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument) from the spec.

> Instrument names MUST conform to the following syntax (described using the
> [Augmented Backus-Naur Form](https://tools.ietf.org/html/rfc5234)):
> 
> ```abnf
> instrument-name = ALPHA 0*62 ("_" / "." / "-" / ALPHA / DIGIT)
> 
> ALPHA = %x41-5A / %x61-7A; A-Z / a-z
> DIGIT = %x30-39 ; 0-9
> ```
> 
> * They are not null or empty strings.
> * They are case-insensitive, ASCII strings.
> * The first character must be an alphabetic character.
> * Subsequent characters must belong to the alphanumeric characters, '_', '.',
>   and '-'.
> * They can have a maximum length of 63 characters.
> 

Additional requirements:

- `unit` must be ASCII with max length of 63 characters
- `description` must be `utf8mb3` and max length of 1023 characters

robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/pull/1004/commits/a9129ba1d87ad32914cf60d8795d9278d66718da

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -0,0 +1,77 @@
...
+        @mutex.synchronize do
+          raise DuplicateInstrumentError if @registry.include? name
+          raise InstrumentNameError if name.nil?
+          raise InstrumentNameError if name.empty?
+          raise InstrumentNameError unless NAME_REGEX.match?(name)
+          raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
+          raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description))
```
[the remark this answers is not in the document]
robertlaurin: Yeah that makes much more sense 😄

fbogsany (commented):
LGTM as a starting point for the Metrics API. It could use a README (or an update to the top-level README) that clearly describes the status of this work. We're not publishing the gem yet, though, so I think we're fine to merge and add the docs in a later PR.

I can't approve, since I started the PR. "Someone" _cough_ @robertlaurin _cough_ will have to 👍 for me.

ericmustin (approved):
:shipit:

robertlaurin (approved):

PR #1005, opened 2021-11-03
chore(http): Add missing example

YanivD:
~~Removed _Examples_ section from HTTP readme file.~~
~~Couldn't find `./example/trace_demonstration.rb`~~

Edit: Added missing example file for HTTP instrumentation.

### DIFF
--- a/instrumentation/http/example/trace_demonstration.rb
+++ b/instrumentation/http/example/trace_demonstration.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'bundler/inline'
+
+gemfile(true) do
+  source 'https://rubygems.org'
+  gem 'opentelemetry-api', path: '../../../api'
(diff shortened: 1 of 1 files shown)


robertlaurin (commented):
Hey instead of deleting the reference in the readme maybe we could add the missing file instead?

https://github.com/open-telemetry/opentelemetry-ruby/tree/a5a0b9c87d2d401717587134b9fe364c32202c2b/instrumentation/net_http/example

Here's an example of how to do it in a single file. https://github.com/open-telemetry/opentelemetry-ruby/blob/a5a0b9c87d2d401717587134b9fe364c32202c2b/instrumentation/resque/example/trace_demonstration.rb

Let me know if you want to take this on, otherwise I can do it and and close this PR.

YanivD:
@robertlaurin thanks for providing examples.

Renamed PR title and added example file

robertlaurin (approved):

PR #1013, opened 2021-11-10
feat: add dalli obfuscation for db_statement

ericmustin:
### Summary

In an attempt to standardize options around database instrumentation `db_statement` settings, this PR adds the `:obfuscate` value to dalli ( a memcached client)'s `db_statement` configuration option. When enabled, it will obfuscate the `db.statement` span attribute. In practice this simply means dropping the arguments and only including the command, which is relatively low value (since it's already available in the span name) but, for the sake of consistency with the other database instrumentations, this allows users to still capture some level of detail in their `db.statement` attribute without potentially exposing sensitive information.


arielvalentin (approved):

mwear (approved):
LGTM

Issue #253, opened 2020-05-12
aws-sdk instrumentation adapter


mwear:
Convert Datadog instrumentation for aws-sdk to OpenTelemetry. 

See
  * [Datadog porting guide](https://github.com/open-telemetry/opentelemetry-ruby/blob/master/adapters/datadog-porting-guide.md)
  * [Semantic Conventions](https://github.com/open-telemetry/opentelemetry-specification/tree/master/specification/trace/semantic_conventions)
  * [Datadog implementation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/aws)

YanivD:
@arielvalentin Thanks for your comment. 
Is there any automated process to test the instrumentation against all `aws-sdk-ruby` versions?

YanivD:
In JS aws-sdk instrumentation there's an option to [suppressInternalInstrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk#aws-sdk-instrumentation-options). When setting true, it hides all underlying http spans. 

Do you think it should be also an option in `aws-sdk-ruby` instrumentation? I couldn't find any ruby instrumentation with such option.

arielvalentin:
> @arielvalentin Thanks for your comment. Is there any automated process to test the instrumentation against all `aws-sdk-ruby` versions?

@YanivD You do not have to test every version, only those you intend on supporting. 

This repository uses the [Appraisals gem](https://github.com/thoughtbot/appraisal) combined with GitHub Actions to test multiple versions of instrumentations. The CI builds will automatically detect your appraisal file and test it against different versions on your behalf. Here is an example in the Rack Instrumentation:

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/rack/Appraisals

arielvalentin:
> In JS aws-sdk instrumentation there's an option to [suppressInternalInstrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk#aws-sdk-instrumentation-options). When setting true, it hides all underlying http spans.
> 
> Do you think it should be also an option in `aws-sdk-ruby` instrumentation? I couldn't find any ruby instrumentation with such option.

I think that option would be useful. 

[...]

PR #1014, opened 2021-11-10
Feat: aws-sdk instrumentation

YanivD:
Related to issue https://github.com/open-telemetry/opentelemetry-ruby/issues/253

During the development I looked on 2 examples:
* [aws-xray instrumentation](https://github.com/aws/aws-xray-sdk-ruby/blob/master/lib/aws-xray-sdk/facets/aws_sdk.rb)
* [datadog instrumentation](https://github.com/DataDog/dd-trace-rb/tree/master/lib/ddtrace/contrib/aws)

This is my first PR in Ruby, so please review this PR carefully 🙏


instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,64 @@
...
+            execute = proc {
+              super(context).tap do |response|
+                if (err = response.error)
+                  span.record_exception(err)
+                  span.status = Trace::Status.error(err)
+                end
+              end
+            }
+
+            if instrumentation_config[:suppress_internal_instrumentation]
```
arielvalentin, line 25: ✨ I really like this! 

@open-telemetry/ruby-maintainers I do realize we don't love using `untraced` but I think this would be great to include as an option for other instrumentation where we may have redundant client instrumentation (if we are not doing so already)

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -0,0 +1,60 @@
...
+          # Cross-check services against loaded AWS constants
+          # Module#const_get can return a constant from ancestors when there's a miss.
+          # If this conincidentally matches another constant, it will attempt to patch
+          # the wrong constant, resulting in patch failure.
+          available_services = ::Aws.constants & SERVICES.map(&:to_sym)
+          available_services.each_with_object([]) do |service, constants|
+            next if ::Aws.autoload?(service)
+
+            # rubocop:disable RescueModifier
+            constants << ::Aws.const_get(service, false).const_get(:Client, false) rescue next
```
arielvalentin, line 53: Consider adding some diagnostic output here using the `OpenTelemetry.error_handler` or `OpenTelemetry.logger` to hint that a specific constant could not be loaded.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/services.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+module OpenTelemetry
+  module Instrumentation
+    # rubocop:disable Metrics/ModuleLength:
+    module AwsSdk
+      SERVICES = %w[
```
arielvalentin, line 7: 😅 This quite a library!

arielvalentin (approved):
This looks good to me! Thank you very much for this excellent contribution. 

I will defer to any other reviewers who have more experience than I using the AWS instrumentation.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,64 @@
...
+        def instrumentation_config
+          AwsSdk::Instrumentation.instance.config
+        end
+      end
+
+      # A Seahorse::Client::Plugin that enables instrumentation for all AWS services
+      class Plugin < Seahorse::Client::Plugin
+        def add_handlers(handlers, config)
+          # run before Seahorse::Client::Plugin::ParamValidator (priority 50)
+          handlers.add Handler, step: :validate, priority: 49
```
arielvalentin, line 59: Question for AWS users... is there anyway to look up the position/priority of an existing Handler like we can do with Rails middleware?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: @arielvalentin 
Agree!
It reduce the amount of redundant spans and making the trace much clearer.

Many nodejs otel instrumentations accept this option, like [opentelemetry-instrumentation-aws-sdk](https://github.com/open-telemetry/opentelemetry-js-contrib/blob/main/plugins/node/opentelemetry-instrumentation-aws-sdk/README.md#aws-sdk-instrumentation-options).
`suppress_internal_instrumentation` option should always be `false` by default to make sure adding new instrumentations won't remove existing spans.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
YanivD: Good idea. Added in [this commit](https://github.com/open-telemetry/opentelemetry-ruby/pull/1014/commits/432902cd6a45ec0b3054baa2eff0f428b059952d)

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,64 @@
...
+              execute.call
+            end
+          end
+        end
+
+        def calculate_span(context)
+          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]
+          span_name = "#{service_name}.#{context.operation.name}"
+          attributes = {
+            'aws.region' => context.config.region,
```
mwear, line 37: The RPC semantic conventions comply with the specification: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md, however, the `aws.region` semantic convention isn't formally specified. OTel JS [uses it](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk#span-attributes) though.

I am wondering if the Resource semantic convention of setting `cloud.provider` and `cloud.region` would be a better option (see https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/cloud.md#cloud). Happy to have others chime in on the best approach here.

mwear (commented):
I just have a couple of comments / questions, but otherwise this looks really good. Thanks for the contribution @YanivD!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,64 @@
...
+            OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
+            OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => context.operation.name,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
+          }
+
+          [span_name, attributes]
```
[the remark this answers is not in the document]
YanivD: Good idea, splitted to 2 methods

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,64 @@
...
+            if instrumentation_config[:suppress_internal_instrumentation]
+              OpenTelemetry::Common::Utilities.untraced(&execute)
+            else
+              execute.call
+            end
+          end
+        end
+
+        def calculate_span(context)
+          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]
```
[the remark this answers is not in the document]
YanivD: I used [xray as inspiration](https://github.com/aws/aws-xray-sdk-ruby/blob/master/lib/aws-xray-sdk/facets/aws_sdk.rb#L21-L22) for service name extraction.
Added safe navigation in [this commit](https://github.com/open-telemetry/opentelemetry-ruby/pull/1014/commits/31ed27fa58aa9a42a45fc1289409caac506e87d6#diff-1723a57ddb33bc724a6da2be5c6a3dffa0a655cb65a43fbe00a78c71f5d2cb89R44).

---

I tried to use safe navigation for every method in the chain, but I got error from `rubocop`. 
@mwear What is the best practice in this case?

### What I tried:
```ruby
def get_service_name(context)
  context&.client&.class&.api&.metadata['serviceId'] || context&.client&.class&.to_s&.split('::')[1]
end
```

### The warning I get:
> lib/opentelemetry/instrumentation/aws_sdk/handler.rb:44:48: W: Lint/SafeNavigationChain: Do not chain ordinary method call after safe navigation operator.
>           context&.client&.class&.api&.metadata['serviceId'] || context&.client&.class&.to_s&.split('::')[1]

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: You're raising a good question. 

I'm not sure resource attributes are a good fit for this attribute. 
As I understand, resource attributes are immutable representation of the running service.
For example, if I'm running a service on EC2, `cloud.provider` and `cloud.region` resource attributes should keep the value of the running environment. If the running service is producing `sqs.send_message` operation to a queue in a different region, `cloud.region` should have the EC2's region and not the SQS destination region.

Maybe we should use `cloud.destination.region` instead of `aws.region`.

The cloud semantic conventions are still in Experimental status. This topic is not covered yet. 
IMO we can keep the behavior as implemented in OTEL JS, and update it in the future when it will have a clear definition.

WDYT?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -0,0 +1,73 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module AwsSdk
+      # Generates Spans for all interactions with AwsSdk
+      class Handler < Seahorse::Client::Handler
+        def call(context)
+          span_name = get_span_name(context)
+          attributes = get_span_attributes(context)
```
ericmustin, line 14: The specification requires that we set `db.system: dynamodb` in the event that this is a request for dynamodb. I think this would be possible to determine based on the context 

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#dynamodb

Could we add support+a test for this as it's required (if there are optional attributes you think are worth adding from dynamoDB, the more the merrier, but this is the only one that's required as per the specification)?

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#dynamodb

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -0,0 +1,62 @@
...
+module OpenTelemetry
+  module Instrumentation
+    module AwsSdk
+      # Instrumentation class that detects and installs the AwsSdk instrumentation
+      class Instrumentation < OpenTelemetry::Instrumentation::Base
+        MINIMUM_VERSION = Gem::Version.new('2.0')
+
+        install do |_config|
+          require_relative './handler'
+          require_relative './services'
```
ericmustin, line 16:
```suggestion
          require_relative 'handler'
          require_relative 'services'
```

ericmustin (changes requested):
A few small nits, generally lgtm nice work!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: @ericmustin Nice catch! 
[Added `db.system` attribute](
https://github.com/open-telemetry/opentelemetry-ruby/pull/1014/commits/cfd81abada4f85dbdbaaa893729b668ec8f7e198)

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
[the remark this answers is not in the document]
ericmustin: chaining after `.to_s` is unnecessary in this case, and likely  `.class` as well. basically if the class exists then it would be reasonable to assume `.class` would return as that method exists on every class, and likewise with `.to_s`. Does that help?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
ericmustin: my 2 cents is, unless we want to put in the effort to make an otep or spec change upstream , stick with what js does. i think `cloud.destination.region` is a completely reasonable choice fwiw, but I just think practically speaking lets lean on the experimental status here and let ppl speak up if they hate the attr name.

ericmustin:
Lgtm pending some of the feedback from @mwear getting addressed/resolved. nice work here ty for the contribution!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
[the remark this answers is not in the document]
YanivD: Thanks @ericmustin!

Can you confirm this chaining is safe?
```rb
context&.client.class.api.metadata['serviceId'] || context&.client.class.to_s.split('::')[1]
```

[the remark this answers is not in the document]
ericmustin: I have no idea off hand, i'd have to look into the implementation 🙃. I can follow up when i have some more bandwidth. 

specifically, are `client`, `api` and `metadata` always guaranteed to return what you expect here? If it's possible they return `nil` then we'd want to chain those with the safe operator

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
mwear: i'm also fine with sticking `aws.region` as is for now.

mwear (approved):
LGTM. Thanks for addressing my feedback!

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -54,7 +54,7 @@
         _(last_span.attributes['rpc.service']).must_equal 'SNS'
         _(last_span.attributes['rpc.method']).must_equal 'Publish'
         _(last_span.attributes['aws.region']).must_equal 'us-stubbed-1'
-        _(last_span.attributes['db.system']).must_equal nil
+        assert_nil last_span.attributes['db.system']
```
arielvalentin, line 57: 💁🏼 I think what we want here is [`must_be_nil`](https://docs.seattlerb.org/minitest/Minitest/Expectations.html#method-i-must_be_nil)
```suggestion
       _(last_span.attributes['db.system']).must_be_nil
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -69,7 +69,7 @@
         _(last_span.attributes['rpc.service']).must_equal 'S3'
         _(last_span.attributes['rpc.method']).must_equal 'ListBuckets'
         _(last_span.attributes['aws.region']).must_equal 'us-stubbed-1'
-        _(last_span.attributes['db.system']).must_equal nil
+        assert_nil last_span.attributes['db.system']
```
arielvalentin, line 72:
```suggestion
       _(last_span.attributes['db.system']).must_be_nil
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -83,7 +83,7 @@
           _(last_span.attributes['rpc.service']).must_equal 'S3'
           _(last_span.attributes['rpc.method']).must_equal 'ListBuckets'
           _(last_span.attributes['aws.region']).must_equal 'us-stubbed-1'
-          _(last_span.attributes['db.system']).must_equal nil
+          assert_nil last_span.attributes['db.system']
```
arielvalentin, line 86:
```suggestion
         _(last_span.attributes['db.system']).must_be_nil
```

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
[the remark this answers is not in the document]
arielvalentin: @ericmustin I was referring to this. In the case context is `nil` it will always results in a `NoMethodError` because `NilClass` does not have a class method `api`. 

```
omoplata ~(:|✔) %
🤘 irb
irb(main):001:0> context = nil
=> nil
irb(main):002:0> context&.client.api
Traceback (most recent call last):
        4: from /Users/arielvalentin/.rbenv/versions/2.7.4/bin/irb:23:in `<main>'
        3: from /Users/arielvalentin/.rbenv/versions/2.7.4/bin/irb:23:in `load'
        2: from /Users/arielvalentin/.rbenv/versions/2.7.4/lib/ruby/gems/2.7.0/gems/irb-1.2.6/exe/irb:11:in `<top (required)>'
        1: from (irb):2
NoMethodError (undefined method `api' for nil:NilClass)
```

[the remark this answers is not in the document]
ericmustin: ah far as i can tell, `context`, `client`, and `api` should be initialized with default values and never return `nil`

- `context` https://github.com/aws/aws-sdk-ruby/blob/b5c46409986c6681b35533e017051d55e733d053/gems/aws-sdk-core/lib/seahorse/client/response.rb#L13

- `client` https://github.com/aws/aws-sdk-ruby/blob/b5c46409986c6681b35533e017051d55e733d053/gems/aws-sdk-core/lib/seahorse/client/request_context.rb#L39-L40

- `api` https://github.com/aws/aws-sdk-ruby/blob/b5c46409986c6681b35533e017051d55e733d053/gems/aws-sdk-core/lib/seahorse/client/base.rb#L171-L173

- `metadata` https://github.com/aws/aws-sdk-ruby/blob/b5c46409986c6681b35533e017051d55e733d053/gems/aws-sdk-core/lib/seahorse/model/api.rb#L8

Technically I guess someone could provide a junk value for context, but many other parts of the aws-sdk would break in that situation anyway, so i'm not too concerned.

tl;dr, i think this is much ado about nothing, 

`context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]` is fine imo

ericmustin (approved):

YanivD:
Is there anything else missing before merge and release?

@open-telemetry/ruby-maintainers

mwear:
I believe everything has been addressed. Any objections to me merging this @robertlaurin and @fbogsany?

robertlaurin (approved):

PR #1015, opened 2021-11-10
fix: resolve Dalli::Server deprecation in 3.0+

tierra:
Dalli 3.0 has deprecated `Dalli::Server` in petergoldstein/dalli#760

I believe the fix is simply to use `Dalli::Protocol::Binary` instead, when Dalli >=3.0 is detected (while still using `Dalli::Server` in <3.0).

I've created a new Appraisal config to test this with Dalli 3.0, while still also testing 2.7.

Related: [Dalli 3.0 Upgrade Guide](https://github.com/petergoldstein/dalli/blob/master/3.0-Upgrade.md)

### DIFF
--- a/instrumentation/dalli/Appraisals
+++ b/instrumentation/dalli/Appraisals
@@ -3,3 +3,7 @@
 appraise 'dalli-2.7' do
   gem 'dalli', '~> 2.7'
 end
+
+appraise 'dalli-3.0' do
+  gem 'dalli', '~> 3.0'
+end
--- a/instrumentation/dalli/gemfiles/dalli_3.0.gemfile
+++ b/instrumentation/dalli/gemfiles/dalli_3.0.gemfile
@@ -4,7 +4,7 @@ source "https://rubygems.org"
 
 gem "opentelemetry-api", path: "../../../api"
 gem "opentelemetry-instrumentation-base", path: "../../base"
-gem "dalli", "~> 2.7"
+gem "dalli", "~> 3.0"
 
 group :test do
   gem "opentelemetry-common", path: "../../../common"
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/instrumentation.rb
@@ -30,7 +30,11 @@ module OpenTelemetry
         end
 
         def add_patches
-          ::Dalli::Server.prepend(Patches::Server)
+          if Gem::Version.new(::Dalli::VERSION) < Gem::Version.new('3.0.0')
+            ::Dalli::Server.prepend(Patches::Server)
+          else
+            ::Dalli::Protocol::Binary.prepend(Patches::Server)
+          end
         end
       end
     end
--- a/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
+++ b/instrumentation/dalli/lib/opentelemetry/instrumentation/dalli/patches/server.rb
@@ -8,7 +8,7 @@ module OpenTelemetry
   module Instrumentation
     module Dalli
       module Patches
-        # Module to prepend to Dalli::Server for instrumentation
+        # Module to prepend to Dalli::Server (or Dalli::Protocol::Binary in 3.0+) for instrumentation
         module Server
           def request(op, *args)
             operation = Utils.opname(op, multi?)
--- a/instrumentation/dalli/opentelemetry-instrumentation-dalli.gemspec
+++ b/instrumentation/dalli/opentelemetry-instrumentation-dalli.gemspec
@@ -31,7 +31,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'appraisal', '~> 2.2.0'
   spec.add_development_dependency 'bundler', '>= 1.17'
-  spec.add_development_dependency 'dalli', '~> 2.7'
+  spec.add_development_dependency 'dalli', '>= 2.7'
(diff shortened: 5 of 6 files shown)


tierra:
I am seeing these failures locally, but was hoping it was just bad local setup/configuration:

```
>> BUNDLE_GEMFILE=/home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/dalli/gemfiles/dalli_3.0.gemfile bundle exec rake test
Run options: --seed 64944

# Running:

...W, [2021-11-10T19:54:06.506858 #7395]  WARN -- : localhost:11211 failed (count: 0) RuntimeError: External timeout
W, [2021-11-10T19:54:06.609232 #7395]  WARN -- : localhost:11211 failed (count: 1) RuntimeError: External timeout
W, [2021-11-10T19:54:06.609384 #7395]  WARN -- : localhost:11211 is down
F...

Finished in 0.115292s, 60.7153 runs/s, 216.8402 assertions/s.

  1) Failure:
OpenTelemetry::Instrumentation::Dalli::Instrumentation::tracing#test_0006_after error [/home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb:87]:
Expected: 1
  Actual: 2

7 runs, 25 assertions, 1 failures, 0 errors, 0 skips
```

I'm not really sure what's broken here. I've done this same change on a different instrumentation tool (not OTEL) that used the same approach to prepend `Dalli::Server`, and that worked flawlessly.

Is there something I'm missing related to configuring Memcached with the second appraisal bundle gemfile maybe?

arielvalentin:
@tierra Has something changed between versions? Does the Binary Protocol version have built in retries where the older versions do not?

tierra:
petergoldstein/dalli#783 made a reconnect change that I think affects `get_multi` / `get_multi_cas` calls. Might be related.

It's possible the difference here is just in how the unit tests mock the connection failure behaviors.

tierra:
As far as I understand it, `Dalli::Server` was always the "binary" protocol version, and it's been renamed (to explicitly state so) in 3.0 since Dalli 4.0 plans to add a new "meta/text" protocol version, and eventually switch to it.

I'm not entirely clear about any of this though.

arielvalentin:
@tierra it does seem like the test is simulating an error https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb#L83

It may be the case that switching to [_optimistic versioning_](https://guides.rubygems.org/patterns/#declaring-dependencies) pulls in a version with retry logic because it will pull in any version greater than 2.7 

https://github.com/open-telemetry/opentelemetry-ruby/pull/1015/files#diff-eec3ea0267d5829c4091540fb309594e7838e340d957930605bfffb936dfbae4R34

tierra:
It probably helps if I mention that I have an application in production that's already upgraded to Dalli 3.0, while also configured to use this instrumentation, and I've already verified that it's still properly recording spans for Dalli calls from this (it's just the addition of the deprecation warnings in logs that has brought me here).

So that's another reason I suspect that the test failures here are likely due to the way the unit tests here mock internal implementation behaviors (and are failing due to Dalli internal implementation changes in 3.0).

ericmustin:
Hello friends,

Took a quick look here and then took a longer look because I too was confused by this. Anyway I think what happens is that actually the underlying behavior of `get_multi` and `Dalli::NetworkError` has changed in a recent 3.0.x release, specifically this PR does legitimately update the behavior here to _retry_ on network errors, https://github.com/petergoldstein/dalli/pull/754. So, the test is failing because our test is too rigid. Rather than try to write multiple test cases depending on the Dalli Gem major/minor version, it's probably quicker to just mock using a different raised error, such as `Dalli::DalliError` which i've done in the diff below. 

Anyway @tierra if you wouldn't mind adding that patch to your branch that should get all the tests working and we can get this merged. ty for the pr here as well!

```
diff --git a/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb b/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
index 05f431de..866a47d9 100644
--- a/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
+++ b/instrumentation/dalli/test/opentelemetry/instrumentation/dalli/instrumentation_test.rb
@@ -80,7 +80,7 @@ describe OpenTelemetry::Instrumentation::Dalli::Instrumentation do
       dalli.set('foo', 'bar')
       exporter.reset
 
-      dalli.instance_variable_get(:@ring).servers.first.stub(:write, ->(_bytes) { raise Dalli::DalliError }) do
         dalli.get_multi('foo', 'bar')
       end
 
@@ -93,8 +93,8 @@ describe OpenTelemetry::Instrumentation::Dalli::Instrumentation do
 
       span_event = span.events.first
       _(span_event.name).must_equal 'exception'
-      _(span_event.attributes['exception.type']).must_equal 'Dalli::DalliError'
-      _(span_event.attributes['exception.message']).must_equal 'Dalli::DalliError'
+      _(span_event.attributes['exception.type']).must_equal 'Dalli::NetworkError'
+      _(span_event.attributes['exception.message']).must_equal 'Dalli::NetworkError'
     end
```

tierra:
Thank you @ericmustin! Your change does fix the tests in local when using Dalli 3.0. It also still works with 2.7 as well. I've updated the PR. I think it requires review to run PR builds, but should be good to go now.

ericmustin (approved):

PR #1022, opened 2021-11-17
feat: deployment resource detector

blumamir:
automatic resource detector for [deployment](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/deployment_environment.md)

I am new to ruby, so please don't hesitate to point me on how to do things correctly if I implemented any anti-pattern.
Also, I am not sure how to write tests for extracting the env from `::Rails::env`. Will appreciate some guidance on this

### DIFF
--- a/resource_detectors/README.md
+++ b/resource_detectors/README.md
@@ -30,6 +30,7 @@ require 'opentelemetry/resource/detectors'
 # For a specific platform
 OpenTelemetry::SDK.configure do |c|
   c.resource = OpenTelemetry::Resource::Detectors::GoogleCloudPlatform.detect
+  c.resource = OpenTelemetry::Resource::Detectors::Deployment.detect
 end
 
 # Or if you would like for it to run all detectors available
--- a/resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
@@ -13,7 +13,7 @@ module OpenTelemetry
 
         DETECTORS = [
           OpenTelemetry::Resource::Detectors::GoogleCloudPlatform,
-          OpenTelemetry::Resource::Detectors::Deployment,
+          OpenTelemetry::Resource::Detectors::Deployment
         ].freeze
 
         def detect
--- a/resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
@@ -12,7 +12,8 @@ module OpenTelemetry
         extend self
 
         DETECTORS = [
-          OpenTelemetry::Resource::Detectors::GoogleCloudPlatform
+          OpenTelemetry::Resource::Detectors::GoogleCloudPlatform,
+          OpenTelemetry::Resource::Detectors::Deployment,
         ].freeze
 
         def detect
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -9,8 +9,8 @@ module OpenTelemetry
 
         def detect
           resource_attributes = {}
-          deployment_environment = rails_env || rack_env
-          resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str if deployment_environment
+          deployment_environment = rails_env || sinatra_env || rack_env
+          resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment if deployment_environment
           OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
         end
 
@@ -20,15 +20,27 @@ module OpenTelemetry
           # rails extract env like this:
           # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
           # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
-          ::Rails.env if defined?(::Rails.env)
+          ::Rails.env.to_str if defined?(::Rails.env)
         end
 
         def rack_env
           ENV['RACK_ENV']
         end
 
-        # TODO: add sinatra:
-        # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
+        def sinatra_env
+          # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
+          # cases:
+          #
+          # 1. if sinatra is "require"d before the detector, then we return the value from the library
+          # this case will return the default "development" fallback if not env variable is set which is good.
+          #
+          # 2. if sinatra is "require"d after the detector, then:
+          # 2.1 if user is setting environment via 'APP_ENV' or 'RACK_ENV' then those value will be picked up and reported
+          # 2.2 else, the sinatra environment will fallback to "development", but detector will return nil for it.
+          # this issue is not covered, as when detector initialize the immutable resource, it has no way
+          # of knowing if "sinatra" will be required later or not.
+          (::Sinatra::Base.environment.to_s if defined?(::Sinatra::Base.environment)) || ENV['APP_ENV']
+        end
       end
     end
   end
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -20,7 +20,7 @@ module OpenTelemetry
           # rails extract env like this:
           # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
           # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
-          ::Rails.env.to_str if defined?(::Rails.env)
+          ::Rails.env.to_s if defined?(::Rails.env)
         end
 
         def rack_env
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -1,38 +1,35 @@
+# frozen_string_literal: true
 
 module OpenTelemetry
-    module Resource
-      module Detectors
-            module Deployment
-                extend self
+  module Resource
+    module Detectors
+      # Deployment contains detect class method for determining deployment resource attributes
+      module Deployment
+        extend self
 
-                def detect
-                    resource_attributes = {}
-                    deployment_environment = get_rails_deployment_environment || get_rack_env
-                    if deployment_environment
-                        resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str
-                    end
-                    OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
-                end
-
-                private
-
-                def get_rails_deployment_environment
-                    # rails extract env like this:
-                    # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
-                    # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
-                    if defined?(::Rails::env)
-                        ::Rails::env
-                    end
-                end
-
-                def get_rack_env
-                    ENV["RACK_ENV"]
-                end
-
-                # TODO: add sinatra:
-                # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
-                
-            end
+        def detect
+          resource_attributes = {}
+          deployment_environment = get_rails_deployment_environment || get_rack_env
+          resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str if deployment_environment
+          OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
         end
+
+        private
+
+        def rails_deployment_environment
+          # rails extract env like this:
+          # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
+          # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
+          ::Rails.env if defined?(::Rails.env)
+        end
+
+        def rack_env
+          ENV['RACK_ENV']
+        end
+
+        # TODO: add sinatra:
+        # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
+      end
     end
+  end
 end
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -1,0 +1,38 @@
+
+module OpenTelemetry
+    module Resource
+      module Detectors
+            module Deployment
+                extend self
+
+                def detect
+                    resource_attributes = {}
+                    deployment_environment = get_rails_deployment_environment || get_rack_env
+                    if deployment_environment
+                        resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str
+                    end
+                    OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
+                end
+
+                private
+
+                def get_rails_deployment_environment
+                    # rails extrace env like this:
+                    # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
+                    # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
+                    if defined?(::Rails::env)
+                        ::Rails::env
+                    end
+                end
+
+                def get_rack_env
+                    ENV["RACK_ENV"]
+                end
+
+                # TODO: add sinatra:
+                # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
+                
+            end
+        end
+    end
+end
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -17,7 +17,7 @@ module OpenTelemetry
                 private
 
                 def get_rails_deployment_environment
-                    # rails extrace env like this:
+                    # rails extract env like this:
                     # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
                     # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
                     if defined?(::Rails::env)
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -9,14 +9,14 @@ module OpenTelemetry
 
         def detect
           resource_attributes = {}
-          deployment_environment = rails_deployment_environment || rack_env
+          deployment_environment = rails_env || rack_env
           resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str if deployment_environment
           OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
         end
 
         private
 
-        def rails_deployment_environment
+        def rails_env
           # rails extract env like this:
           # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
           # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
--- a/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
@@ -9,7 +9,7 @@ module OpenTelemetry
 
         def detect
           resource_attributes = {}
-          deployment_environment = get_rails_deployment_environment || get_rack_env
+          deployment_environment = rails_deployment_environment || rack_env
           resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str if deployment_environment
           OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
         end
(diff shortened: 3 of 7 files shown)


mottibec (approved):

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    let(:detected_resource_attributes) { detected_resource.attribute_enumerator.to_h }
+    let(:expected_resource_attributes) { {} }
+
+    it 'returns an empty resource' do
+      _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
+      _(detected_resource_attributes).must_equal(expected_resource_attributes)
+    end
+
+    describe 'when in a rack environment' do
+      it 'returns a resource with rack environment' do
```
arielvalentin, line 19: There is a test helper [`with_env`](https://github.com/open-telemetry/opentelemetry-ruby/blob/50ce54bd7098c1bc280ec99b486a4f1f51f84ce7/common/lib/opentelemetry/common/test_helpers.rb#L7) that should be available for you to use for test cases.


Here are some examples: https://github.com/open-telemetry/opentelemetry-ruby/search?q=with_env

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
```diff
@@ -0,0 +1,31 @@
...
+        old_env = ENV['RACK_ENV']
+        ENV['RACK_ENV'] = 'env from test'
+        begin
+          _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
+          _(detected_resource_attributes).must_equal('deployment.environment' => 'env from test')
+        ensure
+          ENV['RACK_ENV'] = old_env
+        end
+      end
```
arielvalentin, line 28: Please consider adding a test for the `Rails.env` case. 

You can use helpers like `stub_const` to simulate the existence of classes:

https://github.com/open-telemetry/opentelemetry-ruby/search?q=stub_const

resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
```diff
@@ -12,7 +12,8 @@ module AutoDetector
         extend self
 
         DETECTORS = [
-          OpenTelemetry::Resource::Detectors::GoogleCloudPlatform
+          OpenTelemetry::Resource::Detectors::GoogleCloudPlatform,
+          OpenTelemetry::Resource::Detectors::Deployment
```
arielvalentin, line 16: @open-telemetry/ruby-maintainers I believe we discussed breaking this gem up into separate gems and removing the auto-detector. Is that correct?

Should that be done before we merge this PR?

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
```diff
@@ -0,0 +1,31 @@
...
+    let(:detected_resource) { detector.detect }
+    let(:detected_resource_attributes) { detected_resource.attribute_enumerator.to_h }
+    let(:expected_resource_attributes) { {} }
+
+    it 'returns an empty resource' do
+      _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
+      _(detected_resource_attributes).must_equal(expected_resource_attributes)
+    end
+
+    describe 'when in a rack environment' do
```
arielvalentin, line 18: Please consider adding a test case for when the environment variable is `nil` and the `Rails.env` is undefined

```
irb(main):003:0> ENV['RACK_ENV']
=> nil
````

File: resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb [hunk 5]
```diff
@@ -0,0 +1,35 @@
...
+module OpenTelemetry
+  module Resource
+    module Detectors
+      # Deployment contains detect class method for determining deployment resource attributes
+      module Deployment
+        extend self
+
+        def detect
+          resource_attributes = {}
+          deployment_environment = rails_deployment_environment || rack_env
```
arielvalentin, line 12: Personally speaking I think it would be OK to rely on only using environment variables instead of framework specific helpers. 

So in this case I think it would be fine to use something like `ENV['RACK_ENV'] || ENV['RAILS_ENV'] || ENV['APP_ENV']`

resource_detectors/lib/opentelemetry/resource/detectors/auto_detector.rb
blumamir: I believe it can be done after this PR is merged unless it creates issues

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
blumamir: The issue is that the frameworks have a default environment of "development" in case ENV is missing.
To reflect that in the resource, I will need to write `ENV['RACK_ENV'] || ENV['RAILS_ENV'] || ENV['APP_ENV'] || "development"`, which will work for rails and sinatra but will incorrectly report "development" environment if these frameworks are not being used.

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
blumamir: Much shorter and cleaner, thanks for pointing me to this function :)
I refactored the code to use it

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
blumamir: I meant for the first test "returns an empty resource" to check this case. Is it the case? or should I refactor it to match your suggestion?

arielvalentin: That is fair, although it would mean when executing the test the environment must not have RACK_ENV set otherwise it would result in the test failing.

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
arielvalentin: Totally understand. Will you be using sinatra integration in a separate PR?

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
```diff
@@ -0,0 +1,35 @@
...
+          resource_attributes[::OpenTelemetry::SemanticConventions::Resource::DEPLOYMENT_ENVIRONMENT] = deployment_environment.to_str if deployment_environment
+          OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
+        end
+
+        private
+
+        def rails_deployment_environment
```
arielvalentin, line 19: For symmetry with `rack_env` 😄 

```suggestion
        def rails_env
```

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
blumamir: Added it like this:
```rb
      it 'return rails env when present' do
        allow(::Rails).to receive(:env).and_return('env from rails')
        with_env({}) do
          _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
          _(detected_resource_attributes).must_equal('deployment.environment' => 'env from rails')          
        end
      end
```

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
blumamir: good point :) renamed it to be consistent

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
blumamir: good point, added clearing the env before this test:
```rb
      with_env({}) do
        _(detected_resource).must_be_instance_of(OpenTelemetry::SDK::Resources::Resource)
        _(detected_resource_attributes).must_equal(expected_resource_attributes)
      end
```

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
arielvalentin: Looks like this test is failing in CI.

You'll have to stub the `Rails` constant since it doesn't exist. 

```
NameError: uninitialized constant Rails
5431
    /home/runner/work/opentelemetry-ruby/opentelemetry-ruby/resource_detectors/test/opentelemetry/detectors/deployment_test.rb:22:in `block (4 levels) in <top (required)>'
5432
```

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
blumamir: I wasn't planing to add Sinatra support for this PR.
Tried to extract the env like this:
```rb 
          def sinatra_env
            # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
            ::Sinatra::Base::environment.to_s if defined?(::Sinatra::Base::environment)
          end  
```

Which works fine if the `require 'sinatra'` is placed before the SDK is initialized.
If user configures the SDK and only then require 'sinatra', then this will fail. In addition, it looks like the sinatra instrumentation does not work in this case (we should probably open an issue and debug it as an other PR).

Do you have any suggestion on how to extract the value in this case?

arielvalentin: Thank you for giving it a try and thank you for your patience as at this point I've written more comments that actual code in this PR. 

I think it is reasonable to add development dependencies for testing. Require the dependencies as part of the test case and it will make them available to the system under test. You can do this as part of Bundle.require or explicitly requiring it in the test case.

That combined with mocking and `unstub_cost` helpers should be enough to do the trick.

blumamir: Thank you too for the high-quality review and help on these topics.
did some experimenting with sinatra. I was able to get it to work with the help of @YanivD . If you can take another look at the updated code it could be great :)

As documented in the code, this resource detector will not be able to extract the default enviroment "development" if sinatra is used and it is required after resource detector is run (resource is immutable and there is no way to know if it will be rrequired in the future at the time of resource creation).

This is the relevant part of the code for reference:
```rb
        def sinatra_env
          # https://github.com/sinatra/sinatra/blob/e69b6b9dee7165d3a583fc8a6af10ceee1ea687d/lib/sinatra/base.rb#L1801
          # cases:
          #
          # 1. if sinatra is "require"d before the detector, then we return the value from the library
          # this case will return the default "development" fallback if not env variable is set which is good.
          #
          # 2. if sinatra is "require"d after the detector, then:
          # 2.1 if user is setting environment via 'APP_ENV' or 'RACK_ENV' then those value will be picked up and reported
          # 2.2 else, the sinatra environment will fallback to "development", but detector will return nil for it.
          # this issue is not covered, as when detector initialize the immutable resource, it has no way
          # of knowing if "sinatra" will be required later or not.
          (::Sinatra::Base.environment.to_s if defined?(::Sinatra::Base.environment)) || ENV['APP_ENV']
        end
```

resource_detectors/test/opentelemetry/detectors/deployment_test.rb
blumamir: done. tests should pass now :)

blumamir:
@arielvalentin , thank you again for the detailed review, appreciate it.
CI is not green and I think I addressed all the comments. Can you please resolve the conversations which are solved and let me know if there is anything else missing?

resource_detectors/test/test_helper.rb
```diff
@@ -7,6 +7,8 @@
 require 'simplecov'
 SimpleCov.start
 
+require 'opentelemetry/common/test_helpers'
 require 'opentelemetry/resource/detectors'
 require 'minitest/autorun'
+require 'rspec/mocks/minitest_integration'
```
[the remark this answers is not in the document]
fbogsany: We have preferred not to pull in RSpec related dependencies, despite a strong preference for RSpec on the part of some maintainers (_cough_ @robertlaurin).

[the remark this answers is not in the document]
arielvalentin: hmmm upon further inspection it does not look like the test cases are using `rspec-mocks`. Am I mistaken?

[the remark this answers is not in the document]
blumamir: I added it so I can use the `stub_const`. copied it from other packages in the repo:
image

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
```diff
@@ -0,0 +1,47 @@
...
+          #
+          # 1. if sinatra is "require"d before the detector, then we return the value from the library
+          # this case will return the default "development" fallback if not env variable is set which is good.
+          #
+          # 2. if sinatra is "require"d after the detector, then:
+          # 2.1 if user is setting environment via 'APP_ENV' or 'RACK_ENV' then those value will be picked up and reported
+          # 2.2 else, the sinatra environment will fallback to "development", but detector will return nil for it.
+          # this issue is not covered, as when detector initialize the immutable resource, it has no way
+          # of knowing if "sinatra" will be required later or not.
+          (::Sinatra::Base.environment.to_s if defined?(::Sinatra::Base.environment)) || ENV['APP_ENV']
```
ericmustin, line 42:
```suggestion
          (::Sinatra::Base.environment.to_s if (defined?(::Sinatra) && defined?(::Sinatra::Base))  || ENV['APP_ENV']
```

resource_detectors/test/test_helper.rb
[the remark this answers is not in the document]
arielvalentin: Oh my mistake. I could have sworn that was coming from minitest-stub-const. 😳

Thanks again for your patience

arielvalentin (approved):

ericmustin (changes requested):
Some small feedback on the implementation itself, but broadly I have some hesitations about this as a resource detector. Let me try to lay these out

1. The name, `Deployment` detector is confusing, i think partly because there's nowhere in our repo to let someone add details about what the detector does. If `resource_detectors` is going to grow, I think adding a section to `resource_detectors/README.md`, that briefly describes this resource detector, and improving the code comments here https://github.com/open-telemetry/opentelemetry-ruby/blob/474236b24db7b8894f0f73ff2ce0abc7e63f963e/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb#L6 to inform users that that this resource detector specifically enables detection of the `environment` for Rails, Sinatra, and Rack based applications, would go a long way.

2. Is there any prior art that you're basing this approach on? As a pattern for resource detectors, its unusual to me, I generally expect resource detectors to deal with the platform, like gcp/aws/alibaba/etc, and not so much the application libraries. I'm not sure its necessarily bad or not permitted to take the approach being taken here, but it would help my confidence in maintaining this if there were other languages which have taken a similar approach. I briefly reviewed the js resource detectors and couldnt find anything similar to this one.

arielvalentin:
@ericmustin good catches. Those slipped right past me 😓

blumamir:
Hi @ericmustin , Thanks for the review :)

> The name, Deployment detector is confusing

I implemented the resource detector based on [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/semantic_conventions/deployment_environment.md). I think that as long as it's in the spec, we need to follow the naming and behaviour as described there. Your comment can be a good issue to open in the spec repo.

> i think partly because there's nowhere in our repo to let someone add details about what the detector does

Agree. I was generally following the pattern for the GCP detector in the repo and tried not to introduce any new concepts. Maybe one the detectors are broken down into different gems, we can add a brief details in each gem's README.md, WDYT?

> Is there any prior art that you're basing this approach on? 

I implemented this resource detector for js [here](https://github.com/aspecto-io/opentelemetry-ext-js/tree/master/detectors/node/resource-detector-deployment). This detector is hosted in a separate repo since the native js detectors api is async and we needed a sync detector. It's been included in Aspecto's otel distro for long time and has been working well.

> As a pattern for resource detectors, its unusual to me, I generally expect resource detectors to deal with the platform, like gcp/aws/alibaba/etc, and not so much the application libraries. I'm not sure its necessarily bad or not permitted to take the approach being taken here, but it would help my confidence in maintaining this if there were other languages which have taken a similar approach. I briefly reviewed the js resource detectors and couldnt find anything similar to this one.

I try to look at it from the perspective of the user. When looking at a trace it can be very helpful to know if it's coming from production \ staging \ ci \ tests \ development, as this can aid in filtering relevant traces as well as explaining the behaviour observed in the trace. Unfortunately, the environment configuration in ruby is not integrated into the platform itself (like with node's NODE_ENV variable), but is handled by the frameworks. That is the reason I followed the behavior of the framework to extract this info. I am open to suggestions on alternatives to supplying this resource attribute :)

resource_detectors/README.md
```diff
@@ -30,6 +30,7 @@ require 'opentelemetry/resource/detectors'
 # For a specific platform
 OpenTelemetry::SDK.configure do |c|
   c.resource = OpenTelemetry::Resource::Detectors::GoogleCloudPlatform.detect
+  c.resource = OpenTelemetry::Resource::Detectors::Deployment.detect
```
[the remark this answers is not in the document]
blumamir: The resource setter for the configurator is actually [merging the resource itself](https://github.com/open-telemetry/opentelemetry-ruby/blob/a37637c241cacc1799dd0728248c2d1372986348/sdk/lib/opentelemetry/sdk/configurator.rb#L70):

```rb
      # Accepts a resource object that is merged with the default telemetry sdk
      # resource. The use of this method is optional, and is provided as means
      # to include additional resource information.
      # If a resource key collision occurs the passed in resource takes priority.
      #
      # @param [Resource] new_resource The resource to be merged
      def resource=(new_resource)
        @resource = @resource.merge(new_resource)
      end
```

I was confused myself when writing this, until I looked at the source code for that. Maybe we should consider making it clearer in the configurator so it's less confusing?

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
```diff
@@ -0,0 +1,47 @@
...
+          OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
+        end
+
+        private
+
+        def rails_env
+          # rails extract env like this:
+          # https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74
+          # ActiveSupport::EnvironmentInquirer.new(ENV["RAILS_ENV"].presence || ENV["RACK_ENV"].presence || "development")
+          ::Rails.env.to_str if defined?(::Rails.env)
```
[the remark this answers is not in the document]
blumamir: I was actually starting with just checking the `::Rails` and then something crashed in my playground application when verifying that the detector is working.
Will try to reproduce it and paste here an example so we can analyse it. Is it problematic in any way to check for `defined?(::Rails.env)` instead of `defined?(::Rails)`?

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
blumamir: good catch, for some reason I thought the `defined?` method is able to handle it.
Is there any optional chaining operator similar to `&.` that can work for modules?

resource_detectors/README.md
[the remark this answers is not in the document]
ericmustin: Oh hm, yea i guess it's fine if the configurator handles the merge automagically. Disregard.

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
ericmustin: i did some testing and it actually can handle it just fine! TIL, disregard

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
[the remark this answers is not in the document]
ericmustin: after some testing it seems fine to check `::Rails.env`. 

Would be curious to know what the error was. I just modified my suggestion above, i think we want to be invoking `.to_s` and not `.to_str` here, as there is a difference. `.to_str` returns a "string like object" while `.to_s` returns a string representation of the object. 

Frankly I'm not sure if we even need to invoke `.to_s` at all, since afaik `Rails.env` ought to return a string, but if there's any doubt then `.to_s` would be appropriate (and `.to_str` would likely be incorrect)

ericmustin:
@blumamir thanks for answering my questions. I think I'm fine with this, I do think the deeper issue is that these all ought to be split into separate gems, which would allow us to be a little bit more descriptive about what the resource detectors do, and give users more flexibility about which ones they want to use. I have one small feedback item that needs to be addressed that i just added a comment about, and beside that i think this LGTM

SomalianIvan (changes requested):
This seems like it has good motivation, but basing resource detection on the environment variables seems like guessing to me. I believe the detection of the framework specific attributes on the resource should belong to instrumentation. E.g. `rails_env_check` should be part of rails instrumentation.

resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb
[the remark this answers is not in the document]
blumamir: > i think we want to be invoking .to_s and not .to_str here, as there is a difference. .to_str returns a "string like object" while .to_s returns a string representation of the object.

Thanks for spotting that, I wasn't aware of the difference, was just trying `to_str` and it worked. Changed to use `to_s`

> Frankly I'm not sure if we even need to invoke .to_s at all, since afaik Rails.env ought to return a string, but if there's any doubt then .to_s would be appropriate (and .to_str would likely be incorrect)

The value that is [returned from rails env function is: `ActiveSupport::EnvironmentInquirer`](https://github.com/rails/rails/blob/5647a9c1ced68d20338552d47a3b755e10a271c4/railties/lib/rails.rb#L74).
From the docs:

```
  # Wrapping a string in this class gives you a prettier way to test
  # for equality. The value returned by <tt>Rails.env</tt> is wrapped
  # in a StringInquirer object, so instead of calling this:
  #
  #   Rails.env == 'production'
  #
  # you can call this:
  #
  #   Rails.env.production?
  #
```

This is the reason `to_s` is mandatory for the env value before we store it into the resource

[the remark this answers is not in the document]
blumamir: > Would be curious to know what the error was

So I wasn't able to reproduce it in a real setup, it works good with both `defined?(::Rails.env)` and `defined?(::Rails)`, and also for sinatra.
But if I remove the `env` from the check, current test suite fails with things like:
```
  1) Error:
OpenTelemetry::Resource::Detectors::Deployment::.detect#test_0001_returns an empty resource:
NoMethodError: undefined method `env' for Rails:Module
Did you mean?  end
    /Users/amirblum/repos/opentelemetry-ruby/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb:23:in `rails_env'
    /Users/amirblum/repos/opentelemetry-ruby/resource_detectors/lib/opentelemetry/resource/detectors/deployment.rb:12:in `detect'
    /Users/amirblum/repos/opentelemetry-ruby/resource_detectors/test/opentelemetry/detectors/deployment_test.rb:9:in `block (3 levels) in <top (required)>'
```

Any thought on this? should I update the tests somehow or just leave it the way it is?

blumamir:
> This seems like it has good motivation, but basing resource detection on the environment variables seems like guessing to me. I believe the detection of the framework specific attributes on the resource should belong to instrumentation. E.g. `rails_env_check` should be part of rails instrumentation.

If we detect the environment in the framework instrumentation, then it will be set as span attributes only for this framework spans. I think it's also a valid strategy but has the following cons:
1. only framework spans will be marked with the environment, which means the user will not be able to tell the environment based on downstream service span or trace which started not by the framework.
2. it's less performant, as a resource is immutable for the service and is reported only once per batch, and possibly stored more efficiently in backends. If we publish it as span attribute, it will have to be reported again and again for each framework span.
3. The environment is a value that is static to the run of the service, thus I believe the most suitable place for it is in the resource.
4. Since the Deployment resource detector is defined in the specification, there is a value IMO in supplying such detector for ruby as well. This will give users a unified experience between SDKs in different languages, and backend processing pipelines can rely on common behavior to ingest trace data coming from different sources (they can be written to look for the "deployment.environment" in resource)


> but basing resource detection on the environment variables seems like guessing to me

I agree that this implementation is not 100% robust (as I stated in the comments), and there are edge cases in which it will not work, but IMO, this is the best we can do as long as the resource is an immutable entity in OTEL (which is a problem for many other use-cases as well).

Maybe we can supply a configuration option to additionally capture environment in relevant instrumentations, and then user can choose between installing a resource detector or enabling this option in instrumentation?

YanivD (approved):

blumamir:
As discussed in the SIG - this detector probably belongs to a future contrib repo, since:
1. It only applies to rails, sinatra and rack, and will not report env for other frameworks
2. If multiple frameworks are used, then only one arbitrary value will be used
3. Fetching values from environment variables will not always reflect the real environment set by the framework/tools/ conventions
4. The name "Deployment" can be a bit confusing as the only attribute detected is "environment".

So I'm closing the PR for this repo and will add it in the future to some other repo that better fits it's characters.

Thanks all for the reviews and the discussion

PR #1023, opened 2021-11-19
fix: aws-sdk nitpicks

robertlaurin:


instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
     module AwsSdk
       # Generates Spans for all interactions with AwsSdk
       class Handler < Seahorse::Client::Handler
-        def call(context)
-          span_name = get_span_name(context)
-          attributes = get_span_attributes(context)
+        def call(context) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+          return super unless context
```
robertlaurin, line 13: I noticed that `context&` was being invoked in multiple places.  I think without context we don't have a service name or an operation name, which means we don't have a span name.  That's likely indication of a bigger issue in the call chain here.  Regardless I think it's safer to return a call to super if we cannot collect any of the information to generate a span.  Happy to hear differing opinions here, such as potentially falling back to some ultra generic span naming without any attributes.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
     module AwsSdk
       # Generates Spans for all interactions with AwsSdk
       class Handler < Seahorse::Client::Handler
-        def call(context)
-          span_name = get_span_name(context)
-          attributes = get_span_attributes(context)
+        def call(context) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
```
robertlaurin, line 12: These rubocop rules lead to some a scattering of really shallow methods, it's been a long standing "todo" to clean it up.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
+            end.tap do |response|
+              if (err = response.error)
+                span.record_exception(err)
+                span.status = Trace::Status.error(err)
+              end
             end
           end
         end
 
-        def get_span_attributes(context)
```
robertlaurin, line 34: A common ruby convention is to not name methods in the style of `get_*` , and instead just name them after the thing they're returning.  In this case `get_span_attributes` would be `span_attributes`.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
-          span_attributes.merge(db_attributes)
-        end
-
-        def get_service_name(context)
-          context&.client.class.api.metadata['serviceId'] || context&.client.class.to_s.split('::')[1]
-        end
-
-        def get_operation(context)
-          context&.operation&.name
-        end
```
robertlaurin, line 52: I removed the get_service_name and get_operation methods as they were each being called multiple times and instead opted to storing them once as a variable that we could reuse in the call method.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
+          operation = context.operation&.name
+          attributes = {
+            'aws.region' => context.config.region,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
+            OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
+          }
+          attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
 
+          tracer.in_span("#{service_name}.#{operation}", attributes: attributes, kind: OpenTelemetry::Trace::SpanKind::CLIENT) do |span|
```
robertlaurin, line 25: With service_name and operation available we can just inline it here instead of it having a one off method.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
-            }
+          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]
+          operation = context.operation&.name
+          attributes = {
+            'aws.region' => context.config.region,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
+            OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
+          }
+          attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
```
robertlaurin, line 23: I removed the db helper as this one liner didn't seem like it needed a module.  I think we can abstract this portion away if/when a need arises.   It also prevents from creating an additional hash to be merged.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
-        def get_span_attributes(context)
-          span_attributes = {
-            'aws.region' => context.config.region,
-            OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
-            OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => get_operation(context),
-            OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => get_service_name(context)
-          }
-
-          db_attributes = DbHelper.get_db_attributes(context, get_service_name(context), get_operation(context))
-          span_attributes.merge(db_attributes)
```
robertlaurin, line 43: `.merge` was creating a copy instead of mutating the `span_attributes` hash.

robertlaurin (commented):
Hey @YanivD, I did a few follow up changes to the AWS SDK that I'd like to get merge before the release.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
ericmustin: ideally this instrumentation has many modules with many helpers. IIRC there is sns/sqs additions coming shortly, and we should assume that we'll need a pattern to handle enriching the span attributes with extra information depending on the service, ala node https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk/src/services . That being said, fine to kick the can on that.

ericmustin (approved):

SomalianIvan (approved):
:lgtm:

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,51 +9,34 @@ module Instrumentation
...
+            OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
+            OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
+            OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
+          }
+          attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
 
+          tracer.in_span("#{service_name}.#{operation}", attributes: attributes, kind: OpenTelemetry::Trace::SpanKind::CLIENT) do |span|
             if instrumentation_config[:suppress_internal_instrumentation]
-              OpenTelemetry::Common::Utilities.untraced(&execute)
+              OpenTelemetry::Common::Utilities.untraced { super }
```
YanivD, line 27: @robertlaurin 
When enabling `suppress_internal_instrumentation` config, it should still record errors in `aws-sdk` span.
This part is missing when `suppress_internal_instrumentation` is `True`.
```rb
if (err = response.error)
  span.record_exception(err)
  span.status = Trace::Status.error(err)
end
```

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: I separated it to different module in order to be able to extend the basic client implementation and to add attributes for specific AWS services.
I'm working on another PR to support message attributes for SQS/SNS, including context propagation. 

I'm fine with this implementation for now, soon I'll open another PR with the ability to extend aws services.

YanivD (changes requested):
@robertlaurin Thanks for creating this PR! I'm ok with all code changes you've made. 

There's one change that should be done - https://github.com/open-telemetry/opentelemetry-ruby/pull/1023/files#r753765192

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: Hmm, after a second look, I see it is recording errors in case of `suppress_internal_instrumentation`. 
Please ignore this comment.

YanivD (approved):

PR #1025, opened 2021-11-20
feat: Version Compatibility Based on Gem Specs

arielvalentin:
This changes provides instrumentation authors the ability to use gem specifications for version compatibility checks instead of having to write a custom compatibility block.

In order to take advantage of this feature instrumentation authors must:

1. Declare the library and version constraints as a development dependency
2. Use instrumentation naming conventions e.g. `opentelemetry-instrumentation-ruby_kafka`

Related https://github.com/open-telemetry/opentelemetry-ruby/issues/988

Some Concerns:

- There are certainly many more test scenarios to take into consideration. E.g. When multiple versions of the gem are installed and not yet activated.  
- Naming is strict and requires that the instrumentation name suffix matches the target library. We currently have some cases where the instrumentation names do not match e.g. `ruby-kafka` and `activesupport`

### DIFF
--- a/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
+++ b/instrumentation/base/lib/opentelemetry/instrumentation/base.rb
@@ -300,19 +300,26 @@ module OpenTelemetry
 
       private
 
-      # Checks compatability of `library_name` against the `development_dependency` declared in  instrumentations gemspec
+      # EXPERIMENTAL: Checks compatability of `library_name` against the `development_dependency` declared in instrumentations gemspec
+      #
+      # This method will first search `Gem.loaded_specs` to locate gemspecs that have already been required by RubyGems or Bundler.
+      # In the case the gems were not loaded, or do not use RubyGems/Bundler, it will fallback to searching for installed gems via `Gem::Specification.find_by_name`.
+      #
+      # Instrumentation authors are encouraged to implement a `present` block to ensure the constants are availble to use in monkey patches.
+      #
+      # See https://github.com/DataDog/dd-trace-rb/pull/1510
+      # rubocop:disable Metrics/AbcSize
       def compatible_version?
-        begin
-          instrumentation_spec = Gem.loaded_specs[instrumentation_gem_name] || Gem::Specification.find_by_name(instrumentation_gem_name)
-          library_spec = Gem.loaded_specs[library_name] || Gem::Specification.find_by_name(library_name)
+        instrumentation_spec = Gem.loaded_specs[instrumentation_gem_name] || Gem::Specification.find_by_name(instrumentation_gem_name)
+        library_spec = Gem.loaded_specs[library_name] || Gem::Specification.find_by_name(library_name)
 
-          dependency = instrumentation_spec.development_dependencies.find { |spec| spec.name == library_spec.name }
-          dependency&.requirement&.satisfied_by?(library_spec.version) == true
-        rescue Gem::MissingSpecError => e
-          OpenTelemetry.handle_error(message: 'Instrumentation Compatibility Check Error', exception: e)
-          return false
-        end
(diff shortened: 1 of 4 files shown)


instrumentation/base/lib/opentelemetry/instrumentation/base.rb
```diff
@@ -293,11 +293,15 @@ def enabled?(config = nil)
...
-        dependency = instrumentation_spec&.development_dependencies&.find { |spec| spec.name == library_spec.name }
-        dependency&.requirement&.satisfied_by?(library_spec.version) == true
+        begin
+          instrumentation_spec = Gem.loaded_specs[instrumentation_gem_name] || Gem::Specification.find_by_name(instrumentation_gem_name)
+          library_spec = Gem.loaded_specs[library_name] || Gem::Specification.find_by_name(library_name)
+
+          dependency = instrumentation_spec.development_dependencies.find { |spec| spec.name == library_spec.name }
+          dependency&.requirement&.satisfied_by?(library_spec.version) == true
+        rescue Gem::MissingSpecError
+          return false
```
ericmustin, line 303: A debug or warn log here couldn’t hurt?

arielvalentin: Sure thing.

arielvalentin: I chose to use `handle_error`. Seem reasonable?

arielvalentin:
Ping @open-telemetry/ruby-maintainers @open-telemetry/ruby-approvers 🙏

ericmustin:
@arielvalentin My two cents is i think this looks great and is an improvement on the current system. It feels _ a little_ hacky to me to rely on development dependancies in a gemspec, for something like an instrumentation that doesn't necessarily need to be packaged as a gem to begin with, but since this doesn't override the compatible block, but only acts as a useful fallback/default, I don't see the harm. that being said I'd like some other folks to chime in here, I brought this up in the SIG, would be keen to hear other POVs here.

instrumentation/base/lib/opentelemetry/instrumentation/base.rb
ericmustin: 👍

arielvalentin:
> @arielvalentin My two cents is i think this looks great and is an improvement on the current system. It feels _ a little_ hacky to me to rely on development dependancies in a gemspec, for something like an instrumentation that doesn't necessarily need to be packaged as a gem to begin with, but since this doesn't override the compatible block, but only acts as a useful fallback/default, I don't see the harm. that being said I'd like some other folks to chime in here, I brought this up in the SIG, would be keen to hear other POVs here.

@ericmustin I would think folks that are writing their own auto-instrumentation that are not packaged as a gem are likely either implemented first party or patched in their applications. It's likely they are aware what the compatibility requirements are in their own projects or may not be using instrumentation at all. 

Could you elaborate a bit on what you think makes it _a little hacky_?

ericmustin:
> Could you elaborate a bit on what you think makes it a little hacky?

I don't think that checking [development dependancies in a gemspec from production code](https://github.com/open-telemetry/opentelemetry-ruby/pull/1025/files#diff-7379de806670e7a913ebbe1d8e27945a6367cc83c94d79e55434de15cdd16732R316) is a common convention. Maybe hacky isn't the right term.  That being said I think it's fine imo, just, unusual. Again i'd like some other input here from @open-telemetry/ruby-maintainers @open-telemetry/ruby-approvers

arielvalentin:
@robertlaurin do you have any time this week to take a look?

arielvalentin:
Closing per my offline conversation with @robertlaurin https://github.com/open-telemetry/opentelemetry-ruby/issues/988#issuecomment-1018676853

PR #1026, opened 2021-11-21
feat: SQS / SNS messaging attributes and context propagation

YanivD:
### What's added in this PR?
- Added message attributes for SQS and SNS services
- Added context propagation support for SQS and SNS. For `SQS.SendMessage`, `SQS.SendMessageBatch` and `SNS.Publish` methods

### Context propagation strategy
Currently SQS doesn't natively support propagating the OTEL context. 
I used [JS aws-sdk instrumentation](https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/plugins/node/opentelemetry-instrumentation-aws-sdk) as an inspiration for this PR. It use message metadata to populate propagator fields. 

### Limitations 
[SQS support up to 10 message metadata attributes](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-quotas.html). Context attributes will not be added if there are too many attributes.

### What's missing?
Currently context propagation is working only if the receiver is using JS aws-sdk. I'm planning to add support in another separated PR. 

---

Note: This PR contain file changes from https://github.com/open-telemetry/opentelemetry-ruby/pull/1023


mottibec (approved):

ericmustin:
@mottibec @blumamir @YanivD I'm not quite sure what these approvals from your colleagues are supposed to indicate. The work and motivation is absolutely appreciated, _but_ there are currently multiple errors in this instrumentation and the test suite is clearly and loudly failing and some changes are still required. Approving PRs without reviewing them doesn't do anything except add noise, and erode confidence that you and your colleagues are carefully implementing and testing the changes in this Pull Request. I fully understand that you have product/professional obligations which rely on successfully merging and releasing this work, and I am scheduling time to review this this week. Thank you for your patience and for all the great work so far.

YanivD:
@ericmustin
We're excited for the opportunity to contribute and help the community. We didn't mean to make unnecessary noises and we're sorry for that. I'll make sure it won't happen again 🙏
Folks, we really appreciate all your help, guidance and your professional quick response.
I'll address the failing tests soon and make sure this pr is ready for review. Until then, marking it as draft. Any review and suggestions is welcome.

Again, I'm sorry for unnecessary noises.

YanivD:
[Updated the pr](https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/dc44180554fb8f843f78906ef42a2a511c85aaf1), now the tests are green and ready for review.
The failing tests were from aws-sdk version 2. The context interface changed in version 3. _Appraisal_ tool is great!

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb
```diff
@@ -49,6 +49,10 @@ def apply_sns_attributes(attributes, context, client_method)
             attributes[SemanticConventions::Trace::MESSAGING_DESTINATION_KIND] = 'topic'
             attributes[SemanticConventions::Trace::MESSAGING_DESTINATION] = queue_name(context)
           end
+
+          def params(context, param)
+            defined?(context.metadata[:original_params][param]) ? context.metadata[:original_params][param] : context.params[param]
```
arielvalentin, line 54: Does a `context` have a different interface depending on a specific flow? Why is that?

YanivD: `context` have a different interface depending on `aws-sdk-ruby` gem version.

### aws-sdk-ruby V2
```rb
context.params[:param]
```

### aws-sdk-ruby V3
```rb
context.metadata[:original_params][:param]
```

arielvalentin: Thanks for explaining that. I think we need to come up with some guidance or implementation patterns for supporting different versions of a gem. 

I wonder if we could encode this in some programmatic way. May I ask that you add a method comment that includes links to the AWS SDK in the meantime?

YanivD: @arielvalentin 
After revisiting the source code of 'aws-sdk-ruby', I found out that `context.params[param]` exists both in v2 and v3. 
Updated PR to use `context.params[param]` instead.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
+            inject_context(context, client_method)
             if instrumentation_config[:suppress_internal_instrumentation]
               OpenTelemetry::Common::Utilities.untraced { super }
             else
               super
             end.tap do |response|
               if (err = response.error)
                 span.record_exception(err)
-                span.status = Trace::Status.error(err)
+                span.status = Trace::Status.error(err.to_s)
```
YanivD, line 42: This fix an error caused because `err` is not a string

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
             OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
             OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
           }
           attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
+          MessagingHelper.apply_sqs_attributes(attributes, context, client_method) if service_name == 'SQS'
+          MessagingHelper.apply_sns_attributes(attributes, context, client_method) if service_name == 'SNS'
 
-          tracer.in_span("#{service_name}.#{operation}", attributes: attributes, kind: OpenTelemetry::Trace::SpanKind::CLIENT) do |span|
+          tracer.in_span(span_name(context, client_method), attributes: attributes, kind: span_kind(operation)) do |span|
+            inject_context(context, client_method)
```
johnduhart, line 34: Please provide a way to opt-out of automatic context propagation. We have internal systems which are doing context propagation within the message body and we'd rather not duplicate that in attributes.

arielvalentin: @YanivD take a look at how other instrumentations do this, e.g. ActiveJob provides link, child, or none

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/instrumentation.rb#L55

YanivD: @johnduhart 
Thanks. I think it's a good idea.
I'm thinking how to call the option. Suggestions are very welcome.

@arielvalentin 
In this PR I didn't add support for `process` spans which extract the context propagation. I'm not sure `propagation_style` is the best fit. WDYT?

arielvalentin: My mistake! I did not look closely enough to see the PR only injects context. 

Is there a separate PR for extraction?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
             OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
             OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
             OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
           }
           attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
+          MessagingHelper.apply_sqs_attributes(attributes, context, client_method) if service_name == 'SQS'
+          MessagingHelper.apply_sns_attributes(attributes, context, client_method) if service_name == 'SNS'
 
-          tracer.in_span("#{service_name}.#{operation}", attributes: attributes, kind: OpenTelemetry::Trace::SpanKind::CLIENT) do |span|
+          tracer.in_span(span_name(context, client_method), attributes: attributes, kind: span_kind(operation)) do |span|
```
ericmustin, line 33: I don't think the cardinality of `span_name()` is appropriate for messaging spans., Below are the  `span_name` and `MessagingHelper.queue_name` methods. 

Specifically, it seems possible to me that, SNS will result in incredibly high cardinality span names with a phone number in the span name, like for example, `867 5309 send`. However, the messaging span conventions on span name (and general span name conventions also apply here) [state](https://github.com/open-telemetry/opentelemetry-specification/blob/a1a8676a43dce6a4e447f65518aef8e98784306c/specification/trace/semantic_conventions/messaging.md#span-name):

> The destination name SHOULD only be used for the span name if it is known to be of low cardinality (cf. general span name guidelines). This can be assumed if it is statically derived from application code or configuration. Wherever possible, the real destination names after resolving logical or aliased names SHOULD be used. If the destination name is dynamic, such as a conversation ID or a value obtained from a Reply-To header, it SHOULD NOT be used for the span name. In these cases, an artificial destination name that best expresses the destination, or a generic, static fallback like "(temporary)" for temporary destinations SHOULD be used instead.

Additionally, having something with obvious PII implications in the span name feels gross to me. I realize this is how Node does it currently but, frankly, I think that's incorrect and i'd prefer to see this pattern not extended into Ruby. Or, we should get some clarification from the AWS team on how to represent this.

For now i'd prefer to see Phone number replaced with something either a message attribute of much lower cardinality, or something static like `phone_number send`

```ruby
        def span_name(context, client_method)
          case client_method
          when SQS_SEND_MESSAGE, SQS_SEND_MESSAGE_BATCH, SNS_PUBLISH
            "#{MessagingHelper.queue_name(context)} send"
          when SQS_RECEIVE_MESSAGE
            "#{MessagingHelper.queue_name(context)} receive"
          else
            client_method
          end
        end
```

```ruby
          def queue_name(context) # rubocop:disable Metrics/CyclomaticComplexity
            topic_arn = context.params[:topic_arn]
            target_arn = context.params[:target_arn]
            phone_number = context.params[:phone_number]
            queue_url = context.params[:queue_url]

            if topic_arn || target_arn
              arn = topic_arn || target_arn
              begin
                return arn.split(':')[-1]
              rescue StandardError
                return arn
              end
            end

            return phone_number if phone_number

            return queue_url.split('/')[-1] if queue_url

            'unknown'
          end
```

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
ericmustin: Agreed, we absolutely need to supply

- an `option` for enabling(as child or as link) / disabling propagation as @johnduhart points out. 
- support for both injecting and extracting context as @arielvalentin points out.

I think one of the difficulties here is that `SQS_RECEIVE_MESSAGE` can receive either a single message, or a batch of messages. 

- The simple case, the default, which is we receive, at maximum, `1` message: We can infer this by looking at the [`MaxNumberOfMessages` request parameter as defined here](https://docs.aws.amazon.com/AWSSimpleQueueService/latest/APIReference/API_ReceiveMessage.html), which is how our [good friends at Datadog have approached this](https://github.com/DataDog/dd-trace-js/blob/25264dce67a97384d7f1abc541155cecc2324679/packages/datadog-plugin-aws-sdk/src/services/sqs.js#L45-L48). In the case where we are receiving 1 message, we can relatively easily choose to extract as a child, as a link, or not at all. We can follow the approach taken, as @arielvalentin points out, in [one of our other instruments, like ActiveJob](https://github.com/open-telemetry/opentelemetry-ruby/blob/871ddb4086624db134e003a55e9c57e687bffe25/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb#L30-L55).
- The complex case, which isn't default but practically speaking is the more common pattern since it's more efficient, which is that we can receive _many_ messages in a `RECEIVE_MESSAGE` `handler`, each message potentially containing a different trace-id/parent-span. In this case, because we don't have a seperate span for processing _each message_, we can't effectively extract context in a parent/child relationship since each individual message has it's own parent. but our span can only have 1 parent.  So we're at a bit of an impasse since we need to handle this in some way.  Practically speaking, I think there are three options here:
  - First,  _the datadog way_, where we could simply just _not support_ propagation of trace context when we detect a potential a "batch" of messages ( `MaxNumberOfMessages > 1` ). We aren't really beholden to any specification here since this instrumentation is experimental, so i think if push comes to shove I don't hate this approach, but I think it's practically speaking a _"Half Measure"_, as Mike Ehrmantraut might say.
  - Second, _the node/aspecto way_ , where we create a _span per message processed_, and then because we're back to a 1:1 relationship with span => parent, we can easily choose to propagate as `child/link/not-at-all`. I think this would require additional monkey-patching of the underlying aws-sdk ruby library, so that we can measure each individual message processed, and i haven't reviewed this library closely enough to know if/how-complex that is to do, but that seems like that would match the functionality of Node, and also allow us to expose the option like @arielvalentin is describing and like i've described in the "simple" case.
  - Third, _the Links way_, where we simply do not do any additional monkey patching, but, when we detect that `MaxNumberOfMessages > 1`, we iterate over each message's headers, extract each valid trace context, and attach them _all_ as Span Links of the span.  I actually tend to think this is the "semantically correct" use of [opentelemetry span links](https://github.com/open-telemetry/opentelemetry-specification/blob/e2c2472985b17e37a25a7dc5aa0aa071e6683c98/specification/overview.md#links-between-spans), but since practically speaking many backends don't support Span Links, would still result in disconnected traces in practice, and also wouldn't allow people to maintain parent/child relationships for their spans. This would be a change in convention from both how Datadog and how Node/Aspecto do things currently, but is perhaps a middle ground between the half measure of the 1st option i suggested, and the extra work of the 2nd option i suggested.


So, I think we need to make a determination here around how we want to handle this. I'm open to feedback here, but IMO both of these details (options to enable/disable, and how to support extract propagation in the simple/complex case) I think need to be resolved before we can support sqs propagation and merge this PR.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
+        SQS_SEND_MESSAGE = 'SQS.SendMessage'
+        SQS_SEND_MESSAGE_BATCH = 'SQS.SendMessageBatch'
+        SQS_RECEIVE_MESSAGE = 'SQS.ReceiveMessage'
+        SNS_PUBLISH = 'SNS.Publish'
+
+        def call(context) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           return super unless context
 
-          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]
+          service_name = context.client.class.to_s.split('::')[1]
```
ericmustin, line 20: I'm confused, why are we dropping the check for `serviceId`?

ericmustin (changes requested):
Overall I think this looks very good.

However, I think we have to address `extract` propagation and the options we should expose for `child/link/none` before merging this PR.

I have some other minor feedback but that is more straightforward.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
+          client_method = "#{service_name}.#{operation}"
           attributes = {
             'aws.region' => context.config.region,
             OpenTelemetry::SemanticConventions::Trace::RPC_SYSTEM => 'aws-api',
             OpenTelemetry::SemanticConventions::Trace::RPC_METHOD => operation,
             OpenTelemetry::SemanticConventions::Trace::RPC_SERVICE => service_name
           }
           attributes[SemanticConventions::Trace::DB_SYSTEM] = 'dynamodb' if service_name == 'DynamoDB'
+          MessagingHelper.apply_sqs_attributes(attributes, context, client_method) if service_name == 'SQS'
+          MessagingHelper.apply_sns_attributes(attributes, context, client_method) if service_name == 'SNS'
```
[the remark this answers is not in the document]
ericmustin: Ah i see there's some common attribs documented here, so, i think this can be disregarded https://github.com/open-telemetry/opentelemetry-specification/blob/a1a8676a43dce6a4e447f65518aef8e98784306c/specification/trace/semantic_conventions/instrumentation/aws-sdk.md#common-attributes

YanivD:
Just to update - I'll address the open conversations in the next few days. Sorry for the long respond time.

ericmustin:
@YanivD No worries.

As discussed in the SIG, after some discussion I'm comfortable taking this PR with

1. Config for enabling/disabling `inject` propagation
2. The API for `extract`. Although we've laid out options in [this discussion](https://github.com/open-telemetry/opentelemetry-ruby/discussions/1057), I don't believe we need to provide implementation for `child / link / none` extract method in the instrumentation yet, since it's not clear what the correct use case should be and awaiting that answer could take an indefinite amount of time. However, we know generally _how_ the extraction API should be structured, so let's make sure that's added in this PR. If users followup and request a certain approach to applying that extraction in the instrumentation, then we can provide that.
3. Address the other smaller feedback points (straightforward, mostly nits, some cleanup, etc)

After those are accomplished i'd like to get this approved and merged as, even without `extract` implemented in the instrumentation, it's still incrementally useful (it provides better metadata for sns/sqs spans as well as labels their `kind` and `name` more appropriately)

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: updated 👍
https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/b28178ef1a1677d6d039b5449fa9a35a7febe34b

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: Re-added the check for `serviceId`.

It was part of adjustments for support old versions `v2.0.x`
But it see there are more issues, so it removed `v2.0.x` from supported versions https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/f007501b534d3afd5a166319160cc91d6acdd9e4

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: @arielvalentin
I'm not planning to add support for message context extraction soon. 
There are few open issues with messaging spec for processing SQS messages. [Link for discussion](https://github.com/open-telemetry/opentelemetry-ruby/discussions/1057) about some of the issues. I'll probably implement extraction in the next few months after some messaging spec changes.

YanivD:
@ericmustin 
1. added config `inject_messaging_context` with default value `false` https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/51ba5b33f931ce0f37378e382c2f99b132dd6341
2. added api for extract https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/d401b38549bf00be9cd8da518fbe2d25d4a048c5
3. addressed open points 

Please re-review this PR

ericmustin (approved):
ty for addressing my tedious feedback 😅, appreciate the patience here. This lgtm, approving this. a few housekeeping things before this gets merged

- please rebase i think dalli/kafka failures in CI are unrelated to this? If they still persist ping me i will fix them.
- Why are we bumping from aws-sdk 2.0 to aws-sdk 2.1 min version? i'm fine if there's a valid reason, or frankly i don't think it matters too much since this gem is so new no one is likely using it in conjunction with aws 2.0 at the moment, but if you wouldn't mind explaining why we're bumping and also let's just also drop the appraisal for 2.0 if we do, https://github.com/open-telemetry/opentelemetry-ruby/blob/4216a55be5ad224100f6c2802e8ec5fc04f1e316/instrumentation/aws_sdk/Appraisals#L59-L61
- I don't think it's necessary to move the propagation stuff into it's own repo in the `propagator` folder since this is only partial propagation support and it's experimental, but I defer to @robertlaurin and @fbogsany here.
- there's maybe some bikeshedding on the option name but i think it's fine as is

Nice work @YanivD , thanks again for this

YanivD:
Thanks @ericmustin 
After revisiting v2.0 support, I modified service_name extraction to support 2.0 versions - https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/77074a15156b2baebbac3175e8666b9f7ccb5822.
This PR won't change the supported versions.

@open-telemetry/ruby-maintainers 
Please let me know if more changes are required

ericmustin:
@YanivD lgtm, ty for addressing that, will try to get some other eyes on it from approvers/maintainers this week. Just for reference in US/CA we're a bit light this week and next bc of Christmas holidays, apols if there's any delay here. 🎅 😅

YanivD:
I want to update that I'm working on extracting SQS context, for creating '_process_' spans. 
Do you prefer I'll add the extraction part to this PR? Or to create separated PR for that?

mwear (approved):
I have two minor comments / suggestions, otherwise, LGTM.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -45,6 +54,43 @@ def tracer
         def instrumentation_config
           AwsSdk::Instrumentation.instance.config
         end
+
+        def inject_context(context, client_method)
+          return unless [SQS_SEND_MESSAGE, SQS_SEND_MESSAGE_BATCH, SNS_PUBLISH].include? client_method
```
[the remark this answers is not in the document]
YanivD: Good idea. updated in https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/431deae69b0f265ca9ba6fdd32433ab8176c286f

@mwear thanks for your review and comments

ericmustin:
@YanivD I'd say lets move the extract stuff into a seperate PR if possible so we can get this chunk of work merged  (i see you've already pushed up some commits, apologies for the delayed response), the extract stuff is probably going to take a bit of time to review and get feedback for, and i don't want that to block the work you've already done here from getting released

YanivD:
@open-telemetry/ruby-maintainers
Is there anything left to do before merging this PR to main branch?

habmic:
Do you know when this PR will be merged?

ericmustin:
@habmic @YanivD ack'd, will make sure this gets surfaced at SIG today

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -9,28 +9,37 @@ module Instrumentation
...
+        SQS_SEND_MESSAGE = 'SQS.SendMessage'
+        SQS_SEND_MESSAGE_BATCH = 'SQS.SendMessageBatch'
+        SQS_RECEIVE_MESSAGE = 'SQS.ReceiveMessage'
+        SNS_PUBLISH = 'SNS.Publish'
+
+        def call(context) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
           return super unless context
 
-          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1]
+          service_name = context.client.class.api.metadata['serviceId'] || context.client.class.to_s.split('::')[1] rescue context.client.class.to_s.split('::')[1] # rubocop:disable Style/RescueModifier
```
fbogsany, line 20: What is this intended to do? The `rescue` clause contains the same expression as the right side of `||`, so if that raised an exception, then the `rescue` clause will as well.

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/messaging_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+  module Instrumentation
+    module AwsSdk
+      # MessagingHelper class provides methods for calculating messaging span attributes
+      class MessagingHelper
+        class << self
+          def queue_name(context)
+            topic_arn = context.params[:topic_arn]
+            target_arn = context.params[:target_arn]
+            phone_number = context.params[:phone_number]
+            queue_url = context.params[:queue_url]
```
fbogsany, line 17: I'd consider moving these two down to immediately before the corresponding `return ... if ...`.

YanivD: Thanks, moved it. https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/5f34bc044a18fcfbab6ada684559511b8d76bd6b

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
YanivD: Updated service_name extraction https://github.com/open-telemetry/opentelemetry-ruby/pull/1026/commits/1a59955c5663446a47324c9193d82a7c4475337e

This change was to support `2.0.x` versions. When using `2.0.x`, `metadata` method expect an argument, what raised an exception.

Please re-review this change and approved it's safe.

fbogsany (approved):

ericmustin (approved):

nvolker:
@YanivD just a question on the "What's Missing" part of your description. Specifically "I'm planning to add support in another separated PR", is this to do with extractor propagation on the receiver side? And are you actively looking into this?

YanivD:
@nvolker 
I have implemented the extraction part on my fork, it's still a draft and not fully ready to review. You can see it here - https://github.com/YanivD/opentelemetry-ruby/pull/12

If you have the time to finish the changes and open PR - it will really help and you are very welcome :)
Otherwise, I'll probably get to it and the next few weeks and will open a PR

PR #1032, opened 2021-11-23
Sidekiq messaging operation

habmic:
### What's added in this PR?
Adding `messaging.operation` attribute to sidekiq.
The [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#messaging-attributes) doesn't require it but it could be helpful.

The value must be either `receive` or `process`, in sidekiq case, there isn't a receive span so I have added only for `process`.
For the specification:

"The values allowed for <operation name> are defined in the section Operation names below. If the format above is used, the operation name MUST match the messaging.operation attribute defined for message consumer spans below." 

About tests, I felt it is too small of a change to add test, let me know if you think otherwise

Thanks!
(BTW, my first contribution here. My apologies in advance if I missed some contribution rules.)

### DIFF
--- a/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
+++ b/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/middlewares/client/tracer_middleware.rb
@@ -18,7 +18,8 @@ module OpenTelemetry
                 'messaging.sidekiq.job_class' => job['wrapped']&.to_s || job['class'],
                 'messaging.message_id' => job['jid'],
                 'messaging.destination' => job['queue'],
-                'messaging.destination_kind' => 'queue'
+                'messaging.destination_kind' => 'queue',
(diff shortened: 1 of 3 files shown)


arielvalentin:
@habmic Thank you for the initial contribution and welcome you to the OTEL Ruby project!

Although this is a small change, I would kindly ask you to update the test suite as well.

In addition to that I would encourage you to make similar changes to related instrumentations so that our users have a consistent experience. Small PRs are always welcome so you could submit those changes separately or if you feel comfortable doing so include them as part of this PR.

habmic:
@arielvalentin I have added to the unit tests validations for the new attribute

For the rest of the instrumentations I'll add them in a separate PR, can you point me to the relevant instrumentations?

arielvalentin:
Using code search I am able to find a few results here:

https://github.com/open-telemetry/opentelemetry-ruby/search?q=messaging&type=

fbogsany (approved):

PR #1033, opened 2021-11-23
feat: Instrument Que poller

indrekj:
Que polls the database every few seconds to find new jobs. This
generates postgresql spans (if pg instrumentation library is enabled)
without any parents and without much context. These can get quite noisy.

This commit by default disables these traces. It is however possible to
enable it with `trace_poller` option.

There are also many other places where Que can be noisy, but for now
just starting with the Poller.

A lot of inspiration was taken from the `sidekiq` library.

### DIFF
--- a/instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb
+++ b/instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb
@@ -39,12 +39,14 @@ module OpenTelemetry
         #     direct child of the span that enqueued the job.
         #   - :none - the job's execution will not be explicitly linked to the
         #     span that enqueued the job.
+        # trace_poller: controls whether Que Poller is traced or not.
         #
         # Note that in all cases, we will store Que's Job ID as the
         # `messaging.message_id` attribute, so out-of-band correlation may
         # still be possible depending on your backend system.
         #
         option :propagation_style, default: :link, validate: ->(opt) { %i[link child none].include?(opt) }
+        option :trace_poller,      default: false, validate: :boolean
 
         private
 
@@ -52,6 +54,7 @@ module OpenTelemetry
           require_relative 'tag_setter'
           require_relative 'middlewares/server_middleware'
           require_relative 'patches/que_job'
+          require_relative 'patches/poller'
         end
 
         def gem_version
@@ -60,6 +63,7 @@ module OpenTelemetry
 
         def patch
           ::Que::Job.prepend(Patches::QueJob)
+          ::Que::Poller.prepend(Patches::Poller)
         end
       end
     end
--- a/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
+++ b/instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
@@ -1,0 +1,29 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module Que
+      module Patches
+        # Instrumentation for the Que::Poller module
+        module Poller
+          def poll(*args, **kwargs)
+            # Avoid tracing when should_poll? returns true. This is also used
+            # in Poller#poll to decide if the actual poll should be executed or
+            # not. Without this we would generate a lot of unnecessary spans.
+            return unless should_poll?
+
+            if Que::Instrumentation.instance.config[:trace_poller]
+              Que::Instrumentation.instance.tracer.in_span('Que::Poller#poll') { super(*args, **kwargs) }
+            else
+              OpenTelemetry::Common::Utilities.untraced { super(*args, **kwargs) }
+            end
+          end
+        end
+      end
+    end
+  end
+end
--- a/instrumentation/que/opentelemetry-instrumentation-que.gemspec
+++ b/instrumentation/que/opentelemetry-instrumentation-que.gemspec
@@ -32,6 +32,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency 'appraisal', '~> 2.2.0'
   spec.add_development_dependency 'bundler', '>= 1.17'
   spec.add_development_dependency 'minitest', '~> 5.0'
+  spec.add_development_dependency 'opentelemetry-instrumentation-pg', '~> 0.19.1'
   spec.add_development_dependency 'opentelemetry-sdk', '~> 1.0'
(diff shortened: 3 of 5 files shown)


instrumentation/que/lib/opentelemetry/instrumentation/que/instrumentation.rb
```diff
@@ -60,6 +63,7 @@ def gem_version
 
         def patch
           ::Que::Job.prepend(Patches::QueJob)
+          ::Que::Poller.prepend(Patches::Poller)
```
arielvalentin, line 66: @open-telemetry/ruby-maintainers Should this check the `trace_poller` option here and only mixing if the option was set to `true`?

That would make it symmetric with how the top level instrumentation works, which skips the `install` block when the instrumentation `enabled` is `false`: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L216

indrekj: We want to do this even if this is `false`. Because then we use `OpenTelemetry::Common::Utilities.untraced { super }` in the patched method. This makes sure the inner spans (postgres spans) are not traced.

arielvalentin: Thanks for the explanation. I will follow up with some questions about this inline as part of `que/patches/poller`

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
```diff
@@ -0,0 +1,30 @@
...
+        module Poller
+          def poll(*)
+            # Avoid tracing when should_poll? returns true. This is also used
+            # in Poller#poll to decide if the actuall poll should be executed
+            # or not. Without this we would generate a lot of unnecessary
+            # spans.
+            return unless should_poll?
+
+            if Que::Instrumentation.instance.config[:trace_poller]
+              Que::Instrumentation.instance.tracer.in_span('Que::Poller#poll') { super }
```
arielvalentin, line 21: In this case the code is is going to generate a span for `Que::Poller#poll` as well as the `pg` spans correct?

Is there a case where users would want to granular control and suppress the `pg` spans here as well?

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
```diff
@@ -0,0 +1,30 @@
...
+    module Que
+      module Patches
+        # Instrumentation for the Que::Poller module
+        module Poller
+          def poll(*)
+            # Avoid tracing when should_poll? returns true. This is also used
+            # in Poller#poll to decide if the actuall poll should be executed
+            # or not. Without this we would generate a lot of unnecessary
+            # spans.
+            return unless should_poll?
```
arielvalentin, line 18: I am concerned of the risk here that this will possibly end up changing the implementation/behavior of the target class `Que::Poll#poll`. Is there a way to instrument this class in a way that leaves the behavior in tact and rely on the Gem implementation without reimplementing the logic in the instrumentation?

indrekj: I also wanted to avoid changing this but I didn't find any other way. The [poll](https://github.com/que-rb/que/blob/61e1805deb8161e5c1502e518a9a44fa464d308c/lib/que/poller.rb#L140-L173) function does the SQL query directly in the same function, and without the should_poll guard we'd get way too many spans.

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
indrekj: Correct. Good question. From my perspective, I don't see a need for a poll span without the pg span. The sidekiq instrumentation is following [the same pattern](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/instrumentation/sidekiq/lib/opentelemetry/instrumentation/sidekiq/patches/poller.rb).

instrumentation/que/lib/opentelemetry/instrumentation/que/patches/poller.rb
arielvalentin: Maybe manage the spans manually instead of the `in_span` helper and checking for a `nil` return value from super to see if the instrumentation should record a span?

indrekj: Hmm, interesting idea. I think this could work. The only downside I see is that we'd create a lot of spans that we would discard immediately. About 20 per second. Do you think that would be okay or do you think this is overhead we cannot afford?

arielvalentin: It would not be ideal to
generate unused objects. There also may be cases where pg spans end up missing parents and that would be extremely unpleasant for our users so my suggestion is not helpful.

This makes me want to reach out to the maintainers and get their feedback. I wonder if they would be willing to add or change their internal logger hooks and allow us to leverage that for instrumentation:

https://github.com/que-rb/que/blob/master/lib/que/utils/logging.rb#L35

Is that something you'd be willing to try?

indrekj: I think the existing maintainer is not currently maintaining the project. There's an [Issue open](https://github.com/que-rb/que/issues/313) about adding new maintainers which currently seems to be in progress. I think after that there's a likelihood of getting new improvements/changes into the que repository.

ericmustin (approved):
I guess I also have some minor concerns about the implications discussed here (adding our own `should_poll?` invocation): 
https://github.com/open-telemetry/opentelemetry-ruby/pull/1033/files#diff-13284c8e4aefd61f26792a73f8ff887bf15677a99a7ef580c973bc761aa46a1eR17


But practically speaking, I think if the underlying `poll` method's behavior is exactly the same as our patch (see: https://github.com/que-rb/que/blob/61e1805deb8161e5c1502e518a9a44fa464d308c/lib/que/poller.rb#L145) so i think we actually aren't at risk of changing any behavior, in "this" version at least. Also, an observer effect of generating 20 spans a second that we ultimately discard, is non-trivial and it would be beneficial to avoid if we can, so i'm in favor of this as is.

If we're truly concerned that, perhaps, a future `Que::Poller.poll` method will drift under the hood from our patch and create a delta in the behavior, we could be _extra defensive_ and  enforce `MAX` version in the compatibility block of the `Que` instrumentation patch. But given that this thing isn't even really maintained, I say, let it ride.

fbogsany (approved):

PR #1036, opened 2021-11-29
fix: add max supported version for active record

robertlaurin:
We've observed serious incompatibilities between this instrumentation and AR in the rails 7.X.X alpha releases.  We're going to err to the side of caution and disable the instrumentation patches if rails 7 is detected.

CC: @SomalianIvan

### DIFF
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
@@ -10,6 +10,7 @@ module OpenTelemetry
       # The Instrumentation class contains logic to detect and install the ActiveRecord instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('5.2.0')
+        MAXIMUM_VERSION = Gem::Version.new('7.0.0')
 
         install do |_config|
           require_dependencies
@@ -21,7 +22,7 @@ module OpenTelemetry
         end
 
         compatible do
-          gem_version >= MINIMUM_VERSION
+          gem_version >= MINIMUM_VERSION && gem_version < MAXIMUM_VERSION
         end
 
         private
--- a/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
+++ b/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
@@ -10,7 +10,7 @@ module OpenTelemetry
       # The Instrumentation class contains logic to detect and install the ActiveRecord instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
         MINIMUM_VERSION = Gem::Version.new('5.2.0')
-        MAXIMUM_VERSION = Gem::Version.new('7.0.0')
+        MAX_MAJOR_VERSION = 6
 
         install do |_config|
           require_dependencies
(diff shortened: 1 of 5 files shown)


ericmustin (approved):

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
```diff
@@ -21,7 +22,7 @@ class Instrumentation < OpenTelemetry::Instrumentation::Base
         end
 
         compatible do
-          gem_version >= MINIMUM_VERSION
+          gem_version >= MINIMUM_VERSION && gem_version < MAXIMUM_VERSION
```
arielvalentin, line 25: 💁🏼 In my adventures with adding the default gem compatibility, this functionality is provided to us by  [`Gem::Requirement#satisfied_by?`](https://ruby-doc.org/stdlib-3.0.3/libdoc/rubygems/rdoc/Gem/Requirement.html) and may be a reasonable alternative: 

```ruby
    irb(main):028:0> r = Gem::Requirement.new(">= 5.2.0", "< 7.0.0")
=> #<Gem::Requirement:0x00007ff68714ea30 @requirements=[[">=", #<Gem::Version "5.2.0">], ["<", #<Gem::Version "7.0.0">]]>
irb(main):029:0> r.satisfied_by?(Gem::Version.new("5.1.9"))
=> false
irb(main):030:0> r.satisfied_by?(Gem::Version.new("5.2"))
=> true
irb(main):031:0> r.satisfied_by?(Gem::Version.new("6.9"))
=> true
irb(main):032:0> r.satisfied_by?(Gem::Version.new("7"))
=> false
irb(main):033:0> r.satisfied_by?(Gem::Version.new("7.0"))
=> false
irb(main):034:0> r.satisfied_by?(Gem::Version.new("7.0.0.alpha1"))
=> true
```

arielvalentin (approved):

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb
arielvalentin: Maybe not so funky? I guess it is technically true?
```
irb(main):036:0>  Gem::Version.new("7.0.0.alpha1") < Gem::Version.new("7.0.0")
=> true
```

ericmustin: oh, hm weird, good point @arielvalentin . this doesn't apply to most libs, but for this use case, maybe `::ActiveRecord::VERSION::MAJOR < 7` ([defined here](https://github.com/rails/rails/blob/8a419e8d4d7e15a3b25344d4b727e6410404f093/activerecord/lib/active_record/gem_version.rb#L10)) may work better for us here

robertlaurin: Good callout, this is a case that is very relevant to the value of this PR.  

We can drop the `.alpha/whaterver` from the versioning or check for its presence using release/prerelease?.  But I think I can rework this to be a bit clearer/simpler.

I'll update the PR in a moment.

robertlaurin: Updated.

ericmustin: lgtm

PR #1042, opened 2021-11-30
fix: deprecate api rack env getter

robertlaurin:
Deprecates the api rack env getter, anyone retrieving the getter will receive a warning log message that this code path has been deprecated.

### DIFF
--- a/api/lib/opentelemetry/context/propagation/rack_env_getter.rb
+++ b/api/lib/opentelemetry/context/propagation/rack_env_getter.rb
@@ -19,8 +19,6 @@ module OpenTelemetry
         # Converts key into a rack-normalized key and reads it from the carrier.
         # Useful for extract operations.
         def get(carrier, key)
-          OpenTelemetry.logger.warn('OpenTelemetry::Context::Propagation::RackEnvGetter has been deprecated \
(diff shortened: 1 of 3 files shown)


api/lib/opentelemetry/context/propagation/rack_env_getter.rb
```diff
@@ -7,6 +7,9 @@
 module OpenTelemetry
   class Context
     module Propagation
+      # @deprecated Use the rack env getter found in the
```
fbogsany, line 10: We should add a removal date to this comment. 🤔 

[API support guarantees](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/versioning-and-stability.md#api-support) are more complex than I thought.

> Major versions of the API MUST be supported for a minimum of three years after the release of the next major API version.

Technically, this means we can simply remove the `RackEnvGetter` and bump the major version, but we then need to support both the latest major version and the major version at the time of writing for 3 years from the date of the major version bump.

So maybe the approach for now is:
1. Add the deprecation notice without a removal date.
2. Figure out and document the support process for old major versions (e.g. how to PR against older versions, how CI and releases work).
3. Remove the deprecated class and bump the major version.

fbogsany (approved):

PR #1043, opened 2021-12-01
feat: make ActiveRecord 7 compatible

SomalianIvan:


arielvalentin (approved):

arielvalentin (commented):
Whooops I approved this too soon 😬 did not realize the build was failing.

SomalianIvan:
> Whooops I approved this too soon 😬 did not realize the build was failing.

that's my bad still working on it, should have kept it is draft

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence_insert_class_methods.rb
```diff
@@ -3,6 +3,7 @@
 # Copyright The OpenTelemetry Authors
 #
 # SPDX-License-Identifier: Apache-2.0
+def ruby2_keywords(*) end if RUBY_VERSION < '2.7'
```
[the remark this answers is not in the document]
ericmustin: One last small nit is we might want to protect "future us" and add a max RUBY_VERSION constraint of `>= 3.2` since the 3.2 release will likely coincide with `ruby2_keywords` getting removed, according to [ruby docs](https://www.ruby-lang.org/en/news/2019/12/12/separation-of-positional-and-keyword-arguments-in-ruby-3-0/) (h/t [dd-trace-rb](https://github.com/DataDog/dd-trace-rb/blob/4d77d37f7b94b845791c4695f6b75b9bef5f6256/lib/ddtrace/version.rb#L13-L23))
> If you really worry about the portability, use ruby2_keywords. (Acknowledge that Ruby 2.6 or before themselves have tons of corner cases in keyword arguments. :-) ruby2_keywords might be removed in the future after Ruby 2.6 reaches end-of-life. At that point, we recommend to explicitly delegate keyword arguments (see Ruby 3 code above).

```suggestion
def ruby2_keywords(*) end if RUBY_VERSION < '2.7' || RUBY_VERSION >= '3.2'
```

wdyt?

ericmustin (changes requested):
nice work! generally lgtm, some small feedback but I think we also want to

1. Add an appraisal for  7.0.0.alpha2 (i think? maybe there's a cleaner way to do this so we just pull in _whatever is bleeding edge_)

```ruby
appraise 'activerecord-7.0.0.alpha2' do
  gem 'activerecord', '~> 7.0.0.alpha2'
end
```

2. update the `compatible` block in `ActiveRecord::Instrumentation` so that we can install against rails 7: https://github.com/open-telemetry/opentelemetry-ruby/blob/4922b110c1d34a332bb9449237fcdd3af33cee23/instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/instrumentation.rb#L29

I think beside that this should unblock?

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence_insert_class_methods.rb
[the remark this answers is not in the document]
SomalianIvan: https://github.com/ruby/ruby2_keywords should we use this gem? Its a more sureway to do the right thing here. I was first a little apprehensive to adding a dependency, wdyt?

instrumentation/active_record/test/instrumentation/active_record/patches/querying_test.rb
```diff
@@ -17,7 +17,7 @@
 
   describe 'find_by_sql' do
     it 'traces' do
-      User.first
+      User.find_by_sql('SELECT * FROM users')
```
SomalianIvan, line 20: Rails 7 doesn't call the method we instrument on `first` anymore

SomalianIvan: we can find a better method to instrument once rails 7 gets released.

instrumentation/active_record/lib/opentelemetry/instrumentation/active_record/patches/persistence_insert_class_methods.rb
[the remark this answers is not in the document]
ericmustin: I think we can safely assume `ruby2_keywords` is just part of `ruby` itself as of 2.7, So i think as long as we protect ourselves on `<2.7 || >=3.2` versions, that's maybe a bit easier than adding a dependency? As 3.2 gets closer we can revisit, if they've kicked the can on deprecating it, is my two cents. Also our dependency min version might be differ from what an application has and could cause confusion or issues with installing this gem, so if we can avoid that i think why not?

- https://github.com/ruby/ruby/blob/82ea2870188d66aa75a99f03b4e7fdd1750aa196/vm_method.c#L2742
- https://github.com/ruby/ruby/commit/3b302ea8c95d34d5ef072d7e3b326f28a611e479

instrumentation/active_record/opentelemetry-instrumentation-active_record.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.19.0'
+  spec.add_dependency 'ruby2_keywords'
```
ericmustin, line 30: do we need this still?

SomalianIvan: Yep, the patches are still need to work in ruby 2.5/2.6. We can define our own ruby2_keywords, but there is some logic here to make sure we don't redefine things we don't need etc (https://github.com/ruby/ruby2_keywords/blob/master/lib/ruby2_keywords.rb)

ericmustin: 👍 sounds good

ericmustin (approved):
lgtm (assuming tests are unblocked now)

supaspoida:
Apologies for bumping a merged PR, but when I install the latest version of the active_record instrumentation gem I'm still seeing it fail to instrument my rails 7 app. It looks like maybe something got bungled with the release and this gem was somehow not included? Or is the release cadence distinct from merges? Would love to see a release cut w/this change either way!

plantfansam:
@supaspoida sorry you're having difficulties! Would you mind creating an issue in this repository with steps to reproduce?

supaspoida:
@plantfansam done! https://github.com/open-telemetry/opentelemetry-ruby/issues/1188

plantfansam:
Thanks @supaspoida, will try to take a look soon!

PR #978, opened 2021-10-08
Adds instrumentation for rdkafka

andyfleming:
Closes #919.

Issue #751, opened 2021-05-13
Encoding::UndefinedConversionError in Jaeger's CollectorExporter


chrisholmes:
One of my team's services went live with Otel, but we're seeing frequent but low volume encoding errors in the `Jaeger::CollectorExporter`

Here's an example:
```
Encoding::UndefinedConversionError:  "\xF6" from ASCII-8BIT to UTF-8
/gems/thrift-0.14.1/lib/thrift/bytes.rb:81 in encode
/gems/thrift-0.14.1/lib/thrift/bytes.rb:81 in convert_to_utf8_byte_buffer
/gems/thrift-0.14.1/lib/thrift/protocol/binary_protocol.rb:110 in write_string
/gems/thrift-0.14.1/lib/thrift/serializer/serializer.rb:29 in write
/gems/thrift-0.14.1/lib/thrift/serializer/serializer.rb:29 in serialize
/gems/opentelemetry-exporter-jaeger-0.16.0/lib/opentelemetry/exporter/jaeger/collector_exporter.rb:45 in block in export
/gems/opentelemetry-exporter-jaeger-0.16.0/lib/opentelemetry/exporter/jaeger/collector_exporter.rb:44 in each
/gems/opentelemetry-exporter-jaeger-0.16.0/lib/opentelemetry/exporter/jaeger/collector_exporter.rb:44 in export
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:186 in block in export_batch
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:186 in synchronize
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:186 in export_batch
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:169 in block in work
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:158 in loop
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:158 in work
/gems/opentelemetry-sdk-0.16.0/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb:179 in block in reset_on_fork
```

[...]

PR #1059, opened 2021-12-09
Fix: Don't error on non-string/utf8 kafka message keys

chrisholmes:
The instrumentation for Kafka was assuming that message keys are only
valid utf8 strings, but message keys can be any string of bytes. See:
https://cwiki.apache.org/confluence/display/KAFKA/A+Guide+To+The+Kafka+Protocol#AGuideToTheKafkaProtocol-Responses

This was causing encoding errors in the exporters, and lost spans, as they were trying to encode string into utf8 before transport. These problems were originally described here: https://github.com/open-telemetry/opentelemetry-ruby/issues/751

This fix converts keys that aren't valid utf8 strings into hex strings, which seems to be consistent with how other byte strings are encoded such as with trace ids.

This change may need to be ported onto this PR: https://github.com/open-telemetry/opentelemetry-ruby/pull/978

### DIFF
--- a/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/utils.rb
+++ b/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/utils.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Instrumentation
+    module RubyKafka
+      # Utilities to help with instrumenting kafka
+      module Utils
+        module_function
+
+        def encode_message_key(key)
+          return key if key.encoding == Encoding::UTF_8 && key.valid_encoding?
+
+          key.encode(Encoding::UTF_8)
+        rescue Encoding::UndefinedConversionError
+          key.unpack1('H*')
+        end
+      end
+    end
+  end
+end
--- a/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/utils.rb
+++ b/instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/utils.rb
@@ -11,12 +11,13 @@ module OpenTelemetry
       module Utils
         module_function
 
-        def encode_message_key(key)
-          return key if key.encoding == Encoding::UTF_8 && key.valid_encoding?
+        def extract_message_key(key)
+          # skip encode if already valid utf8
+          return key if key.nil? || (key.encoding == Encoding::UTF_8 && key.valid_encoding?)
 
           key.encode(Encoding::UTF_8)
         rescue Encoding::UndefinedConversionError
-          key.unpack1('H*')
+          nil
         end
       end
     end
--- a/instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
+++ b/instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
@@ -110,7 +110,7 @@ describe OpenTelemetry::Instrumentation::RubyKafka::Patches::Consumer do
 
       # First pair for send and process spans
       first_process_span = process_spans[0]
-      _(first_process_span.attributes['messaging.kafka.message_key']).must_equal('af0fef')
+      _(first_process_span.attributes).wont_include('messaging.kafka.message_key')
 
       second_process_span = process_spans[1]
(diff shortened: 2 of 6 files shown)


instrumentation/ruby_kafka/lib/opentelemetry/instrumentation/ruby_kafka/utils.rb
```diff
@@ -0,0 +1,24 @@
...
+
+module OpenTelemetry
+  module Instrumentation
+    module RubyKafka
+      # Utilities to help with instrumenting kafka
+      module Utils
+        module_function
+
+        def encode_message_key(key)
+          return key if key.encoding == Encoding::UTF_8 && key.valid_encoding?
```
arielvalentin, line 15: This is quite interesting to me. Does Kafka support binary message keys?

It also looks similar to an existing utility method [`OpenTelemetry::Common::Utilities::utf8_encode`](https://github.com/open-telemetry/opentelemetry-ruby/blob/2bf0e6aed060867bc65560a92e2c464d48b26a02/common/lib/opentelemetry/common/utilities.rb#L47) however I see that this implementation interprets the binary as a HEX string. 

I feel like this method should either use or enhance the exiting utilities method.

chrisholmes: Yes, Kafka will accept any collection of bytes as the message key. I have seen:
- UUIDs
- DB row keys
- Integer values

When investigating the error, we found a client producing messages where the message key is a Java Long (derived from DB row key) and this was causing the Otel exporters to errors.

Yes, the method you've shared does look similar. I was unsure about using as it would cause my example to display a message key of `''` as all characters would be replaced. However, I would like to get this fixed so I'm happy to go with whatever the prior art is.

arielvalentin: Thanks for the explanation. 

Is this how other dynamic language instrumentations handle this? 

Would it make sense to always the key as if where binary of the predicate fails in stress of forcing the encoding to raise an exception?

arielvalentin: Oofff my last comment was auto completed into nonsense.

I edited it 🤦

chrisholmes: For this case, I've checked against other Kafka instrumentation implementations and none I've checked so far setting this attribute so unfortunately there's nothing to compare with.

However, I just had another look at the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/messaging.md#apache-kafka) and I've noticed there's a footer for the message key:
> If the key type is not string, it's string representation has to be supplied for the attribute. If the key has no unambiguous, canonical string form, don't include its value.

My reaction to this then is that we should not include keys that can't be encoded rather than try to encode or represent the data in another form. Would you agree then that the correct behaviour should be to not set the attribute if utf8 encoding produces a string with invalid encoding?

instrumentation/ruby_kafka/test/opentelemetry/instrumentation/ruby-kafka/patches/consumer_test.rb
```diff
@@ -93,6 +93,30 @@
...
+          counter += 1
+          break if counter >= 2
+        end
+      end
+
+      process_spans = spans.select { |s| s.name == "#{topic} process" }
+
+      # First pair for send and process spans
+      first_process_span = process_spans[0]
+      _(first_process_span.attributes['messaging.kafka.message_key']).must_equal(nil)
```
arielvalentin, line 113:
```suggestion
      _(first_process_span.attributes.keys).wont_include('messaging.kafka.message_key')
```

arielvalentin (approved):

ericmustin:
Could we rebase? otherwise this lgtm. if tests are still failing after rebasing i will investigate and try to fix those.

ericmustin (approved):

chrisholmes:
@ericmustin I did some investigation and there are some issues with the latest kafka and dalli images, which is stopping the containers from starting so rebasing won’t fix. We might need to pin to a specific version to avoid this.

ericmustin:
@chrisholmes thanks, yea im not sure why it's pulling in dalli 3.1.1, the author dropped a private method we were using in 3.1.1, then re-added as a public method in 3.1.2. I'll look into it and try to get this cleaned up today/tmrw, thanks.

chrisholmes:
@ericmustin looks like the images and gems are fixed now 😅

fbogsany (approved):

PR #1063, opened 2021-12-16
feat: Add Trilogy Auto Instrumentation

arielvalentin:
Provides auto instrumentation for the [trilogy database driver](https://github.com/github/trilogy)

### DIFF
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -1,0 +1,130 @@
+# frozen_string_literal: true
+
+module OpenTelemetry
+  module Instrumentation
+    module Trilogy
+      module Patches
+        # Module to prepend to Trilogy for instrumentation
+        module Client
+          QUERY_NAMES = [
+            'set names',
+            'select',
+            'insert',
+            'update',
+            'delete',
+            'begin',
+            'commit',
+            'rollback',
+            'savepoint',
+            'release savepoint',
+            'explain',
+            'drop database',
+            'drop table',
+            'create database',
+            'create table'
+          ].freeze
+
+          QUERY_NAME_RE = Regexp.new("^(#{QUERY_NAMES.join('|')})", Regexp::IGNORECASE)
+
+          COMPONENTS_REGEX_MAP = {
+            single_quotes: /'(?:[^']|'')*?(?:\\'.*|'(?!'))/,
+            double_quotes: /"(?:[^"]|"")*?(?:\\".*|"(?!"))/,
+            numeric_literals: /-?\b(?:[0-9]+\.)?[0-9]+([eE][+-]?[0-9]+)?\b/,
+            boolean_literals: /\b(?:true|false|null)\b/i,
+            hexadecimal_literals: /0x[0-9a-fA-F]+/,
+            comments: /(?:#|--).*?(?=\r|\n|$)/i,
+            multi_line_comments: %r{\/\*(?:[^\/]|\/[^*])*?(?:\*\/|\/\*.*)}
+          }.freeze
+
+          MYSQL_COMPONENTS = %i[
+            single_quotes
+            double_quotes
+            numeric_literals
+            boolean_literals
+            hexadecimal_literals
+            comments
+            multi_line_comments
+          ].freeze
+
+          FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
+
+          def query(sql)
+            attributes = client_attributes
+
+            case config[:db_statement]
+            when :obfuscate
+              attributes[::OpenTelemetry::SemanticConventions::Trace::DB_STATEMENT] = obfuscate_sql(sql)
+            when :include
+              attributes[::OpenTelemetry::SemanticConventions::Trace::DB_STATEMENT] = sql
+            end
+
+            tracer.in_span(
+              database_span_name(sql),
+              attributes: attributes,
+              kind: :client
+            ) do
+              super(sql)
+            end
+          end
+
+          private
+
+          def obfuscate_sql(sql)
+            if sql.size > 2000
+              'SQL query too large to remove sensitive data ...'
+            else
+              obfuscated = sql.gsub(FULL_SQL_REGEXP, '?')
+              obfuscated = 'Failed to obfuscate SQL query - quote characters remained after obfuscation' if detect_unmatched_pairs(obfuscated)
+              obfuscated
+            end
+          end
+
+          def detect_unmatched_pairs(obfuscated)
+            # We use this to check whether the query contains any quote characters
+            # after obfuscation. If so, that's a good indication that the original
+            # query was malformed, and so our obfuscation can't reliably find
+            # literals. In such a case, we'll replace the entire query with a
+            # placeholder.
+            %r{'|"|\/\*|\*\/}.match(obfuscated)
+          end
+
+          def database_span_name(sql)
+            # Setting span name to the SQL query without obfuscation would
+            # result in PII + cardinality issues.
+            # First attempt to infer the statement type then fallback to
+            # current Otel approach {database.component_name}.{database_instance_name}
+            # https://github.com/open-telemetry/opentelemetry-python/blob/39fa078312e6f41c403aa8cad1868264011f7546/ext/opentelemetry-ext-dbapi/tests/test_dbapi_integration.py#L53
+            # This creates span names like mysql.default, mysql.replica, postgresql.staging etc.
+
+            statement_type = extract_statement_type(sql)
+
+            return statement_type unless statement_type.nil?
+
+            # fallback
+            'mysql'
+          end
+
+          def client_attributes
+            attributes = { ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql' }
+            attributes[::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME] = @connected_host if defined?(@connected_host)
+            attributes
+          end
+
+          def tracer
+            Trilogy::Instrumentation.instance.tracer
+          end
+
+          def config
+            Trilogy::Instrumentation.instance.config
+          end
+
+          def extract_statement_type(sql)
+            QUERY_NAME_RE.match(sql) { |match| match[1].downcase } unless sql.nil?
+          rescue StandardError => e
+            OpenTelemetry.logger.error("Error extracting sql statement type: #{e.message}")
+          end
+        end
+      end
+    end
+  end
+end
--- a/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
+++ b/instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
@@ -48,8 +48,28 @@ module OpenTelemetry
 
           FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
 
+          def initialize(**kwargs)
+            @_otel_net_peer_name = kwargs[:host]
+            super
+          end
+
           def query(sql)
-            attributes = client_attributes
+            tracer.in_span(
+              database_span_name(sql),
+              attributes: client_attributes(sql),
+              kind: :client
+            ) do
+              super(sql)
+            end
+          end
+
+          private
+
+          def client_attributes(sql)
+            attributes = {
+              ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql',
+              ::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => net_peer_name
+            }
 
             case config[:db_statement]
             when :obfuscate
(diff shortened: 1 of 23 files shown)


ericmustin:
just ack'ing this, will try to give this a thorough review otw.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -0,0 +1,130 @@
...
+
+            statement_type = extract_statement_type(sql)
+
+            return statement_type unless statement_type.nil?
+
+            # fallback
+            'mysql'
+          end
+
+          def client_attributes
```
ericmustin, line 107: small nit, would it be possible to support the `peer_service` attribute via a  `config[:peer_service]` option here? Just so there's parity with other db instrumentations? https://github.com/open-telemetry/opentelemetry-ruby/blob/4216a55be5ad224100f6c2802e8ec5fc04f1e316/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb#L129

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -0,0 +1,130 @@
...
+            attributes = { ::OpenTelemetry::SemanticConventions::Trace::DB_SYSTEM => 'mysql' }
+            attributes[::OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME] = @connected_host if defined?(@connected_host)
```
ericmustin, line 109: I peaked into https://github.com/github/trilogy/blob/9380a5844c844dd13ddd6dc761895898b17186a8/contrib/ruby/lib/trilogy.rb a bit and indeed it doesn't look like too much is available via public ruby methods for us to grab extra metadata at present. So it goes (_awesome gem tho, thank you to all the great GH folks for open sourcing it!_ 🎉 ). 

That being said all that's required according to the [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/3e380e249f60c3a5f68746f5e84d10195ba41a79/specification/trace/semantic_conventions/database.md#connection-level-attributes) (which itself is experimental so it leaves us wiggle room) is `db.system` and _either_ `net.peer.name` || `net.peer.ip`. So I think we're covered here. My only incredibly small bikeshedding thing is maybe we should set this to something like `unknown` in the event that `@connected_host` isn't defined, since downstream processors may assume that either `net.peer.name` or `net.peer.ip` will _always_ exist? Idk, not blocking at all, and given that your org is literally the only user of this instrumentation at the moment, I don't think it matters too much and defer to you on this.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
```diff
@@ -0,0 +1,51 @@
...
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = Gem::Requirement.new('>= 2.5.0')
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.0'
+  spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.19.0'
```
ericmustin, line 29: I believe we need to require the semantic conventions gem as a dependency here actually. It's required by the `-sdk` gem but not by the `-api` or `-base`. I think the appropriate way to version it is loosely, ala what the the `-sdk` has done https://github.com/open-telemetry/opentelemetry-ruby/blob/e639d20fbcc355f1df481b9867b922a83c481050/sdk/opentelemetry-sdk.gemspec#L30

Upon inspection of other instrumentations, I think we also need to fix the recent Aws-Sdk instrumentation gem, but I can do that in a seperate PR 😬 😅

ericmustin (changes requested):
Good stuff! A few very small nits, only thing that's blocking is I think we need to add `semantic-conventions` as a dependancy.

I should mention that I do have some hesitation about `:obfuscate` being the default setting here, since it creates a delta between the `mysql2` and the `trilogy` instrumentation, and could add a performance overhead _by default_ that may be unexpected. But as you're the only real user of this at the moment I'd prefer to do what's best for our actual user's use case, which apparently is to `:obfuscate` by default.

Anyway, again awesome stuff ty for this, and your many other, contributions.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
```diff
@@ -0,0 +1,130 @@
...
+            single_quotes
+            double_quotes
+            numeric_literals
+            boolean_literals
+            hexadecimal_literals
+            comments
+            multi_line_comments
+          ].freeze
+
+          FULL_SQL_REGEXP = Regexp.union(MYSQL_COMPONENTS.map { |component| COMPONENTS_REGEX_MAP[component] })
```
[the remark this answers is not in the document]
arielvalentin: I created an issue based on your comment.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: Sure thing I will add it. 

I left this off because I do not use this attribute in any of our instrumentations. Setting it as an instrumentation options makes it difficult to use on a per connection instance since unrelated instances end up getting the same attribute and there isn't an easy way to override it AFAICT.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: > db.system and either net.peer.name || net.peer.ip

TIL! Let me see what I can do. Perhaps I can use the original host provided to the connection and switch to the connected host when it is set.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
arielvalentin: Makes sense. I will fix this. 

Dang, I wish there was a way I could have caught that earlier!!!

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
ericmustin: hmm, let's skip it then. I think we bumped into similar attribution issues for `peer.service` with kafka related instrumentation, and we chose to punt on those.

instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec
arielvalentin: hmmm... looks to me like _none_ of the instrumentations explicitly add the conventions as a dependency and rely on the local path from the Gemfile. 🤔 

I guess we do not have many uses cases were folks are using the instrumentations _without_ the SDK. 

```console
omoplata ~/github/opentelemetry-ruby(arielvalentin/trilogy-instrumentation|✔) %
🤘 rg "add_dependency.*opentelemetry-semantic_conventions" -g '*.gemspec'
sdk/opentelemetry-sdk.gemspec
36:  spec.add_dependency 'opentelemetry-semantic_conventions'
```

```console
omoplata ~/github/opentelemetry-ruby(arielvalentin/trilogy-instrumentation|✔) %
🤘 rg "opentelemetry-semantic_conventions" -g 'Gemfile'

resource_detectors/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../semantic_conventions'
exporter/zipkin/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
exporter/jaeger/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
exporter/otlp/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
sdk/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../semantic_conventions'
instrumentation/concurrent_ruby/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/resque/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/ethon/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/pg/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/delayed_job/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/base/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/excon/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/rails/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/faraday/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/rspec/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/sinatra/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/redis/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/active_job/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/koala/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/restclient/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/mongo/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/action_pack/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/rack/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/net_http/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/trilogy/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/graphql/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/http/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/active_support/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/lmdb/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/all/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/active_record/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/aws_sdk/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/bunny/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/http_client/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/active_model_serializers/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/que/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/ruby_kafka/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/mysql2/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/action_view/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/sidekiq/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
instrumentation/dalli/Gemfile:  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
propagator/ottrace/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
propagator/jaeger/Gemfile:gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
```

ericmustin: I think basically The resource detector and aws-sdk instrumentation, afaik, are the only gems that directly reference the SemanticConventions module . The resource detector depends on the sdk in the gemspec tho, which in turn depends on the semconv get, while instrumentations (or -base) do not. I think the search above is only pulling in the :test dependencies from the Gemfile.

Any way tldr looks good to me now

ericmustin (approved):
I think everything is covered, nice work!

arielvalentin:
Thanks @ericmustin although I think that I'm the case of the UDS the host name will be nil and I don't want to trigger and error so I'm going to set a default there. 

Although not as common, I think that we need some additional coverage for those cases for all of the drivers.

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy/patches/client.rb
arielvalentin: it's all good. I added it anyway.

ericmustin (approved):
Additional updates lgtm

fbogsany (approved):

PR #1064, opened 2021-12-18
fix: Allow OTLP Exporter compression value of `none`

arielvalentin:
> [2]: If no compression value is explicitly specified, SIGs can default to the value they deem most useful among the supported options.
>      This is especially important in the presence of technical constraints,
>      e.g. directly sending telemetry data from mobile devices to backend servers.
>
> Supported values for OTEL_EXPORTER_OTLP_*COMPRESSION options:
>
> * none if compression is disabled.
> * gzip is the only specified compression method for now.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options


ericmustin (approved):
good catch!

fbogsany (approved):

PR #1069, opened 2022-01-04
Add debugging info during Zipkin export

andrewhood125:
Recently ran into an issue trying to integrate Zipkin and used this to help identify the issue. 

After 
Screen Shot 2022-01-03 at 4 14 37 PM

### DIFF
--- a/exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb
+++ b/exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb
@@ -156,29 +156,31 @@ module OpenTelemetry
 
             case response
             when Net::HTTPAccepted, Net::HTTPOK
-              response.body # Read and discard body
               SUCCESS
             when Net::HTTPServiceUnavailable, Net::HTTPTooManyRequests
-              response.body # Read and discard body
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               redo if backoff?(retry_after: response['Retry-After'], retry_count: retry_count += 1, reason: response.code)
               FAILURE
             when Net::HTTPRequestTimeOut, Net::HTTPGatewayTimeOut, Net::HTTPBadGateway
-              response.body # Read and discard body
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               redo if backoff?(retry_count: retry_count += 1, reason: response.code)
               FAILURE
             when Net::HTTPBadRequest, Net::HTTPClientError, Net::HTTPServerError
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               # TODO: decode the body as a google.rpc.Status Protobuf-encoded message when https://github.com/open-telemetry/opentelemetry-collector/issues/1357 is fixed.
-              response.body # Read and discard body
               FAILURE
             when Net::HTTPRedirection
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               @http.finish
               handle_redirect(response['location'])
               redo if backoff?(retry_after: 0, retry_count: retry_count += 1, reason: response.code)
             else
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               @http.finish
               FAILURE
             end
           rescue Net::OpenTimeout, Net::ReadTimeout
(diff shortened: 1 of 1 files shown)


ericmustin (commented):
Sorry for delay in reviewing 😬 . Anyway, I think broadly speaking there's a lot of room for improvement in how failed exporter response information is recorded and surfaced, and how easy that information is for a user to access. This PR improves what we currently have and so i'd like to see it merged. It does differ from the OTLP exporter, which uses the concept of an optional metric reporter for this https://github.com/open-telemetry/opentelemetry-ruby/blob/e58fe948a2552329f3cd64c5cf6256fdddbf427a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L53 which, i think has a lot of benefits but also is a bit cumbersome to setup. But, I don't think we should block adding debug logs to the zipkin exporter just because the otlp exporter does things differently.

However, would it be possible to only add this debug log for non 200 responses? ie before `redo` in `when` non-`SUCCESS` blocks? I worry that users in local dev will have on debug logging and find this library unnecessarily noisy when everything is working as expected (ie, `SUCCESS` responses so there's nothing to debug)

exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb
```diff
@@ -148,27 +148,23 @@ def send_spans(zipkin_spans, timeout: nil) # rubocop:disable Metrics/AbcSize, Me
             @http.start unless @http.started?
 
             response = @http.request(request)
-            response.body # Read and discard body
+            OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path} #{response.body}")
```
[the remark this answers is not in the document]
andrewhood125: I went ahead and dropped the response body.

andrewhood125:
Thanks for the context, I haven't looked at the OTLP exporter but I will. I think logging for non success is the right way to go I've updated the PR.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb
```diff
@@ -148,37 +148,38 @@ def send_spans(zipkin_spans, timeout: nil) # rubocop:disable Metrics/AbcSize, Me
...
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               redo if backoff?(retry_count: retry_count += 1, reason: response.code)
               FAILURE
             when Net::HTTPBadRequest, Net::HTTPClientError, Net::HTTPServerError
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
               # TODO: decode the body as a google.rpc.Status Protobuf-encoded message when https://github.com/open-telemetry/opentelemetry-collector/issues/1357 is fixed.
-              response.body # Read and discard body
               FAILURE
             when Net::HTTPRedirection
+              OpenTelemetry.logger.debug("Zipkin Exporter #{response.code} #{@path}")
```
andrewhood125, line 172: 👍 / 👎  on this block?

ericmustin: hmmm, sure i think so? 👍

exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb
```diff
@@ -148,37 +148,38 @@ def send_spans(zipkin_spans, timeout: nil) # rubocop:disable Metrics/AbcSize, Me
             @http.start unless @http.started?
 
             response = @http.request(request)
-            response.body # Read and discard body
```
ericmustin, line 151: since we've moved the debug log into the non `SUCCESS` blocks and removed the `body` read from those calls, we should still keep this iiuc? other than that this lgtm

ericmustin (approved):
one small nit (i think we still need to keep the `response.body` read/discard in the main hot path since we no longer read it in the debug log) but other than that i think this lgtm, approving (assuming we making that change)

andrewhood125:
Feedback addressed, thanks for the review!

ericmustin:
@andrewhood125 (rubocop complaining about rubocop things)

andrewhood125:
@ericmustin I believe this is ready to go now!

ericmustin:
@andrewhood125 yup, lgtm and i've approved, cc @open-telemetry/ruby-maintainers

PR #1070, opened 2022-01-04
chore: add max faraday version to tests 

ericmustin:
### Summary

Temp workaround for https://github.com/arsduo/koala/issues/659 to unblock CI until upstream Koala itself supports Faraday 2.0

### DIFF
--- a/instrumentation/koala/Gemfile
+++ b/instrumentation/koala/Gemfile
@@ -13,6 +13,7 @@ gem 'opentelemetry-instrumentation-base', path: '../base'
 
 group :test do
   gem 'byebug'
+  gem 'faraday', '<2.0'
   gem 'opentelemetry-common', path: '../../common'
   gem 'opentelemetry-sdk', path: '../../sdk'
   gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'


ericmustin:
cc @SomalianIvan this should unblock https://github.com/open-telemetry/opentelemetry-ruby/pull/1043

SomalianIvan (approved):

SomalianIvan:
Sending my grey approval

mwear (approved):
LGTM

instrumentation/koala/Gemfile
```diff
@@ -13,6 +13,7 @@ gem 'opentelemetry-instrumentation-base', path: '../base'
 
 group :test do
   gem 'byebug'
+  gem 'faraday', '<2.0'
```
arielvalentin, line 16: Would this make more sense to be included in the appraisals file instead? 

That would be true for cases where older versions of faraday would be compatible only with older versions of the koala gem no?

ericmustin: hm, yea that probably makes more sense. practically speaking i think that the outcome of this issue comment https://github.com/arsduo/koala/issues/659#issuecomment-1005192539 will probably resolve this for us in general, but can move into appraisal for now.

PR #1072, opened 2022-01-05
fix: `ActiveSupport` constant conflict in Active Model Serializers instrumentation

flexoid:
When both `ActiveSupport` and `ActiveModelSerializers` instrumentations are loaded,
constant `ActiveSupport::Notifications` inside `ActiveModelSerializers` can be incorrectly
resolved to `OpenTelemetry::Instrumentation::ActiveSupport`.

```
ERROR -- : OpenTelemetry error: Instrumentation: OpenTelemetry::Instrumentation::ActiveModelSerializers unhandled exception during install
...
...
uninitialized constant OpenTelemetry::Instrumentation::ActiveSupport::Notifications
OR
undefined method `subscribe' for Notifications:Module
```

### DIFF
--- a/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
+++ b/instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
@@ -35,7 +35,7 @@ module OpenTelemetry
         end
 
         def register_event_handler
-          ActiveSupport::Notifications.subscribe(event_name) do |_name, start, finish, _id, payload|
+          ::ActiveSupport::Notifications.subscribe(event_name) do |_name, start, finish, _id, payload|
             EventHandler.handle(start, finish, payload)
           end
         end
--- a/instrumentation/all/Gemfile
+++ b/instrumentation/all/Gemfile
@@ -46,6 +46,8 @@ gem 'opentelemetry-instrumentation-sinatra', path: '../sinatra'
 gem 'opentelemetry-instrumentation-trilogy', path: '../trilogy'
 
 group :test do
+  gem 'active_model_serializers'
+  gem 'activesupport'
   gem 'opentelemetry-sdk', path: '../../sdk'
   gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
 end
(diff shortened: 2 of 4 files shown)


instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
```diff
@@ -35,7 +35,7 @@ def require_dependencies
         end
 
         def register_event_handler
-          ActiveSupport::Notifications.subscribe(event_name) do |_name, start, finish, _id, payload|
+          ::ActiveSupport::Notifications.subscribe(event_name) do |_name, start, finish, _id, payload|
```
arielvalentin, line 38: Is there a way to write encode this in an automated test case that will reveal this error and avoid regressions?

flexoid: Well, I thought about it. But it is reproduced only when both `opentelemetry-instrumentation-active_support` and `opentelemetry-instrumentation-active_model_serializers` are included to the app, and these gems have isolated test specs. So it's a bit hard to test the conflict.

arielvalentin: I wonder if this is something we may be able to detect in the `all` test suite? cc: @open-telemetry/ruby-approvers

flexoid: Done! Reproduced with a quite simple test case. Thanks for the suggestion 🙂

arielvalentin: Sweet! Did the tests fail when you removed the top level qualifier `::`?

fbogsany (approved):

instrumentation/active_model_serializers/lib/opentelemetry/instrumentation/active_model_serializers/instrumentation.rb
flexoid: Yep, that's what I started from.

flexoid:
```
Finished in 0.006173s, 161.9958 runs/s, 0.0000 assertions/s.

  1) Error:
OpenTelemetry::Instrumentation::All::ActiveModelSerializers instrumentation#test_0001_does not conflict with ActiveSupport instrumentation:
NameError: uninitialized constant OpenTelemetry::Instrumentation::ActiveSupport::Notifications
```

PR #1076, opened 2022-01-06
chore: update server instrumentation to not reflect 400 status as error

ericmustin:
### Summary

The specification as of late October/early November 2021 has been updated for http server spans so that 400 level responses should not create a span.status of `Error` but instead should leave span status unset. Two instrumentations, rack and sinatra, were off spec and need to be updated, this PR updates them.

see:

- https://github.com/open-telemetry/opentelemetry-specification/pull/1998
- https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#status


fbogsany (approved):

miry:
@ericmustin thank you very much for this change.

PR #1079, opened 2022-01-07
fix: Default scheme for OTLP endpoint

### LINKED ISSUE #1078: Incorrect default url for OTLP HTTP
While doing our due diligence before the 1.0 OTLP Trace Exporter release for JS we noticed that not all exporter implementations follow spec.

[Spec](https://github.com/open-telemetry/opentelemetry-specification/blob/8e29f555b2282a3883b1b57009ad4eb11bfc5228/specification/protocol/exporter.md#configuration-options) states OTLP HTTP should use `http://localhost:4318` by default.

Ruby is using `https://localhost:4318/v1/traces` (https should be http): [link](https://github.com/open-telemetry/opentelemetry-ruby/blob/46020cc0ab9826b362854a6a913f30996ac528a2/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L47)

fbogsany:
https://github.com/open-telemetry/opentelemetry-specification/pull/2014 changed the default scheme for the OTLP endpoint from `https` to `http`. This updates our default.

Fixes #1078

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -44,7 +44,7 @@ module OpenTelemetry
           end
         end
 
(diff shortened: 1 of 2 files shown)


arielvalentin (approved):

ericmustin:
I think we have to also update some of the stub'd urls `https => http` in the exporter test suite and then should be good to go

ericmustin (approved):
:shipit:

PR #1080, opened 2022-01-07
fix: Trilogy Driver Options

arielvalentin:
The trilogy driver does not use keyword arguments in its initializer, which results in an ArgumentError to be raised when an options hash is used instead.

This change rolls back the overridden initializer to accept a hash.


ericmustin (approved):

arielvalentin:
@ericmustin Looks like I should have been a little more careful when I introduced this change: 

https://github.com/open-telemetry/opentelemetry-ruby/pull/1063/commits/027cba10fe485308a1efe8765869e4c017e22bca#diff-6ba5c9beb3eb94883042f29a069003cab7f6296395e612e75b0abe00c98abbb3R51

PR #1083, opened 2022-01-11
refactor: [instrumentation/net/http] Replace strings with constants

miry:
There are semantic convetnion constats exists.
Replaced attributes for `Instrumentation::Net::HTTP` to use them.

### DIFF
--- a/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
+++ b/instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
@@ -11,6 +11,8 @@ module OpenTelemetry
         module Patches
           # Module to prepend to Net::HTTP for instrumentation
           module Instrumentation
+            include OpenTelemetry::SemanticConventions::Trace
+
             HTTP_METHODS_TO_SPAN_NAMES = Hash.new { |h, k| h[k] = "HTTP #{k}" }
             USE_SSL_TO_SCHEME = { false => 'http', true => 'https' }.freeze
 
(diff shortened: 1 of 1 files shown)


instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -11,6 +11,8 @@ module HTTP
         module Patches
           # Module to prepend to Net::HTTP for instrumentation
           module Instrumentation
+            include OpenTelemetry::SemanticConventions::Trace
```
arielvalentin, line 14: One think I worry about with mixing in the constants is the potential for conflicting with ones that _may_ exist in libraries. 

Perhaps mixing in only `OpenTelemetry::SemanticConventions` so that the constants are scoped to `Trace` or `Resource may be good enough?

Not sure how the @open-telemetry/ruby-approvers and @open-telemetry/ruby-maintainers feel about that.

miry: make sense. I can replace with a full path to constants.

Example:
`HTTP_STATUS` -> `OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS`

miry: fixed in https://github.com/open-telemetry/opentelemetry-ruby/pull/1083/commits/c697489673baa5982e2232905eabc5fb54a64674

arielvalentin (approved):

ericmustin (approved):
lgtm, Thanks for the PR!

fbogsany (approved):
This is a good incremental improvement, thanks. Longer term, it would be useful to define instrumentation helper methods with required and optional args that match the required and optional attributes for a given type of instrumentation (HTTP client spans, in this case). This would both help enforce semantic conventions and clean up these attribute hashes.

PR #1084, opened 2022-01-11
chore: add semconv gem to relevant gemspecs and require explicitly

ericmustin:
Followup for https://github.com/open-telemetry/opentelemetry-ruby/pull/1083

We should include a loose semconv gem dependency and explicitly require in the module during instrumentation loading, when it's being used in instrumentation.

Since the `-sdk` takes a loose semconv dependency this has not caused errors when the `SemanticConventions` constant gets used by instrumentations, but technically these instrumentations can be used outside the SDK and must not hold any hard SDK Dependency, so it's necessary that we add the semconv gem to their gemspec

Note: Trilogy already has a semconv gem dependancy in it's gemspec

(alternatively, should we just include this in `-base`?)

### DIFF
--- a/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk.rb
+++ b/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk.rb
@@ -6,6 +6,7 @@
 
 require 'opentelemetry'
 require 'opentelemetry-instrumentation-base'
+require 'opentelemetry/semantic_conventions'
 
 module OpenTelemetry
   module Instrumentation
--- a/instrumentation/aws_sdk/opentelemetry-instrumentation-aws_sdk.gemspec
+++ b/instrumentation/aws_sdk/opentelemetry-instrumentation-aws_sdk.gemspec
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.19.0'
+  spec.add_dependency 'opentelemetry-semantic_conventions'
 
   spec.add_development_dependency 'appraisal', '~> 2.2.0'
   spec.add_development_dependency 'aws-sdk', '>= 2.0'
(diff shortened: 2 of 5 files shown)


instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk.rb
```diff
@@ -6,6 +6,7 @@
 
 require 'opentelemetry'
 require 'opentelemetry-instrumentation-base'
+require 'opentelemetry/semantic_conventions'
```
arielvalentin, line 9:
```suggestion
require 'opentelemetry-semantic_conventions'
```

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http.rb
```diff
@@ -7,6 +7,7 @@
 require 'opentelemetry'
 require 'opentelemetry/common'
 require 'opentelemetry-instrumentation-base'
+require 'opentelemetry/semantic_conventions'
```
arielvalentin, line 10:
```suggestion
require 'opentelemetry-semantic_conventions'
```

instrumentation/trilogy/lib/opentelemetry/instrumentation/trilogy.rb
```diff
@@ -6,6 +6,7 @@
 
 require 'opentelemetry'
 require 'opentelemetry-instrumentation-base'
+require 'opentelemetry/semantic_conventions'
```
arielvalentin, line 9:
```suggestion
require 'opentelemetry-semantic_conventions'
```

instrumentation/aws_sdk/opentelemetry-instrumentation-aws_sdk.gemspec
```diff
@@ -27,6 +27,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
   spec.add_dependency 'opentelemetry-instrumentation-base', '~> 0.19.0'
+  spec.add_dependency 'opentelemetry-semantic_conventions'
```
blumamir, line 30: should there be a version range for that dependency?
[like here](https://github.com/open-telemetry/opentelemetry-ruby/blob/dab6dda9fb1a7fdc2df571f0ad4e2a35cb39b4e0/instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec#L30) for example

ericmustin: If there is a min version of semconv that works for awssdk, then yes we should have a min version, but i wasn't aware of what that was.

plantfansam:

To do that, you can:

PR #1086, opened 2022-01-15
Added ruby 3.1 in ci list

tinchi:
* Add new ruby 3.1 into ci list

### DIFF
--- a/sdk/test/opentelemetry/sdk/trace/span_test.rb
+++ b/sdk/test/opentelemetry/sdk/trace/span_test.rb
@@ -202,7 +202,7 @@ describe OpenTelemetry::SDK::Trace::Span do
     end
 
     it 'honours an explicit timestamp' do
-      ts = Time.new('2021-11-23 12:00:00.000000 -0600')
+      ts = Time.now
       span.add_event('added', timestamp: ts)
       events = span.events
       _(events.size).must_equal(1)
@@ -391,7 +391,7 @@ describe OpenTelemetry::SDK::Trace::Span do
     end
 
     it 'honours an explicit end timestamp' do
-      ts = Time.new('2021-11-23 12:00:00.000000 -0600')
+      ts = Time.now
       _(span.finish(end_timestamp: ts)).must_equal(span)
       _(span.end_timestamp).must_equal(exportable_timestamp(ts))
     end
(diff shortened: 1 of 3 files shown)


sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -202,7 +202,7 @@
     end
 
     it 'honours an explicit timestamp' do
-      ts = Time.new('2021-11-23 12:00:00.000000 -0600')
+      ts = Time.now
```
tinchi, line 205: Had to change it due to, so changes to `Time` class introduced in 3.1
```
OpenTelemetry::SDK::Trace::Span::#finish#test_0002_honours an explicit end timestamp:
2022-01-17T14:36:54.7595272Z ArgumentError: invalid value for Integer(): "2021-11-23 12:00:00.000000 -0600"
2022-01-17T14:36:54.7595701Z     <internal:timev>:310:in `initialize'
2022-01-17T14:36:54.7596384Z     /home/runner/work/opentelemetry-ruby/opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/span_test.rb:394:in `new'
2022-01-17T14:36:54.7597270Z     /home/runner/work/opentelemetry-ruby/opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/span_test.rb:394:in `block (3 levels) in <top (required)>'```

**Comment by SomalianIvan** [comment]
```
  1) Failure:
OpenTelemetry::Instrumentation::ActiveRecord::Patches::PersistenceClassMethods::.create!#test_0002_adds an exception event if it raises [/home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/active_record/test/instrumentation/active_record/patches/persistence_class_methods_test.rb:40]:
--- expected
+++ actual
@@ -1 +1,4 @@
-"unknown attribute 'attreeboot' for User."
+"unknown attribute 'attreeboot' for User.
+
+          raise UnknownAttributeError.new(self, k.to_s)
+          ^^^^^"


47 runs, 61 assertions, 1 failures, 0 errors, 0 skips
```

```
+"unknown attribute 'attreeboot' for User.
+
+          raise UnknownAttributeError.new(self, k.to_s)
+          ^^^^^"
```

sdk/test/opentelemetry/sdk/trace/span_test.rb
arielvalentin: we use static values for time in tests to avoid erratic test results. Please update these to use `Time.parse` instead.

tinchi:
@SomalianIvan I switched `must_equal` to `must_include` as a work around.

sdk/test/opentelemetry/sdk/trace/span_test.rb
robertlaurin: @arielvalentin I double checked some of these tests, and I think I (for reasons now unknown to me) parsed explicit timestamps but it didn't actually matter for the tests.  

The check is just looking at whatever we captured from parse (or in this revision time.now) 
`_(events.first.timestamp).must_equal(exportable_timestamp(ts))`

I think this should be ok.

robertlaurin (approved):

PR #1087, opened 2022-01-18
Remove TIMEOUT timeout status from OTLP exporter

robertlaurin:
In the effort of pushing the OTLP exporter to a 1.0, I'm going through the spec and trying to align the subtle inconsistencies. 

The TIMEOUT status is not explicitly stated, which leads me to believe it should probably be removed.
```
Returns: ExportResult:

ExportResult is one of:

* Success - The batch has been successfully exported. For protocol exporters this typically means that the data is sent over the wire and delivered to the destination server.
* Failure - exporting failed. The batch must be dropped. For example, this can happen when the batch contains bad data and cannot be serialized.
```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#interface-definition-1

However force flush does make reference to notify the caller that the call timed out.

```
ForceFlush SHOULD provide a way to let the caller know whether it succeeded, failed or timed out.
```
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#forceflush-2

In either case, force flush is never going to return a timeout status code so I believe we can safely remove this.
https://github.com/open-telemetry/opentelemetry-ruby/blob/fd24a7b9b096c7a30021bfda1f3b7b1090d8cfce/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L90-L97

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -22,7 +22,6 @@ module OpenTelemetry
       class Exporter # rubocop:disable Metrics/ClassLength
         SUCCESS = OpenTelemetry::SDK::Trace::Export::SUCCESS
         FAILURE = OpenTelemetry::SDK::Trace::Export::FAILURE
-        TIMEOUT = OpenTelemetry::SDK::Trace::Export::TIMEOUT
         private_constant(:SUCCESS, :FAILURE, :TIMEOUT)
 
         # Default timeouts in seconds.
@@ -165,7 +164,7 @@ module OpenTelemetry
             @headers.each { |key, value| request.add_field(key, value) }
 
             remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
-            return TIMEOUT if remaining_timeout.zero?
+            return FAILURE if remaining_timeout.zero?
 
             @http.open_timeout = remaining_timeout
             @http.read_timeout = remaining_timeout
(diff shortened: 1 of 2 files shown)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -22,7 +22,6 @@ module OTLP
       class Exporter # rubocop:disable Metrics/ClassLength
         SUCCESS = OpenTelemetry::SDK::Trace::Export::SUCCESS
         FAILURE = OpenTelemetry::SDK::Trace::Export::FAILURE
-        TIMEOUT = OpenTelemetry::SDK::Trace::Export::TIMEOUT
         private_constant(:SUCCESS, :FAILURE, :TIMEOUT)
```
arielvalentin, line 25:
```suggestion
        private_constant(:SUCCESS, :FAILURE)
```

arielvalentin (approved):

fbogsany:
```
  1) Error:
OpenTelemetry::Exporter::OTLP::Exporter::#export#test_0006_returns TIMEOUT on timeout after retrying:
NameError: uninitialized constant TIMEOUT
Did you mean?  Timeout
    /Users/runner/work/opentelemetry-ruby/opentelemetry-ruby/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb:351:in `block (4 levels) in <top (required)>'
```

robertlaurin:
OpenTelemetry JS does not have the concept of a Timeout status code from what I could see. https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/opentelemetry-core/src/ExportResult.ts#L22-L25

Was not able to find one in Java https://github.com/open-telemetry/opentelemetry-java/blob/6ab88fbc00bf3bf40b033d523aa9d430d08c0cfb/sdk/common/src/main/java/io/opentelemetry/sdk/common/CompletableResultCode.java#L25-L34

Going to go ahead and merge this

PR #1095, opened 2022-01-21
Add HTTP Status Code attribute to AWS SDK span

### LINKED ISSUE #1094: [AWS SDK Instrumentation] Trace does not report HTTP status in attributes
## Description

I was using the `opentelemetry-ruby` SDK to instrument my call to S3 using the `opentelemetry-instrumentation-aws_sdk` gem. I noticed that the populated attributes are as follows:

```ruby
#<struct OpenTelemetry::SDK::Trace::SpanData name="S3.ListBuckets",
kind=:client,
status=
#<OpenTelemetry::Trace::Status:0x00007f9b14253508 @code=1, @description="">,
parent_span_id="\xA1\xA7\x11\xB0\xAEGp\xC5",
total_recorded_attributes=4,
total_recorded_events=0,
total_recorded_links=0,
start_timestamp=1642709302898094000,
end_timestamp=1642709303152641000,
attributes=
 {"aws.region"=>"us-west-2",
  "rpc.system"=>"aws-api",
  "rpc.method"=>"ListBuckets",
  "rpc.service"=>"S3"}, # <- no http.status
```

I noticed that the `attributes` did not have the `http.status` attribute set. In the `opentelemetry-instrumentation-aws_lambda` gem we should be able to popoulate this attribute using `meatadata` from the AWS SDK response here:

https://github.com/open-telemetry/opentelemetry-ruby/blob/f0593e7d97c62b10862c382d14f533dbc49be580/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb#L40-L42

We do [something similar in the OTel Python Instrumentation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/dd72d94615a66105a5275efac08f1fd5c89c54b2/instrumentation/opentelemetry-instrumentation-botocore/src/opentelemetry/instrumentation/botocore/__init__.py#L263-L265) of the AWS SDK:

```pythonresult = None
try:
    result = original_func(*args, **kwargs)
    _apply_response_attributes(span, result)

...

metadata = result.get("ResponseMetadata")

...

status_code = metadata.get("HTTPStatusCode")

if status_code is not None:
    span.set_attribute(SpanAttributes.HTTP_STATUS_CODE, status_code)
```

If this makes sense I can add a PR to populate this attribute 🙂 There might some small differences in AWS SDK Ruby, but should still be accessible.

Thanks.

NathanielRN:
## Description

I looked at the documentation for the AWS Ruby SDK and found that [every `response` has a `successful?` method](https://docs.aws.amazon.com/sdk-for-ruby/v3/api/Seahorse/Client/Response.html#successful%3F-instance_method). It returns a boolean.

However, better than that, [the source code for `successful?`](https://github.com/aws/aws-sdk-ruby/blob/83ac0d4903a4c3f530b5f37c3e34e61cf54a1320/gems/aws-sdk-core/lib/seahorse/client/response.rb#L57-L61) reveals something useful:

```ruby
def successful?
    (200..299).cover?(@context.http_response.status_code) && @error.nil?
end
```

`@context.http_response.status_code` is appropriately populated when the `response` completes!

So I used that to set the `http.status_code` attribute on the span.

I also included the `httpclient` gem so that in the tests we can compare against constants.

Fixes #1094

### DIFF
--- a/instrumentation/aws_sdk/Gemfile
+++ b/instrumentation/aws_sdk/Gemfile
@@ -12,7 +12,6 @@ gem 'opentelemetry-api', path: '../../api'
 gem 'opentelemetry-instrumentation-base', path: '../base'
 
 group :test do
-  gem 'httpclient'
   gem 'opentelemetry-common', path: '../../common'
   gem 'opentelemetry-sdk', path: '../../sdk'
   gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
--- a/instrumentation/aws_sdk/Gemfile
+++ b/instrumentation/aws_sdk/Gemfile
@@ -12,6 +12,7 @@ gem 'opentelemetry-api', path: '../../api'
 gem 'opentelemetry-instrumentation-base', path: '../base'
 
 group :test do
+  gem 'httpclient'
   gem 'opentelemetry-common', path: '../../common'
   gem 'opentelemetry-sdk', path: '../../sdk'
   gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
--- a/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
+++ b/instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
@@ -37,6 +37,9 @@ module OpenTelemetry
             else
               super
             end.tap do |response|
+              span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE,
+                                 context.http_response.status_code)
+
               if (err = response.error)
                 span.record_exception(err)
                 span.status = Trace::Status.error(err.to_s)
--- a/instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
+++ b/instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
@@ -4,7 +4,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'httpclient'
 require 'test_helper'
 
 describe OpenTelemetry::Instrumentation::AwsSdk do
@@ -57,7 +56,7 @@ describe OpenTelemetry::Instrumentation::AwsSdk do
         _(last_span.attributes['aws.region']).must_include 'stubbed'
         _(last_span.attributes['db.system']).must_be_nil
 
-        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::OK
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
 
         _(last_span.status.code).must_equal OpenTelemetry::Trace::Status::UNSET
       end
@@ -75,7 +74,7 @@ describe OpenTelemetry::Instrumentation::AwsSdk do
         _(last_span.attributes['aws.region']).must_include 'stubbed'
         _(last_span.attributes['db.system']).must_be_nil
 
-        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::OK
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
 
         _(last_span.status.code).must_equal OpenTelemetry::Trace::Status::UNSET
       end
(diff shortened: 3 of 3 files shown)


instrumentation/aws_sdk/Gemfile
```diff
@@ -12,6 +12,7 @@ gem 'opentelemetry-api', path: '../../api'
 gem 'opentelemetry-instrumentation-base', path: '../base'
 
 group :test do
+  gem 'httpclient'
```
arielvalentin, line 15:
```suggestion
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'httpclient'
```
arielvalentin, line 7:
```suggestion
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -55,6 +56,9 @@
         _(last_span.attributes['rpc.method']).must_equal 'Publish'
         _(last_span.attributes['aws.region']).must_include 'stubbed'
         _(last_span.attributes['db.system']).must_be_nil
+
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::OK
```
arielvalentin, line 60:
```suggestion
        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -70,6 +74,9 @@
         _(last_span.attributes['rpc.method']).must_equal 'ListBuckets'
         _(last_span.attributes['aws.region']).must_include 'stubbed'
         _(last_span.attributes['db.system']).must_be_nil
+
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::OK
```
arielvalentin, line 78:
```suggestion
        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -87,6 +94,8 @@
           _(last_span.attributes['db.system']).must_be_nil
         end
 
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::BAD_REQUEST
```
arielvalentin, line 97:
```suggestion
        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 400
```

instrumentation/aws_sdk/test/opentelemetry/instrumentation_test.rb
```diff
@@ -99,6 +108,8 @@
 
         _(last_span.attributes['rpc.system']).must_equal 'aws-api'
         _(last_span.attributes['db.system']).must_equal 'dynamodb'
+
+        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal HTTP::Status::OK
```
arielvalentin, line 112:
```suggestion
        _(last_span.attributes[OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE]).must_equal 200
```

arielvalentin (commented):
I would prefer to avoid importing gems that are not related to the instrumented library. 

If you are keen on avoiding using magic values then I would recommend you use the standard library instead. 

```ruby
_(Net::HTTPResponse::CODE_TO_OBJ[code.to_s]).must_equal Net::HTTPOK
```

NathanielRN:
> I would prefer to avoid importing gems that are not related to the instrumented library.

I don't mind at all! In OTel Python we also test the status codes but we use a library so I thought I would be proactive and look for the first popular library that does this in Ruby 🙂 

Thanks for making it easy for me and collecting those suggested changes! I've made the commit!

arielvalentin (approved):

YanivD (approved):

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
```diff
@@ -37,6 +37,9 @@ def call(context) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metri
             else
               super
             end.tap do |response|
+              span.set_attribute(OpenTelemetry::SemanticConventions::Trace::HTTP_STATUS_CODE,
+                                 context.http_response.status_code)
```
blumamir, line 41: Is it guaranteed that `context.http_response` is present in all services and operations?
Should we consider the use of [safe navigation operator](https://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) so this code will no throw if the value is `Nil` for some reason?

blumamir (approved):

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/handler.rb
blumamir: I now read the `successful?` implementation you attached and saw that it's safe to use it 👍

ericmustin (approved):

fbogsany (approved):

PR #1103, opened 2022-01-21
Consistent CRLF in AWS SDK Instrumentation Changelog

NathanielRN:
## Description

Follow up to #1099 which changed the lines breaks to be LF instead of CRLF like they were before. This breaks our regex of the CHANGELOG file that we need for the release:

https://github.com/open-telemetry/opentelemetry-ruby/blob/b7149fb72cad6c45956a6be80acfb2d062aab3a6/.toys/release/.lib/release_utils.rb#L228-L230


fbogsany (approved):

fbogsany:
Thank you @NathanielRN !

PR #1108, opened 2022-01-27
Metrics SDK synchronous instruments

robertlaurin:
First pass implementation of the [Metrics SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md).  

In scope:
- Synchronous instruments
  - Counter
  - Up down counter
  - Histogram
- Default aggregation
  - Sum
  - (optional) ExplicitBucketHistogram
- Meter Provider provisioning
- Meter provisioning
- Synchronous instrument creation

Out of scope:
- Revisions of the API to reflect spec churn
- Views
- Observable instruments
- Exporters (otlp, prom, etc)
- Metric Readers (periodic pull reader)
- Aggregation configuration
  - Aggregation temporality
  - Exponential bucket histogram aggregation, etc...


metrics_sdk/lib/opentelemetry/metrics_sdk.rb
```diff
@@ -0,0 +1,32 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry-metrics-api'
+
+module OpenTelemetry
+  # MetricsSDK provides the reference implementation of the OpenTelemetry Metrics API.
+  module MetricsSDK
```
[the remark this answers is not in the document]
fbogsany: Yep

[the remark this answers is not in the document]
robertlaurin: Updated https://github.com/open-telemetry/opentelemetry-ruby/pull/1108/commits/66c7a0db75c0a0f9006ef78b3dfd89f427034706

metrics_sdk/lib/opentelemetry/sdk/metrics/meter.rb
```diff
@@ -0,0 +1,35 @@
...
+      class Meter < OpenTelemetry::Metrics::Meter
+        # @api private
+        #
+        # Returns a new {Meter} instance.
+        #
+        # @param [String] name Instrumentation package name
+        # @param [String] version Instrumentation package version
+        # @param [TracerProvider] tracer_provider TracerProvider that initialized the tracer
+        #
+        # @return [Tracer]
```
fbogsany, line 22:
```suggestion
        # @return [Meter]
```

metrics_sdk/test/opentelemetry/metrics_sdk_test.rb
```diff
@@ -0,0 +1,53 @@
...
+describe OpenTelemetry::SDK do
+  describe '#configure' do
+    before { reset_metrics_sdk }
+
+    it 'upgrades the API MeterProvider, Meters, and Instruments' do
+      meter_provider = OpenTelemetry.meter_provider
+      meter = meter_provider.meter("test")
+      instrument = meter.create_counter("a_counter")
+
+      # Calls before the SDK is configured return ProxY implementations
```
fbogsany, line 18:
```suggestion
      # Calls before the SDK is configured return Proxy implementations
```

metrics_sdk/lib/opentelemetry/sdk/metrics/export/console_exporter.rb
```diff
@@ -0,0 +1,30 @@
...
+
+          def export(metrics)
+            puts metrics
+          end
+
+          def shutdown
+            SUCCESS
+          end
+
+          def preferred_temporality
```
fbogsany, line 22: Looks like we also need a way to specify aggregation temporality - i.e. an attribute writer or an initializer argument.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/metric_reader.rb
```diff
@@ -0,0 +1,26 @@
...
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class MetricReader
+          def initialize(exporter)
+            @exporter = exporter
+          end
+
+          def collect; end
```
fbogsany, line 16: We *might* want an optional `timeout` parameter here as well. Not sure yet.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,121 @@
...
+        #
+        # @param metric_reader the new MetricReader to be added.
+        def add_metric_reader(metric_reader)
+          @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('calling MetricProvider#add_metric_reader after shutdown.')
+              return
+            end
+
+            @metric_readers = @metric_readers.push(metric_reader)
```
fbogsany, line 106: Don't really need the assignment here, since `push` mutates the receiver.
```suggestion
            @metric_readers.push(metric_reader)
```
Dumb question: this will return the `@metric_readers`, right? Is that a good idea or should we explicitly return `nil`? I know we do the same in `TracerProvider` - I think that might be a mistake as well.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,121 @@
...
+        # or send the metrics. Note: unlike Push Metric Exporter which can send data on
+        # its own schedule, Pull Metric Exporter can only send the data when it is
+        # being asked by the scraper, so ForceFlush would not make much sense.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def force_flush(timeout: nil)
+          @mutex.synchronize do
+            return Export::SUCCESS if @stopped
```
fbogsany, line 83: I'm curious about the style difference between this and `shutdown` above, specifically `return`ing from the block here, and not in `shutdown`. Poking around, it looks like MRI will ensure the mutex is released even on `return`, though this isn't explicitly specified in the docs. We use `return` in the `TracerProvider` as well.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,121 @@
...
+          @views = []
+        end
+
+        # Returns a {Meter} instance.
+        #
+        # @param [String] name Instrumentation package name
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [Meter]
+        def meter(name, version = nil)
```
fbogsany, line 32: Are we holding off on the `schema_url` for now?

robertlaurin: Yeah, I'd like to.  Still needs to be added to the tracing portion of things.  May as well do it at the same time as one consistent PR.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
robertlaurin: I think it's just a stylistic difference.  Do you have a preference around the use of explicit vs implicit returns for a method like this? (I guess the argument could be made that we should just remain consistent with the style already being used in the tracer provider)

metrics_sdk/lib/opentelemetry/sdk/metrics/meter.rb
```diff
@@ -0,0 +1,35 @@
...
+        def initialize(name, version)
+          @mutex = Mutex.new
+          @registry = {}
+          @instrumentation_library = InstrumentationLibrary.new(name, version)
+        end
+
+        def create_counter(name, unit: nil, description: nil)
+          create_instrument(:counter, name, unit, description, nil) { OpenTelemetry::SDK::Metrics::Instrument::Counter.new }
```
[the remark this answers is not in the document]
robertlaurin: I think that makes sense https://github.com/open-telemetry/opentelemetry-ruby/pull/1108/commits/c78ee10acbf6f2ff6cff74f897acb901fab38787

yfractal:
Hi everyone, do we have plan to support this?

robertlaurin:
> Hi everyone, do we have plan to support this?

Hey @yfractal, could you clarify what you mean by support this?  This is a WIP branch for the metrics SDK implementation, when it's no longer a WIP we will release and support it if that is what you are asking.

metrics_sdk/lib/opentelemetry/sdk/metrics/export.rb
```diff
@@ -0,0 +1,29 @@
...
+        ExportError = Class.new(OpenTelemetry::Error)
+
+        # The operation finished successfully.
+        SUCCESS = 0
+
+        # The operation finished with an error.
+        FAILURE = 1
+
+        # The operation timed out.
+        TIMEOUT = 2
```
fbogsany, line 20: Is this part of the spec? For tracing, it's only used as an extra result code for `force_flush` and `shutdown`. The spec is handwavy about whether it should be a code (it could equally be an exception).

metrics_sdk/lib/opentelemetry/sdk/metrics/export/console_metric_exporter.rb
```diff
@@ -0,0 +1,29 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class ConsoleMetricExporter
+          PREFERRED_TEMPORALITY = 'delta'
```
fbogsany, line 12: Since we're using `frozen_string_literal`, is there any benefit to the named constant?

yfractal:
Hi @robertlaurin, yeah. it is, so great! hope it will be merged into master soon.

robertlaurin:
> Hi @robertlaurin, yeah. it is, so great! hope it will be merged into master soon.
> 
> I'm working on Ruby OpenTelemetry metric related thing, like to help on this.

Definitely want this to be out sooner than later too 😄 

Right now the focus is getting the foundational pieces of metrics SDK implemented to make it easier for collaboration on it. So an non exhaustive list would be:
- Metrics sdk initialization
- Meter creation
- Synchronous instrument creation
- Synchronous metric state storage
- Some basic metric readers/exporters (console, in memory)

Once those are at a point where we feel like there is some stability this PR will ideally get merged.

Some of the obvious work to pick up after that will be:
- Various exporters (prometheus for example)
- Adding observable instruments (async instruments)
- Adding support for views
- Many iterations of spec compliance review

yfractal:
> Right now the focus is getting the foundational pieces of metrics SDK implemented to make it easier for collaboration on it.

I read through the specification and Python SDK those days, I fond there are a lot of concepts and not easy to understand.

If we have the "foundational pieces", it will be help a lot !

robertlaurin:
CC: @fbogsany 
image
Corresponding commit https://github.com/open-telemetry/opentelemetry-ruby/pull/1108/commits/9c34b4e826d457b240a5ce1f3a36a195837c407f

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,134 @@
...
+                remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+                if remaining_timeout&.zero?
+                  Export::TIMEOUT
+                else
+                  metric_reader.shutdown(timeout: remaining_timeout)
+                end
+              end
+
+              @stopped = true
+              results.max || Export::SUCCESS
```
ahayworth, line 69: When would we need this fallback case? (And if we didn't have any results, would we want `FAILURE` instead?)

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,134 @@
...
+              results = @metric_readers.map do |metric_reader|
+                remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+                if remaining_timeout&.zero?
+                  Export::TIMEOUT
+                else
+                  metric_reader.force_flush(timeout: remaining_timeout)
+                end
+              end
+
+              results.max || Export::SUCCESS
```
ahayworth, line 98: Same question: When would we need this fallback case? (And if we didn't have any results, would we want FAILURE instead?)

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -0,0 +1,39 @@
...
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module State
+        class MetricStore
+          def initialize
+            @mutex = Mutex.new
+            @epoch_start_time = Time.now.to_i
+            @epoch_end_time = nil;
```
ahayworth, line 15:
```suggestion
            @epoch_end_time = nil
```

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -0,0 +1,39 @@
...
+
+          def collect
+            # this probably needs to take the mutex, take a snapshot of state, (reset state?), release the mutex
+            @mutex.synchronize do
+              [
+                @end_time = Time.now.to_i,
+                @metric_streams,
+                @epoch_start_time = @end_time,
+              ]
+              @metric_streams
```
ahayworth, line 27: I don't understand these 6 lines - did you mean to return the array, or `@metric_streams` ?

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -0,0 +1,39 @@
...
+            end
+          end
+
+          def add_metric_stream(metric_stream)
```
ahayworth, line 31: You might want to take the mutex here too?

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -0,0 +1,60 @@
...
+            @mutex = Mutex.new
+          end
+
+          def update(measurement)
+            @mutex.synchronize do
+              if @data_points[measurement.attributes]
+                @data_points[measurement.attributes] = @data_points[measurement.attributes] + measurement.value
+              else
+                @data_points[measurement.attributes] = measurement.value
+              end
```
ahayworth, line 39: Is this supposed to be an array of metrics updates? Or just a running sum? I think we've implemented the latter here, but I was expecting the former based on the class name.

(Of course maybe you haven't gotten around to it yet. Or maybe I just misunderstand what we're doing entirely, which wouldn't surprise me at all. 😆 )

ahayworth (approved):
I left a few questions, but overall I like the direction!

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
```diff
@@ -68,8 +71,67 @@ def as_etsr(span_data) # rubocop:disable Metrics/MethodLength
...
+                  ),
+                  instrumentation_library_metrics: metric_datas
+                  .group_by(&:instrumentation_library)
+                  .map do |il, mds|
+                    Opentelemetry::Proto::Metrics::V1::InstrumentationLibraryMetrics.new(
+                      instrumentation_library: Opentelemetry::Proto::Common::V1::InstrumentationLibrary.new(
+                        name: il.name,
+                        version: il.version
+                      ),
+                      metrics: mds.group_by.map do |metric_stream|
```
fbogsany, line 95: what does the `group_by` do here?

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
```diff
@@ -68,8 +71,67 @@ def as_etsr(span_data) # rubocop:disable Metrics/MethodLength
...
+        end
+
+        def as_sum_data_point(attributes, value)
+          Opentelemetry::Proto::Metrics::V1::NumberDataPoint.new(
+            attributes: attributes&.map { |k, v| as_otlp_key_value(k, v) },
+            as_int: value
+          )
+        end
+
+        def as_otlp_string_key_value(key, value)
```
fbogsany, line 131: I didn't see any references to this method.

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -24,31 +24,33 @@ class Meter
       InstrumentUnitError = Class.new(OpenTelemetry::Error)
       InstrumentDescriptionError = Class.new(OpenTelemetry::Error)
 
+      attr_reader :instrument_registry
```
fbogsany, line 27: Is this supposed to be public?

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -60,12 +62,14 @@ def create_instrument(kind, name, unit, description, callback)
         raise InstrumentNameError if name.empty?
         raise InstrumentNameError unless NAME_REGEX.match?(name)
         raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
-        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description))
+        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
 
         @mutex.synchronize do
-          raise DuplicateInstrumentError if @registry.include? name
+          if @instrument_registry.include? name
+            OpenTelemetry.logger.warn("duplicate instrument registration occured for instrument #{name}")
```
fbogsany, line 69:
```suggestion
            OpenTelemetry.logger.warn("duplicate instrument registration occurred for instrument #{name}")
```

Do we want to proceed with the registration in this case? I think we'd want to keep the existing instrument and ignore the duplicate one.

fbogsany: I probably missed the change and its associated discussion, but why do we want to log rather than raising here?

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+      module Export
+        class InMemoryMetricPullExporter < MetricReader
+          PREFERRED_TEMPORALITY = 'delta'
+
+          attr_reader :metric_snapshots, :metric_store
+
+          def initialize
+            @metric_snapshots = []
+            @mutex = Mutex.new
+            @metric_store = OpenTelemetry::SDK::Metrics::State::MetricStore.new
```
fbogsany, line 19: The superclass also initializes this instance variable, but you're not calling `super` here.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class InMemoryMetricPullExporter < MetricReader
+          PREFERRED_TEMPORALITY = 'delta'
+
+          attr_reader :metric_snapshots, :metric_store
```
fbogsany, line 14: Why do we expose the `metric_store` directly? Should we just expose the `metric_snapshots`?

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class InMemoryMetricPullExporter < MetricReader
+          PREFERRED_TEMPORALITY = 'delta'
```
fbogsany, line 12: Should this be a `private_constant`? We have a method that returns it, so the constant doesn't seem useful.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
fbogsany: 🤔 the superclass also exposes `metric_store`.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/metric_reader.rb
```diff
@@ -0,0 +1,34 @@
...
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class MetricReader
+          attr_reader :metric_store
+
+          def initialize(exporter)
+            @exporter = exporter
```
fbogsany, line 15:
```suggestion
          def initialize
```
I *think* we don't need to expose the `metric_store` and the `exporter` isn't used. Tell me why I'm wrong.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.rb
```diff
@@ -0,0 +1,34 @@
...
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class PeriodicExportingMetricReader < MetricReader
+          # exporter - the push exporter where the metrics are sent to.
+          # exportIntervalMillis - the time interval in milliseconds between two consecutive exports. The default value is 60000 (milliseconds).
+          # exportTimeoutMillis - how long the export can run before it is cancelled. The default value is 30000 (milliseconds).
+          def initialize(exporter, aggregation_temporality: nil, export_interval_millis: 60_000, export_timeout_millis: 30_000)
+            @exporter = exporter
```
fbogsany, line 16: Superclass also initializes `@exporter`, but we don't call `super` here. The superclass also initializes the `metric_store` - is `collect` going to 💥 ?

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/synchronous_instrument.rb
```diff
@@ -0,0 +1,48 @@
...
+        class SynchronousInstrument
+          def initialize(name, unit, description, instrumentation_library, meter_provider)
+            @name = name
+            @unit = unit
+            @description = description
+            @instrumentation_library = instrumentation_library
+            @metric_streams = []
+
+            meter_provider.metric_readers.each do |metric_reader|
+              register_with_new_metric_store(metric_reader.metric_store)
```
fbogsany, line 20: OIC - so you *do* need `metric_store` exposed on `MetricReader` 🤔 - maybe make it API-private?

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/synchronous_instrument.rb
```diff
@@ -0,0 +1,48 @@
...
+            end
+          end
+
+          def register_with_new_metric_store(metric_store)
```
fbogsany, line 24: Feels like this should be `private`, but I feel like I'm going to discover otherwise 😄

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
fbogsany: Yep - you could end up having this called in multiple threads simultaneously and end up dropping some metric streams.

metrics_sdk/test/integration/in_memory_metric_pull_exporter_test.rb
```diff
@@ -0,0 +1,36 @@
...
+      OpenTelemetry.meter_provider.add_metric_reader(metric_exporter)
+
+      meter = OpenTelemetry.meter_provider.meter('test')
+      instrument = meter.create_counter('b_counter')
+
+      instrument.add(1)
+      instrument.add(2, attributes: { 'a' => 'b' })
+      instrument.add(2, attributes: { 'a' => 'b' })
+      instrument.add(3, attributes: { 'b' => 'c' })
+      instrument.add(4, attributes: { 'd' => 'e' })
```
fbogsany, line 26: Do we want any expectations here about the metric streams?

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,134 @@
...
+          @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('calling MetricProvider#add_metric_reader after shutdown.')
+            else
+              @metric_readers.push(metric_reader)
+              @meter_registry.each do |_k, meter|
+                meter.instrument_registry.each do |_n, instrument|
+                  instrument.register_with_new_metric_store(metric_reader.metric_store)
+                end
+              end
```
fbogsany, line 116: I'd probably move the `meter.instrument_registry ...` bit inside `Meter`, like:
```suggestion
              @meter_registry.each_value { |meter| meter.add_metric_reader(metric_reader) }
```

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
fbogsany: Fallback is hit if `@metric_readers` is empty.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
fbogsany: See below - this is only hit if we have no metric readers.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/metric_exporter.rb
```diff
@@ -0,0 +1,287 @@
...
+            around_request do # rubocop:disable Metrics/BlockLength
+              request = Net::HTTP::Post.new(@path)
+              request.body = if @compression == 'gzip'
+                               request.add_field('Content-Encoding', 'gzip')
+                               Zlib.gzip(bytes)
+                             else
+                               bytes
+                             end
+              request.add_field('Content-Type', 'application/x-protobuf')
+              @headers.each { |key, value| request.add_field(key, value) }
```
fbogsany, line 147: We should hoist all this outside of the `around_request` block.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
robertlaurin: Some left over from previous iterations, I'll clean this up.

I'll be removing all the export work from this PR, and push it up as an entirely separate PR for review.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
robertlaurin: Same comment as above

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/metric_exporter.rb
robertlaurin: Yup, this is now out of date with the released `opentelemetry-exporter-otlp` gem. There's a bit of work to be done here in general.

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -60,12 +62,14 @@ def create_instrument(kind, name, unit, description, callback)
         raise InstrumentNameError if name.empty?
         raise InstrumentNameError unless NAME_REGEX.match?(name)
         raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
-        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description))
+        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
```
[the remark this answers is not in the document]
robertlaurin: I'll have to re-review this, there was some wording in the API specifically around it being a failure if it was not utf8mb3.  That may have changed since adding this.

metrics_api/lib/opentelemetry/metrics/meter.rb
robertlaurin: I should not have brought this into this PR, but the API spec changed.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-type-conflict-detection

>When more than one Instrument of the same name is created for identical Meters, denoted duplicate instrument registration, the implementation MUST create a valid Instrument in every case. Here, "valid" means an instrument that is functional and can be expected to export data, despite potentially creating a [semantic error in the data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model-producer-recommendations).

>It is unspecified whether or under which conditions the same or different Instrument instance will be returned as a result of duplicate instrument registration. The term identical applied to Instruments describes instances where all identifying fields are equal. The term distinct applied to Instruments describes instances where at least one field value is different.

>When more than one distinct Instrument is registered with the same name for identical Meters, the implementation SHOULD emit a warning to the user informing them of duplicate registration conflict(s). The warning helps to avoid the semantic error state described in the [OpenTelemetry Metrics data model](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model-producer-recommendations) when more than one Metric is written for a given instrument name and Meter identity by the same MeterProvider.

metrics_sdk/lib/opentelemetry/sdk/metrics/export.rb
robertlaurin: I got similar hand wavy interpretations, but maybe I'm being too liberal with that and should not codify the timeout result.

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#collect

>Collect does not have any required parameters, however, [OpenTelemetry SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#sdk) authors MAY choose to add parameters (e.g. callback, filter, timeout). [OpenTelemetry SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#sdk) authors MAY choose the return value type, or do not return anything.

metrics_api/lib/opentelemetry/metrics/meter.rb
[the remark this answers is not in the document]
ahayworth: Ah, that would make sense. I may have missed that bit.

robertlaurin:
@ahayworth @fbogsany this is at reviewable stopping point again.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_exporting_metric_reader.rb
robertlaurin: Remove from this PR.

metrics_sdk/test/integration/in_memory_metric_pull_exporter_test.rb
robertlaurin: Added some assertions!

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/metric_exporter.rb
robertlaurin: Removed from the PR.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/synchronous_instrument.rb
robertlaurin: marked as "api private"

metrics_sdk/lib/opentelemetry/sdk/metrics/export/metric_reader.rb
robertlaurin: Removed the exporter, and as you discovered later it is being exposed to allow for registering instruments later.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
robertlaurin: Removed to be addressed in a followup PR.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
robertlaurin: Updated to not inherit from MetricReader, the metric store is required for registering the instruments.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        class InMemoryMetricPullExporter < MetricReader
```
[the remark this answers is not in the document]
robertlaurin: I added a comment to add some description, it is a MetricReader/Exporter combination class to facilitate testing of the SDK. It no longer inherits.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
robertlaurin: Updated to hopefully be more understandable, will resolve the comment to clear up the discussion on the PR.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
robertlaurin: Updated!

metrics_api/lib/opentelemetry-metrics-api.rb
```diff
@@ -17,6 +17,10 @@
 #
 # The OpenTelemetry module provides global accessors for telemetry objects.
 module OpenTelemetry
+
+  @mutex = Mutex.new
```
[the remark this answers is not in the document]
robertlaurin: Removed.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/console_exporter.rb
robertlaurin: Will differ configuration of temporality to a follow up PR.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation.rb
```diff
@@ -0,0 +1,22 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry/sdk/metrics/aggregation/histogram'
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      # The Aggregation module contains the OpenTelemetry metrics reference
+      # aggregation implementations.
+      module Aggregation
```
fbogsany, line 14: Is this meant to be API-private, or does the spec want us to expose aggregations from the SDK?

fbogsany: Looks like the answer is "expose". 👌

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
fbogsany: > the metric store is required for registering the instruments

It's required by the SDK, but not by the spec (I think). Should we mark it API-private? There's a 👃 here...

metrics_sdk/lib/opentelemetry/sdk/metrics/export/metric_reader.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Export
+        # MetricReader provides a minimal example implementation.
+        # It is not required to subclass this class to provide an implementation
+        # of MetricReader, provided the interface is satisfied.
```
fbogsany, line 13: I think it's problematic that we have to expose `metric_store` as part of the interface. That should be a SDK-internal concern, which means we probably *do* have to mark it as API-private and require concrete implementations to subclass this class.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          def initialize
+            @metric_snapshots = []
+            @mutex = Mutex.new
+            @metric_store = OpenTelemetry::SDK::Metrics::State::MetricStore.new
+          end
+
+          def collect
+            @metric_store.collect
+          end
```
fbogsany, line 24: I think it should be something like:
```suggestion
        class InMemoryMetricPullExporter < MetricReader
          attr_reader :metric_snapshots

          def initialize
            super
            @metric_snapshots = []
            @mutex = Mutex.new
          end
```

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -0,0 +1,80 @@
...
+            @unit = unit
+            @instrument_kind = instrument_kind
+            @meter_provider = meter_provider
+            @instrumentation_library = instrumentation_library
+
+            @data_points = {}
+            @mutex = Mutex.new
+          end
+
+          def collect(start_time, end_time)
```
fbogsany, line 37: Does `data_points` keep growing without bound? I'm not clear on the expected behaviour here - part of me thinks we're supposed to only return data points recorded during the interval `start_time .. end_time`, but I'm really not sure.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/counter.rb
```diff
@@ -0,0 +1,47 @@
...
+          end
+
+          # Increment the Counter by a fixed amount.
+          #
+          # @param [numeric] increment The increment amount, which MUST be a non-negative numeric value.
+          # @param [Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}] attributes
+          #   Values must be non-nil and (array of) string, boolean or numeric type.
+          #   Array values must not contain nil elements and all elements must be of
+          #   the same basic type (string, numeric, boolean).
+          def add(increment, attributes: {})
```
ahayworth, line 29:
```suggestion
          def add(increment, attributes: {})
            # TODO: When the metrics SDK stabilizes and is merged into the main SDK,
            # we can leverage the SDK Internal validation classes to enforce this:
            # https://github.com/open-telemetry/opentelemetry-ruby/blob/6bec625ef49004f364457c26263df421526b60d6/sdk/lib/opentelemetry/sdk/internal.rb#L47
```

ahayworth (approved):
I don't have much to add other than what Francis already mentioned!

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
robertlaurin: That needs to be addressed as part of the aggregation temporality work that is not part of this PR, i.e. sum or delta.

In the case of sum, there should be some concept of pruning. It's mentioned in the [guidelines](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/supplementary-guidelines.md#guidelines-for-sdk-authors).

>So here are some suggestions that we encourage SDK implementers to consider:
> * You want to control the memory usage rather than allow it to grow indefinitely / unbounded - regardless of what aggregation temporality is being used.
> * You want to improve the memory efficiency by being able to forget about things that are no longer needed.
> * You probably don't want to keep exporting the same thing over and over again, if there is no updates. You might want to consider [Resets and Gaps](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#resets-and-gaps). For example, if a Cumulative metrics stream hasn't received any updates for a long period of time, would it be okay to reset the start time?

metrics_api/lib/opentelemetry/metrics/meter.rb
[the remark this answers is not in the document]
plantfansam: I read [this](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-description) as "you _must_ support utf8m3" but don't see anything about blowing up if desc is not utfmb3.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
```diff
@@ -0,0 +1,40 @@
...
+
+          # Updates the statistics with the specified amount.
+          #
+          # @param [numeric] amount The amount of the Measurement, which MUST be a non-negative numeric value.
+          # @param [Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}] attributes
+          #   Values must be non-nil and (array of) string, boolean or numeric type.
+          #   Array values must not contain nil elements and all elements must be of
+          #   the same basic type (string, numeric, boolean).
+          def record(amount, attributes: nil)
+            update(OpenTelemetry::Metrics::Measurement.new(amount, attributes))
```
plantfansam, line 30: Do we want to warn if negative numeric value, here?

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -0,0 +1,130 @@
...
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [Meter]
+        def meter(name, version = nil)
+          version ||= ''
+          if @stopped
+            OpenTelemetry.logger.warn 'calling MeterProvider#meter after shutdown, a noop meter will be returned.'
+            OpenTelemetry::Metrics::Meter.new
+          else
+            @mutex.synchronize { @meter_registry[Key.new(name, version)] ||= Meter.new(name, version, self) }
```
plantfansam, line 39: Might explicitly namespace this `Meter` for readability

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -0,0 +1,48 @@
...
+    module Metrics
+      module State
+        # @api private
+        #
+        # The MetricStore module provides SDK internal functionality that is not a part of the
+        # public API.
+        class MetricStore
+          def initialize
+            @mutex = Mutex.new
+            @epoch_start_time = now_in_nano
```
plantfansam, line 18: An `epoch` from start_time to end_time is basically the  duration between calls to `collect`, is that right?

metrics_api/lib/opentelemetry/metrics/meter.rb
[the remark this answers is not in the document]
ahayworth: To be fair, I'm okay with being a little strict and then addressing the case later on when someone uses emoji in their metrics names or something. 😄

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -0,0 +1,80 @@
...
+            @mutex.synchronize do
+              MetricData.new(
```
plantfansam, line 39: I'm not clear on how instantiating this object fulfills the obligations of `collect`, but I'm definitely missing something....

metrics_sdk/lib/opentelemetry/sdk/metrics/meter.rb
```diff
@@ -0,0 +1,51 @@
...
+        # @api private
+        def add_metric_reader(metric_reader)
+          @instrument_registry.each do |_n, instrument|
+            instrument.register_with_new_metric_store(metric_reader.metric_store)
+          end
+        end
+
+        def create_instrument(kind, name, unit, description, callback)
+          super do
+            case kind
```
plantfansam, line 38: The intent is that `super do` will yield the return of the `case` statement to the superclass?

plantfansam (commented):
I have a few questions — feel free to disregard, as most are borne from not fully understanding the spec. Nice work!

metrics_sdk/lib/opentelemetry/sdk/metrics/export.rb
fbogsany: Re-reading:

> Collect SHOULD provide a way to let the caller know whether it succeeded, failed or timed out

So I think this is fine.

metrics_sdk/lib/opentelemetry/sdk/metrics/configuration_patch.rb
```diff
@@ -0,0 +1,25 @@
...
+    module Metrics
+      # The ConfiguratorPatch implements a hook to configure the metrics
+      # portion of the SDK.
+      module ConfiguratorPatch
+        private
+
+        # The metrics_configuration_hook method is where we define the setup process
+        # for metrics SDK.
+        def metrics_configuration_hook
+          OpenTelemetry.meter_provider = Metrics::MeterProvider.new(resource: @resource)
```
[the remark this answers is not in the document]
robertlaurin: The @resource instance var will be available in the configurator, it's a little awkward as were developing this as a separate gem, but once we have stabilized the metrics api/sdk it will get merged into the SDK/API and this patch stuff will go away.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
robertlaurin: The histogram aggregation needs to be added properly, I'm working on that in a separate branch to follow up on this one.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -0,0 +1,48 @@
...
+              @epoch_end_time = now_in_nano
+              snapshot = @metric_streams.map { |ms| ms.collect(@epoch_start_time, @epoch_end_time) }
+              @epoch_start_time = @epoch_end_time
+              snapshot
+            end
+          end
+
+          def add_metric_stream(metric_stream)
+            @mutex.synchronize do
+              @metric_streams = @metric_streams.dup.push(metric_stream)
```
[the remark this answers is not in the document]
robertlaurin: It's a means of making the change atomic, the @metric_streams value won't change mid process. It'll either have the old array, or the new one. It won't change "in flight" so to speak.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
robertlaurin: Think of it as collecting a snapshot of the metrics at a certain point in time. The metrics will accumulate between _collect calls_, so we want to capture all the information necessary to export and the current state of the data points. What's not shown here is that we will also need to configure the temporality of the metrics, so in the case of delta, after grabbing this snapshot we will drop all the datapoints. In cumulative temporality we need the datapoints and their values at the time of collection, but after collection finishes the data points on the metric stream can continue to grow, but we don't want to have that data change once it has been sent off for export. This is intended to be a simple struct to encapsulate this information distinctly from the metric stream.

lisamburns:
This is exciting! Would love to use it, when will be a release cut?

ahayworth:
Hi @lisamburns - not in the _immediate_ future. We need to add aggregation support [in progress here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1346), and then exporter support (not yet formally started, although very rough versions have been written as proof-of-concept stuff).

I don't know that we have a specific timeline to share, but @robertlaurin has been working on getting it to a state where the work can be done in parallel. He could likely share more information!

robertlaurin:
Hey @lisamburns, the metrics isn't quite ready for consumption yet.

We'll cut a release when more of the implementation is complete. In its current state you could simple create synchronous instruments but the export path is not implemented, and the aggregation is incomplete.

You can track the work in flight and the work to do on the metrics project board https://github.com/open-telemetry/opentelemetry-ruby/projects/2

PR #1110, opened 2022-01-29
feat: Add configurable trace metrics reporter

arielvalentin:
These changes make it possible for users to inject a custom trace metrics reporter as part of the configuration block.

- Adds the OTLP Exporter as a test dependency to ensure default metrics reporters are injected by the configurator
- Disables the BSP reporter thread by default to avoid spawning threads for each new instance of the BatchSpanExporter created in test cases


arielvalentin:
ping @open-telemetry/ruby-maintainers @open-telemetry/ruby-approvers

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -210,12 +211,16 @@ def fetch_propagator(name, class_name, gem_suffix = name)
         nil
       end
 
-      def fetch_exporter(name, class_name)
-        Trace::Export::BatchSpanProcessor.new(Kernel.const_get(class_name).new)
+      def fetch_exporter(name, class_name, exporter_options = {})
+        Trace::Export::BatchSpanProcessor.new(Kernel.const_get(class_name).new(**exporter_options), metrics_reporter: metrics_reporter)
```
ericmustin, line 215: If we aren't documenting this feature, i think it's fine to leave as is. Right now i think the goal is, give power users the ability to insert their own metric reporter provider. If we plan to document or make this something a little more user facing, it might be useful to validate that the methods and method signatures for `add_to_counter`, `record_value`, and `observe_value` exist and are correct, since it'll end up creating runtime errors if not. I don't think this is blocking just thinking out loud about ways this could go wrong in the future.

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -210,12 +211,16 @@ def fetch_propagator(name, class_name, gem_suffix = name)
...
       rescue NameError
         OpenTelemetry.logger.warn "The #{name} exporter cannot be configured - please add opentelemetry-exporter-#{name} to your Gemfile, spans will not be exported"
         nil
       end
+
+      def metrics_reporter
+        @trace_metrics_reporter_provider.call
```
ericmustin, line 222: Possible i'm missing some logic in how @trace_metrics_reporter_provider gets defined, but iiuc, if @ `trace_metrics_reporter_provider` is not set and defaults to nil, wouldn't this throw an error?
```suggestion
        @trace_metrics_reporter_provider.call unless @trace_metrics_reporter_provider.nil?
```

ericmustin (changes requested):
Good stuff, small feedback about an edge case but broadly lgtm, i think its fine to keep this pretty loose, ideally we try to deprecate in the future when otel metrics are available, but that shouldn't block this work which is incrementally useful to users.

sdk/lib/opentelemetry/sdk/configurator.rb
arielvalentin: It defaults to a proc that returns a `nil` metrics reporter:

https://github.com/open-telemetry/opentelemetry-ruby/pull/1110/files#diff-e6a05d2b9d8d72a4c1ab589e75a9aa34e484e073af6443d428929c0c554f93eaR44

and in turn BSP and OTLP exporter default to a no-op implementation:

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb#L66

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L73

sdk/lib/opentelemetry/sdk/configurator.rb
ericmustin: i think thats a good idea, if it becomes a pain then we can drop it

arielvalentin: I have added a type check. Reporter implementations will have to extend the module which will provide a minimal set of no-op functions. 

```console
irb(main):001:0> require 'opentelemetry-sdk'
=> true
irb(main):002:1* class Reporter
irb(main):004:1*   def add_to_counter(metric, increment: 1, labels: {}); puts "got #{metric} #{increment} #{labels}" ; end
irb(main):005:0> end
=> :add_to_counter
irb(main):006:0> Reporter.new.add_to_counter("hello", labels: { foo: :bar })
got hello 1 {:foo=>:bar}
irb(main):008:0> Reporter.new.record_value("hello", value: 10, labels: { foo: :bar })
=> nil
```

ericmustin (approved):

plantfansam (approved):
LGTM!

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -210,12 +211,22 @@ def fetch_propagator(name, class_name, gem_suffix = name)
...
+      def fetch_exporter(name, class_name, exporter_options = {})
+        Trace::Export::BatchSpanProcessor.new(Kernel.const_get(class_name).new(**exporter_options), metrics_reporter: metrics_reporter)
       rescue NameError
         OpenTelemetry.logger.warn "The #{name} exporter cannot be configured - please add opentelemetry-exporter-#{name} to your Gemfile, spans will not be exported"
         nil
       end
+
+      def metrics_reporter
+        reporter = @trace_metrics_reporter_provider.call
+        if !reporter.nil? && reporter.is_a?(::OpenTelemetry::SDK::Trace::Export::MetricsReporter)
```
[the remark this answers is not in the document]
fbogsany: You can use safe navigation here:
```suggestion
        if reporter&.is_a?(::OpenTelemetry::SDK::Trace::Export::MetricsReporter)
```
However, won't this end up being a little noisy in the default case? I.e. the default provider returns `nil`, so we'll log the warning below twice if the BSP + OTLP exporter are configured.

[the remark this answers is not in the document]
arielvalentin: Fair. I will look into reducing the number of log messages.

arielvalentin:
@fbogsany please review again at your leisure.

arielvalentin:
ping @open-telemetry/ruby-maintainers

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -181,7 +184,7 @@ def wrapped_exporters_from_env # rubocop:disable Metrics/CyclomaticComplexity, M
               OpenTelemetry.logger.warn "The #{otlp_protocol} transport protocol is not supported by the OTLP exporter, spans will not be exported."
               nil
             else
-              fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter')
+              fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter', metrics_reporter, metrics_reporter: metrics_reporter)
```
fbogsany, line 187: This stuttering feels awkward. There's not a great alternative, though (e.g. just passing the hash and fishing out the `:metrics_reporter` from it is its own brand of awkward). I think we should make the hash explicit, though - I was confused initially thinking this was a keyword argument.

```suggestion
              fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter', metrics_reporter, { metrics_reporter: metrics_reporter })
```

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -170,6 +172,7 @@ def configure_span_processors
       end
 
       def wrapped_exporters_from_env # rubocop:disable Metrics/CyclomaticComplexity, Metrics/AbcSize
+        metrics_reporter = new_metrics_reporter
```
fbogsany, line 175: We're calling `new_metrics_reporter` in `configure` as well, but I don't see us using the `@metrics_reporter` - is that intentional?

sdk/lib/opentelemetry/sdk/configurator.rb
arielvalentin: It does indeed feel awkward. 

How do you feel about extracting BSP arguments into its own keyword arg and making `exporter_options` a keyword?

E.g.: 
```ruby
def fetch_exporter(name, class_name, processor_options: {}, exporter_options: {})
```

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -137,6 +138,7 @@ def add_span_processor(span_processor)
       def configure
         OpenTelemetry.logger = logger
         OpenTelemetry.error_handler = error_handler
+        @metrics_reporter = new_metrics_reporter
```
arielvalentin, line 141:
```suggestion
```

sdk/lib/opentelemetry/sdk/configurator.rb
fbogsany: I mean, it'll be even more stuttering, but I think it'd be a lot clearer:
```suggestion
              fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter',
                             processor_options: { metrics_reporter: metrics_reporter },
                             exporter_options: { metrics_reporter: metrics_reporter })
```

arielvalentin: 👍🏼

sdk/lib/opentelemetry/sdk/configurator.rb
arielvalentin: Done

sdk/lib/opentelemetry/sdk/configurator.rb
arielvalentin: Another option may be to separate span processor instantiation from exporter initialization. Make `fetch_exporter` return an `exporter` instead of a BSP and then instantiate the span processor based on the Exporter type.

Thoughts?

fbogsany: That looks like:
```ruby
          exp = case exporter.strip
          when 'none' then nil
          when 'otlp'
            otlp_protocol = ENV['OTEL_EXPORTER_OTLP_TRACES_PROTOCOL'] || ENV['OTEL_EXPORTER_OTLP_PROTOCOL'] || 'http/protobuf'
            if otlp_protocol != 'http/protobuf'
              OpenTelemetry.logger.warn "The #{otlp_protocol} transport protocol is not supported by the OTLP exporter, spans will not be exported."
              nil
            else
              fetch_exporter(exporter, 'OpenTelemetry::Exporter::OTLP::Exporter', { metrics_reporter: metrics_reporter })
            end
          when 'jaeger' then fetch_exporter(exporter, 'OpenTelemetry::Exporter::Jaeger::CollectorExporter')
          when 'zipkin' then fetch_exporter(exporter, 'OpenTelemetry::Exporter::Zipkin::Exporter')
          when 'console' then next Trace::Export::SimpleSpanProcessor.new(Trace::Export::ConsoleSpanExporter.new)
          else
            OpenTelemetry.logger.warn "The #{exporter} exporter is unknown and cannot be configured, spans will not be exported"
            nil
          end
          Trace::Export::BatchSpanProcessor.new(exp, metrics_reporter: metrics_reporter) unless exp.nil?
```
This seems reasonable to me. Note that we need the `next` for the `console` exporter, since that's wrapped in a SSP instead of a BSP.

arielvalentin: I started experimenting with a few different implementations but the asymmetry of the Jaeger and Zipkin exporters is bugging me so I am going to add the reporter to those exporters as well in a separate PR and then circle back to this.

arielvalentin: Now that the jaeger and zipkin exporters accept a metrics reporter the code is symmetric and should simplify this a little.

sdk/lib/opentelemetry/sdk/configurator.rb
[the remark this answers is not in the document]
arielvalentin: I have set the default value to be a `MetricsReporter` instance so it will no longer emit logs in the default case.

arielvalentin:
Ping @open-telemetry/ruby-approvers

ericmustin (approved):
ah, the ole 1yr+ PR. you're a determined person and i apologize for the delays.

the test failures look related to truffle ruby. i hope that project is doing ok ❤️ . I think this still lgtm, re-approving, do we need to fix those CI errors before merging this tho (not sure so asking, dont have an opinion here).

also, just making sure i understand correctly, the small change in behavior here for ppl not supplying their own metrics reporter is that we now supply the no-op metrics reporter class as a default, right? I think that's fine but just stating that for my own confidence here since it's been awhile and i'm extremely rusty wrt this repo

arielvalentin:
> ah, the ole 1yr+ PR. you're a determined person and i apologize for the delays.
> 

No apologies necessary. I let this PR linger until about 2 weeks ago... right before the holidays. 

> the test failures look related to truffle ruby. i hope that project is doing ok ❤️ . I think this still lgtm, re-approving, do we need to fix those CI errors before merging this tho (not sure so asking, dont have an opinion here).

I am not sure I understand the SimpleCov errors there for Truffle. I am not sure how my changes would have broken that. 

> also, just making sure i understand correctly, the small change in behavior here for ppl not supplying their own metrics reporter is that we now supply the no-op metrics reporter class as a default, right? I think that's fine but just stating that for my own confidence here since it's been awhile and i'm extremely rusty wrt this repo

Correct. The default metrics reporter is a no-op

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -41,6 +41,7 @@ def initialize
         @use_mode = USE_MODE_UNSPECIFIED
         @resource = Resources::Resource.default
         @id_generator = OpenTelemetry::Trace
+        @trace_metrics_reporter_provider = -> { ::OpenTelemetry::SDK::Trace::Export::MetricsReporter }
```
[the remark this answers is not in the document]
arielvalentin: The user provided option is a lambda.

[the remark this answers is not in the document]
arielvalentin: @fbogsany ping

PR #1113, opened 2022-01-31
Fix deprecation warning from rack_env_getter

jonatas:
I was running the example and saw this warning. So, fixing it to use the non-deprecated module.


robertlaurin (approved):
Thanks @jonatas

PR #1116, opened 2022-02-02
fix: prevent redis instrumentation from mutating the command

robertlaurin:
The redis instrumentation has been mutating the command during the parse steps.  This appears to have no impact on redis itself as it parses the command all the same if it's `set` or `SET`.

### DIFF
--- a/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
+++ b/instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
@@ -71,11 +71,10 @@ module OpenTelemetry
               # and return the obfuscated command
               return 'AUTH ?' if command[0] == :auth
 
-              command[0] = command[0].to_s.upcase
               if config[:db_statement] == :obfuscate
-                command[0] + ' ?' * (command.size - 1)
+                command[0].to_s.upcase + ' ?' * (command.size - 1)
               else
-                command.join(' ')
+                command[0].to_s.upcase + " " +  command[1..-1].join(' ')
               end
             end.join("\n")
           end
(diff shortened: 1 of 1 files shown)


instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -71,11 +71,10 @@ def parse_commands(commands) # rubocop:disable Metrics/AbcSize
               # and return the obfuscated command
               return 'AUTH ?' if command[0] == :auth
 
-              command[0] = command[0].to_s.upcase
               if config[:db_statement] == :obfuscate
-                command[0] + ' ?' * (command.size - 1)
+                command[0].to_s.upcase + ' ?' * (command.size - 1)
               else
-                command.join(' ')
+                command[0].to_s.upcase + " " +  command[1..-1].join(' ')
```
fbogsany, line 77: what does `command[1..-1]` return if `command.size == 1`?

robertlaurin: Ah, good catch.  There'd be some padding.
```
irb(main):001:0> command = [:set]
=> [:set]
irb(main):002:0> command[0].to_s.upcase + " " +  command[1..-1].join(' ')
=> "SET "
```

robertlaurin: Updated https://github.com/open-telemetry/opentelemetry-ruby/pull/1116/commits/15339516289ce6da4877f8ef680cedb59edac0f2

SomalianIvan (approved):

fbogsany (approved):

PR #1117, opened 2022-02-02
fix: excessive hash creation on context attr merging

robertlaurin:
Fixes the ordering of attribute merging in some places, switches to using `merge!` to avoid creating an additional hash on merge.


arielvalentin (approved):
I wonder if there is a linter we could leverage for this 🤔

robertlaurin:
>I wonder if there is a linter we could leverage for this 

Yeah I thought about it too, but I don't know how we could make it nuanced enough, we don't also want to use it where we want to preserve the thing that is getting merge.

Like here for example, I think we want to preserve the original hash as we potentially nest various client contexts https://github.com/open-telemetry/opentelemetry-ruby/blob/df9bbb0cf23e537fd5c361cd0ebcd3392ea2ffff/common/lib/opentelemetry/common/http/client_context.rb#L33-L36

fbogsany (approved):

PR #1125, opened 2022-02-21
use rails request's filtered path as http.target attribute

chrisholmes:
This change updates Rack spans so that their `http.target` attributes are consistent with [Rails' filter_parameter config](https://guides.rubyonrails.org/configuring.html#config-filter-parameters) when the ActionPack instrumentation is installed.

I have chosen to not make this optional as I feel it is important that the instrumentation follows the instruction set by Rails' configuration.

### DIFF
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
@@ -18,7 +18,7 @@ module OpenTelemetry
 
                 add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
 
-                rack_span.set_attribute('http.target', request.filtered_path)
+                rack_span.set_attribute('http.target', request.filtered_path) if request.filtered_path != request.fullpath
               end
 
               super(name, request, response)
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
@@ -16,6 +16,9 @@ module OpenTelemetry
               rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
 
               add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
+
+              rack_span.set_attribute('http.target', request.filtered_path)
+
               super(name, request, response)
             end
 
--- a/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
+++ b/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
@@ -13,11 +13,13 @@ module OpenTelemetry
           module Metal
             def dispatch(name, request, response)
               rack_span = OpenTelemetry::Instrumentation::Rack.current_span
-              rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
+              if rack_span.recording?
+                rack_span.name = "#{self.class.name}##{name}" unless request.env['action_dispatch.exception']
 
-              add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
+                add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
 
-              rack_span.set_attribute('http.target', request.filtered_path)
+                rack_span.set_attribute('http.target', request.filtered_path)
+              end
 
               super(name, request, response)
             end
--- a/instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/patches/action_controller/metal_test.rb
+++ b/instrumentation/action_pack/test/opentelemetry/instrumentation/action_pack/patches/action_controller/metal_test.rb
@@ -99,6 +99,13 @@ describe OpenTelemetry::Instrumentation::ActionPack::Patches::ActionController::
     end
   end
 
+  it 'sets filters `http.target`' do
+    get '/ok?param_to_be_filtered=bar&unfiltered_param=baz', {}
+    _(last_response.body).must_equal 'actually ok'
+    _(last_response.ok?).must_equal true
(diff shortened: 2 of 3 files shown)


instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -16,6 +16,9 @@ def dispatch(name, request, response)
               rack_span.name = "#{self.class.name}##{name}" if rack_span.context.valid? && !request.env['action_dispatch.exception']
```
chrisholmes, line 16: Should the following 2 statements also be covered by the conditional above?

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
```diff
@@ -16,6 +16,9 @@ def dispatch(name, request, response)
...
 
               add_rails_route(rack_span, request) if instrumentation_config[:enable_recognize_route]
+
+              rack_span.set_attribute('http.target', request.filtered_path)
```
ericmustin, line 20: I think this is good to have and I agree we should honor it if set.

But, I'd like to try to avoid calling `set_attribute` on a hot path if it isn't necessary, as it's an expensive operation since it calls `mutex.synchronize` under the hood.

Looking into the internals, i think we can safely avoid calling `set_attribute` if either `request.query_string.empty?` (pseudocode but i think this is the right method), or the `action_dispatch.parameter_filter` header is not present? 
https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/actionpack/lib/action_dispatch/http/filter_parameters.rb#L65

wdyt?

ericmustin (changes requested):
Thanks as always for the PR @chrisholmes . Generally i think this is the right thing to add, added some small feedback but broadly speaking i'm ok with having this be non-configurable, imo we should try to honor the User's rails settings as much as possible and they're defensiveness around sensitive params in logs should be extended to tracing if possible.

One thing i'm a bit unsure about is that technically the opentelemetry specification states

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md#common-attributes

```
Following attributes MUST be provided at span creation time (when provided at all), so they can be considered for sampling decisions:

http.method
http.url
http.target
http.host
http.scheme
```

I'm not sure if having a delta btwn the attribute value at span creation time in the sampler, and modifying it downstream, would violate the specification. I don't think it would (and it's experimental anyway so i think we should do whats best for rails/ruby), but would like to clarify at today's SIG mtg. I'll update this PR comment when i ask the maintainers what they think

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
chrisholmes: that sounds reasonable to me

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
chrisholmes: @fbogsany that makes sense and I will adjust to use `rack_span.recording?`

As I understand it, `rack_span.name` defaults to the request path. Should this be set to use the filtered path if the exception is present?

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
chrisholmes: I've implemented this check using a `filtered_path != fullpath` check, which should give us what we need and shouldn't set the attribute if now filters have been applied.

ericmustin: 👍 tyty

ericmustin (approved):
This lgtm

re: whether this violates the spec.

Dug into the spec a bit more and regarding sampling and updating span attributes i found this

(from: https://github.com/open-telemetry/opentelemetry-specification/blob/f5f6a7c8e56bb05c4c55ad0e0ea956d53c28aae8/specification/trace/api.md#set-attributes)
> Setting an attribute with the same key as an existing attribute SHOULD overwrite the existing attribute's value.

> Note that the OpenTelemetry project documents certain ["standard attributes"](https://github.com/open-telemetry/opentelemetry-specification/blob/f5f6a7c8e56bb05c4c55ad0e0ea956d53c28aae8/specification/trace/semantic_conventions/README.md) that have prescribed semantic meanings.

> Note that [Samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/f5f6a7c8e56bb05c4c55ad0e0ea956d53c28aae8/specification/trace/sdk.md#sampler) can only consider information already present during span creation. Any changes done later, including new or changed attributes, cannot change their decisions.


I think this demonstrates that updating attributes that were present at sampling time is permitted and is a path that hasn't been overlooked.

As a user, I think it does introduce tradeoffs, as it may make "debugging" why a sampling decision occurred to be difficult in a case where that sampling decision is based on an "unfiltered" http.target attribute in a custom sampler, but imo this is an issue that can be solved in a custom sampler (by adding debug logs to a user's custom sampler) or with additional logging in the upstream rails instrumentation if it actually becomes problematic. 

Additionally I think we shouldn't let a hypothetical issue block what's a practical improvement. If users reach out and indicate that they are having problems with this, we can add some additional debug logging or expose a config option to resolve this. I think for 99% of users it won't be an issue though, so approving.

not something i feel super duper strong about or anything so if ppl feel like this is a no go i'm open to addressing

robertlaurin (approved):

instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/patches/action_controller/metal.rb
fbogsany: `rack_span.name` will default to the request path, but in general, when an exception occurs, the `rack_span.name` will already have been updated to `<SomeClassName>#<some action>`. We want to avoid updating it again to `ExceptionController#<some action>` when an exception is present. I am not sure if there's a corner case where an exception occurs before the request is routed to a controller+action.

If the corner case exists (I'd want to see evidence of that first), then we might want to change the behaviour. It becomes a little challenging because `span.name` is not supposed to be readable (and isn't, with the API `Span`), so we can't simply parse it as a URL.

Issue #1088, opened 2022-01-18
OTLP Exporter - 1.0 tasks


robertlaurin:
- [ ] Remove TIMEOUT return stats https://github.com/open-telemetry/opentelemetry-ruby/pull/1087
- [ ] [Implement the OTEL_EXPORTER_OTLP_PROTOCOL env var ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md) 
- [x] Ensure protos are current https://github.com/open-telemetry/opentelemetry-ruby/pull/982
- [ ] #1404

PR #1127, opened 2022-02-23
chore: add OTEL_EXPORTER_OTLP_PROTOCOL and OTEL_EXPORTER_OTLP_TRACES_PROTOCOL env var

ericmustin:
Addresses https://github.com/open-telemetry/opentelemetry-ruby/issues/1088, by adding `'OTEL_EXPORTER_OTLP_TRACES_PROTOCOL', 'OTEL_EXPORTER_OTLP_PROTOCOL'` environment variable support for the otlp exporter (and defaulting+only supporting `http/protobuf`)

I think this is the last remaining open item before we can 1.0 the exporter.

The implementation here is kinda of pointless beside for specification purposes, since afaik our plan of action is to have multiple exporter gems for http/protobuf, grpc, http/json, (should we choose to implement those down the line), and there's really no need to expose `protocol` as an instance variable on the exporter at the moment, but it's possible it could be useful in the future, and it keeps the approach in line with the other configuration options.

At the very least what this does is it will loudly raise if the user tries to use a different protocol, which I think is useful, since we should make sure the user's knows what is/isn't possible as an export option for otlp at the moment.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -43,7 +43,7 @@ module OpenTelemetry
           end
         end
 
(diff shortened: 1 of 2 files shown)


ericmustin:
idk why this is giving me issue with cla, closing and cleaning up my local git config then pushing up again :(

PR #1129, opened 2022-02-24
feat: add OTEL_EXPORTER_OTLP_PROTOCOL and OTEL_EXPORTER_OTLP_TRACES_PROTOCOL env var

ericmustin:
trying this one last time...(hooray cla took)

Addresses https://github.com/open-telemetry/opentelemetry-ruby/issues/1088, by adding `'OTEL_EXPORTER_OTLP_TRACES_PROTOCOL', 'OTEL_EXPORTER_OTLP_PROTOCOL'` environment variable support for the otlp exporter (and defaulting+only supporting `http/protobuf`)

I think this is the last remaining open item before we can 1.0 the exporter.

The implementation here is kinda of pointless beside for specification purposes, since afaik our plan of action is to have multiple exporter gems for http/protobuf, grpc, http/json, (should we choose to implement those down the line), and there's really no need to expose `protocol` as an instance variable on the exporter at the moment, but it's possible it could be useful in the future, and it keeps the approach in line with the other configuration options.

At the very least what this does is it will loudly raise if the user tries to use a different protocol, which I think is useful, since we should make sure the user's knows what is/isn't possible as an export option for otlp at the moment.


ericmustin:
@plantfansam makes sense, added a todo about `METRICS` and also future work if/when we add other protocol options.

tsloughter:
Oh good, I was just working on this and went to comment on the issue when I saw this :). The only difference is I included 2 different exceptions depending on what the user supplied protocol was:

```ruby
          raise ArgumentError, "invalid protocol #{protocol}" if invalid_protocol?(protocol)
          raise ArgumentError, "unsupported protocol #{protocol}" if unsupported_protocol?(protocol)
```

```ruby
        def invalid_protocol?(protocol)
          protocol == 'http/protobuf' || protocol == 'grpc' || protocol == 'http/json'
        end

        def unsupported_protocol?(protocol)
          protocol != 'http/protobuf'
        end
```

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -53,6 +53,11 @@ def initialize(endpoint: config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_
           raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" if invalid_url?(endpoint)
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
+          # TODO: add OTEL_EXPORTER_OTLP_METRICS_PROTOCOL protocol env var support when metrics api/sdk is added.
+          # This may need to be moved to sdk/configurator.rb when grpc and http/json options are added, if added as unique gems.
```
tsloughter, line 57: Is it up in the air at all about whether to use separate gems? I'd argue for putting this in the SDK now if that is the plan. Then the code for selecting which gem's exporter to use could be included already as well -- I assume it would be supported to have multiple exporter gems included as deps, so `protocol` in the SDK would be used to select which to use, or error out if the user specifies to use one but hasn't included its gem.

ericmustin: sorry didnt see this comment. sure, yea i mean, i dont have strong opinions but i wasn't sure if we could 1.0 an exporter gem without explicitly supporting this option in the gem itself.

robertlaurin (approved):

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: We can 1.0 this gem without supporting the option. We want to stick the gRPC support in a separate gem, selected by this env var in the SDK. I.e. the code below belongs in the Configurator.

fbogsany (changes requested):

ericmustin:
(@robertlaurin is going to add this bit in the configurator to some other work being done in the short term, closing this for now)

PR #1133, opened 2022-03-01
feat: add hide_query_param option for http client instrumentations

ericmustin:
### Summary
This is a refresh/fresh-start of https://github.com/open-telemetry/opentelemetry-ruby/pull/946 (_which was originally an attempt to get https://github.com/open-telemetry/opentelemetry-ruby/pull/769 merged_). 

it adds `hide_query_params` as a configuration option for http client instrumentations, the default for this option is a boolean `true`, and when `true`, it removes the query parameters from the `http.url` and `http.target` http client attributes. Query params can often contain sensitive PII in typical production applications, like api keys, emails, phone numbers, etc, and so it makes sense to not include these out of the box.

It introduces this through a helper utility module `Common::HTTP::RequestAttributes` in the `common` package which ensures http client spans have consistent attributes, via the method `from_request`. `from_request` returns a hash of consistent http client attributes, and its meant to do so as performantly as possible (avoiding unnecessary use of `URI` when possible).

This has been stuck in a lot of back and forth and architecture discussion, but at this point i think it's easiest and least controversial to just use the `common` package as a place to add this feature.


plantfansam (commented):
This seems reasonable to me! I'm not familiar with our user base so not advocating either way on this, but do you think it's worth having the option to filter out individual query params (instead of the whole querystring)? So `https://google.com?foo=bar&api_key=wat` would yield `OpenTelemetry::SemanticConventions::Trace::HTTP_URL => "https://google.com?foo=bar`? Or perhaps a whitelist, for safety's sake....

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+            OpenTelemetry::SemanticConventions::Trace::HTTP_URL => hide_query_params(url || uri&.to_s, config),
+            OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => hostname || uri&.host,
+            OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => port || uri&.port
+          }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes).compact
+        end
+
+        private
+
+        def hide_query_params(url_or_path, config)
+          return url_or_path unless config.key?(:hide_query_params) && config[:hide_query_params]
```
plantfansam, line 27: I'm missing the probably obviously good reason to not do `return url_or_path unless config[:hide_query_params]`. Enlighten me?

common/test/opentelemetry/common/request_attributes_test.rb
```diff
@@ -0,0 +1,55 @@
...
+      'http.method' => 'GET',
+      'http.scheme' => 'http',
+      'http.url' => 'http://example.com/foo?bar=baz',
+      'http.target' => '/foo?bar=baz',
+      'net.peer.name' => 'example.com',
+      'net.peer.port' => 80
+    )
+  end
+
+  it 'returns http attributes matching the spec given input' do
```
plantfansam, line 24: Is this identical to the above example or have I not had enough coffee?

common/test/opentelemetry/common/request_attributes_test.rb
```diff
@@ -0,0 +1,55 @@
...
+        'http.method' => 'GET',
+        'http.scheme' => 'http',
+        'http.url' => 'http://example.com/foo?',
+        'http.target' => '/foo?',
+        'net.peer.name' => 'example.com',
+        'net.peer.port' => 80
+      )
+    end
+
+    it 'does not alter input uri' do
```
plantfansam, line 49: Maybe clearer to say `does not mutate uri instance provided as argument`? Took me a minute to track down why we were testing this.

instrumentation/faraday/README.md
```diff
@@ -30,6 +30,18 @@ OpenTelemetry::SDK.configure do |c|
...
 ```
 
+### Configuration options
+
+```ruby
+  c.use 'OpenTelemetry::Instrumentation::Faraday', {
+    # Query parameters are hidden by default.
+    # To disable, set hide_query_params to false.
+    hide_query_params: true,
```
plantfansam, line 40: ✨

instrumentation/http/test/instrumentation/http/patches/client_test.rb
```diff
@@ -115,5 +115,23 @@
         headers: { 'Traceparent' => "00-#{span.hex_trace_id}-#{span.hex_span_id}-01" }
       )
     end
+
+    it 'hide query params if enabled' do
```
plantfansam, line 119: `hides`?

instrumentation/http/test/instrumentation/http/patches/client_test.rb
```diff
@@ -115,5 +115,23 @@
...
+      stub_request(:get, 'http://example.com/success?foo=bar').to_return(status: 200)
+      ::HTTP.get('http://example.com/success?foo=bar')
+
+      _(exporter.finished_spans.size).must_equal(1)
+      _(span.attributes['http.target']).must_equal '/success?'
+    end
+
+    it 'show query params if disabled' do
```
plantfansam, line 127: `shows`?

instrumentation/http_client/lib/opentelemetry/instrumentation/http_client/patches/client.rb
```diff
@@ -10,21 +10,15 @@ module HttpClient
       module Patches
         # Module to prepend to HTTPClient for instrumentation
         module Client
+          include Common::HTTP::RequestAttributes
+
           private
 
-          def do_get_block(req, proxy, conn, &block) # rubocop:disable Metrics/AbcSize
```
plantfansam, line 15: ✨

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
```diff
@@ -11,20 +11,35 @@ module HTTP
...
-              attributes = {
-                OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => req.method,
-                OpenTelemetry::SemanticConventions::Trace::HTTP_SCHEME => USE_SSL_TO_SCHEME[use_ssl?],
-                OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET => req.path,
-                OpenTelemetry::SemanticConventions::Trace::NET_PEER_NAME => @address,
-                OpenTelemetry::SemanticConventions::Trace::NET_PEER_PORT => @port
-              }.merge!(OpenTelemetry::Common::HTTP::ClientContext.attributes)
+              method = req.method
+
+              options = if req.uri
```
plantfansam, line 25: Might be worth explaining why a req may/may not have `.uri` in a comment here

chrisholmes:
@ericmustin anything I can do to help move this forward?

ahayworth (commented):
I'm not sure if I've cooked up an unrealistic test or not; but at least Ruby's `URI` seems easily confused:

```ruby
irb(main):001:0> u = URI("https://evil\?:password@host.com/foo?bar=baz").to_s
=> "https://evil?:password@host.com/foo?bar=baz"
irb(main):002:0> u[0, u.index('?') + 1]
=> "https://evil?"
irb(main):003:0> u = URI("https://evil\?:password@host.com/foo?bar=baz").request_uri
=> "/?:password@host.com/foo?bar=baz"
irb(main):004:0> u[0, u.index('?') + 1]
=> "/?"
irb(main):005:0>
```

ericmustin:
@ahayworth I'm confused on what's being demonstrated

For:

> irb(main):003:0> u = URI("https://evil\?:password@host.com/foo?bar=baz").request_uri
=> "/?:password@host.com/foo?bar=baz"

I think it's correctly interpreting the first question mark as the start of the query component. 

https://www.rfc-editor.org/rfc/rfc3986#section-3.4

> The query component contains non-hierarchical data that, along with
   data in the path component ([Section 3.3](https://www.rfc-editor.org/rfc/rfc3986#section-3.3)), serves to identify a
   resource within the scope of the URI's scheme and naming authority
   (if any).  The query component is indicated by the first question
   mark ("?") character and terminated by a number sign ("#") character
   or by the end of the URI.

ericmustin:
@chrisholmes ive surfaced this to the approvers and maintainers a few times, i will bring it up again at todays mtg. There are some style comments here i need to address it looks like.

common/test/opentelemetry/common/request_attributes_test.rb
ericmustin: it was i who needed more coffee

instrumentation/net_http/lib/opentelemetry/instrumentation/net/http/patches/instrumentation.rb
ericmustin: https://github.com/open-telemetry/opentelemetry-ruby/pull/946#discussion_r716972224 will add a comment

ericmustin:
@ahayworth The RFC I linked to says that the first `?` is automatically treated as the start of the query component, and thus also the start of the path. I think everything is working fine.

common/lib/opentelemetry/common/http/request_attributes.rb
ericmustin: I don't know, it's been awhile. I assume there was some intention here but from a glance I have no idea what it was. Let me look into it

instrumentation/http/test/instrumentation/http/patches/client_test.rb
ericmustin: updated

common/test/opentelemetry/common/request_attributes_test.rb
ericmustin: updated

common/test/opentelemetry/common/request_attributes_test.rb
ericmustin: updated (removed dupe). good catch

instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
```diff
@@ -41,21 +42,32 @@
...
 
     it 'has http 200 attributes' do
-      response = client.get('/success')
+      response = client.get('/success?foo=bar')
 
       _(span.name).must_equal 'HTTP GET'
       _(span.attributes['http.method']).must_equal 'GET'
       _(span.attributes['http.status_code']).must_equal 200
-      _(span.attributes['http.url']).must_equal 'http://example.com/success'
+      _(span.attributes['http.url']).must_equal 'http://example.com/success?'
```
fbogsany, line 55: IDK whether this was hashed out on a previous PR... I feel like we should drop the `?` here as well. Alternatively, append something to make it clear the query parameters have been redacted - e.g. `?<redacted>`.

instrumentation/http/lib/opentelemetry/instrumentation/http/patches/client.rb
```diff
@@ -10,18 +10,13 @@ module HTTP
...
+
+          def perform(req, options)
             uri = req.uri
             request_method = req.verb.to_s.upcase
 
-            attributes = {
-              'http.method' => request_method,
-              'http.scheme' => uri.scheme,
-              'http.target' => uri.path,
-              'http.url' => "#{uri.scheme}://#{uri.host}",
```
fbogsany, line 21: This attribute was not checked in the tests, and I *think* its value will change with this PR. We should add the relevant expectations to the tests and then fix the code, if necessary.

fbogsany: In fact, this doesn't match the semantic conventions, so we should definitely add the expectations - the old code would fail.

common/opentelemetry-common.gemspec
```diff
@@ -27,6 +27,13 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
 
+  # Duplicating the approach of the SDK.
+  # This is an intentionally loose dependency, since we want to be able to
+  # release new versions of opentelemetry-semantic_conventions without requiring
+  # a new SDK release. The requirements of the SDK have been satisifed since the
```
fbogsany, line 33:
```suggestion
  # a new release of this gem. The requirements of this gem have been satisfied since the
```

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Common
+    module HTTP
+      # RequestAttributes contains instrumentation helpers for http instrumentation client requests
+      module RequestAttributes
+        QUERY_PARAM_START_KEY = '?'
+        def from_request(method:, config: {}, uri: nil, scheme: nil, target: nil, url: nil, hostname: nil, port: nil)
```
fbogsany, line 13: Can we also add `host: nil` here? The semantic conventions say this should contain the value of the HTTP host header. While there are other things we could add, the `http.host` attribute is one of the attributes that

> **MUST** be provided **at span creation time** (when provided at all), so they can be considered for sampling decisions

It would be great if we could add `ip: nil` as well. IDK if that's available in any of the HTTP client gems, though.

common/lib/opentelemetry/common/http/request_attributes.rb
```diff
@@ -0,0 +1,38 @@
...
+          {
+            OpenTelemetry::SemanticConventions::Trace::HTTP_METHOD => method,
+            OpenTelemetry::SemanticConventions::Trace::HTTP_SCHEME => scheme || uri&.scheme,
+            OpenTelemetry::SemanticConventions::Trace::HTTP_TARGET => hide_query_params(target || uri&.request_uri, config),
+            OpenTelemetry::SemanticConventions::Trace::HTTP_URL => hide_query_params(url || uri&.to_s, config),
```
fbogsany, line 18: Rather than serializing the `uri` and then searching for `?` and stripping the tail, if the `uri` is present, it might be more efficient to:
```ruby
uri = uri.dup
uri.query = nil
uri.to_s
```
or:
```ruby
uri = uri.dup
uri.query = nil
uri.request_uri
```
as appropriate.

Obviously, we also need to handle the case where `uri` is not present.

ericmustin:
This PR has fallen pretty far out of date and don't plan on picking it back up in the short term, IIRC this was my 3rd or 4th iteration of trying to get this feature out the door and little to show for it, probably makes sense to start from scratch for anyone interested.

PR #1137, opened 2022-03-02
feat: Added Google Cloud Function Resource Detection

### LINKED ISSUE #1124: Missing resource detector for Google Cloud Functions
There is already support for resource detection of Google Compute Engine(gce) and Google Kubernetes Engine(gke).
But, resource detection for google cloud function is missing as part of GCP.

I would be happy to contribute if this requirement is valid and needs work.

khyatigandhi6:
Add resource detector for Google Cloud Function.
Closes #1124

### DIFF
--- a/resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
+++ b/resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
@@ -13,7 +13,7 @@ module OpenTelemetry
       module GoogleCloudPlatform
         extend self
 
-        def detect # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength
+        def detect # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
           gcp_env = Google::Cloud::Env.new
           resource_attributes = {}
 
@@ -38,11 +38,28 @@ module OpenTelemetry
             resource_attributes[OpenTelemetry::SemanticConventions::Resource::CONTAINER_NAME] = ENV['CONTAINER_NAME']
           end
 
+          if gcp_env.knative?
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_PROVIDER] = 'gcp'
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_ACCOUNT_ID] = gcp_env.project_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_NAME] = gcp_env.knative_service_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_VERSION] = gcp_env.knative_service_revision
+            zone = gcp_env.instance_zone
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_REGION] = get_region zone
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_AVAILABILITY_ZONE] = zone
+          end
+
           resource_attributes.delete_if { |_key, value| value.nil? || value.empty? }
           OpenTelemetry::SDK::Resources::Resource.create(resource_attributes)
         end
 
         private
+
+        def get_region(zone)
+          return if zone.nil? || zone.empty?
+
+          split_arr = zone.split('-', 3)
+          split_arr[0].concat('-', split_arr[1])
+        end
 
         def safe_gethostname
           Socket.gethostname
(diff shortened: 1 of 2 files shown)


ericmustin (commented):
Broadly lgtm, one small nice to have i mentioned when comparing this to what OpenTelemetry-Go has done for it's cloud run resource detector, but not blocking.

@dazuma Do you have any input here or any feedback? I am deeply familiar with Cloud Functions so if you notice anything missing here or incorrect, it would be appreciated.

ericmustin:
cc @dashpole if you have any qualitative feedback on what resource attributes to include here from the GCP perspective.

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
```diff
@@ -38,12 +38,29 @@ def detect # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metr
...
           end
 
+          if gcp_env.knative?
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_PROVIDER] = 'gcp'
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_ACCOUNT_ID] = gcp_env.project_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_NAME] = gcp_env.knative_service_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_VERSION] = gcp_env.knative_service_revision
+            zone = gcp_env.instance_zone
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_REGION] = get_region zone
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_AVAILABILITY_ZONE] = zone
```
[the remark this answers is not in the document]
dashpole: I think that is getting `service.instance.id`, rather than `faas.instance`.

dashpole:
This looks like the right set of attributes to provide to me

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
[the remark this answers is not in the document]
ericmustin: ah, woops, i am going to resolve this as i'm not sure there's much value to collecting `service.instance.id` at this time (if it proves to be valuable we can add in a separate pr)

ericmustin (approved):
thanks for the contribution @khyatigandhi6 , nice work!

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
```diff
@@ -38,12 +38,29 @@ def detect # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metr
             resource_attributes[OpenTelemetry::SemanticConventions::Resource::CONTAINER_NAME] = ENV['CONTAINER_NAME']
           end
 
+          if gcp_env.knative?
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_PROVIDER] = 'gcp'
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_ACCOUNT_ID] = gcp_env.project_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_NAME] = gcp_env.knative_service_id
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::FAAS_VERSION] = gcp_env.knative_service_revision
+            zone = gcp_env.instance_zone
+            resource_attributes[OpenTelemetry::SemanticConventions::Resource::CLOUD_REGION] = get_region zone
```
robertlaurin, line 47: `get_region` here can return nil, so that means there's scenario where users of this detector will receive a [warning message](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/internal.rb#L47-L58) without the ability to resolve it other than patching this code path.  Let's make sure we only set this key if there is a non-nil value.

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
robertlaurin, line 48: Some issue around possibly setting this as a nil value.

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
khyatigandhi6: Yes if `get_region` returns nil then that field won't get set. This is handled by below piece of code
`resource_attributes.delete_if { |_key, value| value.nil? || value.empty? }`

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
khyatigandhi6: Similarly, if zone is nil then it won't get set. 
`resource_attributes.delete_if { |_key, value| value.nil? || value.empty? }`

resource_detectors/lib/opentelemetry/resource/detectors/google_cloud_platform.rb
robertlaurin: Ah, you're right.  I missed that.

robertlaurin (approved):

mayur-more:
Nice work, @khyatigandhi6

khyatigandhi6:
Thank you! Looking forward to contribute more.

PR #1140, opened 2022-03-03
attribute length limit: only truncate strings and strings in arrays

tsloughter:
This set of patches both adds support for truncating strings in arrays but also tests that a variable is a string before checking its size and truncating. This fixes a bug that an attribute value of type float would crash if `attribute_length_limit` were configured.

### DIFF
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -72,6 +72,17 @@ module OpenTelemetry
         string.is_a?(String) && string.size > size ? "#{string[0...size - 3]}..." : string
       end
 
+      def truncate_attribute_value(value, limit)
+        case value
+        when Array
+          value.map { |x| truncate(x, limit) }
+        when String
+          truncate(value, limit)
+        else
+          value
+        end
+      end
+
       def untraced
         OpenTelemetry::Trace.with_span(OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new)) { yield }
       end
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -69,13 +69,13 @@ module OpenTelemetry
       #
       # @return [String]
       def truncate(string, size)
-        string.is_a?(String) && string.size > size ? "#{string[0...size - 3]}..." : string
+        string.size > size ? "#{string[0...size - 3]}..." : string
       end
 
       def truncate_attribute_value(value, limit)
         case value
         when Array
-          value.map { |x| truncate(x, limit) }
+          value.map { |x| truncate_attribute_value(x, limit) }
         when String
           truncate(value, limit)
         else
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -69,7 +69,7 @@ module OpenTelemetry
       #
       # @return [String]
       def truncate(string, size)
-        string.size > size ? "#{string[0...size - 3]}..." : string
+        string.is_a?(String) && string.size > size ? "#{string[0...size - 3]}..." : string
       end
 
       def untraced
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -347,11 +347,22 @@ module OpenTelemetry
           nil
         end
 
+        def truncate(value, limit)
+          case value
+              when Array
+                value.map { |x| OpenTelemetry::Common::Utilities.truncate(x, limit) }
+              when String
+                OpenTelemetry::Common::Utilities.truncate(value, limit)
+              else
+                value
+          end
+        end
+
         def truncate_attribute_values(attrs)
           return EMPTY_ATTRIBUTES if attrs.nil?
 
           attribute_length_limit = @span_limits.attribute_length_limit
-          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate(value, attribute_length_limit) } if attribute_length_limit
+          attrs.each { |key, value| attrs[key] = truncate(value, attribute_length_limit) } if attribute_length_limit
           attrs
         end
 
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -353,9 +353,7 @@ module OpenTelemetry
           attribute_length_limit = @span_limits.attribute_length_limit
           return attrs if attribute_length_limit.nil?
 
-          attrs.each do |key, value|
-            attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) if value.is_a?(String)
-          end
+          attrs.transform_values! { |value| OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) }
           attrs
         end
 
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -351,7 +351,11 @@ module OpenTelemetry
           return EMPTY_ATTRIBUTES if attrs.nil?
 
           attribute_length_limit = @span_limits.attribute_length_limit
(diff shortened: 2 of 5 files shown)


common/lib/opentelemetry/common/utilities.rb
```diff
@@ -69,7 +69,7 @@ def utf8_encode(string, binary: false, placeholder: STRING_PLACEHOLDER)
       #
       # @return [String]
       def truncate(string, size)
-        string.size > size ? "#{string[0...size - 3]}..." : string
+        string.is_a?(String) && string.size > size ? "#{string[0...size - 3]}..." : string
```
arielvalentin, line 72: I see that there is a type check in the case statement where this function is used. 

Is truncate being used in other places where it does not receive a String? If not then I do not think this needs to change.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -347,11 +347,22 @@ def trim_span_attributes(attrs)
           nil
         end
 
+        def truncate(value, limit)
+          case value
```
arielvalentin, line 351: would it make sense to move this logic into `Utilities.truncate` and then potentially dispatch to different methods based on that type if necessary?

common/lib/opentelemetry/common/utilities.rb
tsloughter: It is called elsewhere, yes, including in instrumentation libraries. So I figured it was better to not require the check be added everywhere. The reason it is also checked in that other case statement is only because there is an additional step to do if the value is an array, I figured if it was already checking if the value was an array there it may as well not even call the method if it wasn't a string.

sdk/lib/opentelemetry/sdk/trace/span.rb
tsloughter: Hm, yes, most likely. I'll make that change.

plantfansam:
Thanks so much for taking this on. I was looking at this the other day too 😂 . Gonna dump some of my research here.

To cover our bases, the [values permitted for an attribute](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute) are `string, boolean, double precision floating point (IEEE 754-1985) or signed 64 bit integer`.

You are [only supposed to truncate strings](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/common.md#attribute-limits) and strings within arrays. Otherwise, you just leave the too-long value in place:

> set an attribute value length limit such that for each attribute value:
> - if it is a string, if it exceeds that limit (counting any character in it as 1), SDKs MUST truncate that value, so that its length is at most equal to the limit,
> - if it is an array of strings, then apply the above rule to each of the values separately,
> otherwise a value MUST NOT be truncated;

SDKs are allowed to log out a configurable warning if they're truncating. I think that logging a warning if we _don't_ truncate a too-big integer or float is in the spirit of the spec, but not the letter of it (`"There MAY be a log emitted to indicate to the user that an attribute was truncated or discarded. To prevent excessive logging, the log MUST NOT be emitted more than once per record on which an attribute is set.`"). But actually I think it's a pretty rare case, so probably not worth introducing log noise.

Anyway tl;dr thank you so much for doing this; could we also add a check for `Float`?

sdk/lib/opentelemetry/sdk/trace/span.rb
tsloughter: Updated.

tsloughter:
@plantfansam thanks, but what would the float check be for? There is no truncating to do so it shouldn't need to know if it is a float, just that it isn't a string or array.

common/lib/opentelemetry/common/utilities.rb
fbogsany: This kind of defensive programming is not typical in Ruby. The method docs state that the first argument is a `String`. Callers should ensure that they only call it with a `String`. We shouldn't be type-checking that argument here.

If another caller is calling it with invalid arguments, that should be considered a bug in the caller, and we should fix the caller.

tsloughter-splunk: Ok, I can change it, but that is also true in Erlang and we still do checks like this more so than normal because as a library for telemetry we want to be as silent and unseen as possible. So I thought it might be similar in otel ruby.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -351,7 +351,7 @@ def truncate_attribute_values(attrs)
           return EMPTY_ATTRIBUTES if attrs.nil?
 
           attribute_length_limit = @span_limits.attribute_length_limit
-          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate(value, attribute_length_limit) } if attribute_length_limit
+          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) } if attribute_length_limit && value.is_a?(String)
```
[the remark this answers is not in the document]
fbogsany: I think it's correct as is if we remove the `&& value.is_a?(String)`
```suggestion
          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) } if attribute_length_limit
```
We type check inside `truncate_attribute_value`. It might be slightly more efficient to check before the call 🤷‍♂️ .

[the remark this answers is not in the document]
tsloughter: Oh wait @fbogsany , hehe, I mixed up my two PRs. I think you are right as well.

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -351,7 +351,7 @@ def truncate_attribute_values(attrs)
           return EMPTY_ATTRIBUTES if attrs.nil?
 
           attribute_length_limit = @span_limits.attribute_length_limit
-          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate(value, attribute_length_limit) } if attribute_length_limit
+          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) if attribute_length_limit && value.is_a?(String) }
```
fbogsany, line 354: This is getting more complicated than a one-liner is worth.
```suggestion
          return attrs if attribute_length_limit.nil?

          attrs.each do |key, value|
            attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) if value.is_a?(String)
          end
```

fbogsany (approved):

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -351,7 +351,11 @@ def truncate_attribute_values(attrs)
           return EMPTY_ATTRIBUTES if attrs.nil?
 
           attribute_length_limit = @span_limits.attribute_length_limit
-          attrs.each { |key, value| attrs[key] = OpenTelemetry::Common::Utilities.truncate(value, attribute_length_limit) } if attribute_length_limit
+          return attrs if attribute_length_limit.nil?
+
+          attrs.each do |key, value|
+            attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) if value.is_a?(String)
```
fbogsany, line 357: 🤦 We missed the `Array` case. IMO there's really little value in checking for either `Array` or `String` here, and also inside `truncate_attribute_value`.
```suggestion
            attrs[key] = OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit)
```

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -351,7 +351,11 @@ def truncate_attribute_values(attrs)
...
+          end
```
fbogsany, line 358: And, in fact, we may as well simplify this to:
```suggestion
          attrs.transform_values! { |value| OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) }
```

tsloughter: Nice, but doesn't it still need the ` if value.is_a?(String)`?

fbogsany: No, because that'll miss the `Array` case, and there's diminishing returns to `if value.is_a?(String) || value.is_a?(Array)` over just dealing with it in `truncate_attribute_value`.

tsloughter: Hehe, like we already discussed and I mixed up :)

PR #1141, opened 2022-03-03
fix warning about Struct initialization in Ruby 3.2+

tsloughter:
I was seeing a lot of warnings like:

```
opentelemetry-ruby/sdk/test/opentelemetry/sdk/trace/export/console_span_exporter_test.rb:14: warning: Passing only keyword arguments to Struct#initialize will behave differently from Ruby 3.2. Please use a Hash literal like .new({k: v}) instead of .new(k: v).
```

on Ruby 3.1 and this patch clears them up.

### DIFF
--- a/sdk/.rubocop.yml
+++ b/sdk/.rubocop.yml
@@ -16,3 +16,5 @@ Naming/FileName:
     - "lib/opentelemetry-sdk.rb"
 Style/ModuleFunction:
   Enabled: false
+Style/BracesAroundHashParameters:
+  Enabled: false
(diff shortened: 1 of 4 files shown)


plantfansam:
Thanks for tackling this! Not sure what Rubocop is yelling about, and the logs in CI seem to cut off during gem install ...mind pushing up an empty commit to try to get better logs?

robertlaurin:
The cop doesn't like the spacing
```
test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb:13:64: C: Layout/SpaceInsideHashLiteralBraces: Space inside { missing.
  let(:span_data2)   { OpenTelemetry::SDK::Trace::SpanData.new({name: 'name2'}) }
```

Nor does it like your curly braces apparently :) 
```
test/opentelemetry/sdk/trace/export/in_memory_span_exporter_test.rb:13:64: C: Style/BracesAroundHashParameters: Redundant curly braces around a hash parameter.
  let(:span_data2)   { OpenTelemetry::SDK::Trace::SpanData.new({name: 'name2'}) }
```

robertlaurin:
Feel free to add the following to the sdk rubocop yaml to get it to quiet down so we can unblock this.
```rb
Style/BracesAroundHashParameters:
  Enabled: false
```

tsloughter:
Thanks @robertlaurin , I'll get that fixed

PR #1143, opened 2022-03-10
[active_job] Fix deserialization of jobs that are missing metadata

hosamaly:
We may get jobs from applications that do not have OpenTelemetry instrumentation, such as during a rolling deployment. In such cases, jobs failed to deserialize with the following error:
```
ActiveJob::DeserializationError: Error while trying to deserialize arguments: undefined method `map' for nil:NilClass
    ~/.gem/ruby/2.7.5/gems/activejob-7.0.2.3/lib/active_job/arguments.rb:42:in `deserialize'
    ~/.gem/ruby/2.7.5/gems/activejob-7.0.2.3/lib/active_job/core.rb:190:in `deserialize_arguments'
    ~/opensource/opentelemetry-ruby/instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/base.rb:30:in `deserialize'
    ~/opensource/opentelemetry-ruby/instrumentation/active_job/test/instrumentation/active_job/patches/base_test.rb:40:in `block (3 levels) in <top (required)>'
```

This pull request fixes this issue.


ericmustin:
Good catch, 

- `deserialize_arguments` calls `Arguments.deserialize` without checking if the argument is nil, https://github.com/rails/rails/blob/3bfe4191a39f064dc5fd2fc37a2cc85f070ad4e9/activejob/lib/active_job/core.rb#L189-L191

- and `Arguments.deserialize` then calls `.map` also making the same assumptions about the argument being an array. https://github.com/rails/rails/blob/3bfe4191a39f064dc5fd2fc37a2cc85f070ad4e9/activejob/lib/active_job/arguments.rb#L42

This PR should help us be more defensive here.

ericmustin (approved):

robertlaurin (approved):
Thanks for the fix @hosamaly

PR #1145, opened 2022-03-11
[active_job] Make the context available in ActiveJob notifications

hosamaly:
Previously, subscribers to `'perform.active_job'` events had no access to the tracing context. Only the job and its callbacks had access to it.

This was a result of adding OpenTelemetry's instrumentation callback after the notifications instrumentation that is configured in `ActiveJob::Logging` (in ActiveJob 6.x).

This pull request fixes the issue by prepending OpenTelemetry's instrumentation callback before all others.

In ActiveJob v7, however, notifications instrumentation happens before any callbacks are called, so the current solution doesn't work, sadly.


ahayworth:
@hosamaly Do you have any thoughts on how we might make this work for ActiveJob v6 and v7? I recognize how useful this would be, but I would hate to create a situation where users of the library could come to rely on this behavior on ActiveJob v6 only to have a regression in functionality when they get around to upgrading to ActiveJob v7. 😢

hosamaly:
@ahayworth: I made another attempt at fixing this. It should now work with all versions of ActiveJob.

Instead of using a callback to propagate the context, we now override `ActiveJob::Base#perform_now`, which is the entrypoint to execution logic. It is called right after the job is deserialized.

This is the same pattern that is used in `ActiveJob::Instrumentation` in v7. Instead of using a callback as it did in previous versions, the instrumentation in v7 [overrides `#perform_now`](https://github.com/rails/rails/pull/40927/files#diff-41ecb8d7126485e207a879c9e3e28f73837309b8370d0db939a23874b97f66b1L12).

I left the indentation looking a bit ugly to keep the file's history intact. Feel free to correct it if you feel it must be.

ahayworth:
> @ahayworth: I made another attempt at fixing this. It should now work with all versions of ActiveJob.

Wonderful!

> I left the indentation looking a bit ugly to keep the file's history intact. Feel free to correct it if you feel it must be.

Thanks - I appreciate that, but I did go ahead and push a whitespace-fixing commit. Since most of this is my ~fault~ handiwork anyways, the appropriate git-blame information is mostly preserved this way.

Conveniently, that will also allow CI to run - the job seemed stuck in a `queued` state despite my clicking the "approve" button for CI to run. That's why I closed/re-opened the PR; I was trying to kick off the jobs again. But no matter, fixing the whitespace also does that. 😄

instrumentation/active_job/lib/opentelemetry/instrumentation/active_job/patches/active_job_callbacks.rb
```diff
@@ -21,45 +21,47 @@ def self.prepended(base) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
...
               end
+            end
+          end
 
-              around_perform do |job, block| # rubocop:disable Metrics/BlockLength
-                span_kind = job.class.queue_adapter_name == 'inline' ? :server : :consumer
-                span_name = "#{otel_config[:span_naming] == :job_class ? job.class : job.queue_name} process"
-                span_attributes = job_attributes(job).merge('messaging.operation' => 'process')
+          def perform_now # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+            job = self
```
ahayworth, line 28: @hosamaly Was this just so that we could avoid changing the references to the various attributes below (eg `job.class`, `job.queue_name`)?

If so - do you mind if I just update the references to `self` instead of setting `job = self` ?

hosamaly: Indeed, I was trying to avoid modifying lines that didn't need to be modified. But feel free to update it as you see fit.

ahayworth: Cool! I will, thank you 😄

ahayworth (approved):

plantfansam (approved):
Wonderful! Appreciate the work that you did to make this compatible with ActiveJob 7!

robertlaurin (approved):

Issue #1139, opened 2022-03-03
PG Instrumentation: Invalid span attribute value type Arel::Nodes::SqlLiteral


danielmbarlow:
## Problem 
Our logs are filled with errors like this:

```
OpenTelemetry error: invalid span attribute value type Arel::Nodes::SqlLiteral for key 'db.statement' on span 'SELECT <database name here>'
```
I expected to just see the statement in the trace (in Jaeger in our case), but it seems to be excluded.

### Runtime
```
ruby -v
=> ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-linux]
```

## How to reproduce
This works:
```
ApplicationRecord.connection.execute("Select 'test'")
```

In Jaeger we get:
![image](https://user-images.githubusercontent.com/7119502/156595470-001c0b06-da65-4ab2-9a5d-28fbccb6f791.png)

This doesn't work:
```
# This is valid code and executes the statement fine
ApplicationRecord.connection.execute(Arel.sql("Select 'test'"))
=> {"?column?"=>"test"}

# ... but console logs show this
=> E, [2022-03-03T16:25:32.573458 #52] ERROR -- : OpenTelemetry error: invalid span attribute value type Arel::Nodes::SqlLiteral for key 'db.statement' on span 'SELECT ***REDACTED**'
```

In Jaeger the `db.statement` tag is missing:

![image](https://user-images.githubusercontent.com/7119502/156596509-d4399a6a-f10a-48e0-bcb9-2508124bba26.png)

### Potential solution 

Maybe the gem could be changed to use `#to_s`:
```
Arel.sql("Select 'test'").to_s
=> "Select 'test'"
```

### Potential Workarounds
* Perhaps use `db_statement: :omit` to reduce log noise
* Use `#to_s` everywhere

ericmustin:
Thanks for the report @danielmbarlow.

It looks like this is the problematic line in our PG instrumentation https://github.com/open-telemetry/opentelemetry-ruby/blob/ad51972e9bdc8ead0e69642056a4935683464dfc/instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb#L87

[...]

PR #1146, opened 2022-03-12
fix: update pg instrumentation to handle non primitive argument

ericmustin:
### Summary

Addresses https://github.com/open-telemetry/opentelemetry-ruby/issues/1139

the pg connection_adapter `.execute` method can be invoked with arguments that are not a  `String` class, like a `Arel::Nodes::SqlLiteral`.

While [`Arel::Nodes::SqlLiteral` is a `String` subclass](https://github.com/rails/rails/blob/f314bae561f7d1d8ff7606ee417fcf0bc55d5ab0/activerecord/lib/arel/nodes/sql_literal.rb#L5), Our otlp exporter doesn't gracefully handle Subclasses of accepted attribute types ([see here](https://github.com/open-telemetry/opentelemetry-ruby/blob/00c3600b99cb4e10a70b6d4625fa40bb7ee51866/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L362-L386)), and instead just returns an error and drops the attribute when attempting to encode a non valid type.

I'm hesitant to fiddle with the portion of our OTLP exporter code that handles encoding, as we'd have to be careful not to introduce a regression.

Instead, since we _know_ which instrumentation is problematic and what attributes, it's more straightforward to just coerce that specific attribute to a`String` class, which is what this PR does (it calls `.to_s` defensively).


instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -84,7 +84,7 @@ def span_attrs(kind, *args) # rubocop:disable Metrics/AbcSize
             end
 
             attrs = { 'db.operation' => validated_operation(operation), 'db.postgresql.prepared_statement_name' => statement_name }
-            attrs['db.statement'] = sql unless config[:db_statement] == :omit
+            attrs['db.statement'] = sql.to_s if config[:db_statement] != :omit && !sql.nil? && sql.respond_to?(:to_s)
```
ahayworth, line 87: While I do ❤️ defensive programming in general, I'm wondering if we need to be _quite_ this defensive. For my own edification, I'm wondering what scenarios we're trying to avoid?

To explain why I'm curious - anything inheriting from `Object` should always have `to_s` defined, so only things explicitly inheriting from `BasicObject` would fail the `respond_to?` check. And, even if `sql` is `nil` (which _could_ happen, depending on how the interactions with the `lru_cache` go) - `to_s` is still defined, although admittedly a blank string as an attribute value seems useless.

It's totally fine either way and I'm approving the PR regardless, but I was just curious what the logic was behind the defensiveness.

ahayworth (approved):
I had a question re: the defensive programming, but I don't think it's anything worth blocking the PR over.

This is a good fix - and thank you to @danielmbarlow for finding the issue! As an avid `Arel.sql` fan myself, we should have been testing for this from the beginning. 🤦 😆

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ericmustin: the nil check is to prevent `""` as an attribute key. agreed the  `to_s` is probably unnecessary. will update

ahayworth: 👍 Cool, thanks for clarifying. I do appreciate it! 😄

arielvalentin, line 87: Would safe navigation do the trick? 

```suggestion
            attrs['db.statement'] = sql.to_s if config[:db_statement] != :omit && sql&.to_s
```

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -84,7 +84,7 @@ def span_attrs(kind, *args) # rubocop:disable Metrics/AbcSize
             end
 
             attrs = { 'db.operation' => validated_operation(operation), 'db.postgresql.prepared_statement_name' => statement_name }
-            attrs['db.statement'] = sql unless config[:db_statement] == :omit
+            attrs['db.statement'] = sql.to_s if config[:db_statement] != :omit && sql&.to_s
```
plantfansam, line 87: Nit, but: thoughts on changing to `attrs['db.statement'] = sql&.to_s if config[:db_statement] != :omit && !sql.nil?` or just `attrs['db.statement'] = sql&.to_s if config[:db_statement] != :omit`

plantfansam (commented):
👌  lovely! Had a in-line nit that you can take or leave as you wish 😄

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
ericmustin: i've updated this to `attrs['db.statement'] = sql&.to_s unless config[:db_statement] == :omit`

I think this is as concise as i can get and handles all the edge cases and doesnt anger rubocop

instrumentation/pg/lib/opentelemetry/instrumentation/pg/patches/connection.rb
```diff
@@ -84,7 +84,7 @@ def span_attrs(kind, *args) # rubocop:disable Metrics/AbcSize
             end
 
             attrs = { 'db.operation' => validated_operation(operation), 'db.postgresql.prepared_statement_name' => statement_name }
-            attrs['db.statement'] = sql unless config[:db_statement] == :omit
+            attrs['db.statement'] = sql&.to_s unless config[:db_statement] == :omit
```
fbogsany, line 87: Is there a circumstance where `sql` would be `nil`? That would have been problematic previously, since attribute values have to be non-nil, and validation would have dropped it and logged an error.

fbogsany: Note: I realize this has been beaten to death, but I really don't think this level of defensiveness is warranted here.

ericmustin: This test case 
https://github.com/open-telemetry/opentelemetry-ruby/blob/06b36a646afb30e86ad426f1be2bc97ed269fa15/instrumentation/pg/test/opentelemetry/instrumentation/pg/instrumentation_test.rb#L137-L138

WIthout the defensiveness it seems it's possible that lru_cache returns nil, which then gets encoded to `""` by `.to_s`. This failed test case is what prompted the defensiveness in the first place https://github.com/open-telemetry/opentelemetry-ruby/runs/5524602649?check_suite_focus=true

I hadn't really looked into the lru cache details or instrumentation implementation other than that it does seem possible, in some cases, for `sql` to be nil, we are currently testing for that case for some reason, and so we should have a `nil` check.

fbogsany: It feels like that's more a case of the cache not working correctly. If the `lru_cache` returns `nil`, we should handle that cache miss and update the cache rather than papering over the problem with a `&.`.

fbogsany: The `lru_cache` is a little unusual here - the expectation is that we write into the cache in the `PREPARE` case and read from it in the `EXECUTE` case. I think we should handle the `sql.nil?` case on line 81, where we read from the cache. By the time we get here, `sql` should be non-`nil`.

ericmustin: I think the idea, (@ahayworth correct me if i'm wrong) is that this specific case is for handling prepared statements. Reviewing the test here, w/an lru cache of size `50`:

- user prepares 51 statements
- each time `prepare` is called, we perform the obfuscation and cache `<statement-name>:<obfuscated-prepared-statement>` k:v pair.
- The 51st call evicts the least recently used `<statement-name>:<obfuscated-prepared-statement>` k:v pair
- when the user later attempts to execute that already-evicted `<statement-name>:<obfuscated-prepared-statement>` k:v pair via `.execute(<statement-name>)`, it causes a cache miss _and_ we can't update the cache because we no longer have access to the `<obfuscated-prepared-statement>` to set as a value, so the lru_cache returns `nil`.

I don't know what the improvement would be here, this seems like an expected behavior, albeit one with tradeoffs.

fbogsany: 🤦 we handle `nil` attributes on the next line. This is complicated and confusing IMO. I think we should `.to_s` the result from `obfuscate_sql`, either in that method or on lines 72 and 77. Then we always cache strings.

ericmustin: ok

ahayworth: @ericmustin your understanding of the LRU cache is correct, both in functionality and in purpose.

fbogsany (approved):

plantfansam (approved):
LGTM

PR #1150, opened 2022-03-17
Add ability to report set/retrieved value size in bytes in Redis instrumentation

plantfansam:
~~This patch adds a configuration option that, when set, collects the number of bytes sent and received in an individual Redis `process` call.~~

~~The implementation here is a little clunky, but I did my best to avoid allocating too many strings / re-invoking the same methods.~~

This patch adds a Redis instrumentation configuration option `record_value_size` (defaults to `false`), which will:

* set a span attribute `db.set_value_size_bytes` to the size of the value being set during `SET` Redis operations
* set a span attribute `db.retrieved_value_size_bytes` to the size of the _returned_ value during `GET` and `MGET` redis operations.

As is visible in the commit history, I started off by tracking sent and received bytes (which was simpler). Given the large number of [Redis commands](https://redis.io/commands/) out there, sent/received bytes might be a simpler implementation and more flexible to users, but confining this instrumentation to `SET`, `GET`, and `MGET` does give us the opportunity to use more meaningful attribute tags. Curious to hear folks' thoughts on this one...

**Feel free to stop reading here**, but so it's not lost to the sands of time, here was me trying to categorize different redis commands based on whether or not we want to track bytes sent to the redis server or bytes received:

```ruby
          SET_VALUE_SIZE_COMMANDS = %w[HSET HSETNX HMSET LINSERT LPUSH LPUSHX
            LSET RPUSH RPUSHX SADD ZADD XADD APPEND MSET MSETNX SET SETEX SETNX
            SETRANGE GEOADD]
          RECEIVED_VALUE_SIZE_COMMANDS = %w[DUMP KEYS RANDOMKEY HGET HGETALL
            HKEYS HMGET HVALS BLMOVE BLMPOP BRPOP BRPOPLPUSH LINDEX LMOVE LMPOP
            LPOP LRANGE  RPOP RPOPLPUSH EVAL EVALSHA EVALSHA_RO FCALL FCALL_RO
            SDIFF SPOP RANDMEMBER SUNION BZMPOP BZPOPMAX BZPOPMIN ZINTER ZMPOP
            ZPOPMAX ZPOPMIN ZRANDMEMBER ZRANGE ZRANGEBYLEX ZRANGEBYSCORE
            ZREVRANGE ZREVRANGEBYLEX ZREVRANGEBYSCORE ZUNION XREAD XREADGROUP
            XREVRANGE GET GETDEL GETEX GETRANGE GETSET MGET SUBSTR GEODIST
            GEOHASH GEOPOS GEORADIUS GEORADIUS_RO GEORADIUSBYMEMBER
            GEORADIUSBYMEMBER_RO GEOSEARCH]
```


instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -69,16 +96,31 @@ def parse_commands(commands) # rubocop:disable Metrics/AbcSize
               # If we receive an authentication request command
               # we want to short circuit parsing the commands
               # and return the obfuscated command
-              return 'AUTH ?' if command[0] == :auth
+              if command[0] == :auth
+                ret = "AUTH ?"
+                command_bytes += "AUTH ".bytesize + command[1].bytesize
```
plantfansam, line 101: Need to add the password chars to accurately count the `bytesize` here.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -29,13 +29,19 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
+
+            # Parse commands if we need to derive span attributes from them
+            if config[:db_bytes_transmitted] || config[:db_statement] != :omit
+              bytes_sent, parsed_commands = parse_commands(commands) 
+            end
+
             unless config[:db_statement] == :omit
-              parsed_commands = parse_commands(commands)
               parsed_commands = OpenTelemetry::Common::Utilities.truncate(parsed_commands, MAX_STATEMENT_LENGTH)
-              parsed_commands = OpenTelemetry::Common::Utilities.utf8_encode(parsed_commands, binary: true)
```
plantfansam, line 35: Moved this  to body of `parsed_commands`

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -69,16 +96,31 @@ def parse_commands(commands) # rubocop:disable Metrics/AbcSize
...
                 command_copy = command.dup
                 command_copy[0] = command_copy[0].to_s.upcase
-                command_copy.join(' ')
+                cmd = command_copy.join(' ')
+                command_bytes += cmd.bytesize
+                cmd
               end
             end.join("\n")
+
+            ret = OpenTelemetry::Common::Utilities.utf8_encode(ret, binary: true)
```
plantfansam, line 120: This is now in `parse_commands` because  the next line's `count` call will error out if the command is an invalid byte sequence in UTF-8.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -29,11 +31,15 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
+
             unless config[:db_statement] == :omit
-              parsed_commands = parse_commands(commands)
-              parsed_commands = OpenTelemetry::Common::Utilities.truncate(parsed_commands, MAX_STATEMENT_LENGTH)
-              parsed_commands = OpenTelemetry::Common::Utilities.utf8_encode(parsed_commands, binary: true)
-              attributes['db.statement'] = parsed_commands
+              attributes['db.statement'] = OpenTelemetry::Common::Utilities.truncate(parsed_commands, MAX_STATEMENT_LENGTH)
+            end
+
+            if config[:record_value_size] && commands.length == 1 && SET_VALUE_SIZE_COMMANDS.include?(commands[0][0])
```
plantfansam, line 41: pretty gross 😦

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
 
+          def calculate_bytesize(value) # rubocop:disable Metrics/CyclomaticComplexity
+            case value
+            when ::Redis::CommandError
+              # Don't report on these; call it zero.
+              0
+            when String
+              value.bytesize
+            when Array
+              value.sum { |s| calculate_bytesize(s) }
```
plantfansam, line 76: This makes me a little nervous 😄

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
                   s.record_exception(reply)
                   s.status = Trace::Status.error(reply.message)
                 end
+                next unless config[:record_value_size] && commands.length == 1
+
+                if RETRIEVED_VALUE_SIZE_COMMANDS.include?(commands[0][0])
+                  s['db.retrieved_value_size_bytes'] = calculate_bytesize(reply)
```
fbogsany, line 60: IMO there's we should use the same attribute name for both `SET` and `RETRIEVE` cases - `db.value_size_bytes` seems reasonable (or `db.value_size` - similar semantic conventions are inconsistent).

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -10,6 +10,8 @@ module Redis
       module Patches
         # Module to prepend to Redis::Client for instrumentation
         module Client
+          SET_VALUE_SIZE_COMMANDS = %i[set]
+          RETRIEVED_VALUE_SIZE_COMMANDS = %i[get mget]
```
fbogsany, line 14: If this set gets large, we should consider using a `Set`. Given the small size and use of symbols, the current implementation is likely more efficient.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
+              # Don't report on these; call it zero.
+              0
+            when String
+              value.bytesize
+            when Array
+              value.sum { |s| calculate_bytesize(s) }
+            when Integer
+              value.size
+            when Float
+              # Redis stores floats as strings, so get bytesize of stringified
```
fbogsany, line 80: Technically, I think it stores floats as floats, but the wire representation is a string. That begs the question: are we interested in the byte size of the wire representation or the stored representation? I think the wire representation is more interesting - the most common analysis is something like "are we sending/receiving a lot of large objects, exhausting network capacity?"

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
+              # float
+              value.to_s.bytesize
```
fbogsany, line 82: I'm always sad when we have to serialize something to calculate its size in instrumentation, and then throw away the serialized representation. I don't suppose there's a way to estimate the size without actually allocating a string and serializing the float?

Is there any difference in the output of `.to_s` vs the wire representation? E.g. rounding or truncation?

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
+            set_value_bytes = nil
+            calculate_set_bytes = false
+
+            ret = commands.map do |command|
               # We are checking for the use of Redis#queue command, if we detect the
               # extra level of array nesting we return the first element so it
               # can be parsed.
               command = command[0] if command.is_a?(Array) && command[0].is_a?(Array)
 
+              if SET_VALUE_SIZE_COMMANDS.include?(command[0])
```
fbogsany, line 103:
```suggestion
              if config[:record_value_size] && SET_VALUE_SIZE_COMMANDS.include?(command[0])
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
               # and return the obfuscated command
-              return 'AUTH ?' if command[0] == :auth
+              if command[0] == :auth
+                return nil, 'AUTH ?'
+              end
 
               if config[:db_statement] == :obfuscate
-                command[0].to_s.upcase + ' ?' * (command.size - 1)
+                cmd = command[0].upcase.to_s + ' ?' * (command.size - 1)
+                set_value_bytes += calculate_bytesize(command[-1]) if calculate_set_bytes && config[:record_value_size]
```
fbogsany, line 117:
```suggestion
                set_value_bytes += calculate_bytesize(command[-1]) if calculate_set_bytes
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -48,37 +54,78 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
               else
                 command_copy = command.dup
                 command_copy[0] = command_copy[0].to_s.upcase
-                command_copy.join(' ')
+                cmd = command_copy.join(' ')
+                set_value_bytes += calculate_bytesize(command_copy[-1]) if calculate_set_bytes && config[:record_value_size]
```
fbogsany, line 122:
```suggestion
                set_value_bytes += calculate_bytesize(command_copy[-1]) if calculate_set_bytes
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
fbogsany: Moving it has swapped the order of truncation and encoding. I'm not sure that's the right thing to do here. 🤔 @robertlaurin do you remember why we did it in this order, and whether it matters in practice?

robertlaurin: The ordering was driven by "don't waste time/resources encoding all of it if you are going to truncate it"

robertlaurin: We could do some benchmarks to see how much it matters in practice. Similar approach to here would probably be reasonable https://github.com/open-telemetry/opentelemetry-ruby/pull/760

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -29,11 +31,15 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
             attributes['peer.service'] = config[:peer_service] if config[:peer_service]
             attributes.merge!(OpenTelemetry::Instrumentation::Redis.attributes)
 
+            # Parse commands if we need to derive span attributes from them
+            set_value_size_bytes, parsed_commands = parse_commands(commands) if config[:record_value_size] || (config[:db_statement] != :omit)
```
fbogsany, line 35: Given the complexity of `parse_commands` with this change, and the requirement to allocate an array for the return value, I think we're better off calculating the size separately. I think this looks something like:

```ruby
value_size_bytes = value_size(commands) if config[:record_value_size]

# ...

def value_size(commands)
  value_size = nil
  commands.map do |command|
    command = command[0] if command.is_a?(Array) && command[0].is_a?(Array)
    return nil if command[0] == :auth
    if SET_VALUE_SIZE_COMMANDS.include?(command[0])
      value_size ||= 0
      value_size += calculate_bytesize(command[-1])
    end
  end
  value_size
end
```

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
fbogsany: Technically, this omits `SET` enclosed in `Redis#queue`, right? Do we want that?

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: Yeah I think you're spot on, here. `parse_commands` has gotten too thorny.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: Yeah, the comment for sure lacks appropriate rigor. FWIW the Redis docs do not represent `Float` as a [first class data type](https://redis.io/docs/manual/data-types/), but who knows how it's storing these `Float`-y "Strings" behind the scenes. 

When working on this, I was caught off guard by the consideration you bring up - do we care about byte size of wire representation or how much space it takes up on the Redis side (and wait, how does Redis store Floats, and how would we calculate that from Ruby without knowledge of which Redis version etc etc etc). 

I think you're right that we care about the wire representation more than however much disk space Redis is using. Not to mention that figuring out Redis' disk utilization and calculating from this library seems like a bit of a rabbit hole. Also, I've got to imagine that someone running analysis on big keys will end up discovering values that are not `Floats` (or Redis' string representations thereof).

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: 🤔 this is a great question - I will 👀

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: 8147d43dc144003c5462f6b9b52bc67a230c6427

plantfansam: Upon reflection, this makes telemetry on pipelined commands a little odd. Take e.g.:

```ruby
redis.pipelined do |p|
  p.set("X", "abc123")
  p.get("X")
end
```

If we have one attribute to store `SET`ted and `GET`ted values, then `span["value_size_bytes"]` would equal `("abc123" + "abc123").bytesize`. I think having `db.set_value_size_bytes` and `db.get_value_size_bytes`  could be more useful...

plantfansam: 🤔  I think we also will need to patch [`Redis::Client.call_pipelined`](https://github.com/redis/redis-rb/blob/master/lib/redis/client.rb#L214-L249) in order to access the response from pipelined calls 😅

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: I undid the relocation, so did not feel it necessary to benchmark as part of this PR

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -29,30 +76,41 @@ def process(commands) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength, M
...
           end
 
-          private
+          def calculate_bytesize(value) # rubocop:disable Metrics/CyclomaticComplexity
+            case value
+            when ::Redis::CommandError
+              # Don't report on these; call it zero.
+              0
+            when String
+              value.b.bytesize
```
plantfansam, line 100: Could do more research w/r/t if this is actually necessary — it is an extra string allocation. Per https://github.com/open-telemetry/opentelemetry-ruby/pull/1150/files#r847575538, redis is calling `.b` under the hood before sending over the wire.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
```diff
@@ -81,6 +139,37 @@ def parse_commands(commands) # rubocop:disable Metrics/AbcSize
...
+              command = command[0] if command.is_a?(Array) && command[0].is_a?(Array)
+              return nil if command[0] == :auth
+              if commands_to_record.include?(command[0])
+                value_size += calculate_bytesize(command[-1])
+              end
+            end
+            value_size
+          end
+
+          def retrieved_value_size(reply, commands, commands_to_record)
```
plantfansam, line 154: Need to document parameters here.

instrumentation/redis/lib/opentelemetry/instrumentation/redis/patches/client.rb
plantfansam: This is a few iterations ago - I think we're all set in `sent_value_size` and `retrieved_value_size` now

plantfansam:
Moved here https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/50

PR #1151, opened 2022-03-18
[Bunny] Accept both string and symbol topic names

hosamaly:
Specifying the topic as a symbol used to cause the following error:
```
OpenTelemetry error: invalid span attribute value type Symbol for key 'messaging.destination' on span '<topic> send'
```

This happens, for example, when one uses `ActiveJob::Base.queue_as :topic_name`.


arielvalentin (commented):
Thank you for this contribution! 

One request I would like to make it for you to please resubmit these commits using conventional commits:

https://www.conventionalcommits.org/en/v1.0.0/

Our tooling requires this format to generate change logs during a release.

hosamaly:
I updated the commit messages.

arielvalentin (approved):
I think it would be good to have test cases that cover both scenarios where the topic is configured as a string or a symbol. 

If you have time to follow up to add another scenario that would be great. 

Thank you again for your contribution!

fbogsany:
From the failures, it looks very much like `bunny` itself does not support symbols as topic names. E.g.:
```
opentelemetry-instrumentation-bunny: test
>> BUNDLE_GEMFILE=/home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/bunny/gemfiles/bunny_2.gemfile bundle exec rake test
Run options: --seed 60180

# Running:

.E.....

Finished in 1.047769s, 6.6809 runs/s, 21.9514 assertions/s.

  1) Error:
OpenTelemetry::Instrumentation::Bunny::Patches::Channel#test_0001_traces produce and consuming:
NoMethodError: undefined method `gsub' for :"topic-58882225-2006-4afc-9f7a-35f0b2a247ec":Symbol
    /opt/hostedtoolcache/Ruby/2.6.9/x64/lib/ruby/gems/2.6.0/gems/bunny-2.19.0/lib/bunny/channel.rb:1180:in `exchange_declare'
    /opt/hostedtoolcache/Ruby/2.6.9/x64/lib/ruby/gems/2.6.0/gems/bunny-2.19.0/lib/bunny/exchange.rb:254:in `declare!'
    /opt/hostedtoolcache/Ruby/2.6.9/x64/lib/ruby/gems/2.6.0/gems/bunny-2.19.0/lib/bunny/exchange.rb:87:in `initialize'
    /opt/hostedtoolcache/Ruby/2.6.9/x64/lib/ruby/gems/2.6.0/gems/bunny-2.19.0/lib/bunny/channel.rb:352:in `new'
    /opt/hostedtoolcache/Ruby/2.6.9/x64/lib/ruby/gems/2.6.0/gems/bunny-2.19.0/lib/bunny/channel.rb:352:in `topic'
    /home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/bunny/test/opentelemetry/instrumentation/bunny/patches/channel_test.rb:23:in `block (2 levels) in <top (required)>'

7 runs, 23 assertions, 0 failures, 1 errors, 0 skips
```

arielvalentin:
@hosamaly I have removed my approval after @fbogsany pointed out the tests are failing. 

Is it possible this is an issue in ActiveJob and not the Bunny instrumentation?

hosamaly:
I'm surprised to see these failures. The CI tests pass [on MacOS](https://github.com/open-telemetry/opentelemetry-ruby/runs/5604834256?check_suite_focus=true#step:5:2464) and [on Windows](https://github.com/open-telemetry/opentelemetry-ruby/runs/5604834344?check_suite_focus=true#step:5:2205) but fail on all Linux tests!

The only difference that I can see is that Linux tests have an actual RabbitMQ service that can be used. This probably means that the test runs through a different code path...

arielvalentin:
There are some instrumentations that the build skips on OSX and Windows since the service dependencies are or native bindings unavailable for those platforms.

hosamaly:
I was slightly mistaken. The exchange name caused the issue, not the topic (which ActiveJob stringifies). I've updated the tests and the PR description.

I hope it passes CI this time. It passes on my machine, but it had also passed locally the first time!

hosamaly:
It [failed again](https://github.com/open-telemetry/opentelemetry-ruby/runs/5628678629?check_suite_focus=true#step:6:2645). Is there a way to reproduce the failure locally?

hosamaly:
I did that and it still passed locally. Turned out that Bundler had been using an old version of `bunny` that was compatible with Symbols. I was able to reproduce the failure after updating the bundle.

Using Symbols [became invalid](https://github.com/ruby-amqp/bunny/pull/566/files#diff-df2c93872027d73da6f9de309fd0d79ff4e3b24209d7c3c0483d8d92be09a326R1178) in v2.12 of the bunny gem.

I still think that the fix is valid even if it's not tested, but I can also add a test that is only run when `Bunny::VERSION <= '2.11'` (and an appropriate Appraisal option) if desired.

hosamaly:
> Using Symbols [became invalid](https://github.com/ruby-amqp/bunny/pull/566/files#diff-df2c93872027d73da6f9de309fd0d79ff4e3b24209d7c3c0483d8d92be09a326R1178) in v2.12 of the bunny gem.
> 
> I still think that the fix is valid even if it's not tested, but I can also add a test that is only run when `Bunny::VERSION <= '2.11'` (and an appropriate Appraisal option) if desired.

Hi, @arielvalentin. Do you have a preference as to which way to go? I can see three options:
* Keep the fix without a test
* Keep the fix and add a test that only runs on older versions of Bunny
* Close the merge request

fbogsany:
I would prefer to close the PR. v2.11 is almost 4 years old. v2.19 is the latest release. Our policy is to limit support to versions of the instrumented gem that are supported upstream. In this case, `bunny` does not have an official support policy, however there is no evidence that old releases are supported (no out of order releases since 2018, except for one that appears erroneous).

At this point, the test is more work than the fix, which isn't needed for recent releases. I would prefer it if we added a `compatible` constraint similar to https://github.com/open-telemetry/opentelemetry-ruby/blob/51b6e8ad465ba81fa743224b5ecfd4ab871bea51/instrumentation/action_pack/lib/opentelemetry/instrumentation/action_pack/instrumentation.rb#L24-L26
that limits support to v2.12+.

PR #1152, opened 2022-03-18
Log status in otlp exporter

fbogsany:
This logs the status message returned from the collector for failures. This can help with debugging unexpected errors during export.


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -231,6 +232,17 @@ def handle_redirect(location)
           # TODO: figure out destination and reinitialize @http and @path
         end
 
+        def log_status(body)
+          status = Google::Rpc::Status.decode(body)
+          details = status.details.map do |detail|
+            klass_or_nil = ::Google::Protobuf::DescriptorPool.generated_pool.lookup(detail.type_name).msgclass
+            detail.unpack(klass_or_nil) if klass_or_nil
+          end.compact
```
fbogsany, line 240: This strives to decode any types that we know about. Example log output:
```
E, [2022-03-18T12:46:00.897332 #22203] ERROR -- : OpenTelemetry error: OTLP exporter received rpc.Status{message=bad request, details=[<Google::Protobuf::StringValue: value: \"you are a bad request\">]}
```
I'm not sure we can practically dig any deeper than this, although `.to_json` is an option 🤔

fbogsany: One thing to note is that we won't decode unknown messages - we'll just drop them. Would it be better to log the encoded messages in that case?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -231,6 +232,17 @@ def handle_redirect(location)
...
+          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}}")
```
ericmustin, line 241: Not blocking, but do we want to add `status.code` as a label that we pass into the @metrics_reporter? 

https://cloud.google.com/tasks/docs/reference/rpc/google.rpc#google.rpc.Code

ericmustin (approved):
Lgtm, had one question but not blocking. Imo this will help the majority of users improving their debugging experience.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
arielvalentin: 💭 Have you considered checking the `Content-Type` response header before attempting to parse the body as protobuf?

In the case where there is a decoding error, will the Exception include enough information in the message to provide us with enough useful information to see detailed information about the response body? Would reporting the raw response body in that case help?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: Not really. The OTLP spec says it doesn't use `status.code`. We "MAY record it for troubleshooting purposes", but it seems a bit pointless. I can add it if you're curious to see what we get.

ericmustin: zero curiosity, this can be resolved

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
fbogsany: > 💭 Have you considered checking the `Content-Type` response header before attempting to parse the body as protobuf?

Not really. The spec says we're supposed to get a `Status` back, encoded as protobuf. If we don't, I think it's fair to say that's a bug. That said, I'm sure this will be noisy for us when we have a LB in the way that returns a 500 🤔 

> 
> In the case where there is a decoding error, will the Exception include enough information in the message to provide us with enough useful information to see detailed information about the response body? Would reporting the raw response body in that case help?

Wow, you're sure asking a lot! I was hoping to knock this out quick & bask in the adulation :trollface:.

arielvalentin: In this scenario am I the :trollface: ?

fbogsany: I'll think about this some more. We had this on the backburner for awhile because the collector didn't respond with the required status message. They fixed it recently, and we had some errors where it would have been useful to have some more detailed logs, so I tackled this. I suspect we may need a couple of iterations to get it "right" in terms of controlling verbosity while also providing insights.

I think my path forward here will be to ensure it doesn't 💥 when we get a decoding error, but otherwise limit the complexity until we've had a chance to let this soak in production & see if we want more detail or less noise.

fbogsany: @arielvalentin would you be OK with **NOT** parsing the `details` and instead just logging the raw data? Then we can return to this and refine it if we find something useful in the logs.

arielvalentin: That sounds fine to me.

Incremental improvements > bug changes

arielvalentin (approved):

PR #1153, opened 2022-03-18
[API] Reduce exception handling in `TextMapPropagator#extract`

hosamaly:
Previous code used to throw an error from `TraceParent.from_string` when given a `nil`, which happens when the `carrier` is empty.

This pull request changes the implementation to return early instead. It follows the same pattern that was introduced in `Baggage::TextMapPropagator` in PR #835.


robertlaurin:
Hey @hosamaly just would like to understand the issue this is meant to correct a little better.

> Previous code used to throw an error from `TraceParent.from_string` when given a nil, which happens when the carrier is empty.

I believe we are handling nil cases in that method https://github.com/open-telemetry/opentelemetry-ruby/blob/8a538e7f60f2ad38600c3d8d83718290f185ea70/api/test/opentelemetry/trace/tracestate_test.rb#L14-L17

api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb
```diff
@@ -45,8 +45,12 @@ def inject(carrier, context: Context.current, setter: Context::Propagation.text_
...
           #   if extraction fails
           def extract(carrier, context: Context.current, getter: Context::Propagation.text_map_getter)
-            tp = TraceParent.from_string(getter.get(carrier, TRACEPARENT_KEY))
-            tracestate = Tracestate.from_string(getter.get(carrier, TRACESTATE_KEY))
+            trace_parent_value = getter.get(carrier, TRACEPARENT_KEY)
+            trace_state_value = getter.get(carrier, TRACESTATE_KEY)
+            return context if trace_parent_value.nil? || trace_state_value.nil?
+
+            tp = TraceParent.from_string(trace_parent_value)
+            tracestate = Tracestate.from_string(trace_state_value)
```
robertlaurin, line 53: I believe I understand the intention of this PR now.

This change is intends to stop relying on exceptions for control flow in the (potentially common) case where the TRACEPARENT key is not present in the carrier, so instead we return the context early if we failed to extract any value from that key.

So this is not resolving any bug but instead meant to improve performance, is that correct @hosamaly?

If we do take this approach, we should *not* return early if the trace_state_value is nil, as the `Tracestate.from_string` can handle that case.  

```suggestion
            tracestate = Tracestate.from_string(getter.get(carrier, TRACESTATE_KEY))
            return context unless trace_parent_value = getter.get(carrier, TRACEPARENT_KEY)
            tp = TraceParent.from_string(trace_parent_value)
```

api/test/opentelemetry/baggage/propagation/text_map_propagator_test.rb
```diff
@@ -184,5 +184,9 @@ def assert_value(context, key, value, metadata = nil)
   entry = OpenTelemetry::Baggage.raw_entries(context: context)[key]
   _(entry).wont_be_nil
   _(entry.value).must_equal(value)
-  _(entry.metadata).must_equal(metadata)
+  if metadata.nil?
+    _(entry.metadata).must_be_nil
```
robertlaurin, line 188: Am I correct that this change is entirely unrelated to the changes to the text map propagator?  It seems this is addressing the mini test complaints about assert_equal on nil values.

api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb
```diff
@@ -45,8 +45,12 @@ def inject(carrier, context: Context.current, setter: Context::Propagation.text_
           # @return [Context] context updated with extracted baggage, or the original context
           #   if extraction fails
           def extract(carrier, context: Context.current, getter: Context::Propagation.text_map_getter)
-            tp = TraceParent.from_string(getter.get(carrier, TRACEPARENT_KEY))
-            tracestate = Tracestate.from_string(getter.get(carrier, TRACESTATE_KEY))
+            trace_parent_value = getter.get(carrier, TRACEPARENT_KEY)
+            trace_state_value = getter.get(carrier, TRACESTATE_KEY)
+            return context if trace_parent_value.nil? || trace_state_value.nil?
```
fbogsany, line 50: The [W3C Trace Context](https://www.w3.org/TR/trace-context/#tracestate-header) states that:
> If the vendor failed to parse `traceparent`, it _MUST NOT_ attempt to parse `tracestate`. Note that the opposite is not true: failure to parse `tracestate` _MUST NOT_ affect the parsing of `traceparent`.

so we must not return early if `trace_state` is `nil`.
```suggestion
            trace_parent_value = getter.get(carrier, TRACEPARENT_KEY)
            trace_state_value = getter.get(carrier, TRACESTATE_KEY)
            return context if trace_parent_value.nil?
```

fbogsany:
This is both a perf optimization and an attempt to reduce the use of exceptions for "normal" flow control. In general, I think it would be worth investing the effort to remove the use of exceptions in `TraceParent` parsing as well, but this is valid regardless.

api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb
fbogsany: Unfortunately, we can't parse `tracestate` before parsing `traceparent`:
> If the vendor failed to parse `traceparent`, it _MUST NOT_ attempt to parse `tracestate`.

api/test/opentelemetry/baggage/propagation/text_map_propagator_test.rb
hosamaly: Yes, exactly. That's why it's in a separate commit.

api/lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb
hosamaly: Indeed, you're absolutely right. I've committed your suggestion and reworded the commit to indicate that it's a performance improvement rather than a fix.

robertlaurin (approved):
Just waiting on CI, will merge when complete.

Thanks for the PR!

arielvalentin:
Linter error https://github.com/open-telemetry/opentelemetry-ruby/runs/5628071090?check_suite_focus=true#step:5:64

```
lib/opentelemetry/trace/propagation/trace_context/text_map_propagator.rb:48:54: W: Lint/AssignmentInCondition: Use == if you meant to do a comparison or wrap the expression in parentheses to indicate you meant to assign in a condition.
[65](https://github.com/open-telemetry/opentelemetry-ruby/runs/5628071090?check_suite_focus=true#step:5:65)
            return context unless trace_parent_value = getter.get(carrier, TRACEPARENT_KEY)
```

hosamaly:
> [Linter error](https://github.com/open-telemetry/opentelemetry-ruby/runs/5628071090?check_suite_focus=true#step:5:64)

Fixed

PR #1157, opened 2022-03-22
Move class-scoped attribute logic to shared module

plantfansam:
We're duplicating the logic for the very handy `with_attributes` functionality in two places, and I was [about to add in a third](https://github.com/plantfansam/opentelemetry-ruby/compare/plantfansam/extract-with-attributes...plantfansam:plantfansam/mysql-with-attrs)! As such, it seemed like time to extract the common parts into a shared module, tentatively named `OpenTelemetry::Common::ClassScopedAttributes`. Very happy to change the name if anyone's got something better ...

This PR:

* Moves the former contents of `OpenTelemetry::Common::HTTP::ClientContext` to a `ClassScopedAttributes` module, leaving them mostly unchanged. The new version is the same as the old, except that sets a context key that's namespaced to the class or module that's extending `ClassScopedAttributes` (to prevent one extending class from overwriting another).
* Moves tests from `common/test/opentelemetry/common/http_client_context_test.rb` to `common/test/opentelemetry/common/class_scoped_attributes_test.rb`.
* Adds tests to verify that attributes are actually namespaced by extending class. 
* Includes the `ClassScopedAttributes` functionality in `Http::ClientContext` and `Instrumentation::Redis`, preserving existing functionality.


arielvalentin:
@plantfansam what a coincidence!!! @robbkidd and I were pairing on doing this very thing only we added a custom span processor that would share the attributes across all child spans for a given context. 

https://cloud-native.slack.com/archives/C01NWKKMKMY/p1648582974571549

plantfansam:
> @plantfansam what a coincidence!!! @robbkidd and I were pairing on doing this very thing only we added a custom span processor that would share the attributes across all child spans for a given context.
> 
> https://cloud-native.slack.com/archives/C01NWKKMKMY/p1648582974571549

Whoa. It's the same!

common/lib/opentelemetry/common/class_scoped_attributes.rb
```diff
@@ -0,0 +1,54 @@
...
+
+module OpenTelemetry
+  module Common
+    # Attributecontains common helpers for atribute propagation
+    # Various classes should extend this module to get `with_attributes`
+    # functionality scoped to that class.
+    module ClassScopedAttributes
+      extend self
+
+      def current_attributes_hash
```
fbogsany, line 15: This should be `private`. We should not expose the context key outside of this module.

common/lib/opentelemetry/common/class_scoped_attributes.rb
```diff
@@ -0,0 +1,54 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Common
+    # Attributecontains common helpers for atribute propagation
```
fbogsany, line 9:
```suggestion
    # ClassScopedAttributes contains common helpers for attribute propagation
```

common/lib/opentelemetry/common/class_scoped_attributes.rb
```diff
@@ -0,0 +1,54 @@
...
+    # functionality scoped to that class.
+    module ClassScopedAttributes
+      extend self
+
+      def current_attributes_hash
+        @current_attributes_hash ||= Context.create_key("#{self.name}-current-attributes-hash")
+      end
+
+      # Returns the attributes hash representing the context found
+      # in the optional context or the current context if none is provided.
```
fbogsany, line 20:
```suggestion
      # Returns the attributes hash found in the optional context or the
      # current context if none is provided.
```

common/lib/opentelemetry/common/class_scoped_attributes.rb
plantfansam: 28c1499689fef32d8ee98abbf44b5faef4f9068e

plantfansam: > This should be `private`. We should not expose the context key outside of this module.

bf98ea86e1817c5f4095aab4310219da5c0a648d

fbogsany:
I grilled @robertlaurin about this & he confessed that his threshold for duplication is 5. :trollface:
Screen Shot 2022-04-05 at 1 15 22 PM

PR #1159, opened 2022-03-22
fix: Allow HTTP method to be passed as string

### LINKED ISSUE #1158: It's not possible to track requests made with excon using string as HTTP Method
**Description of the bug**
The excon instrumentation shows an error if an string is used as value for the method. According to the [excon docs](https://www.rubydoc.info/gems/excon/0.79.0#:~:text=request%20takes%20a%20method%20option%2C%20accepting%20either%20a%20symbol%20or%20string) this is allowed.

The error on the logs is
`[2022-03-22T17:31:14.333447 #85108] ERROR -- : OpenTelemetry error: invalid span attribute value type NilClass for key 'http.method' on span 'HTTP '`

The code only checks for HTTP Methods set as symbols [here](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares#L28)

Operating system details: MacOS, Monterey 12.2.1
RUBY_ENGINE: "ruby"
RUBY_VERSION: "2.7.5"
RUBY_DESCRIPTION: "ruby 2.7.5p203 (2021-11-24 revision f69aeb8314) [x86_64-darwin20]"

```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-instrumentation-all'
  gem 'excon'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'
require 'excon'

  c.use_all
end

Excon.new('http://google.com').request(:path => '/', :method => 'GET')
```

nickrisaro:
According to [Excon docs](https://www.rubydoc.info/gems/excon/0.79.0) is possible to set the HTTP Method with a string but when using excon instrumentation this fails with the following error `[2022-03-22T17:31:14.333447 #85108] ERROR -- : OpenTelemetry error: invalid span attribute value type NilClass for key 'http.method' on span 'HTTP '`

On this PR we add a test case for the described scenario and a fix to use the string if no symbol was provided.

Fixes #1158

### DIFF
--- a/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
+++ b/instrumentation/excon/lib/opentelemetry/instrumentation/excon/middlewares/tracer_middleware.rb
@@ -26,6 +26,7 @@ module OpenTelemetry
             begin
               unless datum.key?(:otel_span)
                 http_method = HTTP_METHODS_SYMBOL_TO_STRING[datum[:method]]
+                http_method = datum[:method] if http_method.nil?
                 attributes = span_creation_attributes(datum, http_method)
                 tracer.start_span(
                   "HTTP #{http_method}",
--- a/instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
+++ b/instrumentation/excon/test/opentelemetry/instrumentation/excon/instrumentation_test.rb
@@ -143,5 +143,23 @@ describe OpenTelemetry::Instrumentation::Excon::Instrumentation do
 
       _(span.attributes['peer.service']).must_equal 'example:custom'
     end
+
+
+    it 'accepts string method' do
+      ::Excon.new('http://example.com').request(:path => '/success', :method => 'GET')
+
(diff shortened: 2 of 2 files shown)


plantfansam:
Thanks so much for the PR @nickrisaro! At first glance, I _believe_ that this should be addressed by @hosamaly's work in https://github.com/open-telemetry/opentelemetry-ruby/pull/1147, which will [create a lookup hash with both string and symbol keys](https://github.com/open-telemetry/opentelemetry-ruby/pull/1147/files#diff-1661c74feeade9ce3285977d3efe553e692078d5a60216eceffa3cc1a4161d8aR13-R30).

nickrisaro:
Thanks for the comment @plantfansam! I didn't find that PR yesterday when looking for this. I agree with you, @hosamaly's solution is better than mine.

I'll close this PR in favor of #1147

plantfansam:
Thanks for your contribution @nickrisaro 😄!

PR #1161, opened 2022-03-25
fix: RubyGems Fallback

arielvalentin:
The GitHub Monolith bypasses RubyGems when loading dependencies, which results in a `nil` variable being returned by `Gem.loaded_specs`.
    
This change fallsback to using `Gem::Version.new(<VERSION>)` where possible.
    
See https://github.com/open-telemetry/opentelemetry-ruby/issues/947
 
Related https://github.com/open-telemetry/opentelemetry-ruby/issues/988#issuecomment-1018676853

PR #1166, opened 2022-03-30
chore: AWS SDK min version 3.0

arielvalentin:
Bumps auto-instrumentation version to 3.0, which is the minimum supported version as of 2021-11-21

> We are announcing that version 2 of the AWS SDK For Ruby will enter maintenance mode on 11/20/2020.
> Support for version 2 will end on 11/21/2021.

https://aws.amazon.com/blogs/developer/deprecation-schedule-for-aws-sdk-for-ruby-v2/

cc: @YanivD


instrumentation/aws_sdk/gemfiles/aws_sdk_3.0.gemfile
```diff
@@ -10,6 +10,7 @@ group :test do
   gem "opentelemetry-common", path: "../../../common"
   gem "opentelemetry-sdk", path: "../../../sdk"
   gem "opentelemetry-semantic_conventions", path: "../../../semantic_conventions"
+  gem "webrick", "~> 1.7"
```
plantfansam, line 13: Why add webrick now?

instrumentation/aws_sdk/lib/opentelemetry/instrumentation/aws_sdk/instrumentation.rb
```diff
@@ -9,7 +9,7 @@ module Instrumentation
     module AwsSdk
       # Instrumentation class that detects and installs the AwsSdk instrumentation
       class Instrumentation < OpenTelemetry::Instrumentation::Base
-        MINIMUM_VERSION = Gem::Version.new('2.0')
+        MINIMUM_VERSION = Gem::Version.new('3.0')
```
plantfansam, line 12: 🎉

plantfansam (approved):
I had an informational question but this LGTM

instrumentation/aws_sdk/gemfiles/aws_sdk_3.0.gemfile
arielvalentin: Regenerating the appraisal files brought in `webrick` because if you take a look at the Gemfile you will see it is an explicit dependency:

https://github.com/open-telemetry/opentelemetry-ruby/blame/main/instrumentation/aws_sdk/Gemfile#L18

`webrick` was added to the main Gemfile _after_ the appraisals were already generated but not updated at the time to include it.

This is the commit that introduced the appraisals: 2021-11-04
https://github.com/open-telemetry/opentelemetry-ruby/pull/1014/commits/32335be10b61f82d65f208476ecbc67f098b21c9

This is commit is were `webrick` was added but the appraisals were not updated along with it: 2021-11-07
https://github.com/open-telemetry/opentelemetry-ruby/pull/1014/commits/d57cd30a25372179fbb5933b485764f6723c69fe

YanivD:
@arielvalentin Thanks for the mention. 

What is the motivation to drop support for versions `2.x.x`? As far as I know the instrumentation works well also on versions 2.
I think if we can support those versions without heavy effort - it's worth it.

arielvalentin:
This came up because someone requested we support Rails 4 and we made it clear that we would only support instrumentation for gems that receive maintenance updates.

In this case since the AWS-SDK instrumentation already demonstrated support for 2.0, they should lock the instrumentation to a specific version in their gemfile.

Once this PR is merged I can change how we perform version lookup. I will have some questions for you in a different PR.

YanivD:
I'm not sure I understand what is the motivation to drop support. Users might still use a deprecated version in real life projects. 

IMO, it's valuable information and we should support version 2 if it's possible.

ericmustin:
I think i'd like to get some clarification on this from the other maintainers or the GC on what policy we should enforce and what sort of discretion we have. This is something I've seen pop up on other tracing sdk's i've been involved with and I think it can be contentious, and I'd like to make sure everyone feels like they're heard.

Imo, if we have have some discretion here then i wouldn't be opposed to _not_ restricting compatibility (don't bump min version), but perhaps dropping support (drop appraisals and tests against 2.x). There are tradeoffs to this (_encouraging security anti patterns, potentially increased support tickets, tech debt for maintaining all the 2.x code, etc_), but also benefits (_more adoption_). This is a tradeoff consideration I think is useful to think about on a case by case basis, and the considerations of rails 4.x are perhaps different than aws-sdk 2.x (just my two cents).

I also wouldn't be opposed to merging this PR but with the caveat of adding a support matrix in our documentation somewhere so users would know _"if i'm running 2.x, i should pin to xyz instrumentation version, if i'm running 3.x, i should pin to a different xyz instrumentation version, and so on"_ This is an approach that worked well in my anecdotal experience, such as https://github.com/DataDog/dd-trace-rb/blob/master/docs/GettingStarted.md#integration-instrumentation. I think without this documentation it could be a very confusing experience to users.

Even if we don't have discretion and _must_ drop support, that compatibility matrix is probably a good thing to add.

Additionally, if we don't have discretion, then i think a followup item for us is to figure out a way to track this sort of thing across our instrumentations programmatically.

arielvalentin:
> The policy we discussed previously:
> 
> 1. Our auto-instrumentation will only explicitly support and test versions of the instrumented gem that are themselves officially maintained.
> 2. We will feel free to use new features (e.g. Ruby language features) when they're supported by the minimum supported version.
> 3. If a vendor (e.g. a Cloud vendor or an Observability vendor) commits to owning and maintaining auto-instrumentation for an older, unmaintained gem version, we'll support them in that. If they don't commit or don't address issues in a timely fashion, we will withdraw that support.
> 
> I think this is a pretty reasonable policy. It strikes the right balance for expectations and load on OpenTelemetry Ruby maintainers, and delegates ownership of "old" gem instrumentation to community members that benefit from it.
> 
> Without a clear policy like this, and strict adherence to it, we will have to re-litigate this on a case-by-case basis.

👍🏼 In this case then I think then the AWS SDK would fall under **3** since Aspecto (@YanivD and @blumamir) are effectively the maintainers and I think they have been good about keeping up with it. 

Of course if AWS wouldn't mind we would appreciate it _hint hint wink wink @NathanielRN_

I will close this PR and preserve version compatibility in https://github.com/open-telemetry/opentelemetry-ruby/pull/1161

NathanielRN:
@arielvalentin Sorry for the delay! Would be happy to facilate setting up a meeting between you and the AWS SDK team if you’re interested but I don’t know the answer myself 🙂 Feel free to message me on Slack!

PR #1167, opened 2022-04-03
refactor: Extract Instrumentation Registry Gem

arielvalentin:
This change is the first in many steps that will allow us to move instrumentation gems into the contrib repo.
It copies the existing `Registry` into its own gem and will allow us to release it without breaking the SDK
or Instrumentations.

Once published we can update existing dependencies to use the `opentelemetry-registry` gem.

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1
See https://github.com/open-telemetry/opentelemetry-ruby/discussions/776

### DIFF
--- a/registry/README.md
+++ b/registry/README.md
@@ -1,0 +1,52 @@
+# OpenTelemetry Registry Instrumentation
+
+Todo: Add a description.
+
+## How do I get started?
+
+Install the gem using:
+
+```
+gem install opentelemetry-instrumentation-registry
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-instrumentation-registry` in your `Gemfile`.
+
+## Usage
+
+To use the instrumentation, call `use` with the name of the instrumentation:
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use 'OpenTelemetry::Instrumentation::Registry'
+end
+```
+
+Alternatively, you can also call `use_all` to install all the available instrumentation.
+
+```ruby
+OpenTelemetry::SDK.configure do |c|
+  c.use_all
+end
+```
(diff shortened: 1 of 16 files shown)


arielvalentin:
Leaving the base gem in the main repo means any changes to the Base class would require coordination with the contrib repo. That isn't necessarily true for the registry, which hasn't changed much. 

The tight coupling between the instrumentations and the Base class made me think isolating the registry would make maintenance easier for us.

robertlaurin (approved):

robertlaurin:
Why this change makes sense from my perspective
* Having the registry decoupled from the sdk allows a hypothetical third party sdk to make us of it, and by extension all the instrumentation.
* The registry has little to no churn, and is more closely related to the SDK than anything else. 
* I think it'll improve ease of development for instrumentation packages (in the contrib repo) if some changes need to be made in the base class and you want to see if they’re compatible immediately as part of ci, and not when you do a gem bump.

registry/README.md
```diff
@@ -0,0 +1,52 @@
+# OpenTelemetry Registry Instrumentation
+
+Todo: Add a description.
```
robertlaurin, line 3: Whoops, missed these.  Will wait till Todos are finished before merging.

arielvalentin: My bad

arielvalentin: @robertlaurin Updated the README. Please take another look at your earliest convenience.

arielvalentin:
@robertlaurin It looks like there are intermittent test failures in Linux JRuby. Has that come up before? 

https://github.com/open-telemetry/opentelemetry-ruby/runs/5853687205?check_suite_focus=true

PR #1170, opened 2022-04-04
Refactor: Add test helpers gem

robertlaurin:
Extracts test helpers from common which is a run time dependency of many gems.  Instead let us put any convenience methods related to testing OpenTelemetry Ruby into its own gem that is only included as a test or development dependency.

This may be a lot of boilerplate setup for a couple repeated methods, but I wanted to create a designated place to put things like this.

I opted to use the full namespaces instead of structuring this as a module to include. There is a trade off between brevity of helper method name, and being able to discover where these helper methods are located now that they are extracted to a new gem.


ericmustin (approved):
lgtm i think this makes things much more DRY...and the tests pass so 🚀

PR #1172, opened 2022-04-05
fix: warn on unsupported otlp transport protocols

robertlaurin:
Follow up from https://github.com/open-telemetry/opentelemetry-ruby/pull/1129

https://github.com/open-telemetry/opentelemetry-specification/blob/v1.10.0/specification/protocol/exporter.md#specify-protocol


fbogsany (approved):

plantfansam:
Prob not worth the time it takes me to type this, but we could do something like `APPROPRIATE_PREFIX_EXPORTER_PROCOLS=[''http/protobuf'].freeze` and check for inclusion in that array.

PR #1173, opened 2022-04-05
Only allow certain types of Numeric values as attribute values.

plantfansam:
We ran into a pretty odd failure mode, which this patch will hopefully help address (h/t @ericmustin for sussing this out):

In an app, we set a span attribute value to an `ActiveSupport::Duration` instance.  So imagine something like `.in_span("foobar", { "baz" => 1.minute }`. When we viewed the span in various destinations, it the attribute showed up with the value as `nil` (or `null` or what have you). We saw the symptom (null attribute value, which is not allowed per the otel spec) before we sussed out the problem.

Turns out that, when exporting, the otlp exporter checks for a whitelist of exportable value classes that are pretty basic (code [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L382-L398)). As a result, a value with class `ActiveSupport::Duration` will be `nil`.

This patch should ensure that users will get a warning if they use an unsupported `Numeric` class, and that the invalid attribute k/v pair will not be exported.

I didn't see the need to do this, but we could also potentially:

* log a warning from the OTLP exporter about this
* allow other numeric classes (BigDecimal, etc.)

### DIFF
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -21,7 +21,7 @@ module OpenTelemetry
       end
 
       def numeric?(value)
-        value.is_a?(Integer) || value.is_a?(Float)
+        value.instance_of?(Integer) || value.instance_of?(Float)
       end
 
       def valid_simple_value?(value)
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'BigDecimal'
 module OpenTelemetry
   module SDK
     # @api private
@@ -20,8 +21,12 @@ module OpenTelemetry
         key.instance_of?(String)
       end
 
+      def valid_numeric?(value)
+        value.is_a?(Integer) || value.is_a?(Float) || value.is_a?(Rational) || value.is_a?(Complex) || value.is_a?(BigDecimal)
+      end
+
       def valid_simple_value?(value)
-        value.instance_of?(String) || value == false || value == true || value.is_a?(Numeric)
+        value.instance_of?(String) || value == false || value == true || valid_numeric?(value)
       end
 
       def valid_array_value?(value)
@@ -34,7 +39,7 @@ module OpenTelemetry
         when TrueClass, FalseClass
           value.all? { |v| boolean?(v) }
         when Numeric
-          value.all? { |v| v.is_a?(Numeric) }
+          value.all? { |v| valid_numeric?(v) }
         else
           false
         end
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -20,8 +20,12 @@ def valid_key?(key)
         key.instance_of?(String)
       end
 
+      def valid_numeric?(value)
```
fbogsany, line 23: Maybe try to be consistent with `boolean?` above?
```suggestion
      def numeric?(value)
```

sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -20,8 +20,12 @@ def valid_key?(key)
...
+        value.is_a?(Integer) || value.is_a?(Float)
```
fbogsany, line 24: Two things:
1. I wonder if we should explicitly use `.instance_of?` to restrict values to instances of these specific classes, in case someone tries to subclass them & that 💥 in `protobuf` serialization later.
2. What's the current state of `BigInteger`? IIRC it was merged with `Integer` some releases of MRI ago. I'm not sure whether we should check for it here or if we can now comfortably assume `value.is_a?(Integer)` covers that.

sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -20,8 +20,12 @@ def valid_key?(key)
...
+      end
+
       def valid_simple_value?(value)
-        value.instance_of?(String) || value == false || value == true || value.is_a?(Numeric)
+        value.instance_of?(String) || value == false || value == true || valid_numeric?(value)
```
fbogsany, line 28: Further to the `instance_of?` vs `is_a?` question, we are using `==` here for booleans, which *may* be too loose a constraint. We also have `boolean?` defined above in terms of `is_a?` and used in `valid_array_value?` below. Should we instead use `instance_of?` in `boolean?` and use `boolean?(value)` here?

sdk/lib/opentelemetry/sdk/internal.rb
plantfansam: 100% agree with moving to `instance_of?` - that's a great idea.

plantfansam: I think `Bignum` and `Fixnum` are kaput as of 2.4.0: https://github.com/ruby/ruby/blob/master/doc/ChangeLog-2.4.0#L4927-L4963

plantfansam: changed from `is_a?` to `instance_of?` here 0c1f281cb7980f63e037a50dd809f4fdf627b1ed

robertlaurin (approved):

PR #1175, opened 2022-04-07
Add `with_attributes` context propagation for mysql2 instrumentation

plantfansam:
As in `Redis` and `HTTP::ClientContext`, I'd like to use `with_attributes` to assign arbitrary attributes to spans generated within the `mysql2` instrumentation. Test coverage is light on the `*attributes` methods in `Mysql2` because it's a lightly modified copy/paste of [the redis instrumentation's implementation](https://github.com/open-telemetry/opentelemetry-ruby/blob/b7149fb72cad6c45956a6be80acfb2d062aab3a6/instrumentation/redis/lib/opentelemetry/instrumentation/redis.rb#L14). Happy to fill in test coverage if folks would prefer.

### DIFF
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2.rb
@@ -11,6 +11,45 @@ module OpenTelemetry
   module Instrumentation
     # Contains the OpenTelemetry instrumentation for the Mysql2 gem
     module Mysql2
+      extend self
+
+      CURRENT_ATTRIBUTES_KEY = Context.create_key('mysql-attributes-hash')
+
+      private_constant :CURRENT_ATTRIBUTES_KEY
+
+      # Returns the attributes hash representing the Mysql2 context found
+      # in the optional context or the current context if none is provided.
+      #
+      # @param [optional Context] context The context to lookup the current
+      #   attributes hash. Defaults to Context.current
+      def attributes(context = nil)
+        context ||= Context.current
+        context.value(CURRENT_ATTRIBUTES_KEY) || {}
+      end
+
+      # Returns a context containing the merged attributes hash, derived from the
+      # optional parent context, or the current context if one was not provided.
+      #
+      # @param [optional Context] context The context to use as the parent for
+      #   the returned context
+      def context_with_attributes(attributes_hash, parent_context: Context.current)
+        attributes_hash = attributes(parent_context).merge(attributes_hash)
+        parent_context.set_value(CURRENT_ATTRIBUTES_KEY, attributes_hash)
+      end
+
+      # Activates/deactivates the merged attributes hash within the current Context,
+      # which makes the "current attributes hash" available implicitly.
+      #
+      # On exit, the attributes hash that was active before calling this method
+      # will be reactivated.
+      #
+      # @param [Span] span the span to activate
+      # @yield [Hash, Context] yields attributes hash and a context containing the
+      #   attributes hash to the block.
+      def with_attributes(attributes_hash)
+        attributes_hash = attributes.merge(attributes_hash)
+        Context.with_value(CURRENT_ATTRIBUTES_KEY, attributes_hash) { |c, h| yield h, c }
+      end
     end
   end
 end
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -58,9 +58,10 @@ module OpenTelemetry
             when :obfuscate
               attributes['db.statement'] = obfuscate_sql(sql)
             end
+
             tracer.in_span(
               database_span_name(sql),
-              attributes: attributes,
+              attributes: attributes.merge(OpenTelemetry::Instrumentation::Mysql2.attributes),
               kind: :client
             ) do
               super(sql, options)
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -58,7 +58,6 @@ module OpenTelemetry
             when :obfuscate
               attributes['db.statement'] = obfuscate_sql(sql)
             end
-
             tracer.in_span(
               database_span_name(sql),
               attributes: attributes.merge(OpenTelemetry::Instrumentation::Mysql2.attributes),
--- a/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
+++ b/instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
@@ -60,7 +60,7 @@ module OpenTelemetry
             end
             tracer.in_span(
               database_span_name(sql),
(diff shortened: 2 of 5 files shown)


instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2/patches/client.rb
```diff
@@ -60,7 +60,7 @@ def query(sql, options = {})
             end
             tracer.in_span(
               database_span_name(sql),
-              attributes: attributes,
+              attributes: attributes.merge(OpenTelemetry::Instrumentation::Mysql2.attributes),
```
robertlaurin, line 63:
```suggestion
              attributes: attributes.merge!(OpenTelemetry::Instrumentation::Mysql2.attributes),
```

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2.rb
```diff
@@ -11,6 +11,45 @@ module OpenTelemetry
   module Instrumentation
     # Contains the OpenTelemetry instrumentation for the Mysql2 gem
     module Mysql2
+      extend self
+
+      CURRENT_ATTRIBUTES_KEY = Context.create_key('mysql-attributes-hash')
```
robertlaurin, line 16: I wonder if we should change the redis context key to be more specific as well. 

This is what context storage looks like in a block like this
```rb
OpenTelemetry::Instrumentation::Redis.with_attributes({ 'foo' => 'bar' }) do
  OpenTelemetry::Common::HTTP::ClientContext.with_attributes({'bar' => 'foo'}) do
    binding.pry
  end
end
```

```rb
=> [#<OpenTelemetry::Context:0x00000001071447b0
  @entries=
   {#<OpenTelemetry::Context::Key:0x000000012729d448
     @name="current-attributes-hash">=>{"foo"=>"bar"}}>,
 #<OpenTelemetry::Context:0x0000000107144698
  @entries=
   {#<OpenTelemetry::Context::Key:0x000000012729d448
     @name="current-attributes-hash">=>{"foo"=>"bar"},
    #<OpenTelemetry::Context::Key:0x000000012711f6e8
     @name="current-attributes-hash">=>{"bar"=>"foo"}}>]
```

I've never actually had to debug this ever, but it might be a kindness to future us if we ever find ourselves in that predicament. 

(I don't expect that change to be included in this PR just wanted to raise that thought here where its relevant.)

plantfansam: Yeah, I think this is a great point! Will address in separate PR.

arielvalentin:
Is the direction we are headed here having a `with_attributes` for each instrumentation? 

Have you all considered adding this method dynamically to each child that derives from the `Base` class and registering a Context Key using the `instrumentation_name`?

plantfansam:
@arielvalentin great question! AFAIK we're not planning on having `with_attributes` for every instrumentation package, but I don't think that's a bad idea. 

When I first was working on this, I implemented a 👋 metaprogrammy approach vaguely along the lines you suggested 👋 , but after talking with @robertlaurin, we decided to hold off on implementing the metaprogrammy style until we were sure that more instrumentation packages were going to use `with_attributes` (see discussion [here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1157#issuecomment-1088949729)). 

I think the approach you describe will likley be optimal if/when we decide to add the `with_attributes` functionality to every instrumentation class! Requiring an `extend` on every class (which my previous PR did) seems unnecessary if we're gonna do this everywhere...

arielvalentin:
Ok. Then maybe I am bringing this up in the wrong forum. 

@robertlaurin mind if I break this discussion off into an issue?

robertlaurin (approved):

robertlaurin:
> Ok. Then maybe I am bringing this up in the wrong forum.
> 
> @robertlaurin mind if I break this discussion off into an issue?

By all means!

robertlaurin:
```
opentelemetry-instrumentation-mysql2: rubocop
RuboCop failed!
Running RuboCop...
Inspecting 15 files
.............C.

Offenses:

test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:68:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.
test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:71:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.
test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:84:24: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
          client.query("SELECT 1")
                       ^^^^^^^^^^
test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:87:55: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
        _(span.attributes['db.statement']).must_equal "foobar" 
                                                      ^^^^^^^^
test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:87:63: C: Layout/TrailingWhitespace: Trailing whitespace detected.
        _(span.attributes['db.statement']).must_equal "foobar" 
                                                              ^
```

plantfansam:
> ```
> opentelemetry-instrumentation-mysql2: rubocop
> RuboCop failed!
> Running RuboCop...
> Inspecting 15 files
> .............C.
> 
> Offenses:
> 
> test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:68:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.
> test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:71:1: C: Layout/TrailingWhitespace: Trailing whitespace detected.
> test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:84:24: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
>           client.query("SELECT 1")
>                        ^^^^^^^^^^
> test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:87:55: C: Style/StringLiterals: Prefer single-quoted strings when you don't need string interpolation or special symbols.
>         _(span.attributes['db.statement']).must_equal "foobar" 
>                                                       ^^^^^^^^
> test/opentelemetry/instrumentation/mysql2/instrumentation_test.rb:87:63: C: Layout/TrailingWhitespace: Trailing whitespace detected.
>         _(span.attributes['db.statement']).must_equal "foobar" 
>                                                               ^
> ```

🤦  my bad. 075c27b022fd0f41be2053a4b470955b421d4357 and a86b21a7dce1d9420511c9da80875bb62987ae11
 _should_ fix it.

fbogsany:
> Is the direction we are headed here having a `with_attributes` for each instrumentation?
> 
> Have you all considered adding this method dynamically to each child that derives from the `Base` class and registering a Context Key using the `instrumentation_name`?

The pattern so far has **not** been `with_attribute` for each instrumentation. We have instead done it for categories of instrumentation, like "things that use Redis" and "HTTP clients". The first of those is instrumentation-specific. The second is very much not instrumentation-specific.

It is also not yet clear that this is something we want everywhere. At this point, we have 3 examples. We should consider bringing it up at the spec level, perhaps in the instrumentation WG (if that's still a thing @robertlaurin?), before generalizing it across all instrumentation.

instrumentation/mysql2/lib/opentelemetry/instrumentation/mysql2.rb
plantfansam: https://github.com/open-telemetry/opentelemetry-ruby/pull/1186

PR #1179, opened 2022-04-11
Split otlp exporter

robertlaurin:
This PR:
- Creates a new gem for the otlp exporter explicitly using http protocol
- Extracts the OTLP encoding methods into a common gem, with the intention of pulling out other common behaviour here
- Sets us up to introduce a otlp over grpc exporter https://github.com/open-telemetry/opentelemetry-ruby/pull/1180


exporter/otlp-common/README.md
```diff
@@ -0,0 +1,102 @@
+# opentelemetry-exporter-otlp
```
fbogsany, line 1: `opentelemetry-exporter-otlp-common` needs its own readme.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
```diff
@@ -0,0 +1,144 @@
...
+        extend self
+
+        def encode(span_data) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+          Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.encode(as_etsr(span_data))
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: 'unexpected error in OTLP::Exporter#encode')
+          nil
+        end
+
+        def as_etsr(span_data)
```
fbogsany, line 30: Does this need to be public? For tests? Could we mark this API private with a comment?

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
```diff
@@ -0,0 +1,144 @@
...
+  module Exporter
+    module OTLP
+      # Contains common functionality between the different OTLP export protocols
+      module Common # rubocop:disable Metrics/ModuleLength
+        extend self
+
+        def encode(span_data) # rubocop:disable Metrics/AbcSize, Metrics/MethodLength
+          Opentelemetry::Proto::Collector::Trace::V1::ExportTraceServiceRequest.encode(as_etsr(span_data))
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: 'unexpected error in OTLP::Exporter#encode')
```
fbogsany, line 26:
```suggestion
          OpenTelemetry.handle_error(exception: e, message: 'unexpected error in OTLP::Common#encode')
```

exporter/otlp-http/CHANGELOG.md
```diff
@@ -0,0 +1,141 @@
+# Release History: opentelemetry-exporter-otlp
```
fbogsany, line 1: Same thing here - let's start with a fresh changelog whose first entry refers to the old one.

exporter/otlp-http/Gemfile
```diff
@@ -0,0 +1,21 @@
...
+
+source 'https://rubygems.org'
+
+gemspec
+
+# Use the opentelemetry-api gem from source
+gem 'opentelemetry-api', path: '../../api'
+gem 'opentelemetry-common', path: '../../common'
+gem 'opentelemetry-exporter-otlp-common', path: '../otlp-common'
+gem 'opentelemetry-instrumentation-base', path: '../../instrumentation/base'
```
fbogsany, line 15: Is this necessary for this gem? Is `opentelemetry-common` needed?

exporter/otlp-http/README.md
```diff
@@ -0,0 +1,102 @@
+# opentelemetry-exporter-otlp
```
fbogsany, line 1: `opentelemetry-exporter-otlp-http` needs its own README, or at least we need to replace all the instances of `opentelemetry-exporter-otlp` in this file.

exporter/otlp-http/Rakefile
```diff
@@ -0,0 +1,45 @@
...
+  'collector/trace/v1/trace_service.proto'
+].freeze
+
+task :update_protobuf do
+  system('git clone https://github.com/open-telemetry/opentelemetry-proto')
+  PROTOBUF_FILES.each do |file|
+    system("protoc --ruby_out=lib/ --proto_path=opentelemetry-proto opentelemetry/proto/#{file}")
+  end
+  system('rm -rf opentelemetry-proto')
+end
```
fbogsany, line 45: Do we need this anymore? I think this functionality is in the `otlp-common` gem now.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
```diff
@@ -0,0 +1,303 @@
...
+          # Default timeouts in seconds.
+          KEEP_ALIVE_TIMEOUT = 30
+          RETRY_COUNT = 5
+          WRITE_TIMEOUT_SUPPORTED = Gem::Version.new(RUBY_VERSION) >= Gem::Version.new('2.6')
+          private_constant(:KEEP_ALIVE_TIMEOUT, :RETRY_COUNT, :WRITE_TIMEOUT_SUPPORTED)
+
+          ERROR_MESSAGE_INVALID_HEADERS = 'headers must be a String with comma-separated URL Encoded UTF-8 k=v pairs or a Hash'
+          private_constant(:ERROR_MESSAGE_INVALID_HEADERS)
+
+          def self.ssl_verify_mode
```
fbogsany, line 35: Was this method always public? Why?

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
```diff
@@ -0,0 +1,303 @@
...
+          #
+          # @param [Enumerable<OpenTelemetry::SDK::Trace::SpanData>] span_data the
+          #   list of recorded {OpenTelemetry::SDK::Trace::SpanData} structs to be
+          #   exported.
+          # @param [optional Numeric] timeout An optional timeout in seconds.
+          # @return [Integer] the result of the export.
+          def export(span_data, timeout: nil)
+            return FAILURE if @shutdown
+
+            send_bytes(OpenTelemetry::Exporter::OTLP::Common.encode(span_data), timeout: timeout)
```
fbogsany, line 86: Do you prefer this to including the `Common` module in this class?

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
robertlaurin: Maybe I'll merge the GRPC PR back into this one so it's easier to see. But GRPC handles the encoding/decoding itself, so we need to expose the proto format for the GRPC exporter.

https://github.com/open-telemetry/opentelemetry-ruby/pull/1180/files#diff-8f036f2cfa248f9a798edf1f55e2d7f19fbdac08dcc8625c9a980ef8df026d99R52

exporter/otlp-http/Rakefile
robertlaurin: Will remove

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
robertlaurin: It appears to be currently public, https://github.com/open-telemetry/opentelemetry-ruby/blob/c521837879e38629b2facb8d01bc19cc784dbfc3/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L38-L46

Not sure why.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
robertlaurin: I do prefer this over including and having the implicit availability of an `encode` method.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common.rb
fbogsany: Perhaps we can rename `encode` to `encode_spans` or something similar (if we end up wanting to add `as_emsr` and `encode_metrics` for metrics support)?

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
fbogsany: It should be private IMO - there's no reason I can see in the original PR for making it public. There's talk of extracting it to some common place for various exporters to use, but that hasn't been done. Let's make it private for now.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb
fbogsany: Dummy question: which one is more performant? D'you know if this is encoded as 4 constant lookups followed by a send?

fbogsany: Conclusion is that it doesn't matter, both because this isn't a hot path and because the chain of constant lookups is cached.

exporter/otlp-http/Gemfile
robertlaurin: They get pulled in implicitly from the SDK regardless.  But no we should not need them explicitly here with the decoupling changes being made for extracting instrumentation in the contrib repo.

robertlaurin:
### Other languagues

Open question about package structure.

If we look at [Otel Go](https://github.com/open-telemetry/opentelemetry-go/tree/main/exporters/otlp) they appear to have four distinct packages for metrics, traces, and otlp over grpc/http.

```
go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetrichttp
go.opentelemetry.io/otel/exporters/otlp/otlpmetric/otlpmetricgrpc

go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc
go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracehttp
```

[Otel JS](https://github.com/open-telemetry/opentelemetry-js/tree/main/experimental/packages) seems to be doing something similar.
```
@opentelemetry/exporter-trace-otlp-grpc
@opentelemetry/exporter-trace-otlp-http
@opentelemetry/exporter-trace-otlp-proto

@opentelemetry/exporter-metrics-otlp-grpc
@opentelemetry/exporter-metrics-otlp-http
@opentelemetry/exporter-metrics-otlp-proto
```

### Our implementation

At risk of having to rename a bunch of things on this PR. What do we want to see for our exporters?

Option A) We have gems that follow the structure
- `opentelemetry-exporter-otlp-http` (which includes both trace and metrics exporters)
- `opentelemetry-exporter-otlp-grpc` (which includes both trace and metrics exporters)
- `opentelemetry-exporter-otlp-common` (which includes encoding support for metrics and tracing)

In this scenario you would have a classes like:
- `OpenTelemetry::Exporter::OTLP::HTTP::TraceExporter`
- `OpenTelemetry::Exporter::OTLP::HTTP::MetricsExporter`


Option B)
- `opentelemetry-exporter-trace-otlp-http` (which includes trace exporter)
- `opentelemetry-exporter-trace-otlp-grpc` (which includes trace exporter)
- `opentelemetry-exporter-trace-otlp`      (which includes encoding support for tracing)

- `opentelemetry-exporter-metrics-otlp-http` (which includes metrics exporter)
- `opentelemetry-exporter-metrics-otlp-grpc` (which includes metrics exporter)
- `opentelemetry-exporter-metrics-otlp`      (which includes encoding support for tracing)

In this scenario you would have a class like:
- `OpenTelemetry::Exporter::OTLP::HTTP::Trace::Exporter`
- or `OpenTelemetry::Exporter::OTLP::Trace::HTTP::Exporter`

I originally was more inclined to pursue Option A, however if many of the other languague implementations are following a similar structure I think that's reason enough for us to be consistent.

#### So what do we think?
@open-telemetry/ruby-approvers @open-telemetry/ruby-maintainers @open-telemetry/ruby-contrib-maintainers

fbogsany:
A couple of 🤔 

1. We're going to end up with exporters for logging as well, and any other signals that might be added in future. This has the potential to 🤯 the cardinality of OTLP exporter packages with Option B.
2. With Option A, we need to support metrics export while developing the metrics SDK. It isn't clear to me whether we can do this in the proposed packages - we'll end up bumping the OTLP exporter package versions as we figure out the SDK's interfaces, and we'd be introducing a dependency between the exporter and the metrics SDK even if the user only wanted trace export.

I'm strongly in favour of as few packages as possible, but I think other languages have chosen Option B because of the signal stability concern.

robertlaurin:
> 1. We're going to end up with exporters for logging as well, and any other signals that might be added in future. This has the potential to 🤯 the cardinality of OTLP exporter packages with Option B.

That's a valid concern. I earnestly didn't even consider the logging exporters (sorry logs).

> With Option A, we need to support metrics export while developing the metrics SDK. It isn't clear to me whether we can do this in the proposed packages - we'll end up bumping the OTLP exporter package versions as we figure out the SDK's interfaces, and we'd be introducing a dependency between the exporter and the metrics SDK even if the user only wanted trace export.

It may end up being more effort but I think we should take the Option A approach, but we can address the signal stability concerns by having the metrics/logs exporter be a separate package that is merged into the "main" one once the signal is stable.  Similar to how we have the metrics api/sdk as a distinct gems until they stabilize.

So I think for further changes to this PR would be:
- Refine the exporter namespace `OpenTelemetry::Exporter::OTLP::HTTP::Exporter` -> `OpenTelemetry::Exporter::OTLP::HTTP::TraceExporter` or `OpenTelemetry::Exporter::OTLP::HTTP::Trace`

exporter/otlp-common/CHANGELOG.md
```diff
@@ -0,0 +1,141 @@
+# Release History: opentelemetry-exporter-otlp
```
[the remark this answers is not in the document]
robertlaurin: Done 👍

ahayworth:
> It may end up being more effort but I think we should take the Option A approach, but we can address the signal stability concerns by having the metrics/logs exporter be a separate package that is merged into the "main" one once the signal is stable. Similar to how we have the metrics api/sdk as a distinct gems until they stabilize.

Between the two options you listed, this is how I would personally prefer it - Option A, and we put metrics in an `experimental` package of some kind until we're ready for it to be stable.

exporter/otlp-http/README.md
```diff
@@ -0,0 +1,102 @@
+# opentelemetry-exporter-otlp-http
+
+The `opentelemetry-exporter-otlp-http` gem provides an [OTLP](https://github.com/open-telemetry/opentelemetry-proto) over http exporter for OpenTelemetry for Ruby. Using `opentelemetry-exporter-otlp-http`, an application can configure OpenTelemetry to export collected tracing data to [the OpenTelemetry Collector][opentelemetry-collector-home].
```
fbogsany, line 3:
```suggestion
The `opentelemetry-exporter-otlp-http` gem provides an [OTLP](https://github.com/open-telemetry/opentelemetry-proto) over HTTP exporter for OpenTelemetry for Ruby. Using `opentelemetry-exporter-otlp-http`, an application can configure OpenTelemetry to export collected tracing data to [the OpenTelemetry Collector][opentelemetry-collector-home].
```

exporter/otlp-http/README.md
```diff
@@ -0,0 +1,102 @@
...
+gem install opentelemetry-exporter-otlp-http
+```
+
+Or, if you use [bundler][bundler-home], include `opentelemetry-sdk` in your `Gemfile`.
+
+Then, configure the SDK to use the OTLP exporter as a span processor, and use the OpenTelemetry interfaces to produces traces and other information. Following is a basic example.
+
+```ruby
+require 'opentelemetry/sdk'
+require 'opentelemetry/exporter/otlp'
```
fbogsany, line 36: Does this still work or should it be:
```suggestion
require 'opentelemetry/exporter/otlp/http'
```
I suppose the next question would be: if the old example doesn't work anymore, that seems like a breaking change - do we want that or should we make the old way work (but still fix the example)?

plantfansam:
> 

I agree with @ahayworth here in support of option A, with any unstable signals cordoned off into an experimental package of some sort.

exporter/otlp-http/README.md
fbogsany: Ignore the second part of my comment. I think the suggestion is relevant, tho

fbogsany (approved):

PR #1213, opened 2022-05-03
chore: Remove Local Path Gem Dependencies

arielvalentin:
This is the first of many steps to decouple the contrib packages from the main repo.

It only covers the instrumentation directory and does not attempt to address inconsistencies in Gemfiles and gemspecs on how it uses dependencies like pry. 

See https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1


ahayworth (approved):
I mostly just verified that there don't seem to be immediately wrong things, but I think this looks fine to me.

arielvalentin:
@chrisholmes If you have some time I could use your help looking into the rspec auto-instrumentation. It seems that in CI mode it runs _without_ colors but _always_ colorizes output when run locally. I could not figure out the incantation to disable coloring and was hoping you had some insight into why this is failing all of the sudden.

https://github.com/open-telemetry/opentelemetry-ruby/actions/runs/2266028975

instrumentation/all/opentelemetry-instrumentation-all.gemspec
```diff
@@ -61,6 +61,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-instrumentation-sinatra', '~> 0.19.4'
   spec.add_dependency 'opentelemetry-instrumentation-trilogy', '~> 0.50.0'
 
+  spec.add_development_dependency 'active_model_serializers'
+  spec.add_development_dependency 'activesupport'
```
robertlaurin, line 65: These are showing up above in the run time dependencies and here in the dev dependencies.  Is there a reason for that?  Just want to make sure I'm not missing something nuanced here.

robertlaurin (approved):
One question, and then it's good to go.

instrumentation/all/opentelemetry-instrumentation-all.gemspec
arielvalentin: I relocated them from the `Gemfile` to the `gemspec` so I assumed they were only needed for tests but now that you point it out I am going to double check if they are needed at all. 

I do not see them in the runtime dependencies. Where did you see them?

https://github.com/arielvalentin/opentelemetry-ruby/blob/remove-local-path-deps/instrumentation/all/opentelemetry-instrumentation-all.gemspec

robertlaurin: 🤦‍♂️ , ignore me.  As it turns out `opentelemetry-instrumentation-active_support != activesupport`

instrumentation/action_pack/Gemfile
```diff
@@ -8,15 +8,8 @@ source 'https://rubygems.org'
 
 gemspec
 
-gem 'opentelemetry-api', path: '../../api'
-
 group :test do
-  gem 'byebug'
-  gem 'opentelemetry-common', path: '../../common'
   gem 'opentelemetry-instrumentation-base', path: '../../instrumentation/base'
   gem 'opentelemetry-instrumentation-rack', path: '../../instrumentation/rack'
```
plantfansam, line 13: I'm sure I'm missing something obvious, but why leave the relative import of `instrumentation/rack`?

arielvalentin: It is non-obvious IMO. The reason is since `action_pack` depends on the `rack` instrumentation it is easier to test changes against the latest version in the repo instead of only relying on a released version.

PR #1218, opened 2022-05-04
Drop support for MRI Ruby 2.5 and 2.6 (EOL)

ahayworth:
MRI Ruby 2.5 and 2.6 have reached EOL, and we can thus remove support for them. However, jruby still targets `2.6.8` for compatibility. Thus, we officially remove support for MRI Ruby 2.5 and 2.6, but maintain best-effort 2.6 compatibility for jruby's sake. We may revisit this decision in the future if this becomes too difficult.

This PR accomplishes the following:
- Updating the minimum ruby requirements in our Gemfiles to `2.6.0` throughout.
- Updating our Rubocop configuration to target `2.6.0` for style.
- Removing a small compatibility shim around `Random.bytes`.
- Updating docs in a few places to note this change.

Ruby 2.6 final release and EOL notice: https://www.ruby-lang.org/en/news/2022/04/12/ruby-2-6-10-released/
Ruby 2.5 final release and EOL notice: https://www.ruby-lang.org/en/news/2021/04/05/ruby-2-5-9-released/


arielvalentin (approved):
🧨

ahayworth:
Ah .. well, this would force us to drop `jruby` altogether...

PR #1222, opened 2022-05-04
chore: Stabilize ActiveJob Instrumentation Tests

arielvalentin:
Prior to this change tests would intermittently fail due to threads executing out of order.

This change ensures that the AsyncAdapter runs in the main thread and blocks execution to
ensure jobs are completed before making assertions on spans.


instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
```diff
@@ -179,22 +169,16 @@
         TestJob.perform_later
 
         [send_span, process_span].each do |span|
-          _(span.attributes['messaging.system']).must_equal('inline')
+          _(span.attributes['messaging.system']).must_equal('async')
```
plantfansam, line 172: I think we should do 

```ruby
      ::ActiveJob::Base.queue_adapter.shutdown rescue nil
      ::ActiveJob::Base.queue_adapter = :inline
```

here to retain the test's meaning.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
```diff
@@ -205,20 +189,10 @@
...
 
       it 'tracks correctly for jobs that do retry' do
-        ::ActiveJob::Base.queue_adapter = :async
-
-        RetryJob.perform_now
-        ::ActiveJob::Base.queue_adapter.shutdown
-
-        # 1 enqueue, 2 perform
-        _(spans.count).must_equal(3)
+        RetryJob.perform_later rescue nil
```
plantfansam, line 192: Why `perform_later` instead of `perform_now`?

instrumentation/active_job/test/test_helper.rb
```diff
@@ -78,4 +78,5 @@ def initialize(*)
 OpenTelemetry::SDK.configure do |c|
   c.use 'OpenTelemetry::Instrumentation::ActiveJob'
   c.add_span_processor span_processor
+  c.logger = Logger.new(File::NULL)
```
plantfansam, line 81: Is this meant to be in the PR?

plantfansam (commented):
I haven't waded too deeply into the ActiveJob docs, yet, so feel free to tell me to buzz off and read those! A few questions:

* why set the queue adapter to `async` before, then reset to `inline` after?
* why rescue nil after `queue_adapter.shutdown`? 
* why rescue nil after `RetryJob.perform_later`?

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
arielvalentin: `perform_now` executed the job in the main thread the first time it ran and then every subsequent run was being added to the queue being operated on by the thread pool. 

That use case did not make sense to me anymore now that I am testing it as if the job were enqueued and all instances were being picked up by a downstream worker.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
arielvalentin: uggg I got too aggressive here. I will roll it back!

instrumentation/active_job/test/test_helper.rb
arielvalentin: Yes. The diagnostic logs in the test case are unhelpful and noisy. It think we should make this the default for all tests.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
arielvalentin: Done!

arielvalentin:
> why set the queue adapter to async before, then reset to inline after?

Since the queue_adapter is a global variable, I reset it to `:inline` so it does not bleed into other tests that may run after this one completes. 

> why rescue nil after queue_adapter.shutdown?

I do not want the test suite to halt if concurrent ruby gets an error. It is not significant to me. Also, the method does not exist on the inline adapter so if I switch it for another test this will fail with a `NoMethodError`.

> why rescue nil after RetryJob.perform_later?

Because when running in the `main` thread the last execution will raise an error. In most queuing systems the exception is handled and thrown into a dead letter queue. When running the ImmediateExecutor it propagates the error to the caller. In this case it is the test case.

instrumentation/active_job/test/test_helper.rb
ahayworth: > It think we should make this the default for all tests.

I would support that if it were behind some kind of flag to turn it on or off globally (for tests). They are noisy and not usually useful, but then sometimes when the tests are broken you do want them. That tripped us up in https://github.com/open-telemetry/opentelemetry-ruby/pull/1136, for example.

instrumentation/active_job/test/instrumentation/active_job/patches/active_job_callbacks_test.rb
```diff
@@ -205,20 +204,14 @@
...
-        _(span.attributes['messaging.active_job.executions']).must_equal(2)
+        begin
+          RetryJob.perform_later
+        rescue StandardError
+          nil
+        end
 
-      ensure
-        ::ActiveJob::Base.queue_adapter = :inline
+        executions = spans.filter { |s| s.kind == :consumer }.map { |s| s.attributes['messaging.active_job.executions'] }.sum
```
ahayworth, line 213: An absolutely minor nit for your consideration:

```suggestion
        executions = spans.filter { |s| s.kind == :consumer }.sum { |s| s.attributes['messaging.active_job.executions'] }
```

```ruby
[1] pry(main)> a = [ { foo: 1 }, { foo: 2 }, { foo: 3 } ]
=> [{:foo=>1}, {:foo=>2}, {:foo=>3}]
[2] pry(main)> a.sum { |s| s[:foo] }
=> 6
[3] pry(main)>
```

ahayworth (approved):
I left a minor suggestion, mostly because I just learned recently that you can actually pass a block to `sum` and I wanted to share the knowledge.

Otherwise though, I'm absolutely fine with this! I truly appreciate that you came in and cleaned up the flakiness here. (Which was 100% my fault 😢)

This kind of work is important for the long-term health of a project, and your willingness to do it says a lot about how much you care about it. Thank you! ❤️

instrumentation/active_job/test/test_helper.rb
arielvalentin: That would be OTEL_LOG_LEVEL 🤦

arielvalentin: Fixed in 9e429b11146a12dda68274fa63a428a242d52275

plantfansam: Ty for explaining @arielvalentin. Setting `OTEL_LOG_LEVEL` seems reasonable to me!

plantfansam:
> Because when running in the `main` thread the last execution will raise an error. In most queuing systems the exception is handled and thrown into a dead letter queue. When running the ImmediateExecutor it propagates the error to the caller. In this case it is the test case.

Thanks so much for the detailed explanation! This all makes sense to me 😄

robertlaurin (approved):

PR #1230, opened 2022-05-06
fix: SpanLimits setting event attributes length limit

robertlaurin:
The spec allows for generic and more specific configuration of attribute length limits. https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/README.md#attribute-limits

However we are currently applying legacy and span specific attribute environment variables to the event attribute length limit.

`OTEL_EVENT_ATTRIBUTE_VALUE_LENGTH_LIMIT` should take priority over `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT`.

`OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` and `OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` should not configure the event attributes length limit.

Both span attribute length limit and event attribute length limit should use the more general `OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT` in the event a more specific configuration is not supplied.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -137,7 +137,7 @@ module OpenTelemetry
         #
         # @return [self] returns itself
         def add_event(name, attributes: nil, timestamp: nil)
-          event = Event.new(name, truncate_attribute_values(attributes), relative_timestamp(timestamp))
+          event = Event.new(name, truncate_attribute_values(attributes, @span_limits.event_attribute_length_limit), relative_timestamp(timestamp))
 
           @mutex.synchronize do
             if @ended
@@ -349,14 +349,12 @@ module OpenTelemetry
 
           excess = attrs.size - @span_limits.attribute_count_limit
           excess.times { attrs.shift } if excess.positive?
-          truncate_attribute_values(attrs)
+          truncate_attribute_values(attrs, @span_limits.attribute_length_limit)
           nil
         end
 
-        def truncate_attribute_values(attrs)
+        def truncate_attribute_values(attrs, attribute_length_limit)
           return EMPTY_ATTRIBUTES if attrs.nil?
-
-          attribute_length_limit = @span_limits.attribute_length_limit
           return attrs if attribute_length_limit.nil?
 
           attrs.transform_values! { |value| OpenTelemetry::Common::Utilities.truncate_attribute_value(value, attribute_length_limit) }
--- a/sdk/lib/opentelemetry/sdk/trace/span_limits.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span_limits.rb
@@ -24,6 +24,9 @@ module OpenTelemetry
         # The global default max number of attributes per {OpenTelemetry::SDK::Trace::Event}.
         attr_reader :event_attribute_count_limit
 
+        # The global default max length of attribute value per {OpenTelemetry::SDK::Trace::Event}.
+        attr_reader :event_attribute_length_limit
+
         # The global default max number of attributes per {OpenTelemetry::Trace::Link}.
         attr_reader :link_attribute_count_limit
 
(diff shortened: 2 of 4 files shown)


sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -343,14 +343,12 @@ def trim_span_attributes(attrs)
...
-          truncate_attribute_values(attrs)
+          truncate_attribute_values(attrs, @span_limits.attribute_length_limit)
           nil
         end
 
-        def truncate_attribute_values(attrs)
+        def truncate_attribute_values(attrs, attribute_length_limit)
           return EMPTY_ATTRIBUTES if attrs.nil?
-
-          attribute_length_limit = @span_limits.attribute_length_limit
```
robertlaurin, line 353: I believe this should/could be considered a breaking change. However I don't know how flexible that is when it was doing the incorrect thing previously.  This line was applying span specific limits against the event attribute length.

sdk/lib/opentelemetry/sdk/trace/span_limits.rb
```diff
@@ -33,23 +36,26 @@ class SpanLimits
         # @raise [ArgumentError] if any of the max numbers are not positive.
         def initialize(attribute_count_limit: Integer(ENV.fetch('OTEL_SPAN_ATTRIBUTE_COUNT_LIMIT', 128)), # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
                        attribute_length_limit: ENV.fetch('OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT',
-                                                         ENV['OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT']),
+                                                         ENV['OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT'] || ENV['OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT']),
```
plantfansam, line 39: I _believe_ we want to switch the order of the fallback values in `fetch` here. So if it's currently `ENV['OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT'] || ENV['OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT']),`, we want it to be `ENV['OTEL_RUBY_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT']) || ENV['OTEL_ATTRIBUTE_VALUE_LENGTH_LIMIT']`. 

See:

```ruby
irb(main):001:0> nil || "foo" || "bar"
=> "foo"
```

This is based on the assumption that we want the `OTEL_RUBY_*` env vars to take precedence over `OTEL_*` env vars. Maybe we can add an explicit test case for this?

Side note: the Go SDK has a function dedicated to reconciling env vars in priority order that we _might_ want to emulate? https://github.com/open-telemetry/opentelemetry-go/pull/2677/files#diff-603faa80af2db68f66a19ef9a299eb128cd2e2d99c53506852d5107eb2a92a24R77-R97. I find the `ENV.fetch("foo", ENV["bar"] || ENV["baz"])` convention a bit hard to process mentally.

sdk/lib/opentelemetry/sdk/trace/span.rb
plantfansam: Yeah, for some definition of breaking 😂. Technically we broke the API defined by environment variables. IMO a minor version bump with a note in the changelog would be fine.

plantfansam (commented):
Seems great! Thanks for picking this up! I had a couple nits and one question about the ordering of env vars that is added as a code comment.

sdk/lib/opentelemetry/sdk/trace/span_limits.rb
fbogsany: In this case, the specced env var should trump the Ruby-specific one IMO.

sdk/lib/opentelemetry/sdk/trace/span.rb
plantfansam: I stand by my comment on May 9, which was approx 300 years ago. I think a minor version bump is fine!

plantfansam (approved):
LGTM!

sdk/lib/opentelemetry/sdk/trace/span_limits.rb
plantfansam: Either way is fine by me, so long as we document 😄

fbogsany (approved):

PR #1232, opened 2022-05-06
fix: Strip console codes from RSpec messages and better handle multiple failures

chrisholmes:
The instrumentation was including console codes in failure messages,
which was leading to some slightly brittle tests that were quick fixed [here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1219). We don't
need console codes so we can strip them out.

RSpec also wraps multiple failures and errors into a single exception. This
PR also handles these exceptions to provide clearer reporting of the
failures and errors.


robertlaurin (approved):

instrumentation/rspec/lib/opentelemetry/instrumentation/rspec/formatter.rb
```diff
@@ -96,6 +110,19 @@ def pop_and_finalize_span
...
+          error_message = strip_console_codes(error.message)
+          span.record_exception(error, attributes: { 'exception.message' => error_message })
+        end
+
+        def strip_console_codes(string)
+          string.gsub(/\e\[([;\d]+)?m/, '')
+        end
+
+        def multiple_failure_message(exception)
+          exception.failures.map(&:message).map(&method(:strip_console_codes)).join("\n\n")
```
plantfansam, line 124: TIL `&method(:methodname)`! 🎉

Issue #1234, opened 2022-05-09
DelayedJob Test Errors: `undefined method `silence_warnings' for Delayed::Worker:Class (NoMethodError)`


arielvalentin:
This appears to be an issue with `activesupport/lib/active_support/core_ext/kernel` `silence_warnings` not being loaded before setting up `Delayed::Worker.backed` 

https://github.com/collectiveidea/delayed_job_active_record/blob/v4.1.7/lib/delayed_job_active_record.rb#L7

~It is unclear to me if this is due to a change in recent release of Rails 7.~ 

The error also appears upstream when running tests against Rails Edge https://github.com/collectiveidea/delayed_job/runs/5790491817?check_suite_focus=true#step:4:11

https://github.com/open-telemetry/opentelemetry-ruby/runs/6353172039

```
opentelemetry-instrumentation-delayed_job: test
>> BUNDLE_GEMFILE=/home/runner/work/opentelemetry-ruby/opentelemetry-ruby/instrumentation/delayed_job/gemfiles/delayed_job_4.1.gemfile bundle exec rake test
/opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/delayed_job-4.1.10/lib/delayed/worker.rb:288: warning: method redefined; discarding old max_attempts
/opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/delayed_job-4.1.10/lib/delayed/worker.rb:22: warning: previous definition of max_attempts was here
/opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/delayed_job-4.1.10/lib/delayed/worker.rb:292: warning: method redefined; discarding old max_run_time
/opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/delayed_job-4.1.10/lib/delayed/worker.rb:22: warning: previous definition of max_run_time was here
/opt/hostedtoolcache/Ruby/3.0.4/x64/lib/ruby/gems/3.0.0/gems/delayed_job-4.1.10/lib/delayed/worker.rb:71:in `backend=': undefined method `silence_warnings' for Delayed::Worker:Class (NoMethodError)
[...]

PR #1236, opened 2022-05-09
fix: Rails 7.0.3 test suite incompatibility

arielvalentin:
It is possible that a transitive require statement was removed in this PR https://github.com/rails/rails/pull/44340/files

It is likely that in most Rails production workloads this will not be a problem since the Rails Railtie requires `kernel/reporting`.

https://github.com/rails/rails/blob/8ace32c4ccdc760d17548de777b08c13bc9e07bd/railties/lib/rails.rb#L8

Fixes: https://github.com/open-telemetry/opentelemetry-ruby/issues/1234
cc: https://github.com/collectiveidea/delayed_job/issues/1168

### DIFF
--- a/instrumentation/delayed_job/test/test_helper.rb
+++ b/instrumentation/delayed_job/test/test_helper.rb
@@ -4,6 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'active_support/core_ext/kernel/reporting'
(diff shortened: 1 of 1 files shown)


ahayworth (approved):
Nice! I was just gearing up to figure out why my PR wasn't building, and now I know. 😄

PR #1237, opened 2022-05-09
feat: Add basic support for schema_url

ahayworth:
This adds basic schema URL support in the API + SDK, specifically in and
around Resources, Tracers, Tracer Provider, and the OTLP exporter.

Some notes, in no particular order:

- I did not add schema URLs to any of our instrumentation, because that
  was a lot of work.
- The _API_ ProxyTracer tests around "asking for a tracer multiple times
  returns the same thing" don't actually work, because we don't do
  anything with the name+version(+schema url, after this change) that
  you pass in. I don't think that matters, but I'm not really sure.
- I didn't add the schema information to the jaeger/zipkin exporters,
  after surveying what Go and Python were doing. I don't see an easy
  place to add it in, either.

Questions:

1. The behavior of merging two resources with different, non-empty
   schema URLs is implementation dependent. I chose to _drop_ the schema
   URL in that case and continue merging resources as before. *Is this
   what we want?* Other SDKs do things differently (Go returns an error,
   Python logs an error and returns the old resource, etc).
2. I did *not* add support to the Configurator for setting the
   schema_url on the default Resource that gets initialized. That seemed
   like the kind of thing folks would do incorrectly, but you can still
   do such a thing by creating and assigning an entire Resource if you
   wish. *Is this what we want?*


ahayworth:
Tests should pass after #1236 is merged.

api/lib/opentelemetry/internal/proxy_tracer_provider.rb
```diff
@@ -37,21 +37,22 @@ def delegate=(provider)
...
 
       # Returns a {Tracer} instance.
       #
       # @param [optional String] name Instrumentation package name
       # @param [optional String] version Instrumentation package version
+      # @param [optional String] schema_url Schema URL to be recorded with traces
       #
       # @return [Tracer]
-      def tracer(name = nil, version = nil)
+      def tracer(name = nil, version = nil, schema_url = nil)
```
fbogsany, line 51: Can we make `schema_url` a named parameter rather than positional? At this point, I wish we'd made `version` an optional named parameter 😞

api/test/opentelemetry/trace/tracer_provider_test.rb
```diff
@@ -10,9 +10,10 @@
   let(:tracer_provider) { OpenTelemetry::Trace::TracerProvider.new }
 
   describe '.tracer' do
+    # TODO: This test doesn't really do anything for the API tracer
```
fbogsany, line 13: That's intentional 😄 - it just tests against the spec.

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -15,19 +15,22 @@ class << self
...
           # Returns a newly created {Resource} with the specified attributes
           #
-          # @param [Hash{String => String, Numeric, Boolean} attributes Hash of key-value pairs to be used
+          # @param [Hash{String => String, Numeric, Boolean}] attributes Hash of key-value pairs to be used
           #   as attributes for this resource
+          # @param [String] Specifies the Schema URL that should be recorded in the emitted resource.
           # @raise [ArgumentError] If attribute keys and values are not strings
           # @return [Resource]
-          def create(attributes = {})
+          def create(attributes = {}, schema_url = '')
```
fbogsany, line 23: I kind of wish we just used `nil` here instead of the empty string. The `schema_url` default for `InstrumentationScope` is `nil` and the difference in behaviour is troubling.

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -15,19 +15,22 @@ class << self
...
+
             frozen_attributes = attributes.each_with_object({}) do |(k, v), memo|
               raise ArgumentError, 'attribute keys must be strings' unless k.is_a?(String)
               raise ArgumentError, 'attribute values must be (array of) strings, integers, floats, or booleans' unless Internal.valid_value?(v)
 
               memo[-k] = v.freeze
             end.freeze
 
-            new(frozen_attributes)
+            new(frozen_attributes, schema_url)
```
fbogsany, line 33: We should `freeze` the `schema_url` as well.
```suggestion
            new(frozen_attributes, schema_url.freeze)
```

api/lib/opentelemetry/internal/proxy_tracer_provider.rb
ahayworth: Yes - and given the upcoming [instrumentation attributes stuff](https://github.com/open-telemetry/opentelemetry-ruby/issues/1262) I think that's a wise idea; will change.

> At this point, I wish we'd made version an optional named parameter 😞

That would be nice, yes. We could do something like this, I think:

```ruby
      def tracer(_name = nil, _version = nil, name: nil, version: nil, schema_url: nil)
        ...
        name ||= _name
        version ||= _version
```

...but that might be even more distasteful. Regardless, I'll make `schema_url` a named param.

api/test/opentelemetry/trace/tracer_provider_test.rb
ahayworth: Oh excellent; I'll remove the TODO. :D

sdk/lib/opentelemetry/sdk/resources/resource.rb
ahayworth: I'll look into it!

ahayworth: @fbogsany I changed that behavior; but I chose to interpret a blank string as if it were opaque in the merging behavior (that is `schema_url: nil` and `schema_url: ''` are not treated as equivalent in our merge, and we fall through to the unspecified behavior section of the spec). Let me know if that's a bad idea in your eyes.

api/lib/opentelemetry/internal/proxy_tracer_provider.rb
ahayworth: Update: I went through and changed this to kwargs just about everywhere; except for the internal _SDK_ tracer initialization. We could do it there as well, but I was lazy and didn't feel like updating anything else since that's `@api private` anyways.

(But if you feel strongly, I can go further).

sdk/lib/opentelemetry/sdk/resources/resource.rb
ahayworth: Also: I was going to make it a kwarg on the `Resource` class as well, but since we have `attributes = {}` we ran into failures like:

```
/Users/ahayworth/projects/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/resources/resource.rb:24:in `create': unknown keyword: "service.name" (Argument
Error)
        from /Users/ahayworth/projects/opentelemetry-ruby/sdk/lib/opentelemetry/sdk/resources/resource.rb:38:in `default'
```

...which makes sense of course, it would be treating everything passed as an attribute as part of the kwargs.

I chose to just keep it a positional param there, but if there's a more elegant way to fix that up, we can do so.

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -15,19 +15,23 @@ class << self
...
           #
-          # @param [Hash{String => String, Numeric, Boolean} attributes Hash of key-value pairs to be used
+          # @param [Hash{String => String, Numeric, Boolean}] attributes Hash of key-value pairs to be used
           #   as attributes for this resource
+          # @param [String] Specifies the Schema URL that should be recorded in the emitted resource.
           # @raise [ArgumentError] If attribute keys and values are not strings
+          # @raise [ArgumentError] If the schema URL is given but it is not a string.
           # @return [Resource]
-          def create(attributes = {})
+          def create(attributes = {}, schema_url = nil)
```
fbogsany, line 24: Can we make this a named parameter as well?
```suggestion
          def create(attributes = {}, schema_url: nil)
```

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -93,18 +99,36 @@ def attribute_enumerator
...
+        def merge(other) # rubocop:disable Metrics/AbcSize
           return self unless other.is_a?(Resource)
 
-          self.class.send(:new, attributes.merge(other.attributes).freeze)
+          # This is slightly verbose, but tries to follow the definition in the spec closely.
+          new_schema_url = if schema_url.nil?
+                             other.schema_url
+                           elsif other.schema_url.nil? || schema_url == other.schema_url
+                             schema_url
+                           elsif schema_url != other.schema_url
```
fbogsany, line 117: Can just be `else`
```suggestion
                           else
```

sdk/lib/opentelemetry/sdk/resources/resource.rb
```diff
@@ -93,18 +99,36 @@ def attribute_enumerator
...
+                               "Merging resources with schema version '#{schema_url}' and '#{other.schema_url}' is undefined."
+                             )
+
+                             nil
+                           end
+
+          self.class.send(:new, attributes.merge(other.attributes).freeze, new_schema_url.freeze)
         end
 
+        attr_reader :schema_url
```
fbogsany, line 130: It is interesting to note that the spec doesn't say anything about a getter for `schema_url`. Obviously, we need one, since exporters need to be able to read it. We should open a PR against the spec to add this after https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/resource/sdk.md#retrieve-attributes

sdk/lib/opentelemetry/sdk/resources/resource.rb
ahayworth: It [made the tests unhappy](https://github.com/open-telemetry/opentelemetry-ruby/pull/1237#discussion_r879757292), but I do _want_ to. I'll try a little harder.

ahayworth: @fbogsany I implemented that here: 0966e37b - take a look and let me know if you hate it or not. I don't know quite how I feel.

ahayworth:
```
Style/BracesAroundHashParameters: Redundant curly braces around a hash parameter.
```

It's hard to overstate just how annoying the cop can be. I'll fix that later.

PR #1249, opened 2022-05-11
Split CI builds by gems at top-level

ahayworth:
This PR splits our builds up by gem at the top-level, so that we can actually see what has failed more easily (and, we can individually re-run gem builds!).
It's slightly faster, and I think probably more useful - but it's a big change, so please take a look at the CI output and see what you think.

The PR can be reviewed commit-by-commit if you'd like and there are descriptive messages at each commit. However, the bulk of it is described in one particular commit, so I will paste that here for your reviewing convenience:

---

ci: replace toys.rb functionality with an explosion of actions + yaml

This replaces the "test it all in a loop" approach that `toys/ci.rb` was
taking, by leveraging some more advanced features of GitHub Actions.

To start, we construct a custom Action (not a workflow!) that can run
all the tests we were doing with `toys/ci.rb`. It takes a few different
To start, we construct a custom Action (not a workflow!) that can run
all the tests we were doing with `toys/ci.rb`. It takes a few different
inputs: gem to test, ruby version to use, whether or not to do rubocop,
etc. Then, it figures out where in the repo that gem lives, sets up ruby
(including appraisals setup, if necessary), and runs rake tests (and
then conditionally runs YARD, rubocop, etc).

Then, over in `ci.yml`, we list out all of the gems we currently have
and chunk them up into different logical groups:

- base (api, sdk, etc)
- exporters
- propagators
- instrumentation that requires sidecar services to test
- instrumentaiton that doesn't require anything special to test

For most groups, we set up a matrix build of operating systems (ubuntu,
macos, and windows) - except for the "instrumentation_with_services"
group, because sidecar services are only supported on linux.

For each matrix group (gem + os), we then have a build that has multiple
steps - and each step calls the custom Action that we defined earlier,
passing appropriate inputs. Each step tests a different ruby version:
3.1, 3.0, 2.7, or jruby - and we conditionally skip the step based on
the operating system (we only run tests against ruby 3.1 for mac /
windows, because the runners are slower and we can't launch as many at
once).

Notably, we have a few matrix exclusions here: things that wont build on
macos or windows, but there aren't many.

Finally, each group also maintains a "skiplist" of sorts for jruby -
it's ugly, but some instrumentation just doesn't work for our Java
friends. So we have a step that tests whether or not we should build the
gem for jruby, and then the jruby step is skipped depending on the
answer. We can't really use a matrix exclusion here because we don't use
the ruby version in the matrix at all - otherwise we'd have a *huge*
explosion of jobs to complete, when in reality we can actually install +
test multiple ruby versions on a single runner, if we're careful.

The net effect of all of this is that we end up having many different
builds running in parallel, and if a given gem fails we can *easily* see
that and get right to the problem. Builds are slightly faster, too.

The major downsides are:
- We need to add new gems to the build list when we create them.
- We can't cache gems for appraisals, which adds a few minutes onto the
  build times (to be fair, we weren't caching anything before)
- It's just kinda unwieldy.
- I didn't improve anything around the actual release process yet.

Future improvements could be:
- Figuring out how to cache things with Appraisals, because I gave up
  after a whole morning of fighting bundler.
- Dynamically generating things again, because it's annoying to add gems
  to the build matrices.


.github/actions/test_gem/action.yml
```diff
@@ -0,0 +1,130 @@
...
+        TEST_REDIS_PORT: 6379
+
+    - name: YARD
+      shell: bash
+      if: "${{ inputs.yard == 'true' }}"
+      run: |
+        cd "${{ steps.setup.outputs.dir }}"
+        bundle exec rake yard
+
+    - name: Rubocop
```
arielvalentin, line 118: I do not think we need to run this for every gem version. 

Would it be OK to run Rubocop only for the minimal version of Ruby we intend on supporting?

.github/actions/test_gem/action.yml
```diff
@@ -0,0 +1,130 @@
...
+      run: |
+        cd "${{ steps.setup.outputs.dir }}"
+        bundle exec rake rubocop
+
+    - name: Build Gem
+      shell: bash
+      if: "${{ inputs.build == 'true' }}"
+      run: |
+        cd "${{ steps.setup.outputs.dir }}"
+        gem build ${{ inputs.gem }}.gemspec
```
arielvalentin, line 130: ✨ 

I wonder if installing the gem also will give us an indication of failure if there are un-released incompatible versions?

Or maybe that should be part of the release build?

.github/workflows/ci.yml
```diff
@@ -9,33 +10,308 @@ on:
...
+        os:
+          - ubuntu-latest
+          - macos-latest
+          - windows-latest
+    name: ${{ matrix.gem }} / ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: "Test Ruby 3.1"
+        uses: ./.github/actions/test_gem
```
arielvalentin, line 35: 🎉 Why hasn't this ever occurred to me to do?

bin/instrumentation_generator
```diff
@@ -2,4 +2,14 @@
 
 require_relative '../.instrumentation_generator/instrumentation_generator'
 
+puts <<-EOF
+
+🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨🚨
+
+ WARNING: Your gem will *NOT* be tested until you add it to the CI workflows in `.github/workflows/ci.yml`!!
```
arielvalentin, line 9: follow up PR, maybe this script can edit the action and add the new instrumentation to the list?

arielvalentin (approved):
I am mixed on the naming/grouping whether it should be `matix.os/matrix.gem` and have some other suggestions but otherwise I am on board!!!

.github/actions/test_gem/action.yml
ahayworth: > Would it be OK to run Rubocop only for the minimal version of Ruby we intend on supporting?

Yes, and

> I do not think we need to run this for every gem version.

we aren't! One of the downsides of this change is that it doesn't make our CI system _much_ more clear - although I believe it helps clarify that to a small degree.

The relevant bits of code are [here, defining a default 'rubocop' input to 'false'](https://github.com/open-telemetry/opentelemetry-ruby/pull/1249/files#diff-e018a7f25c2bd16ae50ee7acbd551cb6a4afa64811f45451b454e0896e0d6e0cR17-R21); [here, where we only run the rubocop step if the variable is in fact 'true'](https://github.com/open-telemetry/opentelemetry-ruby/pull/1249/files#diff-e018a7f25c2bd16ae50ee7acbd551cb6a4afa64811f45451b454e0896e0d6e0cR120); and finally [here, where we only set that variable to true for Ruby 2.7 builds](https://github.com/open-telemetry/opentelemetry-ruby/pull/1249/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR45-R53) (and that last pattern is repeated for every build group unless I messed it up somehow).

So the net effect is "we are only running Rubocop (and the YARD / `gem build` smokescreen tests) for Ruby 2.7".

.github/actions/test_gem/action.yml
ahayworth: It might? I understand what you're getting at. I wonder if it's possible to actually do so before all the gems are pushed to rubygems? For example, if `opentelemetry-instrumentation-foo` depends on `opentelemetry-sdk` v2, but only `opentelemetry-sdk` v1 is currently released, it would necessarily fail.

That _may_ just be an edge case not worth considering, though (specifically when we are bumping the version of one gem in a PR and bumping the dependencies of another gem in the _same_ PR) ? 

I'm interested in what you think about that. If the barriers don't seem insurmountable, we could do that fairly easily in a follow-up PR to this one. Or in this one, depending on how my waning motivation for the day holds up. 😆

.github/workflows/ci.yml
ahayworth: Honestly I didn't think much about doing it until @veraticus showed me how to do so at `$PREV_JOB`, so all credit goes to him.

More practically, it's a useful tool but limited, so it's not always useful (in the same way that re-using workflows via `workflow_call` is also useful, but limited). It worked out pretty well here though!

.github/actions/test_gem/action.yml
arielvalentin: Thank you for explaining that!!! Do I even work at GH?

.github/actions/test_gem/action.yml
arielvalentin: It would make more sense as a follow up PR. It is unrelated to this problem and this PR got my mental juices flowing.

bin/instrumentation_generator
ahayworth: Yes, I think that's best in a follow-up. As this PR stands, it's not very easy to know _where_ to put it into the workflow file.

In the future, I think a nice evolution might be:
- One YAML file (modified by `instrumentation_generator`) that is the source of truth not only for gem releases, but also for CI in general.
- That YAML file could have things like `sidecar_services: false` or `appraisals: true` or `jruby: false` that inform how things are to be built.
- By consuming that YAML file, we could then dynamically generate the build matrices and do more impressive things. At minimum, it could help us reduce some of the indirection and hacks this PR introduces around skipping things for jruby, and it also could present a way out of the "we can't cache gems for `Appraisals`" problem too.

But ... in the interests of "stop iterating on this and ship at least a modest improvement," I decided to wait. 😆 

cc @plantfansam, with whom I've been chatting about this stuff ☝️

arielvalentin: > But ... in the interests of "stop iterating on this and ship at least a modest improvement," I decided to wait. 😆

💯

.github/actions/test_gem/action.yml
ahayworth: It's absolutely a good idea though, it's another way to stop our release 🦶 🔫  problems.

ahayworth:
> I am mixed on the naming/grouping whether it should be matix.os/matrix.gem

@arielvalentin do you have a different naming idea? I'm open to suggestions - I just couldn't come up with anything particularly _inspired_ on my own 😆

.github/actions/test_gem/action.yml
ahayworth: 🤣

arielvalentin:
> > I am mixed on the naming/grouping whether it should be matix.os/matrix.gem
> 
> @arielvalentin do you have a different naming idea? I'm open to suggestions - I just couldn't come up with anything particularly _inspired_ on my own 😆

Me naming things? lol you know how bad I am at that

bin/instrumentation_generator
plantfansam: Yeah, I think it's great to merge in the instrumentation generator change as-is and then do a follow-on that can update the workflow file as necessary. I'm a little bit wary of the fanciness of deriving things from a single YAML file (very tough to trace a code path that ends in a YAML file), but will reserve judgement until we try 😇

plantfansam (approved):
LGTM! Thank you SO MUCH for taking this on. I'd love for someone that understands the release process to confirm that this won't break that, but it's a 🚢 from me.

bin/instrumentation_generator
ahayworth: True. It could be quite difficult if we're too clever.

.github/workflows/ci.yml
```diff
@@ -9,33 +10,308 @@ on:
...
-      JRUBY_OPTS: --debug
+    steps:
+      - uses: actions/checkout@v2
+      - name: "Test Ruby 3.1"
+        uses: ./.github/actions/test_gem
+        with:
+          gem: "${{ matrix.gem }}"
+          ruby: "3.1"
+      - name: "Test Ruby 3.0"
+        if: "${{ matrix.os == 'ubuntu-latest' }}"
```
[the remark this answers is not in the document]
ahayworth: > So here we're testing ruby 3.1 on everything

Yes

> and only testing older versions on macos and windows?

The other way around: we're only testing older versions on linux (unless I accidentally inverted the conditional further down or something).

We have more stringent limits around [Mac runners](https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration), and Windows runners just seem somewhat anemic and seem to take longer to pick up the jobs (anecdotally speaking), so we try to hurry through the builds and not delay things too much. Testing one version of ruby is _likely_ enough to catch platform specific issues.

Previously we were only testing ruby 3.0 on mac + windows; I changed it to 3.1 here just to keep up with the times.

.github/workflows/ci.yml
```diff
@@ -9,33 +10,308 @@ on:
...
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-graphql"                  ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-http"                     ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-http_client"              ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-koala"                    ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-lmdb"                     ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-rack"                     ]] && echo "::set-output name=skip::true"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-rails"                    ]] && echo "::set-output name=skip::true"
+          # This is essentially a bash script getting evaluated, so we need to return true or the whole job fails.
```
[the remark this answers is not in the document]
ahayworth: It took me a few tries to figure out why the jobs started failing and figured people might run into the same confusion later :P

.github/workflows/ci.yml
```diff
@@ -9,33 +10,308 @@ on:
...
+          gem: "${{ matrix.gem }}"
+          ruby: "2.7"
+          yard: true
+          rubocop: true
+          build: true
+      - name: "JRuby Filter"
+        id: jruby_skip
+        shell: bash
+        run: |
+          echo "::set-output name=skip::false"
```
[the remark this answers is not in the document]
ahayworth: Today I learned what sarcasm from @plantfansam looks like! 😝 

To be fair, it definitely could have been a lot uglier, that's for sure.

.github/workflows/ci.yml
```diff
@@ -9,33 +10,308 @@ on:
...
+          true
+
+      - name: "Test JRuby"
+        if: "${{ matrix.os == 'ubuntu-latest' && steps.jruby_skip.outputs.skip == 'false' }}"
+        uses: ./.github/actions/test_gem
+        with:
+          gem: "${{ matrix.gem }}"
+          ruby: "jruby"
+
+  instrumentation_with_services:
```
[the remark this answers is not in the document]
ahayworth: 💥 b1cdf483

.github/workflows/ci.yml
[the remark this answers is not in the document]
plantfansam: Got it, thanks. Makes sense to me!

.github/workflows/ci.yml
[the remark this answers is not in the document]
plantfansam: Thanks!!!

robertlaurin (approved):

PR #1261, opened 2022-05-13
fix: handle OTLP exporter 404s discretely

chrisholmes:
The otel-collecor uses a 404 with a plain/text content-type when the
exporter path doesn't match the one configured by the colletor. This
commit adds a check to handle this content type.


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -232,15 +232,27 @@ def handle_redirect(location)
...
-        rescue StandardError => e
-          OpenTelemetry.handle_error(exception: e, message: 'unexpected error decoding rpc.Status in OTLP::Exporter#log_status')
+        def log_status(response) # rubocop:disable Metrics/AbcSize
+          content_type = response['Content-Type']
+          code = response.code
+          body = response.body
+
+          http_status_message = "http.code=#{code} http.content_type='#{content_type}'"
+          case content_type
+          when 'text/plain; charset=utf-8'
```
arielvalentin, line 242: I am a little torn here. Should this code explicitly check for the protobuf content type instead and treat anything else as an error?

chrisholmes: Yes, I've been thinking about this some more and I think it would be better to take a slightly different approach. This change is meant to be about logging 404 errors cleanly so really it should checking for that code and creating a log message that is 404 specific.

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -418,6 +418,25 @@
       OpenTelemetry.logger = logger
     end
 
+    it 'logs a specific message when there is a 404' do
+      log_stream = StringIO.new
+      logger = OpenTelemetry.logger
+      OpenTelemetry.logger = ::Logger.new(log_stream)
+
+      stub_request(:post, 'http://localhost:4318/v1/traces').to_return(status: 404, body: "Not Found\n")
+      span_data = create_span_data
```
robertlaurin, line 427:
```suggestion
      span_data = OpenTelemetry::TestHelpers.create_span_data
```

fbogsany (approved):

PR #1180, opened 2022-04-11
feat: Otlp grpc

robertlaurin:
Implements a starting point for a OTLP via GRPC exporter.  This implementation currently has no error handling, and must not be considered production ready.

Follow on from https://github.com/open-telemetry/opentelemetry-ruby/pull/1179

PR #1263, opened 2022-05-17
Feat: Add url, and config_opt helpers

robertlaurin:
https://github.com/open-telemetry/opentelemetry-ruby/pull/1180#discussion_r874159643

- config_opt is repeated in 4 places
- invalid_url? is repeated in 5 places

### DIFF
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -101,6 +101,36 @@ module OpenTelemetry
         url
       end
 
+      # Returns the first non nil environment variable requested,
+      # or the default value if provided.
+      #
+      # @param [String] env_vars The environment variable(s) to retrieve
+      # @param default The fallback value to return if the requested
+      #  env var(s) are not present
+      #
+      # @returns [String]
+      def config_opt(*env_vars, default: nil)
+        env_vars.each do |env_var|
+          val = ENV[env_var]
+          return val unless val.nil?
+        end
+        default
+      end
+
+      # Returns a true if the provided url is invalid
+      #
+      # @param [String] url the URL string to test validity
+      #
+      # @return [boolean]
+      def invalid_url?(url)
+        return true if url.nil? || url.strip.empty?
+
+        URI(url)
+        false
+      rescue URI::InvalidURIError
+        true
+      end
+
       # Returns true if exporter is a valid exporter.
       def valid_exporter?(exporter)
         exporter && %i[export shutdown force_flush].all? { |m| exporter.respond_to?(m) }
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -113,7 +113,7 @@ module OpenTelemetry
         ENV.values_at(*env_vars).compact.fetch(0, default)
       end
 
-      # Returns a true if the provided url is invalid
+      # Returns a true if the provided url is valid
       #
       # @param [String] url the URL string to test validity
       #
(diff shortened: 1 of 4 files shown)


common/lib/opentelemetry/common/utilities.rb
```diff
@@ -101,6 +101,36 @@ def cleanse_url(url)
...
+        end
+        default
+      end
+
+      # Returns a true if the provided url is invalid
+      #
+      # @param [String] url the URL string to test validity
+      #
+      # @return [boolean]
+      def invalid_url?(url)
```
arielvalentin, line 125: I am inclined to lean towards positive method names instead of negative ones, i.e. `valid_url?` as opposed to `invalid_url?`. 

For convenience if we wanted to support the inverse I would appreciates a `valid_url?` combined with `def invalid_url?(url); !valid_url?(url); done`

Also it would really be good to capture something about the reason why the URL is invalid. That could give a clue to someone debugging this.

common/lib/opentelemetry/common/utilities.rb
```diff
@@ -101,6 +101,36 @@ def cleanse_url(url)
...
+      # @param default The fallback value to return if the requested
+      #  env var(s) are not present
+      #
+      # @returns [String]
+      def config_opt(*env_vars, default: nil)
+        env_vars.each do |env_var|
+          val = ENV[env_var]
+          return val unless val.nil?
+        end
+        default
```
arielvalentin, line 117: I am inclined to want to do this is a more functional way like this

```suggestion
        ENV.values_at(env_vars).compact.fetch(0, default)
```

However I am not certain if there is a guarantee that the Array returned from `values_at` will maintain the precedent order.

common/lib/opentelemetry/common/utilities.rb
robertlaurin: No strong feelings towards positive vs negative methods here, so I can switch that.

Re: Capturing something about the reason.

Technically I think we could just let the error bubble up from creating the URI out of the url provided, and do away with this method entirely.  However I think the error message provided is a lot less human friendly in most cases, even if the stack trace sends you to the right place.

Here is a random example I used in the tests
```
pry(main)> URI('1113:1113')
URI::InvalidURIError: bad URI(is not URI?): "1113:1113"
from /opt/rubies/2.7.5/lib/ruby/2.7.0/uri/rfc3986_parser.rb:67:in `split'
```
vs

`"invalid url for OTLP::Exporter 1113:1113"`

common/lib/opentelemetry/common/utilities.rb
robertlaurin: 🤷 this seems fine to me. 

It appears that the ordering should be considered predictable as the [method source](https://ruby-doc.org/core-2.7.1/Hash.html#method-i-values_at) shows it enumerating the provided args.
```
[9] pry(main)> ENV.values_at('LINES', 'COLUMNS')
=> ["37", "181"]
[10] pry(main)> ENV.values_at('COLUMNS', 'LINES')
=> ["181", "37"]
```

common/lib/opentelemetry/common/utilities.rb
robertlaurin: Follow up thought, if we wanted to provide more hints we can do that as a follow on. I'm just shuffling some code around right now :)

common/lib/opentelemetry/common/utilities.rb
```diff
@@ -101,6 +101,32 @@ def cleanse_url(url)
...
+        ENV.values_at(*env_vars).compact.fetch(0, default)
+      end
+
+      # Returns a true if the provided url is valid
+      #
+      # @param [String] url the URL string to test validity
+      #
+      # @return [boolean]
+      def valid_url?(url)
+        return true if url.nil? || url.strip.empty?
```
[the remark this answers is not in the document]
robertlaurin: https://github.com/open-telemetry/opentelemetry-ruby/pull/1263/commits/31137b9ed31264f5a925064cd2a1c98e1f76e215

arielvalentin (approved):

PR #1267, opened 2022-05-18
fix: add otlp grpc/http to ci

robertlaurin:
I think they got missed during the CI refactoring


arielvalentin (approved):

.github/workflows/ci.yml
```diff
@@ -65,6 +65,9 @@ jobs:
         gem:
           - opentelemetry-exporter-jaeger
           - opentelemetry-exporter-otlp
+          - opentelemetry-exporter-otlp-common
+          - opentelemetry-exporter-otlp-grpc
```
robertlaurin, line 69: @arielvalentin or @ahayworth oh keepers of ci, what's the right way to skip jruby for this specific gem?

robertlaurin: I see a stanza for the instrumentation gems that allows skipping jruby, should I mimic that?

ahayworth: You are all correct. There are ways to DRY it out but I'd rather not, because I think we should continue to evolve the CI system in a slightly different way.

I would copy and paste your way to victory here, and I can help out if you need.

PR #1273, opened 2022-05-20
Validate Gem dependencies before release

ahayworth:
This PR adds a way to check that the gems we are releasing will not be broken upon said release.

We run this check as a GitHub action that only fires on pushes to `release/**` branches, and we accomplish it by:
- Getting a list of changed gemspecs in this PR (ie: the things we will be releasing)
- Loading *all* gemspecs in the repo
- Loading the latest released version of every gem in the repo, directly from rubygems.org
- For each gemspec in the repo, we then check the dependencies:
  - If the dependency is included in this release request, we check that the dependency will be satisifed by interrogating the loaded gemspecs directly.
  - If the dependency is *not* included in this release request, we then construct a fake gemspec with the info on rubygems.org, and then see if it's a match.

We'll fail if there is a gem dependency that doesn't satisfy the version constraints based on what we're about to release or what's already available on rubygems.org - or if it's not published at all.

It's kinda hard to test this locally, so I'm not 100% if the `changed_gems` logic with `FETCH_HEAD` works correctly.

### DIFF
--- a/.github/workflows/release-request.yml
+++ b/.github/workflows/release-request.yml
@@ -23,6 +23,8 @@ jobs:
         uses: actions/checkout@v2
       - name: Install Toys
         run: "gem install --no-document toys"
+      - name: Validate OpenTelemetry internal gem dependencies
+        run: "rake each:validate_otel_dependencies"
       - name: Open release pull request
         env:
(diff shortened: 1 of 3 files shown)


ahayworth:
I'm going to add a little more sanity checking here.

PR #1291, opened 2022-06-04
docs: Add additional warning about attributes and span_creation

### LINKED ISSUE #1258: Clarify documentation around adding attributes at span creation
Added in: https://github.com/open-telemetry/opentelemetry-specification/pull/2383

I think our docs kinda say this today, so it would be a rather minor update on our end.

ahayworth:
Fixes #1258 by noting that samplers only consider attributes present at span creation.


plantfansam (approved):
This makes sense to me. Do you think we should add a comment to the `api` gem's `set_attribute` and `set_attributes` methods indicating preference for setting attributes at span creation? And _maybe_ the SDK gem?

fbogsany (approved):

PR #1294, opened 2022-06-04
feat: support OTEL_PROPAGATORS=none

### LINKED ISSUE #1253: Add `none` as a value for `OTEL_PROPAGATORS`
Added in: https://github.com/open-telemetry/opentelemetry-specification/pull/2052

ahayworth:
Fixes #1253. We were already doing this if you passed an invalid value,
so I guess *technically* we were already in compliance. But now it's
explicit and without a spurious warning.


sdk/test/opentelemetry/sdk/configurator_test.rb
```diff
@@ -163,7 +163,17 @@
         _(OpenTelemetry.propagation).must_equal(OpenTelemetry::Baggage::Propagation.text_map_propagator)
       end
 
-      it 'defaults to none with invalid env var' do
+      it 'supports "none" as an environment variable' do
+        OpenTelemetry::TestHelpers.with_env('OTEL_PROPAGATORS' => 'none') do
```
arielvalentin, line 167: I think this test will pass even if you remove L206 wouldn't it? 
Is it essential for us to suppress the log output? Should that be checked as part of this test?

arielvalentin (approved):

sdk/test/opentelemetry/sdk/configurator_test.rb
ahayworth: The only reason the tests are failing is because I angered the rubocop. Like usual. 😂

That said, we shouldn't log a warning when you pass 'none', since the spec considers that a valid value now.

Will fix up the linter momentarily.

ahayworth: Oh whoops... I see what you mean. It isn't essential to suppress the log output, but I think in terms of spec compliance it's kinda "meh" to log a warning for what should be an explicitly supported option.

Regarding the tests that would pass if we removed L206 - good point. I'll fix that right away.

ahayworth: Fixed in 7fd8c6be

arielvalentin (approved):

fbogsany (approved):

PR #1295, opened 2022-06-05
feat: support truffleruby

ahayworth:
Does what it says on the tin. By "support" I mean "turn it on in CI and
see if everything works."

The compatibility list is actually a bit better than JRuby!

Notably, I needed to exclude the OTLP exporter to make tests pass. I'm
not certain it's a real bug; I just haven't tried to fix those test
failures yet.

---

For discussion: Do we actually even want this? I went down a rabbithole today and landed on truffleruby, and thought it might be neat to support it. It seems like it's broadly more compatible than JRuby to me. Digging through past issues and PRs, it looks like we've at least attempted to support this in the past, and then maybe accidentally dropped it over time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,6 +56,12 @@ jobs:
         with:
           gem: "${{ matrix.gem }}"
           ruby: "jruby"
+      - name: "Test truffleruby"
+        if: "${{ matrix.os == 'ubuntu-latest' }}"
+        uses: ./.github/actions/test_gem
+        with:
+          gem: "${{ matrix.gem }}"
+          ruby: "truffleruby"
 
   exporters:
     strategy:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -306,13 +350,30 @@ jobs:
...
         uses: ./.github/actions/test_gem
         with:
           gem: "${{ matrix.gem }}"
           ruby: "jruby"
+      - name: "truffleruby Filter"
+        id: truffleruby_skip
+        shell: bash
+        run: |
+          echo "::set-output name=skip::false"
+          [[ "${{ matrix.gem }}" == "opentelemetry-instrumentation-que"        ]] && echo "::set-output name=skip::true"
```
plantfansam, line 364: So these gems fail against truffleruby, is that right?

plantfansam (approved):
LGTM! If the gems enumerated in `truffleruby filter` don't pass when run in CI, what do you think about making GH issues for each of them? These could make great tasks for newcomers to the project. I am happy to make those tickets if you think it's a good idea.

.github/workflows/ci.yml
ahayworth: Indeed.

ahayworth:
@plantfansam That's a great idea - and perhaps doing it for `jruby` too is also a good idea... I'm not sure how much effort the different gems would be, though.

plantfansam:
> @plantfansam That's a great idea - and perhaps doing it for `jruby` too is also a good idea... I'm not sure how much effort the different gems would be, though.

SGTM. Will make tickets for truffleruby and jruby

fbogsany (approved):

plantfansam:
Hi @ahayworth! Sorry for the trouble, but with the `opentelemetry-ruby-contrib` split, I think we'll need to chop this PR up into contrib-relevant and this-repo-relevant pieces. There will probably be a merge conflict once https://github.com/open-telemetry/opentelemetry-ruby/pull/1319 makes it into `main`.

ahayworth:
> Hi @ahayworth! Sorry for the trouble, but with the opentelemetry-ruby-contrib split, I think we'll need to chop this PR up into contrib-relevant and this-repo-relevant pieces. There will probably be a merge conflict once https://github.com/open-telemetry/opentelemetry-ruby/pull/1319 makes it into main.

It's no trouble at all; I'll update this.

PR #1308, opened 2022-06-09
fix: Stop leaking basic authentication credentials in Faraday instrumentation

gregolsen:
Some Ruby gems (e.g. [`elasticsearch-transport`](https://github.com/elastic/elastic-transport-ruby/blob/391ed956b23b9b3be3fd962f8fc5366e410006eb/lib/elastic/transport/transport/http/faraday.rb#L49-L54) use [`Faraday::Connection#run_request`](https://www.rubydoc.info/github/lostisland/faraday/Faraday%2FConnection:run_request) directly. This method accepts full URL along with the basic auth credentials as part of it.
As a result, credentials leak via `http.url` span attribute.
This PR modifies the specs to fail by using `run_request` and also fixes the code to "cleanse" the URL with the existing helper method.

<details>
<summary>How to reproduce</summary>

```ruby
# Gemfile
gem 'faraday', '= 0.15.4'
gem 'opentelemetry-sdk'
gem 'opentelemetry-instrumentation-all'
gem 'elasticsearch', '= 6.1.0'
```

```ruby
# script.rb
require 'opentelemetry/sdk'
require 'opentelemetry/exporter/otlp'
require 'opentelemetry/instrumentation/all'
require 'elasticsearch'

ENV['OTEL_TRACES_EXPORTER'] = 'console'

  c.service_name = 'test_app'
  c.use 'OpenTelemetry::Instrumentation::Faraday'
end

client = Elasticsearch::Client.new(hosts: "http://elastic:top-secret@localhost:9200");
client.search
```

Output:
```
I, [2022-06-09T16:41:13.061317 #13360]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Faraday was successfully installed with the following options {:peer_service=>nil}
#<struct OpenTelemetry::SDK::Trace::SpanData
 name="HTTP GET",
 kind=:client,
 status=
  #<OpenTelemetry::Trace::Status:0x0000000114a10328 @code=1, @description="">,
 parent_span_id="\x00\x00\x00\x00\x00\x00\x00\x00",
 total_recorded_attributes=4,
 total_recorded_events=0,
 total_recorded_links=0,
 start_timestamp=1654789273069515000,
 end_timestamp=1654789273093570000,
 attributes=
  {"http.method"=>"GET",
   "http.url"=>"http://elastic:top-secret@localhost:9200/_search",
   "net.peer.name"=>"localhost",
   "http.status_code"=>200},
 links=nil,
 events=nil,
 resource=
  #<OpenTelemetry::SDK::Resources::Resource:0x00000001249dcf40
   @attributes=
    {"service.name"=>"test_app",
     "process.pid"=>13360,
     "process.command"=>"app.rb",
     "process.runtime.name"=>"ruby",
     "process.runtime.version"=>"2.6.5",
     "process.runtime.description"=>
      "ruby 2.6.5p114 (2019-10-01 revision 67812) [-darwin20]",
     "telemetry.sdk.name"=>"opentelemetry",
     "telemetry.sdk.language"=>"ruby",
     "telemetry.sdk.version"=>"1.1.0"}>,
 instrumentation_library=
  #<struct OpenTelemetry::SDK::InstrumentationLibrary
   name="OpenTelemetry::Instrumentation::Faraday",
   version="0.20.1">,
 span_id="v\x9D\xBF;\xFE\x9B\xD9\xD4",
 trace_id="\xB3\x8ASX\x90NA\xDF\x9E\x1ET\xC9\xBE[+\x7F",
 trace_flags=#<OpenTelemetry::Trace::TraceFlags:0x0000000114913d80 @flags=1>,
 tracestate=#<OpenTelemetry::Trace::Tracestate:0x0000000114973000 @hash={}>>
```

</details>

### DIFF
--- a/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
+++ b/instrumentation/faraday/lib/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware.rb
@@ -44,7 +44,7 @@ module OpenTelemetry
           def span_creation_attributes(http_method:, url:)
             instrumentation_attrs = {
               'http.method' => http_method,
-              'http.url' => url.to_s,
+              'http.url' => OpenTelemetry::Common::Utilities.cleanse_url(url.to_s),
               'net.peer.name' => url.host
             }
             config = Faraday::Instrumentation.instance.config
--- a/instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
+++ b/instrumentation/faraday/test/opentelemetry/instrumentation/faraday/middlewares/tracer_middleware_test.rb
@@ -13,9 +13,10 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
   let(:instrumentation) { OpenTelemetry::Instrumentation::Faraday::Instrumentation.instance }
   let(:exporter) { EXPORTER }
   let(:span) { exporter.finished_spans.first }
+  let(:base_url) { 'http://username:password@example.com' }
 
   let(:client) do
-    ::Faraday.new('http://username:password@example.com') do |builder|
+    ::Faraday.new(base_url) do |builder|
       builder.adapter(:test) do |stub|
         stub.get('/success') { |_| [200, {}, 'OK'] }
         stub.get('/failure') { |_| [500, {}, 'OK'] }
@@ -43,7 +44,7 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
     end
 
     it 'has http 200 attributes' do
-      response = client.get('/success')
+      response = client.run_request(:get, "#{base_url}/success", nil, {})
 
       _(span.name).must_equal 'HTTP GET'
       _(span.attributes['http.method']).must_equal 'GET'
@@ -56,7 +57,7 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
     end
 
     it 'has http.status_code 404' do
-      response = client.get('/not_found')
+      response = client.run_request(:get, "#{base_url}/not_found", nil, {})
 
       _(span.name).must_equal 'HTTP GET'
       _(span.attributes['http.method']).must_equal 'GET'
@@ -69,7 +70,7 @@ describe OpenTelemetry::Instrumentation::Faraday::Middlewares::TracerMiddleware 
     end
 
     it 'has http.status_code 500' do
-      response = client.get('/failure')
+      response = client.run_request(:get, "#{base_url}/failure", nil, {})
 
       _(span.name).must_equal 'HTTP GET'
       _(span.attributes['http.method']).must_equal 'GET'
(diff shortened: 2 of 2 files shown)


arielvalentin (approved):

gregolsen:
@arielvalentin unfortunately had to fix a rubcop issue about using double-quotes unnecessarily. Could you please re-review? 🙏🏻

gregolsen:
@arielvalentin sorry to bother you again but it looks like some unrelated segfault in rdkafka is blocking the build to be green https://github.com/open-telemetry/opentelemetry-ruby/runs/6829513244?check_suite_focus=true

ahayworth (changes requested):
I have a few concerns:

1. We shouldn't change every test to use `run_request` directly, because that's not how you're supposed to actually use Faraday. We want to make sure that the core functionality - tracing - works when users follow the same methods they'll see in the upstream docs. If our tests work by poking the internals of Faraday directly, they risk becoming brittle or inadvertently becoming useless when the official entry points to the library change.
2. I don't actually see any tests that directly assert the new behavior. By changing every test method to use `run_request` directly (which is what triggers the behavior in the first place), you're actually testing your change _implicitly_ all over the place and never _explicitly_. We can't trust our tests if they're heavily relying on unstated side-effects to assert something.
3. While we _should_ sanitize this attribute, we might also want to raise this with upstream Faraday. You can reproduce this if you call `Faraday.get("http://secret:password@foo.com")` directly, but _not_ if you do `conn = Faraday.new("http://secret:password@foo.com"); conn.get("/")`. That seems like unintentional behavior, since Faraday goes through [great trouble](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L366) to actually make sure this doesn't leak out otherwise. I think it might be a bug.

gregolsen:
@ahayworth thank you so much for the great feedback! Just want to explain my rationale:

1. I agree with this one in general but Faraday is a bit tricky. It is a fairly common building block for many other Ruby gems that deal with API requests. For example, our app has Faraday as a dependency for 10 other Ruby gems. Some of those gems use Faraday to send arbitrary requests and they can't use the built-in helper methods (e.g. `Faraday.get`). So they either resort to `run_request` (2 out of 10 use it including elasticsearch-transport) or just use `Object#send` (3 of 10 in our case) which is terrible and they would be better of with `run_request` instead. Faraday also doesn't have a helper shortcut for `OPTIONS` HTTP method too. So, while `run_request` is not surfaced in the getting-started docs, it is part of the public interface. Given the Faraday ubiquitousness as a building block for generic API requests I wouldn't be too concerned about this method being removed. One benefit of relying on it in tests is that all the other helpers are just a [meta-programmed wrappers](https://github.com/lostisland/faraday/blob/main/lib/faraday/connection.rb#L197-L206) invoking `run_request`. I believe using `run_request` in tests provides a wider coverage against regressions. All that being said, my main objective was to make the test suite fail w/o the applied fix. I can add a separate test-case specifically to validate that `run_request` doesn't result in a leak and leave the rest of the tests unchanged - would that be a better option?
2. I think you are correct about not having an explicit test case. However I was mainly following the approach from [this PR](https://github.com/open-telemetry/opentelemetry-ruby/pull/648/files#diff-8bc6849ebbe4b97f7c4edc9402d2375e99f0246c5afc8b35d5551d8f8ed2f0e2) where we made the same call on implicitly relying on Faraday to wipe out credentials from the URL while leaving existing assertions unchanged. I'm happy to add an explicit separate test if you feel strongly about this.
3. This is an excellent point! I'll open an issue or a PR to get the fix into Faraday itself. I do agree that, whether of Faraday doing the right thing or not, we should sanitize to be extra-safe.

To sum up:  I'm more than happy to add a separate specific test case leaving the rest of the tests unchanged. But I do believe that specific case should use `run_request` for us to make the test fail w/o the fix. Please let me know if that makes sense and I'll do the modifications.

plantfansam:
Hello @gregolsen, and thank you for your contribution! Interrupting the conversation here for a moment....

To do that, you can:

gregolsen:
Moved to https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/53

PR #1318, opened 2022-06-15
exporter/http: gzip request body outside of retryable block

tanner-bruce:
Fixes #1302

Issue #1302, opened 2022-06-08
Trace exporter request body gzipped on each attempt


tanner-bruce:
The request body is gzipped on [each retry attempt](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb#L139-L144) when exporting traces via otlp-http

plantfansam:
To make things more explicit: you're saying we ought to cache the gzipped contents to avoid the overhead of re-gzipping, right?

tanner-bruce:
Fixed by #1322

PR #1322, opened 2022-06-17
Report bundle size stats in exporter; also don't re-gzip unnecessarily

plantfansam:
- It can be helpful for operators to see the size of the batches going over the wire
- We are unnecessarily re-gzipping in the exporter (see https://github.com/open-telemetry/opentelemetry-ruby/pull/1318 and https://github.com/open-telemetry/opentelemetry-ruby/issues/1302)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -133,21 +133,25 @@ def around_request
 
         def send_bytes(bytes, timeout:) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/MethodLength, Metrics/PerceivedComplexity
           return FAILURE if bytes.nil?
+          @metrics_reporter.record_value('otel.otlp_exporter.bundle_bytes', value: bytes.bytesize)
```
arielvalentin, line 136: I lean towards naming this attribute closer to how we name things using semantic conventions. I realize that this is not the _entire_ size of the request but perhaps something like this:

`otel.otlp_exporter.bundle.uncompressed_size`
`otel.otlp_exporter.bundle.compressed_size`

WDYT?

https://opentelemetry.io/docs/reference/specification/trace/semantic_conventions/rpc/#events

fbogsany (approved):

PR #1326, opened 2022-06-21
fixup: remove active_model_serializers from releases.yml

ahayworth:
This was missed in the great -contrib migration, and causes the "Update open releases" hook to fail on pushes to `main`.


arielvalentin (approved):

plantfansam (approved):
👍

robertlaurin (approved):

PR #1332, opened 2022-06-24
feat: Consistent probability sampler

fbogsany:
Implements the [TraceState: Probability Sampling spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md) and the associated [TraceState Handling spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-handling.md).

Note: these will probably move to an `-experimental` or `-unstable` SDK extension gem.

TODO:
- [ ] [Sampler composition rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#composition-rules)
- [ ] [Statistical test requirements](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#appendix-statistical-test-requirements)


sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,74 @@
...
+          def generate_r(trace_id)
+            i = 8
+            trace_id.each_byte do |b|
+              return i - (b&3).bit_length if i == 64
+              return i - b.bit_length unless b.zero?
+              i = i + 8
+            end
+          end
+
+          def y_u_pp?
```
fbogsany, line 63: https://twitter.com/Loudwire/status/1459247159942033408

sdk/lib/opentelemetry/sdk/trace/samplers.rb
```diff
@@ -85,6 +88,25 @@ def self.trace_id_ratio_based(ratio)
...
+
+          ConsistentProbabilityBased.new(ratio)
+        end
+
+        # Returns a new sampler.
+        #
+        # @param [Sampler] root The sampler to which the sampling
+        #   decision is delegated for spans with no parent (root spans).
+        def self.parent_consistent_probability_based(root:)
+          ParentConsistentProbabilityBased.new(root)
```
ahayworth, line 108: Should we raise `ArgumentError` if you don't pass a root span sampler?

From the spec:
> The ParentConsistentProbabilityBased Sampler constructor SHOULD take a single Sampler argument, which is the Sampler to use in case the ParentConsistentProbabilityBased Sampler is called for a root span.

So I certainly understand why we might not want to enforce the argument, but as the code stands now it'll just blow up at runtime (which is definitely not what we want). Alternately we could address the `nil` root-sampler case directly in `ParentConsistentProbabilityBased` and just ... idk, always sample or something?

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,89 @@
...
+
+          # @api private
+          #
+          # See {Samplers}.
+          def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:) # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity
+            parent_span_context = OpenTelemetry::Trace.current_span(parent_context).context
+            if !parent_span_context.valid?
+              r = generate_r(trace_id)
+              p = probabilistic_p
+              if p < r
```
ahayworth, line 42: Should this be `p <= r` ? From the spec: `The span would be sampled because p-value 2 is less than or equal to r-value 5.` If so, we should update everywhere.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
```diff
@@ -0,0 +1,49 @@
...
+            end
+          end
+
+          def invariant(p, r, sampled) # rubocop:disable Naming/UncommunicativeMethodParamName
+            ((p <= r) == sampled) || (sampled && (p == 63))
+          end
+
+          def decimal(str)
+            str.to_i if !str.nil? && DECIMAL.match?(str)
+          end
```
ahayworth, line 44: I always find myself surprised that ruby has never implemented this as a method on `String` directly. That said, should this really be called `decimal` (see also, the `DECIMAL` const)? I guess that's technically correct, but honestly I usually expect to see a decimal separator when we call something a "decimal" (which is entirely because I did not major in mathematics and I'm in fact quite bad at math).

My gut feeling is that we should call this `integer` (and rename the constant). Although I should state that I don't actually care much, and will reserve my bikeshedding for the place where you explicitly requested it. 😄

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
```diff
@@ -0,0 +1,49 @@
...
+      module Samplers
+        # @api private
+        #
+        # The ConsistentProbabilityTraceState module implements common TraceState
+        # validation and manipulation for the consistent probability-based samplers.
+        module ConsistentProbabilityTraceState
+          DECIMAL = /\A\d+\z/.freeze
+          private_constant(:DECIMAL)
+
+          def validate_tracestate(span_context) # rubocop:disable Metrics/PerceivedComplexity, Metrics/CyclomaticComplexity
```
ahayworth, line 19: The method naming here is interesting - we're not really returning boolean indicating if it's valid; we're actually parsing the `ot` bits out of the trace state, validating them, allowing you to modify them, and then serializing it all back to the trace state.

Maybe `with_validated_tracestate`? Not sure really.

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Trace
+      # TraceState represents the OpenTelemetry::Trace::Tracestate values that
+      # are part of the OpenTelemetry ecosystem, contained in a single entry
+      # using the `ot` key. The format and handling requirements are specified
+      # at https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-handling.md
+      class TraceState
```
ahayworth, line 14: The overloading of `TraceState` here made me scratch my head for awhile. We're really defining a programmatic way to interact with the part of the API TraceState that corresponds to the special, reserved `ot` bits.

It's a little wordy, but it would be more clear to me if we called it `SDK::Trace::TraceState::OpenTelemetryConcern` or something like that, so it's clear that it's a subset. And then you don't end up with interesting calls like `ot = OpenTelemetry::SDK::Trace::Tracestate.from_tracestate(tracestate)` 😄

(Part of me almost wants to suggest that we could somehow fold all of this in as an SDK extension to the API Tracestate class - intercepting calls to get / set the `ot` key, etc ... but that feels equally problematic somehow).

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+
+          # Returns a newly created TraceState parsed from the Tracestate provided.
+          #
+          # @param [OpenTelemetry::Trace::Tracestate] tracestate
+          # @return [TraceState] A new TraceState instance or DEFAULT
+          def from_tracestate(tracestate) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+            return DEFAULT if tracestate.empty?
+
+            ot = tracestate.value('ot')
+            return DEFAULT if ot.nil? || ot.length > MAX_LIST_LENGTH
```
ahayworth, line 26: We should probably warn here, like we do within `from_hash`.

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+
+            hash = ot.split(';').each_with_object({}) do |member, memo|
+              member.strip!
```
ahayworth, line 29: We could have spaces in a W3C trace context value (apparently, from reading the spec), but nothing in the `ot` key should ever have a space ... and yet I know it's wise to be defensive here.

I have no useful comment, I just hate computers and specs sometimes.

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+              kv = member.split(':')
+              k, v = *kv
+              # TODO: "the used keys MUST be unique." - do we validate the header during parsing or just during formatting?
+              next unless kv.length == 2 && VALID_KEYS.include?(k) && VALID_VALUE.match?(v)
```
ahayworth, line 33:
```suggestion
              k, v, remainder = member.strip.split(':')
              # TODO: "the used keys MUST be unique." - do we validate the header during parsing or just during formatting?
              next unless remainder.nil? && VALID_KEYS.include?(k) && VALID_VALUE.match?(v)
```

```
irb(main):001:0> a = "foo:bar"
=> "foo:bar"
irb(main):002:0> k, v, remainder = a.strip.split(":")
=> ["foo", "bar"]
irb(main):003:0> remainder.nil?
=> true
irb(main):004:0> a = "foo:bar:baz"
=> "foo:bar:baz"
irb(main):005:0> k, v, remainder = a.strip.split(":")
=> ["foo", "bar", "baz"]
irb(main):006:0> remainder.nil?
=> false
irb(main):007:0> RUBY_VERSION
=> "2.7.2"
irb(main):008:0>
```

ahayworth: I guess I didn't bother to demonstrate but `k` and `v` get destructured correctly in those examples, as well.

ahayworth: For the `TODO` comment about uniqueness validation ... I'm not sure if we ever really need to do that?

Consider:

```
irb(main):006:1* hash = "foo:bar;foo:baz".split(";").each_with_object({}) do |member, memo|
irb(main):007:1*   k, v, remainder = member.strip.split(":")
irb(main):008:1*   next unless remainder.nil? # not doing the key or value validation for the example
irb(main):009:1*   memo[k] = v
irb(main):010:0> end
=> {"foo"=>"baz"}
irb(main):011:0>
```

Similar thoughts apply for the `from_hash` method as well; it would be very hard indeed to construct a hash that actually resulted in duplicate keys once serialized.

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+        end
+
+        # Returns this TraceState encoded according to the specification at
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-handling.md,
+        #
+        # @return [String] this TraceState encoded as a string.
+        def to_s
+          @hash.inject(+'') do |memo, (k, v)|
+            memo << k << ':' << v << ';'
+          end.chop! || ''
```
ahayworth, line 157: Is the `+''` argument to `inject` something to do with frozen string literals? I've never seen that before.

While I'm at it, another way to do this might be:

```
irb(main):001:0> h = { 'foo' => 'bar', 'baz' => 'bat' }
=> {"foo"=>"bar", "baz"=>"bat"}
irb(main):002:0> h.map { _1.join(':') }.join(';')
=> "foo:bar;baz:bat"
irb(main):003:0> RUBY_VERSION
=> "2.7.2"
irb(main):004:0>
```

But it's a style thing I guess.

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
```diff
@@ -0,0 +1,186 @@
...
+
+        # Sets the value of the `ot` key in tracestate to this TraceState
+        # encoded as a string.
+        #
+        # @param [Tracestate] tracestate The Tracestate to update.
+        # @return [Tracestate] tracestate, if unchanged, or a new Tracestate
+        #   containing this TraceState under the `ot` key, encoded as a string.
+        def this_is_a_deliberately_terrible_name_please_bikeshed(tracestate)
+          tracestate.set_value('ot', to_s)
+        end
```
ahayworth, line 148: `encode_to_tracestate!` ?

ahayworth:
Funny that GH interprets every comment as an independent review. Also, I should have just clicked 'start a review' but oh well.

So overall it looks pretty good to me - I think the `TODO` about the testing stuff is a definite must. I was _mostly_ able to follow the spec and it _looks_ like it does what it's supposed to but I don't think I grasp it well enough to be able to spot problems without running tests on it.

Not sure how we'd do the composition stuff. Do we even have support for composite samplers? I don't see anything in the SDK spec about that in general ... 🤔

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
ahayworth: Update: `with_validated_tracestate` also doesn't make a ton of sense, given that we don't always _expect_ it to be called with a block... maybe `sanitize_tracestate!` ? 🤔

sdk/lib/opentelemetry/sdk/trace/samplers.rb
fbogsany: The code matches what we do with the required `root` parameter to `parent_based`: https://github.com/open-telemetry/opentelemetry-ruby/blob/6bec625ef49004f364457c26263df421526b60d6/sdk/lib/opentelemetry/sdk/trace/samplers.rb#L67-L73

Maybe I'm holding it wrong, but I think the required named parameter is the right approach here. To pass `nil`, someone would have to explicitly configure it as `parent_consistent_probability_based(root: nil)` which seems like "if you do this, you deserve the 💥 ". You're not going to end up with this by accident.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,89 @@
...
+            end
+          end
+
+          private
+
+          def generate_r(trace_id)
+            x = trace_id[8, 8].unpack1('Q>') | 0x3
+            clz = 64 - x.bit_length
+            clz
+          end
```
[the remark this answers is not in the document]
fbogsany: Yeah - I experimented with this a bunch in the interpreter. It's not intuitive, but it *is* efficient.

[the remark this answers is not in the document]
fbogsany: To be clear, `clz` is "count leading zero bits", which is what the spec recommends.
```ruby
            x = trace_id[8, 8].unpack1('Q>') | 0x3
```
Take the same 8 bytes (64 bits) of the randomly generated trace ID that we use in the trace ID ratio-based sampler. Set the 2nd bit (well, 1st and 2nd). This ensures that when we count leading zeros, we get a max of 62 (64 - 2).
```ruby
            clz = 64 - x.bit_length
```
`bit_length` returns the minimum number of bits required to represent the value. I.e. a value like `0001 0011` has a bit length of 5 because we can represent it as `1 0011`. Since we have 64 bits and `bit_length` returns the highest set bit, `64 - x.bit_length` is the count of leading zeros.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
fbogsany: Decimal is the correct term (contrast with "hexadecimal", which is also an integer, but includes non-decimal characters). The spec also calls this "decimal encoding".

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
```diff
@@ -0,0 +1,49 @@
...
+            sampled = span_context.trace_flags.sampled?
+            tracestate = span_context.tracestate
+            ot = OpenTelemetry::SDK::Trace::Tracestate.from_tracestate(tracestate)
+            r = decimal(ot.value('r'))
+            p = decimal(ot.value('p'))
+            new_ot = ot
+            new_ot = ot.delete('p') if !p.nil? && p > 63
+            new_ot = ot.delete('p').delete('r') if !r.nil? && r > 62
+            r = nil if !r.nil? && r > 62 # TODO: clean up this logic
+            new_ot = ot.delete('p') if !p.nil? && !r.nil? && !invariant(p, r, sampled)
```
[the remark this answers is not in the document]
fbogsany: It is hard to follow. I explicitly asked Robert for his opinion on this, since I found it messy, but it captures the requirements and we had a hard time finding a clearer way to express it in code. Suggestions welcome.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
fbogsany: It's rough because the spec refers to this as validating tracestate. I totally agree that it doesn't make sense, and `sanitize_tracestate` is clearer, but we'd then need to add comments to tie this back to the spec (so someone reading the spec alongside the code can understand what we're implementing).

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
fbogsany: Yeah - the overloading is unpleasant. Again, I'm trying to adhere closely to the spec terms. `OpenTelemetryConcern` is reasonable. The spec uses that term in describing the `ot` [Key](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-handling.md#key).

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
fbogsany: https://ruby-doc.org/core-3.1.2/String.html#method-i-2B-40

> +string → new_string or self
> Returns `self` if `self` is not frozen.
> 
> Otherwise. returns `self.dup`, which is not frozen.

This gives us a mutable string, whereas string literals are usually frozen.

fbogsany: The alternative style `h.map { _1.join(':') }.join(';')` will (probably) allocate more temporary strings if there are many keys. It's unlikely to matter much in practice, but this mimics what we do in the "other" `Tracestate` class, which is where I stole it from: https://github.com/open-telemetry/opentelemetry-ruby/blob/672739120a605c5c14afc48e5aa07d1320443355/api/lib/opentelemetry/trace/tracestate.rb#L127-L131

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
fbogsany: The suggestion has one more allocation (for the `strip` vs `strip!`) per iteration.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
fbogsany: Thanks! Good catch.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
fbogsany: Technically, we're not mutating the tracestate - we're returning a validated (really, sanitized, as you said) copy of the tracestate extracted from the `span_context`. So the `!` doesn't apply here.

[The implementation for Go](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/samplers/probability/consistent/tracestate.go) is interesting in that they haven't built a generic representation of the OpenTelemetry `tracestate` concern, probably because these are the only keys currently defined - they've built something very specific - essentially just the pair of values. I'm trying to decide if we should do something similar, and if that will simplify the code a little (e.g. get rid of the `Tracestate`/`OpenTelemetryConcern` class).

sdk/lib/opentelemetry/sdk/trace/samplers/parent_consistent_probability_based.rb
```diff
@@ -0,0 +1,62 @@
...
+
+          # @api private
+          #
+          # See {Samplers}.
+          def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:)
+            parent_span_context = OpenTelemetry::Trace.current_span(parent_context).context
+            if !parent_span_context.valid?
+              @root.should_sample?(trace_id: trace_id, parent_context: parent_context, links: links, name: name, kind: kind, attributes: attributes)
+            else
+              tracestate = validate_tracestate(parent_span_context)
```
fbogsany, line 46: Interestingly, the [implementation for Go](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/efbcbfcb06541ef690398898f5d281068b92581b/samplers/probability/consistent/parent.go#L40-L67) always cleans up the `tracestate` even for root spans:
```go
	// Note: We do not check psc.IsValid(), i.e., we repair the tracestate
	// with or without a parent TraceId and SpanId.
	state := psc.TraceState()


	otts, err := parseOTelTraceState(state.Get(traceStateKey), psc.IsSampled())


	if err != nil {
		otel.Handle(err)
		value := otts.serialize()
		if len(value) > 0 {
			// Note: see the note in
			// "go.opentelemetry.io/otel/trace".TraceState.Insert(). The
			// error below is not a condition we're supposed to handle.
			state, _ = state.Insert(traceStateKey, value)
		} else {
			state = state.Delete(traceStateKey)
		}


		// Fix the broken tracestate before calling the delegate.
		params.ParentContext = trace.ContextWithSpanContext(params.ParentContext, psc.WithTraceState(state))
```

I couldn't find any requirement for this in the spec, which only states:

> It is required to first validate the `tracestate` and then respect the `sampled` flag in the W3C traceparent.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,89 @@
...
+          def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:) # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity
+            parent_span_context = OpenTelemetry::Trace.current_span(parent_context).context
+            if !parent_span_context.valid?
+              r = generate_r(trace_id)
+              p = probabilistic_p
+              if p < r
+                tracestate = TraceState.from_hash({ 'p' => p.to_s, 'r' => r.to_s })
+                Result.new(decision: Decision::RECORD_AND_SAMPLE, tracestate: tracestate)
+              else
+                tracestate = TraceState.from_hash({ 'r' => r.to_s })
```
fbogsany, line 46: And here.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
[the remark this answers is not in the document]
ahayworth: Ahhhhh this makes it much more clear. I really appreciate that, thank you. 😄

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
ahayworth: That's totally fair, and like I said I don't really feel _that_ strongly. Objection withdrawn! 😄

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
ahayworth: > The suggestion has one more allocation (for the strip vs strip!) per iteration.

And we do need to be careful with allocations, so ... never mind! 😄 

> so in this case it is very much possible to have duplicates in the input (but not in the output).

Ah yeah, good point. Maybe we just `warn` if we see a duplicate key and keep going.

> are we required to validate uniqueness when parsing the ot tracestate entry, and reject invalid entries, or do we just adopt "last one wins"?

😬 My vote is "warn during parse, allow last one to win" ... and maybe we should bring this up in the spec directly then?

sdk/lib/opentelemetry/sdk/trace/trace_state.rb
ahayworth: Interesting - I went looking for that in the docs but must have overlooked it entirely.

Regarding the alternate style - if you suspect it'll allocate more, than never mind. Like I just commented elsewhere it _is_ important for us to be miserly with our allocations so I definitely prefer the less-expensive version. And, consistency with how we do it elsewhere (as you also said) is super good 👍. 😄

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,89 @@
...
+          #
+          # See {Samplers}.
+          def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:) # rubocop:disable Metrics/MethodLength, Metrics/PerceivedComplexity
+            parent_span_context = OpenTelemetry::Trace.current_span(parent_context).context
+            if !parent_span_context.valid?
+              r = generate_r(trace_id)
+              p = probabilistic_p
+              if p < r
+                tracestate = TraceState.from_hash({ 'p' => p.to_s, 'r' => r.to_s })
+                Result.new(decision: Decision::RECORD_AND_SAMPLE, tracestate: tracestate)
```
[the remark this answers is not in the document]
ahayworth: 🦶 🔫 💥

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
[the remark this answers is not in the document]
fbogsany: I have a substantial refactoring (after reading the Go implementation) that might clean this up a bit. I'll try to push that up today.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
fbogsany: I'm refactoring at the moment based on the Go implementation. I *think* it'll be a little more efficient, and we can revert the refactoring later if/when we add more otel-specific concerns in the tracestate. I hope, when we do that, that we have a clear and consistent API spec for working with those concerns.

fbogsany:
I mentioned in a few comments above: I'm in the process of refactoring this based on insights gleaned from reading the Go implementation. 🤞 it'll be more efficient and easier to read & understand ("obviously correct" rather than "not obviously wrong"). I hope to have that pushed up today, or early next week at the latest.

sdk/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
ahayworth: That makes sense to me!

fbogsany:
@ahayworth I pushed up the refactor. This is good for another 👀 . Still needs tests & 🤔 about composition.

ahayworth (approved):
I can follow that a lot better, yeah. I'm still holding out to see how this performs against an actual test suite since I'm just not convinced I understand the spec very well myself ... but it's looking great!

I have no useful opinions on how to compose the samplers, unfortunately. :(

sdk_experimental/test/opentelemetry/sdk/trace/samplers/consistent_probability_based_test.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    it 'generates a new r if r is invalid in the parent tracestate' do
+      tid = trace_id(1)
+      ctx = parent_context(trace_id: tid, ot: 'p:1;r:63')
+      result = call_sampler(subject, trace_id: tid, parent_context: ctx)
+      _(result.tracestate['ot']).must_equal('p:1;r:62')
+      _(result).must_be :sampled?
+    end
+
+    # TODO: statistical tests
```
fbogsany, line 63: See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#appendix-statistical-test-requirements

sdk_experimental/lib/opentelemetry/sdk/trace/samplers/parent_consistent_probability_based.rb
```diff
@@ -0,0 +1,87 @@
...
+          # tracestate probability sampling spec.
+          #
+          # @param span_context [OpenTelemetry::Trace::SpanContext] the parent span context
+          # @return [OpenTelemetry::Trace::Tracestate] the sanitized tracestate
+          def sanitized_tracestate(span_context)
+            sampled = span_context.trace_flags.sampled?
+            tracestate = span_context.tracestate
+            parse_ot_vendor_tag(tracestate) do |p, r, rest|
+              if !r.nil? && r > 62
+                p = r = nil
```
fbogsany, line 72: This isn't happening, according to the tests. Need to figure out why.

sdk_experimental/test/opentelemetry/sdk/trace/samplers/parent_consistent_probability_based_test.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    it 'sanitizes input tracestate' do
+      _(call_sampler(subject, parent_context: parent_context(ot: 'junk;p:1;r:1')).tracestate['ot']).must_equal('r:1;junk')
+      _(call_sampler(subject, parent_context: parent_context(ot: 'p:64;r:1')).tracestate['ot']).must_equal('r:1')
+      _(call_sampler(subject, parent_context: parent_context(sampled: false, ot: 'p:2;r:1')).tracestate['ot']).must_equal('p:2;r:1')
+      _(call_sampler(subject, parent_context: parent_context(sampled: true, ot: 'p:2;r:1')).tracestate['ot']).must_equal('r:1')
+      _(call_sampler(subject, parent_context: parent_context(sampled: true, ot: 'p:63;r:1')).tracestate['ot']).must_equal('p:63;r:1')
+      # TODO: figure out why this fails
+      # _(call_sampler(subject, parent_context: parent_context(ot: 'p:1;r:63')).tracestate['ot']).must_be_nil
```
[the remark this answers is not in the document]
ahayworth: Does the behavior we're not-testing here imply that the sampling isn't correct? I know this is experimental, but is this a show-stopper to even an experimental release?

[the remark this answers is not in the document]
fbogsany: I wrote some tests, found some 🐛 and fixed most of them. But not this one. 👁️ the code, it *should* work (see https://github.com/open-telemetry/opentelemetry-ruby/pull/1332#discussion_r923898330 below), but it doesn't. I'd like to figure out why before we 🚢 . But yeah, maybe we don't have to 🤷

fbogsany:
I'm going to propose implementing (or porting the [Go implementation](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/samplers/probability/consistent/statistical_test.go) of) the [statistical tests](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#appendix-statistical-test-requirements) in a subsequent PR.

I haven't seen an implementation of the [composition rules](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/tracestate-probability-sampling.md#composition-rules) and I'm inclined to ignore that part of the spec for now.

@ahayworth would you mind taking another 👀 at this before I 🚢 ?

sdk_experimental/lib/opentelemetry/sdk/trace/samplers/consistent_probability_based.rb
```diff
@@ -0,0 +1,89 @@
...
+          #
+          # See {Samplers}.
+          def should_sample?(trace_id:, parent_context:, links:, name:, kind:, attributes:)
+            parent_span_context = OpenTelemetry::Trace.current_span(parent_context).context
+            p = probabilistic_p
+            if parent_span_context.valid?
+              tracestate = parent_span_context.tracestate
+              parse_ot_vendor_tag(tracestate) do |_, in_r, rest|
+                r = if in_r.nil? || in_r > 62
+                      # TODO: warn the user that a potentially inconsistent trace is being produced
```
ahayworth, line 49:
```suggestion
                      OpenTelemetry.logger.debug("ConsistentProbabilitySampler: potentially inconsistent trace detected - r: #{in_r.inspect}")
```

ahayworth (approved):
I've been following it pretty closely, and tracking the changes over time - I don't see anything in the latest set of changes that looks concerning! I did leave one suggestion to warn the user about a potentially inconsistent trace, but that's certainly not a blocking suggestion.

There are I guess a few outstanding things:
- We haven't really decided on whether or not we must validate tracestate keys and raise a fuss if they're duplicated. I think (if I read things right) that "last duplicate key wins" in our implementation which may not technically be spec-compliant ... but I'm not super concerned. We can open a follow-up issue for that one.
- Re: statistical tests - yeah, I'm fine with shipping that as a next-step kind of thing.
- Re: composing samplers - I'm also fine with punting on that for now. Another thing we can open an issue for re: spec-compliance and come back to it later.

Overall, LGTM!

robertlaurin (approved):

ahayworth:
Thanks @robertlaurin ❤️

PR #1336, opened 2022-06-29
feat: Upgrade SemConv to v1.12.0

arielvalentin:
https://github.com/open-telemetry/opentelemetry-specification/tree/v1.12.0/semantic_conventions

### DIFF
--- a/semantic_conventions/Rakefile
+++ b/semantic_conventions/Rakefile
@@ -10,7 +10,7 @@ require 'yard'
 require 'rubocop/rake_task'
 require 'tmpdir'
 
-SPEC_VERSION = '1.10.0'
+SPEC_VERSION = '1.12.0'
 
 RuboCop::RakeTask.new
 
--- a/semantic_conventions/lib/opentelemetry/semantic_conventions/resource.rb
+++ b/semantic_conventions/lib/opentelemetry/semantic_conventions/resource.rb
@@ -7,6 +7,19 @@
 module OpenTelemetry
   module SemanticConventions
     module Resource
+      # Array of brand name and version separated by a space
+      # @note This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.brands)
+      BROWSER_BRANDS = 'browser.brands'
+
+      # The platform on which the browser is running
+      # @note This value is intended to be taken from the [UA client hints API](https://wicg.github.io/ua-client-hints/#interface) (navigator.userAgentData.platform). If unavailable, the legacy `navigator.platform` API SHOULD NOT be used instead and this attribute SHOULD be left unset in order for the values to be consistent.
+      #  The list of possible values is defined in the [W3C User-Agent Client Hints specification](https://wicg.github.io/ua-client-hints/#sec-ch-ua-platform). Note that some (but not all) of these values can overlap with values in the [os.type and os.name attributes](./os.md). However, for consistency, the values in the `browser.platform` attribute should capture the exact value that the user agent provides
+      BROWSER_PLATFORM = 'browser.platform'
+
+      # Full user-agent string provided by the browser
+      # @note The user-agent value SHOULD be provided only from browsers that do not have a mechanism to retrieve brands and platform individually from the User-Agent Client Hints API. To retrieve the value, the legacy `navigator.userAgent` API can be used
+      BROWSER_USER_AGENT = 'browser.user_agent'
+
       # Name of the cloud provider
       CLOUD_PROVIDER = 'cloud.provider'
 
@@ -96,24 +109,41 @@ module OpenTelemetry
       DEVICE_MANUFACTURER = 'device.manufacturer'
 
       # The name of the single function that this runtime instance executes
-      # @note This is the name of the function as configured/deployed on the FaaS platform and is usually different from the name of the callback function (which may be stored in the [`code.namespace`/`code.function`](../../trace/semantic_conventions/span-general.md#source-code-attributes) span attributes)
+      # @note This is the name of the function as configured/deployed on the FaaS
+      #  platform and is usually different from the name of the callback
+      #  function (which may be stored in the
+      #  [`code.namespace`/`code.function`](../../trace/semantic_conventions/span-general.md#source-code-attributes)
+      #  span attributes).
+      #  
+      #  For some cloud providers, the above definition is ambiguous. The following
+      #  definition of function name MUST be used for this attribute
+      #  (and consequently the span name) for the listed cloud providers/products:
+      #  
+      #  * **Azure:**  The full name `<FUNCAPP>/<FUNC>`, i.e., function app name
+      #    followed by a forward slash followed by the function name (this form
+      #    can also be seen in the resource JSON for the function).
+      #    This means that a span attribute MUST be used, as an Azure function
+      #    app can host multiple functions that would usually share
+      #    a TracerProvider (see also the `faas.id` attribute)
       FAAS_NAME = 'faas.name'
 
       # The unique ID of the single function that this runtime instance executes
-      # @note Depending on the cloud provider, use:
+      # @note On some cloud providers, it may not be possible to determine the full ID at startup,
+      #  so consider setting `faas.id` as a span attribute instead.
+      #  
+      #  The exact value to use for `faas.id` depends on the cloud provider:
       #  
       #  * **AWS Lambda:** The function [ARN](https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html).
-      #  Take care not to use the "invoked ARN" directly but replace any
-      #  [alias suffix](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html) with the resolved function version, as the same runtime instance may be invokable with multiple
-      #  different aliases.
+      #    Take care not to use the "invoked ARN" directly but replace any
+      #    [alias suffix](https://docs.aws.amazon.com/lambda/latest/dg/configuration-aliases.html)
+      #    with the resolved function version, as the same runtime instance may be invokable with
+      #    multiple different aliases.
       #  * **GCP:** The [URI of the resource](https://cloud.google.com/iam/docs/full-resource-names)
-      #  * **Azure:** The [Fully Qualified Resource ID](https://docs.microsoft.com/en-us/rest/api/resources/resources/get-by-id).
-      #  
-      #  On some providers, it may not be possible to determine the full ID at startup,
-      #  which is why this field cannot be made required. For example, on AWS the account ID
-      #  part of the ARN is not available without calling another AWS API
-      #  which may be deemed too slow for a short-running lambda function.
-      #  As an alternative, consider setting `faas.id` as a span attribute instead
+      #  * **Azure:** The [Fully Qualified Resource ID](https://docs.microsoft.com/en-us/rest/api/resources/resources/get-by-id) of the invoked function,
+      #    *not* the function app, having the form
+      #    `/subscriptions/<SUBSCIPTION_GUID>/resourceGroups/<RG>/providers/Microsoft.Web/sites/<FUNCAPP>/functions/<FUNC>`.
+      #    This means that a span attribute MUST be used, as an Azure function app can host multiple functions that would usually share
+      #    a TracerProvider
       FAAS_ID = 'faas.id'
 
       # The immutable version of the function being executed
--- a/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb
+++ b/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb
@@ -55,6 +55,7 @@ module OpenTelemetry
       DB_OPERATION = 'db.operation'
 
       # Remote hostname or similar, see note below
+      # @note `net.peer.name` SHOULD NOT be set if capturing it would require an extra DNS lookup
       NET_PEER_NAME = 'net.peer.name'
 
       # Remote address of the peer (dotted decimal for IPv4 or [RFC5952](https://tools.ietf.org/html/rfc5952) for IPv6)
--- a/semantic_conventions/lib/opentelemetry/semantic_conventions/version.rb
+++ b/semantic_conventions/lib/opentelemetry/semantic_conventions/version.rb
@@ -6,6 +6,6 @@
 
 module OpenTelemetry
   module SemanticConventions
-    VERSION = '1.10.0'
+    VERSION = '1.12.0'
   end
 end


ahayworth:
@arielvalentin This is fine as-is, but it looks like the semantic conventions generator isn't pulling in the `trace/instrumentation` subdirectory. I think that might be able to be fixed with some hacks to the `Rakefile` but I'm not sure.

arielvalentin:
@ahayworth I mentioned this in slack already but I was wondering if we could chat to discuss some possible changes to the layout of the source. 

One thing we have done at GH for our internal schema is to segment the SemConv attributes into an additional layer of submodules. 

This restructuring has improved the overall development experience for users that rely on auto-complete in IDEs, text editors and IRB. E.g.

```ruby
module GitHub
  module SemConv
    module Trace
      # The `aws` conventions apply to operations using the AWS SDK. They map request or response
      # parameters in AWS SDK API calls to attributes on a Span. The conventions have been collected over
      # time based on feedback from AWS users of tracing and will continue to evolve as new interesting
      # conventions are found.
      # Some descriptions are also provided for populating general OpenTelemetry semantic conventions
      # based on these APIs.
      module Aws
      end
      # Attributes that exist for multiple DynamoDB request types.
      module AwsDynamodb
        # The keys in the `RequestItems` object field
        #
        # Required: No
        #
        # Examples:
        #   'Users', 'Cats'
        TABLE_NAMES = "aws.dynamodb.table_names"
        # The JSON-serialized value of each item in the `ConsumedCapacity` response field
        #
        # Required: No
        #
        # Examples:
        #   '{ "CapacityUnits": number, "GlobalSecondaryIndexes": { "string" : { "CapacityUnits": number,
        #  "ReadCapacityUnits": number, "WriteCapacityUnits": number } }, "LocalSecondaryIndexes": {
        #  "string" : { "CapacityUnits": number, "ReadCapacityUnits": number, "WriteCapacityUnits": number }
        #  }, "ReadCapacityUnits": number, "Table": { "CapacityUnits": number, "ReadCapacityUnits": number,
        #  "WriteCapacityUnits": number }, "TableName": "string", "WriteCapacityUnits": number }'
        CONSUMED_CAPACITY = "aws.dynamodb.consumed_capacity"
        # The JSON-serialized value of the `ItemCollectionMetrics` response field
        #
        # Required: No
        #
        # Examples:
        #   '{ "string" : [ { "ItemCollectionKey": { "string" : { "B": blob, "BOOL": boolean, "BS": [ blob ],
        #  "L": [ "AttributeValue" ], "M": { "string" : "AttributeValue" }, "N": "string", "NS": [ "string"
        #  ], "NULL": boolean, "S": "string", "SS": [ "string" ] } }, "SizeEstimateRangeGB": [ number ] } ]
        #  }'
        ITEM_COLLECTION_METRICS = "aws.dynamodb.item_collection_metrics"
        # The value of the `ProvisionedThroughput.ReadCapacityUnits` request parameter
        #
        # Required: No
        #
        # Examples:
        #   1.0, 2.0
        PROVISIONED_READ_CAPACITY = "aws.dynamodb.provisioned_read_capacity"
        # The value of the `ProvisionedThroughput.WriteCapacityUnits` request parameter
        #
        # Required: No
        #
        # Examples:
        #   1.0, 2.0
        PROVISIONED_WRITE_CAPACITY = "aws.dynamodb.provisioned_write_capacity"
        # The value of the `ConsistentRead` request parameter
        #
        # Required: No
        CONSISTENT_READ = "aws.dynamodb.consistent_read"
        # The value of the `ProjectionExpression` request parameter
        #
        # Required: No
        #
        # Examples:
        #   'Title', 'Title, Price, Color', 'Title, Description, RelatedItems, ProductReviews'
        PROJECTION = "aws.dynamodb.projection"
        # The value of the `Limit` request parameter
        #
        # Required: No
        #
        # Examples:
        #   10
        LIMIT = "aws.dynamodb.limit"
        # The value of the `AttributesToGet` request parameter
        #
        # Required: No
        #
        # Examples:
        #   'lives', 'id'
        ATTRIBUTES_TO_GET = "aws.dynamodb.attributes_to_get"
        # The value of the `IndexName` request parameter
        #
        # Required: No
        #
        # Examples:
        #   'name_to_group'
        INDEX_NAME = "aws.dynamodb.index_name"
        # The value of the `Select` request parameter
        #
        # Required: No
        #
        # Examples:
        #   'ALL_ATTRIBUTES', 'COUNT'
        SELECT = "aws.dynamodb.select"
        # The JSON-serialized value of each item of the `GlobalSecondaryIndexes` request field
        #
        # Required: No
        #
        # Examples:
        #   '{ "IndexName": "string", "KeySchema": [ { "AttributeName": "string", "KeyType": "string" } ],
        #  "Projection": { "NonKeyAttributes": [ "string" ], "ProjectionType": "string" },
        #  "ProvisionedThroughput": { "ReadCapacityUnits": number, "WriteCapacityUnits": number } }'
        GLOBAL_SECONDARY_INDEXES = "aws.dynamodb.global_secondary_indexes"
        # The JSON-serialized value of each item of the `LocalSecondaryIndexes` request field
        #
        # Required: No
        #
        # Examples:
        #   '{ "IndexArn": "string", "IndexName": "string", "IndexSizeBytes": number, "ItemCount": number,
        #  "KeySchema": [ { "AttributeName": "string", "KeyType": "string" } ], "Projection": {
        #  "NonKeyAttributes": [ "string" ], "ProjectionType": "string" } }'
        LOCAL_SECONDARY_INDEXES = "aws.dynamodb.local_secondary_indexes"
        # The value of the `ExclusiveStartTableName` request parameter
        #
        # Required: No
        #
        # Examples:
        #   'Users', 'CatsTable'
        EXCLUSIVE_START_TABLE = "aws.dynamodb.exclusive_start_table"
        # The the number of items in the `TableNames` response parameter
        #
        # Required: No
        #
        # Examples:
        #   20
        TABLE_COUNT = "aws.dynamodb.table_count"
        # The value of the `ScanIndexForward` request parameter
        #
        # Required: No
        SCAN_FORWARD = "aws.dynamodb.scan_forward"
        # The value of the `Segment` request parameter
        #
        # Required: No
        #
        # Examples:
        #   10
        SEGMENT = "aws.dynamodb.segment"
        # The value of the `TotalSegments` request parameter
        #
        # Required: No
        #
        # Examples:
        #   100
        TOTAL_SEGMENTS = "aws.dynamodb.total_segments"
        # The value of the `Count` response parameter
        #
        # Required: No
        #
        # Examples:
        #   10
        COUNT = "aws.dynamodb.count"
        # The value of the `ScannedCount` response parameter
        #
        # Required: No
        #
        # Examples:
        #   50
        SCANNED_COUNT = "aws.dynamodb.scanned_count"
        # The JSON-serialized value of each item in the `AttributeDefinitions` request field
        #
        # Required: No
        #
        # Examples:
        #   '{ "AttributeName": "string", "AttributeType": "string" }'
        ATTRIBUTE_DEFINITIONS = "aws.dynamodb.attribute_definitions"
        # The JSON-serialized value of each item in the the `GlobalSecondaryIndexUpdates` request field
        #
        # Required: No
        #
        # Examples:
        #   '{ "Create": { "IndexName": "string", "KeySchema": [ { "AttributeName": "string", "KeyType":
        #  "string" } ], "Projection": { "NonKeyAttributes": [ "string" ], "ProjectionType": "string" },
        #  "ProvisionedThroughput": { "ReadCapacityUnits": number, "WriteCapacityUnits": number } }'
        GLOBAL_SECONDARY_INDEX_UPDATES = "aws.dynamodb.global_secondary_index_updates"
      end
      # Span attributes used by AWS Lambda (in addition to general `faas` attributes).
      module AwsLambda
        # The full invoked ARN as provided on the `Context` passed to the function (`Lambda-Runtime-Invoked-Function-Arn` header on the `/runtime/invocation/next` applicable)
        #
        # Required: No
        #
        # Examples:
        #   'arn:aws:lambda:us-east-1:123456:function:myfunction:myalias'
        # @note This may be different from `faas.id` if an alias is involved
        INVOKED_ARN = "aws.lambda.invoked_arn"
      end
# and so on
end
```

Does this sound reasonable to you?

ahayworth:
@arielvalentin In general, yes I think that's reasonable. @fbogsany and @robertlaurin raised concerns about overly-lengthy namespaces in the [original PR review](https://github.com/open-telemetry/opentelemetry-ruby/pull/858#issuecomment-874286984) that added auto-generation, and I think those concerns are still valid and something we should keep in mind.

Given a cassandra example (as in the original PR comments), what you're suggesting would take us from `OpenTelemetry::SemanticConventions::Trace::DB_CASSANDRA_TABLE` to `OpenTelemetry::SemanticConventions::Trace::DB::CASSANDRA_TABLE`? If so I think that's probably okay, but I would be curious what others think. Improved support for editor autocomplete is nice, but this is definitely one of those changes where people might look at it and say "WTF? Why do I have to go update my constant references everywhere, do the maintainers just hate me?" 😆

In general, I do not personally have incredibly strong feelings on the topic and would generally support whatever most people think is the best option.

Also as mentioned in Slack, I have some free time this afternoon to sync up about it if you'd like, and we can bring any interesting ideas from that conversation back to the PR for people to consider. ❤️

fbogsany:
> Given a cassandra example (as in the original PR comments), what you're suggesting would take us from `OpenTelemetry::SemanticConventions::Trace::DB_CASSANDRA_TABLE` to `OpenTelemetry::SemanticConventions::Trace::DB::CASSANDRA_TABLE`?

I think it becomes `OpenTelemetry::SemanticConventions::Trace::DB::Cassandra::TABLE`. As mentioned previously, this is getting a little long, but then again it's already quite long. If it helps someone, I'm not opposed. I'd love it if we could come up with some coding conventions to make it less unwieldy to use the constants.

arielvalentin:
The generator groups by prefix, so it only generates one additional level because in many instances a group has fewer than 10 attributes. We also shorted the length of the `SemanticConventions` module name to `SemConv`. 

We did not feel it was necessary to dig any deeper than that. 

```ruby
      # Call-level attributes for SQL databases
      module DbSql
        # The name of the primary table that the operation is acting upon, including the database name (if applicable)
        #
        # Required: Recommended if available.
        #
        # Examples:
        #   'public.users', 'customers'
        # @note It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set
        TABLE = "db.sql.table"
      end

# Call-level attributes for Cassandra
      module DbCassandra
        # The fetch size used for paging, i.e. how many rows will be returned at once
        #
        # Required: No
        #
        # Examples:
        #   5000
        PAGE_SIZE = "db.cassandra.page_size"
        # The consistency level of the query. Based on consistency values from [CQL](https://docs.datastax.com/en/cassandra-oss/3.0/cassandra/dml/dmlConfigConsistency.html)
        #
        # Required: No
        CONSISTENCY_LEVEL = "db.cassandra.consistency_level"
        # The name of the primary table that the operation is acting upon, including the keyspace name (if applicable)
        #
        # Required: Recommended if available.
        #
        # Examples:
        #   'mytable'
        # @note This mirrors the db.sql.table attribute but references cassandra rather than sql. It is not recommended to attempt any client-side parsing of `db.statement` just to get this property, but it should be set if it is provided by the library being instrumented. If the operation is acting upon an anonymous table, or more than one table, this value MUST NOT be set
        TABLE = "db.cassandra.table"
        # Whether or not the query is idempotent
        #
        # Required: No
        IDEMPOTENCE = "db.cassandra.idempotence"
        # The number of times a query was speculatively executed. Not set or `0` if the query was not executed speculatively
        #
        # Required: No
        #
        # Examples:
        #   0, 2
        SPECULATIVE_EXECUTION_COUNT = "db.cassandra.speculative_execution_count"
        # The ID of the coordinating node for a query
        #
        # Required: No
        #
        # Examples:
        #   'be13faa2-8574-4d71-926d-27f16cf8a7af'
        COORDINATOR_ID = "db.cassandra.coordinator.id"
        # The data center of the coordinating node for a query
        #
        # Required: No
        #
        # Examples:
        #   'us-west-2'
        COORDINATOR_DC = "db.cassandra.coordinator.dc"
      end
```

The Golang and C# teams use enums and KeyValue types, which they are happy with but did not feel idiomatic to me in Ruby.

We also use mixins to mitigate very lengthy usage in our classes like so:

```ruby
  class Thing
    include GitHub::SemConv

    def call(user)
      logger.info("thing-called", { Trace::Enduser::ID => user.login )
    end
  end
```

> I'd love it if we could come up with some coding conventions to make it less unwieldy to use the constants.

We tried a few things but didn't really love anything. I've thought about creating a set of classes that implement enumerable or quack like a hash (inspired by [Faraday Headers](https://github.com/lostisland/faraday/blob/main/lib/faraday/utils/headers.rb) that would allow users to set attributes but it did not resonate with me:

E.g. 

```ruby
# this
db_attributes = OTel::SemConv::Trace::DB.new(system: "mysql", connection_string: conn_string, user: dbusername, statement: sql)

tracer.in_span("select", db_attributes) do
   # some some things
end

# vs this

db_attributes = { 
                             Trace::DB::SYSTEM =>"mysql", 
                             Trace::DB::CONNECTION_STRING => conn_string
                              race::DB::USER => dbusername,
                              race::DB::SQL => sql
                            }
tracer.in_span("select", db_attributes) do
   # some some things
end
```

Thoughts?

ahayworth:
> Thoughts?

Yes, my primary galaxy-brain idea here is that none of these are particularly thrifty when it comes to characters:

```
~ via 💎 v2.7.2 took 2s
zsh ❯ wc -c <<< "OpenTelemetry::SemanticConventions::Trace::DB_CASSANDRA_TABLE"
      62

~ via 💎 v2.7.2
zsh ❯ wc -c <<< "OpenTelemetry::SemanticConventions::Trace::Db::Cassandra::Table"
      64

~ via 💎 v2.7.2
zsh ❯ wc -c <<< "OpenTelemetry::SemanticConventions::Trace::DbCassandra::Table"
      62

~ via 💎 v2.7.2
zsh ❯
```

If we really care about that, the biggest impact would come from shortening `SemanticConventions` to `SemConv`, for a reduction in 12 characters. 😆 

---

On an unrelated note, I don't think this gem provides an exceptional amount of value today - mostly, it just makes us compliant with the standard. Maybe a bit of gentle encouragement for users to not make mistakes when they hard-code the values everywhere; maybe a little bit better autocomplete in some editors (but I don't really see the difference between namespaced constants and a global list of constants in that department; any editor worth it's salt should generate similar autocompletes for either case).

We _could_ do more, though - if we leveraged schema definitions alongside it, we could do nifty things like automatically translating removed/deprecated calls to their replacements, and log warnings when appropriate.

---

I kinda like something about this approach:

```ruby
db_attributes = OTel::SemConv::Trace::DB.new(system: "mysql", connection_string: conn_string, user: dbusername, statement: sql)
```

Specifically, we could validate what people are providing there and either warn or reject values that do _not_ comply with a given spec version. That could be nifty.

---

You could do something wild like make this a DSL:

```ruby

attributes = {
  OpenTelemetry::Semconv.db.cassandra.table => 'foo'
}

```

I'm not entirely sure if that's a _good_ idea, though...

---

After having said all of that, I wouldn't be opposed to shortening `SemanticConventions` to `SemConv`. 😄

arielvalentin:
> Specifically, we could validate what people are providing there and either warn or reject values that do not comply with a given spec version. That could be nifty.

That was the idea. Since it was namespaced users could set properties on the object or use constructor injection. 

The object would behave like a hash. The class name would be the namespace. It would have a limited set of keys it supported which would be the suffix, but it would have to me mergeable with user defined fields. 

For symmetry I looked to the Resource SDK but found that to be so _different_ from a hash. 

I asked myself what value would that give me than throwing a bunch of key-value pairs into a hash?

In the end I dropped the experiment. 

If you're up for it let's pair up. I'd be willing to give this another try.

ahayworth:
Honestly those are good points ... I feel like I need to think about it for awhile but I would otherwise be happy to pair on it once we have some solid ideas lined up.

(And in any case, it's not blocking _this_ PR :D )

ahayworth:
> I am also curious if you all would be annoyed if I shorted the parent namespace to be OTel.

Personally I don't actually want to do that. But - I wouldn't be opposed if it were optional somehow. I can't find a way to _not_ include some file in a `.gemspec`, but I could almost see something like `lib/shortener.rb`:

```ruby
OTel = OpenTelemetry
```

Screen Shot 2022-06-30 at 8 10 09 PM

plantfansam:
None of these comments should be interpreted as blocking the PR, which I think is splendid...

And now onto the bikeshedding:

I'm totally happy adding an additional layer of modules and am not worried about overly long constant names. I'd imagine that most folks will be tab-completing them, either via a purpose-built IDE or a vim/JEDI. Shortening to `SemConv` seems like a great thing to do, regardless of what happens, IMO.

If we do decide to add another layer of modules (and maybe if we don't), it could be nice to add a comment directly above each constant with its full namespace; this would allow easier copy/paste and easier codebase searches. So if we were to do that against the current nesting, it'd be like:

```ruby
# OpenTelemetry::SemanticConventions::Resource
BROWSER_USER_AGENT = 'browser.user_agent'
```

In the potential, more nested world:

```ruby
# OpenTelemetry::SemConv::Trace::AwsDynamoDb::TABLE_NAMES
TABLE_NAMES = "blahblahblah"
```

ahayworth:
@plantfansam that should be possible; there's a template in this directory that controls all of it.

Other than just being verbose, I think the other concern with additional layers of modules is lookup time: I don't know (but could maybe test) how much additional time it would take for ruby to look through additional layers of modules.

I think it's okay if it increases slightly with nested modules, but we should be careful not to make it a performance bottleneck of any kind. It's probably going to be fine but I think we owe it to ourselves to do a simple benchmark.

I may try to do that today at the airport if I get bored (and I suspect I will get bored).

ahayworth:
Benchmarking results are actually interesting - see gist: https://gist.github.com/ahayworth/f7a5a0697d79dd9a09c73d3943852730

However the tl;dr from my perspective is "differences are basically negligible, especially on 3.1.2" so I think we should not worry about any kind of performance impact.

plantfansam:
Thanks for the benchmarking @ahayworth! I've put in my 2 cents and feel OK with any course of action described in this PR discussion.

fbogsany:
I don't think we should change either namespace. Longer term (well, I'd prefer it to be shorter term, but no-one has the energy to take this on), we should build a library of helpers of the form:
```ruby
def http_client_span_attributes(a_required_attribute:, an_optional_attribute: nil, ...)
```
that would provide a ton more value and mostly abstract the use of the semantic convention constants defined here.

PR #1339, opened 2022-07-07
fix: `OTEL_EXPORTER_OTLP_ENDPOINT` appends the correct path with a trailing slash

pkanal:
Addresses #1340 

## Changes
- Added a check to see if the `OTEL_EXPORTER_OTLP_ENDPOINT` ends in a trailing `/` and if it does append `v1/traces` appropriately
- Unit tests to test whether endpoints work with or without the trailing `/`
- Extracted `v1/traces` into a constant variable

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -58,7 +58,7 @@ module OpenTelemetry
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
           @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
-                   URI(endpoint.end_with?('/') ? "#{endpoint}#{TRACES_PATH}" : "#{endpoint}/#{TRACES_PATH}")
+                   URI(File.join(endpoint, TRACES_PATH))
                  else
                    URI(endpoint)
                  end
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -58,7 +58,7 @@ module OpenTelemetry
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
           @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
-                   URI(File.join(endpoint, TRACES_PATH))
+                   URI.join(endpoint, 'v1/traces')
                  else
                    URI(endpoint)
                  end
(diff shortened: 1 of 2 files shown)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -56,7 +58,7 @@ def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPOR
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
           @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
-                   URI("#{endpoint}/v1/traces")
+                   URI(endpoint.end_with?('/') ? "#{endpoint}#{TRACES_PATH}" : "#{endpoint}/#{TRACES_PATH}")
```
plantfansam, line 61: What do you think about using `File.join` here? I think that would eliminate the ternary and the string interpolation

```ruby
irb(main):001:0> File.join("http://localhost:4318", "v1/traces")
=> "http://localhost:4318/v1/traces"
irb(main):002:0> File.join("http://localhost:4318/", "v1/traces")
=> "http://localhost:4318/v1/traces"
```

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -56,7 +58,7 @@ def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPOR
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
           @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
-                   URI("#{endpoint}/v1/traces")
+                   URI(File.join(endpoint, TRACES_PATH))
```
fbogsany, line 61: We can just use `URI.join` here:
```
irb(main):002:0> uri = URI.join('https://localhost:1234/', 'v1/traces')
=> #<URI::HTTPS https://localhost:1234/v1/traces>
irb(main):003:0> uri = URI.join('https://localhost:1234', 'v1/traces')
=> #<URI::HTTPS https://localhost:1234/v1/traces>
```
```suggestion
                   URI.join(endpoint, 'v1/traces')
```

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -35,6 +35,8 @@ class Exporter # rubocop:disable Metrics/ClassLength
         ERROR_MESSAGE_INVALID_HEADERS = 'headers must be a String with comma-separated URL Encoded UTF-8 k=v pairs or a Hash'
         private_constant(:ERROR_MESSAGE_INVALID_HEADERS)
 
+        TRACES_PATH = 'v1/traces'
```
fbogsany, line 38: There's no real value in this constant.
```suggestion
```

fbogsany (approved):

fbogsany:
🤦 this modified the old, deprecated exporter package instead of the [new `otlp-http` package](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/exporter/otlp-http). We'll have to port it over, and make sure we're doing the right thing for metrics as well (and maybe gRPC).

robertlaurin:
> 🤦 this modified the old, deprecated exporter package instead of the [new `otlp-http` package](https://github.com/open-telemetry/opentelemetry-ruby/tree/main/exporter/otlp-http). We'll have to port it over, and make sure we're doing the right thing for metrics as well (and maybe gRPC).

This is ok for now! I have not actually deprecated this gem, and the new package isn't published. There's a few changes to port over already, so one more doesn't hurt at this point :)

PR #1346, opened 2022-07-22
Metrics SDK Synchronous aggregation

robertlaurin:
In scope of PR:
- Sum aggregation
- Explicit Bucket Histogram aggregation

- Added aggregation classes for sum and explicit buckets
- Added structs for data Number Data Points, and Histogram Data Points
image


metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+    module Metrics
+      module Aggregation
+        # Contains the implementation of the ExplicitBucketHistogram aggregation
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#explicit-bucket-histogram-aggregation
+        class ExplicitBucketHistogram
+          # The Default Value represents the following buckets:
+          # (-inf, 0], (0, 5.0], (5.0, 10.0], (10.0, 25.0], (25.0, 50.0],
+          # (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0],
+          # (500.0, 1000.0], (1000.0, +inf)
+          DEFAULT_BOUNDARIES = [0, 5, 10, 25, 50, 75, 100, 250, 500, 1000].freeze
```
fbogsany, line 18: This should be a private constant. The documentation for it belongs to `initialize` or a class comment.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+              hdp.start_time_unix_nano = start_time
+              hdp.time_unix_nano = end_time
+              hdp
+            end
+            @data_points.clear if @aggregation_temporality == :delta
+            hdps
+          end
+
+          def update(amount, attributes)
+            hdp = @data_points[attributes] || @data_points[attributes] = HistogramDataPoint.new(
```
fbogsany, line 42: This is difficult to understand. I'd prefer the use of `fetch`. E.g.:
```suggestion
            hdp = @data_points.fetch(attributes) { @data_points[attributes] = HistogramDataPoint.new(
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+          # (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0],
+          # (500.0, 1000.0], (1000.0, +inf)
+          DEFAULT_BOUNDARIES = [0, 5, 10, 25, 50, 75, 100, 250, 500, 1000].freeze
+
+          def initialize(
+            boundaries: DEFAULT_BOUNDARIES,
+            record_min_max: true
+          )
+            @data_points = {}
+            @aggregation_temporality = :delta
```
fbogsany, line 25: We need a way to configure this. I'm still trying to understand in what order things are created and wired up. Ideally, this would be an argument to `initialize`, with a default of `:delta`, and it would be immutable (no `attr_writer`).

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+
+          def collect(start_time, end_time)
+            hdps = @data_points.map do |_key, hdp|
+              hdp.count = hdp.bucket_counts.sum
+              hdp.start_time_unix_nano = start_time
+              hdp.time_unix_nano = end_time
+              hdp
+            end
+            @data_points.clear if @aggregation_temporality == :delta
+            hdps
```
fbogsany, line 38: Since we're mutating the data points here, it feels more honest and clearer to use `each_value`:
```suggestion
            @data_points.each_value do |hdp|
              hdp.count = hdp.bucket_counts.sum
              hdp.start_time_unix_nano = start_time
              hdp.time_unix_nano = end_time
            end
            hdps = @data_points.values
            @data_points.clear if @aggregation_temporality == :delta
            hdps
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+          # (-inf, 0], (0, 5.0], (5.0, 10.0], (10.0, 25.0], (25.0, 50.0],
+          # (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0],
+          # (500.0, 1000.0], (1000.0, +inf)
+          DEFAULT_BOUNDARIES = [0, 5, 10, 25, 50, 75, 100, 250, 500, 1000].freeze
+
+          def initialize(
+            boundaries: DEFAULT_BOUNDARIES,
+            record_min_max: true
+          )
+            @data_points = {}
```
fbogsany, line 24: I'm concerned about mutability of the data points. (Where) do we protect against concurrent calls to `update` and `collect`? `collect` returns an array of mutable `HistogramDataPoint`, which we could reasonably document as immutable for callers, but in the `@aggregation_temporality != :delta` case, we subsequently and concurrently mutate them in `update`.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+            @aggregation_temporality = :delta
+            @boundaries = boundaries
```
fbogsany, line 26: I *think* we should verify that `boundaries` is sorted.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+              -Float::INFINITY     # :max
+            )
+
+            if @record_min_max
+              hdp.max = amount if amount > hdp.max
+              hdp.min = amount if amount < hdp.min
+            end
+
+            hdp.sum += amount
+            hdp.bucket_counts[bisect_left(@boundaries, amount)] += 1
```
fbogsany, line 62: Does this work instead?
```suggestion
            bucket_index = @boundaries.bsearch_index { _1 >= amount } || @boundaries.size
            hdp.bucket_counts[bucket_index] += 1
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
ahayworth: There is an implicit assumption that boundaries are sorted, so yes - that should be done.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -0,0 +1,48 @@
...
+    module Metrics
+      module Aggregation
+        # Contains the implementation of the Sum aggregation
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#sum-aggregation
+        class Sum
+          def initialize
+            @aggregation_temporality = :delta
+
+            @data_points = {}
+            @mutex = Mutex.new
```
ahayworth, line 18: I didn't see where this was being used, but maybe I'm missing something. I mean, I think the idea of a mutex around updates might be desirable (see @fbogsany's comments on the histogram aggregation), don't get me wrong. 😄

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -0,0 +1,48 @@
...
+
+          def update(increment, attributes)
+            ndp = @data_points[attributes] || @data_points[attributes] = NumberDataPoint.new(
+              attributes,
+              nil,
+              nil,
+              0,
+              nil,
+              nil
+            )
```
ahayworth, line 39: How do we feel about keyword init for structs?

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -43,21 +44,15 @@ def collect(start_time, end_time)
...
-          def update(measurement, aggregation)
-            @mutex.synchronize do
-              @data_points[measurement.attributes] = if @data_points[measurement.attributes]
-                                                       aggregation.call(@data_points[measurement.attributes], measurement.value)
-                                                     else
-                                                       measurement.value
-                                                     end
-            end
+          def update(value, attributes)
+            @mutex.synchronize { @aggregation.update(value, attributes) }
```
ahayworth, line 55: Ah - is this what protects multiple callers from colliding when we update during aggregation?

ahayworth (commented):
No huge feedback from me, other than things @fbogsany already said (which serves me right, not bothering to review this in a timely manner 😆). Overall it looks fairly good, seems to do what it says on the tin and all that! 😄

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+              nil,                 # :start_time_unix_nano
+              nil,                 # :time_unix_nano
+              0,                   # :count
+              0,                   # :sum
+              empty_bucket_counts, # :bucket_counts
+              @boundaries,         # :explicit_bounds
+              nil,                 # :exemplars
+              nil,                 # flags
+              Float::INFINITY,     # :min
+              -Float::INFINITY     # :max
```
fbogsany, line 53: I know it makes things cumbersome, but should `min` and `max` be `nil` if `@record_min_max` is `false`?

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,90 @@
...
+          end
+
+          def update(amount, attributes)
+            hdp = @data_points[attributes] || @data_points[attributes] = HistogramDataPoint.new(
+              attributes,
+              nil,                 # :start_time_unix_nano
+              nil,                 # :time_unix_nano
+              0,                   # :count
+              0,                   # :sum
+              empty_bucket_counts, # :bucket_counts
```
fbogsany, line 48: `bucket_counts` is documented as optional. Under what conditions would this not exist? If no boundaries are defined (not even the defaults)? How should we model that?

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
```diff
@@ -0,0 +1,25 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Aggregation
+        HistogramDataPoint = Struct.new(:attributes,            # optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}
+                                        :start_time_unix_nano,  # optional Integer nanoseconds since Epoch
```
fbogsany, line 12: Is `start_time_unix_nano` really optional? How would we end up with it missing?

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
```diff
@@ -0,0 +1,25 @@
...
+                                        :time_unix_nano,        # Integer nanoseconds since Epoch
+                                        :count,                 # Integer count is the number of values in the population. Must be non-negative.
+                                        :sum,                   # Integer sum of the values in the population. If count is zero then this field then this field must be zero
+                                        :bucket_counts,         # optional Array[Integer] field contains the count values of histogram for each bucket.
+                                        :explicit_bounds,       # Array[Float] specifies buckets with explicitly defined bounds for values.
+                                        :exemplars,             # optional List of exemplars collected from measurements that were used to form the data point
+                                        :flags,                 # Integer Flags that apply to this specific data point.
```
fbogsany, line 19: What are the flags and how are they set?

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
fbogsany: 🤷 we didn't use it for `SpanData`: https://github.com/open-telemetry/opentelemetry-ruby/blob/6fed21de005917f1900a8b8dc879f89623e4c61a/sdk/lib/opentelemetry/sdk/trace/span.rb#L255-L276

I do like keyword arguments in general. I'm ambivalent about them for internal things like this that are called only in one place. I would expect our tests to quickly catch a bug here. 

If folks have strong feelings about this, we should change `to_span_data` as well, for consistency.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
```diff
@@ -33,6 +31,12 @@ def record(amount, attributes: nil)
             OpenTelemetry.handle_error(exception: e)
```
fbogsany, line 31: I can't comment on the line, but I think the call to `update` should change (the signature and other callers have changed).

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/up_down_counter.rb
```diff
@@ -36,6 +34,12 @@ def add(amount, attributes: nil)
             OpenTelemetry.handle_error(exception: e)
```
fbogsany, line 34: Same thing here - the call to `update` should change.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
fbogsany: 👍 it protects against concurrent reads and writes during `update` and `collect`. It does not protect against writes to metric data points while (metric) readers, such as exporters, are reading from the data points returned by `collect`. This happens when the aggregation temporality is not `:delta`. I think in that case, we should return a clone of each data point. The depth of cloning required will be interesting (e.g. we need to clone the buckets of a histogram data point, but not the boundaries or attributes).

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -116,6 +116,15 @@ def add_metric_reader(metric_reader)
           end
         end
 
+        # @api private
+        def register_synchronous_instrument(instrument)
+          @mutex.synchronize do
+            @metric_readers.each do |mr|
+              instrument.register_with_new_metric_store(mr.metric_store)
```
fbogsany, line 123: How do we envisage setting a non-default aggregation?

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
robertlaurin: It is really option in the OTLP protos

```
  // StartTimeUnixNano is optional but strongly encouraged, see the
  // the detailed comments above Metric.
  //
  // Value is UNIX Epoch time in nanoseconds since 00:00:00 UTC on 1 January
  // 1970.
  fixed64 start_time_unix_nano = 2;
```
https://github.com/open-telemetry/opentelemetry-proto/blob/v0.19.0/opentelemetry/proto/metrics/v1/metrics.proto#L391-L396

```
// # TimeUnixNano
//
// This field is required, having consistent interpretation across
// DataPoint types.  TimeUnixNano is the moment corresponding to when
// the data point's aggregate value was captured.
//
// Data points with the 0 value for TimeUnixNano SHOULD be rejected
// by consumers.
//
// # StartTimeUnixNano
//
// StartTimeUnixNano in general allows detecting when a sequence of
// observations is unbroken.  This field indicates to consumers the
// start time for points with cumulative and delta
// AggregationTemporality, and it should be included whenever possible
// to support correct rate calculation.  Although it may be omitted
// when the start time is truly unknown, setting StartTimeUnixNano is
// strongly encouraged.
```

https://github.com/open-telemetry/opentelemetry-proto/blob/v0.19.0/opentelemetry/proto/metrics/v1/metrics.proto#L144-L161

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
robertlaurin: I just mirrored the OTLP definitions when creating this structs, so they're not being used or set by any code within the project right now.

https://github.com/open-telemetry/opentelemetry-proto/blob/v0.19.0/opentelemetry/proto/metrics/v1/metrics.proto#L314-L330

```
// DataPointFlags is defined as a protobuf 'uint32' type and is to be used as a
// bit-field representing 32 distinct boolean flags.  Each flag defined in this
// enum is a bit-mask.  To test the presence of a single flag in the flags of
// a data point, for example, use an expression like:
//
//   (point.flags & FLAG_NO_RECORDED_VALUE) == FLAG_NO_RECORDED_VALUE
//
enum DataPointFlags {
  FLAG_NONE = 0;

  // This DataPoint is valid but has no recorded value.  This value
  // SHOULD be used to reflect explicitly missing data in a series, as
  // for an equivalent to the Prometheus "staleness marker".
  FLAG_NO_RECORDED_VALUE = 1;

  // Bits 2-31 are reserved for future use.
}
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
robertlaurin: Good catch, this is not being used. It is remnants of a previous iteration.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
ahayworth: I don't have super strong feelings, other than "it's hard to remember what the positional args mean". I don't care enough to block over it though, just thought I'd take the temperature of the room. :)

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
robertlaurin: updated c9bcc465a7bc78f9e1e2c95b65e21db6f449ffbd

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/up_down_counter.rb
robertlaurin: updated 8b149a742fa97eca626ed3a06967d96a7ab93408

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
robertlaurin: Updated 0dee2a29410ecee88b6c3569e9a2c65c63078e49

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
fbogsany: Do we anticipate producing the "no recorded value" flag? If not, I'd remove the `:flags` for now.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
fbogsany: For the most part, we've had 2 callers (the `to_span_data` method and a test helper, although the latter has been copied a bunch of times). I think that's true for metrics as well. With only 2 callers, positional args don't seem that onerous. I'm not averse to keyword args, though, unless there's runtime overhead associated with them.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
robertlaurin: Removed e42584135ba9e15a9bafeb18a58a2f266f6e71f1

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
robertlaurin: Updated bc133ca6315a65585e67d9d3c2168a7a5edbb6ba

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
robertlaurin: Updated 1fa04e5afd729b0081c66d6f1bf8a4284c2b8920

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
robertlaurin: Removed d3e5eea855b9a60a7f0233374fdb97685d5849a9

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -0,0 +1,46 @@
...
+
+            @data_points = {}
+          end
+
+          def collect(start_time, end_time)
+            ndps = @data_points.map do |_key, ndp|
+              ndp.start_time_unix_nano = start_time
+              ndp.time_unix_nano = end_time
+              ndp
+            end
```
fbogsany, line 25:
```suggestion
            @data_points.each_value do |ndp|
              ndp.start_time_unix_nano = start_time
              ndp.time_unix_nano = end_time
            end
            ndps = @data_points.values
```

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
robertlaurin: There are a few ways aggregation could be configured, some examples via view here.

>[Note: Implementors MAY choose the best idiomatic practice for their language to represent the semantic of an Aggregation and optional configuration parameters.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#aggregation)

```
# Use Histogram with custom boundaries
meter_provider
  .add_view(
    "X",
    aggregation="ExplicitBucketHistogram",
    aggregation_params={"Boundaries": [0, 10, 100]}
    )
```

```
// Use Histogram with custom boundaries
meterProviderBuilder
  .AddView(
    instrumentName: "X",
    aggregation: new ExplicitBucketHistogramAggregation(
      boundaries: new double[] { 0.0, 10.0, 100.0 }
    )
  );
```

I think in the trace implementation we have more examples of the latter where we allow consumers to instantiate a configurable something and pass it into the respective attribute writer. 

Something not totally clear to me is whether the following line could be interpreted to mean that the metric reader can also configure the aggregation. That doesn't seem to follow prior wording in the SDK spec. 

>[The default output aggregation (optional), a function of instrument kind. If not configured, the [default aggregation](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#default-aggregation) SHOULD be used.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#metricreader)

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
robertlaurin: I think duping the datapoints on collection is probably the right approach here. cbb93bbac95c345ee0be564542bb8fc40cef7e17

Do you think it makes sense to only dup **if** it's not using delta temporality (that way we get more of a performance win from using delta).

fbogsany: > Do you think it makes sense to only dup **if** it's not using delta temporality (that way we get more of a performance win from using delta).

Yep.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,82 @@
...
+            @record_min_max = record_min_max
+          end
+
+          def collect(start_time, end_time)
+            @data_points.each_value do |hdp|
+              hdp.count = hdp.bucket_counts.sum
+              hdp.start_time_unix_nano = start_time
+              hdp.time_unix_nano = end_time
+            end
+            hdps = @data_points.values(&:dup)
```
fbogsany, line 37: We need to `dup` the `bucket_counts` as well.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/number_data_point.rb
```diff
@@ -0,0 +1,19 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Aggregation
+        NumberDataPoint = Struct.new(:attributes,            # Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}
+                                     :start_time_unix_nano,  # optional Integer nanoseconds since Epoch
```
fbogsany, line 12:
```suggestion
                                     :start_time_unix_nano,  # Integer nanoseconds since Epoch
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/histogram_data_point.rb
fbogsany:
```suggestion
                                        :start_time_unix_nano,  # Integer nanoseconds since Epoch
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,82 @@
...
+        class ExplicitBucketHistogram
+          DEFAULT_BOUNDARIES = [0, 5, 10, 25, 50, 75, 100, 250, 500, 1000].freeze
+          private_constant :DEFAULT_BOUNDARIES
+
+          # The default value for boundaries represents the following buckets:
+          # (-inf, 0], (0, 5.0], (5.0, 10.0], (10.0, 25.0], (25.0, 50.0],
+          # (50.0, 75.0], (75.0, 100.0], (100.0, 250.0], (250.0, 500.0],
+          # (500.0, 1000.0], (1000.0, +inf)
+          def initialize(
+            boundaries: DEFAULT_BOUNDARIES,
```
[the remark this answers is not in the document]
fbogsany: Relevant text:
> Bucket counts are optional. A Histogram without buckets conveys a population in terms of only the sum and count, and may be interpreted as a histogram with single bucket covering `(-Inf, +Inf)`.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,82 @@
...
+
+          def collect(start_time, end_time)
+            @data_points.each_value do |hdp|
+              hdp.count = hdp.bucket_counts.sum
+              hdp.start_time_unix_nano = start_time
+              hdp.time_unix_nano = end_time
+            end
+            hdps = @data_points.values(&:dup)
+            @data_points.clear if @aggregation_temporality == :delta
+            hdps
```
fbogsany, line 39:
```suggestion
            hdps = @data_points.values
            if @aggregation_temporality == :delta
              @data_points.clear
            else
              hdps = hdps.map do |hdp|
                hdp = hdp.dup
                hdp.bucket_counts = hdp.bucket_counts.dup
                hdp
              end
            end
            hdps
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -0,0 +1,46 @@
...
+            @data_points = {}
+          end
+
+          def collect(start_time, end_time)
+            @data_points.each_value do |ndp|
+              ndp.start_time_unix_nano = start_time
+              ndp.time_unix_nano = end_time
+            end
+            ndps = @data_points.values(&:dup)
+            @data_points.clear if @aggregation_temporality == :delta
```
fbogsany, line 26:
```suggestion
            ndps = @data_points.values
            if @aggregation_temporality == :delta
              @data_points.clear
            else
              ndps = ndps.map(&:dup)
            end
```

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
fbogsany: Discussed IRL: aggregations can be configured either in the `MetricReader` or in `View`s. We'll be tackling those in follow-up PRs, in that order, so we'll handle the wiring at that time.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
fbogsany: Will be resolved with suggestions elsewhere.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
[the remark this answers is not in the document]
robertlaurin: Added a commit to handle the nil case 2cc2761914a7475c5b4787e55ab0c3cb5de28eb8
Modified the empty array to match the nil case 11d42c1057eb16336a0832f0b2bff322db5863ea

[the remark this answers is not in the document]
robertlaurin: Should be keeping track of the total `count` value when we're not keeping track of the `bucket_counts`.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -0,0 +1,50 @@
...
+              ndp.start_time_unix_nano = start_time
+              ndp.time_unix_nano = end_time
+            end
+            ndps = @data_points.values
+            if @aggregation_temporality == :delta
+              @data_points.clear
+            else
+              ndps = ndps.map(&:dup)
+            end
+            ndps
```
fbogsany, line 31: The logic here is tricky. I think we need to distinguish `@aggregation_temporality == :delta` at the top-level and handle the two cases separately. Otherwise there'll be a bunch of messy conditionals. Here's my attempt:
```suggestion
            if @aggregation_temporality == :delta
              ndps = @data_points.map do |_, ndp|
                ndp.start_time_unix_nano = start_time
                ndp.time_unix_nano = end_time
                ndp
              end
              @data_points.clear
              ndps
            else
              @data_points.map do |_, ndp|
                ndp.start_time_unix_nano ||= start_time
                ndp.time_unix_nano = end_time
                ndp.dup
              end
            end
```
The main difference is that we don't need to `dup` the `ndp` in the `delta` case because we `clear` the hash, and in the non-`delta` case we need to only set the `start_time` once for each `ndp` *and then* `dup`. Unless I misunderstand aggregation temporality, non-`delta` data points should have a `start_time` of when they were first observed.

fbogsany: Alternative:
```suggestion
            if @aggregation_temporality == :delta
              # Set timestamps and 'move' data point values to result.
              ndps = @data_points.values.each do |ndp|
                ndp.start_time_unix_nano = start_time
                ndp.time_unix_nano = end_time
              end
              @data_points.clear
              ndps
            else
              # Update timestamps and take a snapshot.
              @data_points.values.map! do |ndp|
                ndp.start_time_unix_nano ||= start_time # Start time of a data point is from the first observation.
                ndp.time_unix_nano = end_time
                ndp.dup
              end
            end
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,93 @@
...
+                max                  # :max
+              )
+            end
+
+            if @record_min_max
+              hdp.max = amount if amount > hdp.max
+              hdp.min = amount if amount < hdp.min
+            end
+
+            hdp.sum += amount
```
fbogsany, line 76:
```suggestion
            hdp.sum += amount
            hdp.count += 1
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -0,0 +1,93 @@
...
+            if @aggregation_temporality == :delta
+              @data_points.clear
+            else
+              hdps = hdps.map do |hdp|
+                hdp = hdp.dup
+                hdp.bucket_counts = hdp.bucket_counts.dup
+                hdp
+              end
+            end
+            hdps
```
fbogsany, line 47: This has the same issue with start time in non-`delta` as the `sum` aggregation. I think the same approach would be good here, distinguishing the two cases up front.

```suggestion
            if @aggregation_temporality == :delta
              # Set timestamps and 'move' data point values to result.
              hdps = @data_points.values.each do |hdp|
                hdp.start_time_unix_nano = start_time
                hdp.time_unix_nano = end_time
              end
              @data_points.clear
              hdps
            else
              # Update timestamps and take a snapshot.
              @data_points.values.map! do |hdp|
                hdp.start_time_unix_nano ||= start_time # Start time of a data point is from the first observation.
                hdp.time_unix_nano = end_time
                hdp = hdp.dup
                hdp.bucket_counts = hdp.bucket_counts.dup
                hdp
              end
            end
```

Note that this suggestion assumes we increment `hdp.count` in `update`, as suggested below.

Sidenote: there's an alternative approach where we extract the 4 cases into separate classes: `delta` or not, and buckets or not. Probably not worth doing that yet, but it's something to consider for later.

metrics_sdk/test/integration/in_memory_metric_pull_exporter_test.rb
```diff
@@ -17,28 +17,35 @@
...
+      _(last_snapshot[0].data_points[0].attributes).must_equal({})
+
+      _(last_snapshot[0].data_points[1].value).must_equal(4)
+      _(last_snapshot[0].data_points[1].attributes).must_equal('a' => 'b')
+
+      _(last_snapshot[0].data_points[2].value).must_equal(3)
+      _(last_snapshot[0].data_points[2].attributes).must_equal('b' => 'c')
+
+      _(last_snapshot[0].data_points[3].value).must_equal(4)
+      _(last_snapshot[0].data_points[3].attributes).must_equal('d' => 'e')
```
fbogsany, line 48: The reliance on ordering here feels brittle. Not something we have to fix just now, but I'd like to leave open the possibility that the ordering could change. I'm more worried about someone 👀 this test and assuming the ordering is guaranteed.

metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram_test.rb
```diff
@@ -0,0 +1,262 @@
...
+      _(hdps[1].bucket_counts).must_equal([1, 1, 0, 1, 0, 1, 1, 0, 0, 0, 0])
+      _(hdps[1].explicit_bounds).must_equal([0, 5, 10, 25, 50, 75, 100, 250, 500, 1000])
+    end
+
+    it 'sets the timestamps' do
+      ebh.update(0, {})
+      hdp = ebh.collect(now_in_nano, now_in_nano)[0]
+      _(hdp.start_time_unix_nano).must_equal(now_in_nano)
+      _(hdp.time_unix_nano).must_equal(now_in_nano)
+    end
```
fbogsany, line 60: It'd be good to have different start time and time stamps here.

metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram_test.rb
```diff
@@ -0,0 +1,262 @@
...
+
+    it 'calculates the bucket count' do
```
fbogsany, line 62:
```suggestion
    it 'calculates the count' do
```

metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram_test.rb
```diff
@@ -0,0 +1,262 @@
...
+      ebh.update(0, {})
+      ebh.update(0, {})
+      ebh.update(0, {})
+      ebh.update(0, {})
+      hdp = ebh.collect(now_in_nano, now_in_nano)[0]
+      _(hdp.count).must_equal(4)
+    end
+
+    it 'does not aggregate between collects' do
```
fbogsany, line 71:
```suggestion
    it 'does not aggregate between collects with default delta aggregation' do
```

metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram_test.rb
```diff
@@ -0,0 +1,262 @@
...
+
+        hdps = ebh.collect(now_in_nano, now_in_nano)
+        # Assert that we are accumulating values
+        # and not just capturing the delta since
+        # the previous collect call
+        _(hdps[0].count).must_equal(10)
+        _(hdps[0].sum).must_equal(44)
+        _(hdps[0].min).must_equal(0)
+        _(hdps[0].max).must_equal(10)
+        _(hdps[0].bucket_counts).must_equal([2, 4, 4, 0, 0, 0, 0, 0, 0, 0, 0])
```
fbogsany, line 134: I'd like us to also assert that the previously collected metrics were not modified by the most recent `collect` call.
```suggestion
        hdps1 = ebh.collect(now_in_nano, now_in_nano)
        # Assert that we are accumulating values
        # and not just capturing the delta since
        # the previous collect call
        _(hdps1[0].count).must_equal(10)
        _(hdps1[0].sum).must_equal(44)
        _(hdps1[0].min).must_equal(0)
        _(hdps1[0].max).must_equal(10)
        _(hdps1[0].bucket_counts).must_equal([2, 4, 4, 0, 0, 0, 0, 0, 0, 0, 0])

        # Assert that the recent collect does not
        # impact the already collected metrics
        _(hdps[0].count).must_equal(5)
        _(hdps[0].sum).must_equal(22)
        _(hdps[0].min).must_equal(0)
        _(hdps[0].max).must_equal(10)
        _(hdps[0].bucket_counts).must_equal([1, 2, 2, 0, 0, 0, 0, 0, 0, 0, 0])
```

metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram_test.rb
robertlaurin: Updated 422aa9dd64c2931579540162a9c97aa181e1b9c6

ahayworth (approved):
👍 I have no major concerns with the PR at this point (of course, it may be that I just don't know enough to see problems with it, but 🤷 ).

LGTM!

fbogsany (approved):

PR #1349, opened 2022-07-26
refactor: code-style refactor on Exporter::OTLP

shaul75:


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -36,43 +36,42 @@ class Exporter # rubocop:disable Metrics/ClassLength
...
+            certificate_file: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE', 'OTEL_EXPORTER_OTLP_CERTIFICATE'),
+            compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: DEFAULT_COMPRESSION),
+            headers: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS', default: DEFAULT_HEADERS),
+            timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: DEFAULT_TIMEOUT),
+            ssl_verify_mode: ssl_verify_mode,
+            metrics_reporter: nil
+          }
+        end
+
+        def initialize(opts = {})
```
robertlaurin, line 62: Speaking from personal experience with a variety of ruby codebases, option hashes can become quite unwieldily over time. I would not support the introduction of them in this codebase.

robertlaurin (commented):
Hey, so first I want to say thanks for taking the time to open this PR.

I think code style oriented refactor PRs are particularly "interesting" (read challenging) because there is so much room for subjectivity.

So at a high level I would expect some clarification in the PR description regarding what this refactor is meant to accomplish. Or more simply put, what problem is being solved by these changes.

I think there is a pattern of breaking apart the methods in an attempt to reduce complexity, however it is just shuffling the complexity around into smaller methods. The end result is that the burden is put on the reader to jump to those smaller single use methods and stitch the context back together inline.

While I appreciate some of the aesthetic appeal to these changes I'm not convinced this is an improvement to the code.

I won't close this PR because I'm happy to have you or some the other maintainers/approvers change my view point here.

cc: @open-telemetry/ruby-approvers @open-telemetry/ruby-maintainers

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -111,7 +110,19 @@ def shutdown(timeout: nil)
...
 
-        def http_connection(uri, ssl_verify_mode, certificate_file)
+        def validate!(args)
+          raise ArgumentError, "invalid url for OTLP::Exporter #{args[:endpoint]}" unless OpenTelemetry::Common::Utilities.valid_url?(args[:endpoint])
+          raise ArgumentError, "unsupported compression key #{args[:compression]}" unless args[:compression].nil? || %w[gzip none].include?(args[:compression])
+          raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS unless [String, Hash].include?(args[:headers].class)
+        end
+
+        def uri
+          @uri ||= URI.join(@endpoint, 'v1/traces') if @endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
```
shaul75, line 120: I can't make sense of this condition. Maintainers what do you think?

```ruby
[4] pry(main)> ENV['OTEL_EXPORTER_OTLP_ENDPOINT']='http://localhost:4318/v1/traces'
=> "http://localhost:4318/v1/traces"
[5] pry(main)> OpenTelemetry::Exporter::OTLP::Exporter.new(endpoint: ENV['OTEL_EXPORTER_OTLP_ENDPOINT'])        
=> #<OpenTelemetry::Exporter::OTLP::Exporter:0x000000011a16d328
...
 @path="/v1/v1/traces", # <---- this is obviously a bug

```

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
shaul75: I agree with you on that, what bothers me is the poor readability though. how about using named parameters as before, but defining the default values in a private method,
Something like this?
```ruby
        def initialize(endpoint: config[:endpoint], ...)
        ... 
        end

        private

        def config
        { endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: DEFAULT_ENDPOINT), ... }
        end
```

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -36,43 +36,42 @@ class Exporter # rubocop:disable Metrics/ClassLength
...
-            OpenSSL::SSL::VERIFY_PEER
-          elsif ENV.key?('OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_NONE')
-            OpenSSL::SSL::VERIFY_NONE
-          else
-            OpenSSL::SSL::VERIFY_PEER
-          end
+          return OpenSSL::SSL::VERIFY_PEER if ENV.key?('OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_PEER')
+          return OpenSSL::SSL::VERIFY_NONE if ENV.key?('OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_NONE')
+
+          OpenSSL::SSL::VERIFY_PEER
```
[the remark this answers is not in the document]
shaul75: It looks cleaner, easier to read. Might be subjective :)
regardless of that, I opened a discussion about this method. https://github.com/open-telemetry/opentelemetry-ruby/discussions/1350

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -36,43 +36,42 @@ class Exporter # rubocop:disable Metrics/ClassLength
...
-                     else
-                       raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS
-                     end
-          @timeout = timeout.to_f
-          @compression = compression
-          @metrics_reporter = metrics_reporter || OpenTelemetry::SDK::Trace::Export::MetricsReporter
+        DEFAULT_ENDPOINT = 'http://localhost:4318/v1/traces'
+        DEFAULT_COMPRESSION = 'gzip'
+        DEFAULT_HEADERS = {}.freeze
+        DEFAULT_TIMEOUT = 10
```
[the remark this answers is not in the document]
shaul75: Agree. I'll drop it.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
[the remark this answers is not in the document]
ahayworth: I like this style when it’s short, but this hits around 80 characters. When it gets this long, I find it’s too much to hold in my head before I know if I can proceed to the next line or if I must jump back up the call stack. The previous version of this code let me know if I could progress through the file more quickly as I read, and thus feels nicer (for the same reasons people like to feel as though they are “making progress” on a trip via the scenic route, even if sitting in slow traffic is technically faster because it’s more direct).

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
ahayworth: Apologies, are you asking if you don’t understand the conditional in the code you wrote, or as it was prior to this change?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
ahayworth: It was just as readable to me before the refactor as it is after.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -390,6 +401,8 @@ def as_otlp_any_value(value)
         end
 
         def parse_headers(raw)
+          return raw if raw.is_a?(Hash)
```
ahayworth, line 404: Moving this logic so far from where it was used obscures it a little.

ahayworth (changes requested):
Thank you for showing interest in the project! I appreciate the time that you took to offer these refactors, but at this time I would personally decline to accept them.

What is it isn’t readable is a subjective thing, and to me the styles adopted here are less readable than what came before (even though I don’t personally care for the current style either). It’s largely a matter of taste, and so the change is just as likely to make it better for some and worse for others.

One unwritten principle I’ve seen applied throughout the current codebase is that it is often written to be easier to understand in terms of “how much state must I hold in my head at once”. I personally find this even more valuable than aesthetic style, because this is a large codebase already and I’m not intimately familiar with all parts. Because of this (somewhat inconsistently applied) preference for “clarity of state,” I feel more confident working in the codebase and that my assumptions of how it works hold true.

But again, even that is subjective. Others might think that’s nonsense and that I am speaking gibberish.

If you choose to continue to contribute (and I really hope you do, because we want help!), working on the style of the codebase is something I would hope you discuss with all of us over time. We haven’t had a lot of discussion on it overall, and coming to formal consensus about what we want this code to look like (and why) is a valuable thing for a community of maintainers. It’s a good thing to discuss at the SIG meetings, so we can more easily talk about the non-technical aspects of the styles we prefer.

However - I really appreciate the effort this took on your part. I would encourage you to pick up another task in the project first, and that would be a better way to start collaborating! There are issues marked as “good first issue” (and some bugs!) and one of those might a really good candidate to work on. ❤️

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
shaul75: I didn't change the condition, I extracted it from the initializer.

```ruby
                 else
                   URI(endpoint)
                 end
```

shaul75: what I don't understand is the idea behind it, why would both `endpoint` and `ENV['OTEL_EXPORTER_OTLP_ENDPOINT']` have to be set to the same value for the URI to have `v1/traces` appended to it?

shaul75:
@ahayworth 
No worries, I should've known a subjective thing like a code-style change is not a good first contribution, especially when it's on a small part of an already fairly large project.
Appreciate your review. I will stick around and contribute with perhaps more practical things. 😄

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
ahayworth: Ah, gotcha. So this is a little subtle - OpenTelemetry defines various environment variables that SDKs must respect, and the ones we're dealing with here are for [the OTLP exporter, specifically](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#configuration-options). Here's the entire blob about endpoint selection:

> Endpoint (OTLP/HTTP): Target URL to which the exporter is going to send spans or metrics. The endpoint MUST be a valid URL with scheme (http or https) and host, MAY contain a port, SHOULD contain a path and MUST NOT contain other parts (such as query string or fragment). A scheme of https indicates a secure connection. When using OTEL_EXPORTER_OTLP_ENDPOINT, exporters MUST construct per-signal URLs as [described below](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp). The per-signal endpoint configuration options take precedence and can be used to override this behavior (the URL is used as-is for them, without any modifications). See the [OTLP Specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/otlp.md#otlphttp-request) for more details.

> Default: http://localhost:4318 [1]
> Env vars: OTEL_EXPORTER_OTLP_ENDPOINT OTEL_EXPORTER_OTLP_TRACES_ENDPOINT OTEL_EXPORTER_OTLP_METRICS_ENDPOINT

Crucially, our OTLP exporter *only* supports traces right now. So if someone configures `OTEL_EXPORTER_OTLP_ENDPOINT` and gives it `http://my-fancy-exporter.com:4318` as a _generic_ OTLP endpoint, we *must* construct a signal-specific endpoint for traces, which must have the `/v1/traces` bits appended (which is also dictated by the spec).

However, signal-specific endpoints take precedence, so the best way for us to detect this condition is to see if our env-var resolving code (which considers all the relevant env vars) selected the _same_ thing as what is present in `ENV['OTEL_EXPORTER_OTLP_ENDPOINT']`? If so, we assume that the user configured a generic OTLP endpoint and that we must construct the signal-specific endpoint.

ahayworth:
>  I will stick around and contribute with perhaps more practical things

@shaul75 Wonderful! We're happy to have you. 😄 😄

PR #1366, opened 2022-08-24
Rename internal_create_span

fbogsany:
This PR renames the API-internal `TracerProvider#internal_create_span` method to `TracerProvider#internal_start_span`. It is an attempt to address concerns around the change in method signature in #1281. That PR changed the signature of `internal_create_span` from:
```ruby
def internal_create_span(name, kind, trace_id, parent_span_id, attributes, links, start_timestamp, parent_context, parent_span, instrumentation_library)
```
to
```ruby
def internal_create_span(name, kind, attributes, links, start_timestamp, parent_context, instrumentation_scope)
```
and also moved some functionality between it and its caller.

While this method is marked API-internal, "someone" may have chosen to monkey-patch it to change its behaviour, and if that "someone" didn't pin the SDK version, then consumers bumping the SDK dependency may 💥.

### TODO: describe possible impact

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/tracer.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/tracer.rb
@@ -32,7 +32,7 @@ module OpenTelemetry
           kind ||= :internal
           with_parent ||= Context.current
 
-          @tracer_provider.internal_create_span(name, kind, attributes, links, start_timestamp, with_parent, @instrumentation_scope)
+          @tracer_provider.internal_start_span(name, kind, attributes, links, start_timestamp, with_parent, @instrumentation_scope)
         end
       end
     end
--- a/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
@@ -126,7 +126,7 @@ module OpenTelemetry
         end
 
         # @api private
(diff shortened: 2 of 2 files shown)


ahayworth (approved):
I don't think there is any real fall-out from this change; it really only affects those who might be monkey-patching private internal SDK methods. Whomever that may be.

arielvalentin (approved):

fbogsany:
The "someone" may have monkey-patched it to re-use the parent span ID if the returned span was non-recording, like (hypothetically):
```ruby
        def internal_create_span(name, kind, trace_id, parent_span_id, attributes, links, start_timestamp, parent_context, parent_span, instrumentation_library)
          span = super
          return span if span.recording?

          OpenTelemetry::Trace.non_recording_span(
            OpenTelemetry::Trace::SpanContext.new(
              trace_id: span.context.trace_id,
              span_id: parent_span_id || OpenTelemetry::Trace::INVALID_SPAN_ID,
              tracestate: span.context.tracestate
            )
          )
        end
```
This would likely 💥 by passing the wrong args to `super`.

Changing the name of the method means this monkey-patch will not be called, and this "someone" will get the standard SDK behaviour. That's presumably not what they want, but at least it won't 💥 (it'll just potentially produce broken traces).

plantfansam (approved):

Issue #1321, opened 2022-06-17
Update to latest rubocop


plantfansam:
Our various gems are on a pretty old version of Rubocop (see @dazuma's comment [here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1096#issuecomment-1018193322) for context). The SDK gem is at [`0.73.0`](https://github.com/open-telemetry/opentelemetry-ruby/blob/e533817ba387bbfd6270e5c4d0ae42452dd7d9dc/sdk/opentelemetry-sdk.gemspec#L46). We should update to the latest rubocop.

Now that instrumentation has been moved to `-contrib`, this should be a lighter lift.

- [x] API
- [ ] Common
- [ ] SDK
- [ ] Exporters
- [ ] Propagators
- [ ] Registry
- [ ] Semantic conventions

richardmcmillen:
@ahayworth thanks for the reviews on those PR's. I take note of your comment about not enabling cops by default. Would you be ok to have new cops enabled if it doesn't result in too many autofixes?

I also note this issue wishes to standardise some more: https://github.com/open-telemetry/opentelemetry-ruby/issues/1334
I am thinking upgrading gems individually first and then consolidating would be a reasonable approach to get there.

Looking at the current configuration there aren't many unique entries throughout the .rubocop.yml files.

```
grep --include='.rubocop.yml' -REho "^[A-Z][a-zA-Z/]+" | grep -v AllCops | sort | uniq -c | sort -r
  27 Metrics/AbcSize
  19 Metrics/LineLength
  19 Metrics/BlockLength
  17 Style/ModuleFunction
  17 Naming/FileName
  17 Metrics/ParameterLists
  17 Metrics/MethodLength
  17 Lint/UnusedMethodArgument
   7 Layout/LineLength
   5 Style/StringLiterals
   5 Style/StringConcatenation
   5 Style/FrozenStringLiteralComment
   5 Lint/MissingSuper
   5 Lint/ConstantDefinitionInBlock
   5 Bundler/OrderedGems
   4 Metrics/PerceivedComplexity
   4 Metrics/CyclomaticComplexity
   3 Style/MethodCallWithoutArgsParentheses
   3 Style/BlockDelimiters
   2 Metrics/ModuleLength
   2 Gemspec/RequireMFA
   1 Style/BracesAroundHashParameters
   1 Layout/TrailingWhitespace
   1 Layout/TrailingEmptyLines
[...]

PR #1367, opened 2022-08-28
Update rubocop on multiple items to 1.3.x

scotthain:
Signed-off-by: Scott Hain <elejia@gmail.com>

Helps add to [opentelemetry-ruby 1321](https://github.com/open-telemetry/opentelemetry-ruby/issues/1321)

### DIFF
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -83,8 +83,8 @@ module OpenTelemetry
         end
       end
 
-      def untraced
-        OpenTelemetry::Trace.with_span(OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new)) { yield }
+      def untraced(&block)
+        OpenTelemetry::Trace.with_span(OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new), &block)
       end
 
       # Returns a URL string with userinfo removed.
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -83,8 +83,11 @@ module OpenTelemetry
         end
       end
 
-      def untraced(&block)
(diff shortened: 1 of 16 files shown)


common/lib/opentelemetry/common/utilities.rb
```diff
@@ -84,7 +84,10 @@ def truncate_attribute_value(value, limit)
       end
 
       def untraced
+        # Unsure how this could be cleanly refactored based on the cop examples.
+        # rubocop:disable Style/ExplicitBlockArgument
         OpenTelemetry::Trace.with_span(OpenTelemetry::Trace.non_recording_span(OpenTelemetry::Trace::SpanContext.new)) { yield }
```
plantfansam, line 89: Can we disable using an end-of-line comment on line 89 instead of wrapping the line? See https://docs.rubocop.org/rubocop/configuration.html#disabling-cops-within-source-code

plantfansam (approved):
One nit, but this LGTM. Thanks so much for the contribution!

common/lib/opentelemetry/common/utilities.rb
scotthain: Sure thing, I had that first and IIRC that made it so I had to also add the _Layout/LineLength_ rule as well 🤣 
I'll double check!

scotthain:
@ahayworth yeah I was noticing that as I went through some of those. At a $Job-n we had some kind of standard for this... maybe @robbkidd remembers.

common/lib/opentelemetry/common/utilities.rb
scotthain: I was wrong! I think I prematurely added the LineLength because I assumed it would need it. Good catch :)

robbkidd:
New gem "otel-style" in the style of "[chefstyle](https://github.com/chef/chefstyle/blob/main/config/upstream.yml)". 

Absolutely … but not really.

robertlaurin (approved):

PR #1392, opened 2022-10-24
chore: Upgrade rubocop 0.73 -> 1.3 for TestHelpers

richardmcmillen:
* enable all new cops by default
* Style/BracesAroundHashParameters was removed - see https://github.com/rubocop/rubocop/issues/7641
* LineLength changed group from Metrics to Layout
* disable Gemspec/RequireMFA
* autocorrect explicit block over yield

Partially addresses: https://github.com/open-telemetry/opentelemetry-ruby/issues/1321

### DIFF
--- a/test_helpers/.rubocop.yml
+++ b/test_helpers/.rubocop.yml
@@ -1,11 +1,12 @@
 AllCops:
   TargetRubyVersion: "2.6.0"
+  NewCops: enable
 
 Lint/UnusedMethodArgument:
   Enabled: false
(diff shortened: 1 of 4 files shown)


ahayworth (approved):
I'm not 100% sure about enabling new cops (we may find ourselves annoyed with that someday), but I don't think that's a blocking concern - we can always turn it back off. 😄 

LGTM, thanks!

robertlaurin (approved):

Issue #1250, opened 2022-05-11
Extract `otel.foo` to constants throughout the codebase


ahayworth:
We have a sprinkling of `otel.foo` kinds of tags being used throughout the codebase:

```
opentelemetry-ruby on  ahayworth/add-dropped-counts-zipkin-jaeger via 💎 v3.1.1
zsh ❯ git grep 'otel\.' | egrep -v '\/test\/'
exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb:          tags << encoded_tag('otel.dropped_attributes_count', dropped_attributes_count) if dropped_attributes_count.positive?
exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb:          tags << encoded_tag('otel.dropped_events_count', dropped_events_count) if dropped_events_count.positive?
exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb:          tags << encoded_tag('otel.dropped_links_count', dropped_links_count) if dropped_links_count.positive?
exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb:          tags << encoded_tag('otel.library.name', instrumentation_library.name) if instrumentation_library.name
exporter/jaeger/lib/opentelemetry/exporter/jaeger/encoder.rb:          tags << encoded_tag('otel.library.version', instrumentation_library.version) if instrumentation_library.version
exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb:                @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response.code })
exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb:              @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => e.class.to_s })
exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb:              @metrics_reporter.record_value('otel.otlp_exporter.request_duration',
exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/exporter.rb:            @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => reason })
exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb:              @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response.code })
[...]

PR #1403, opened 2022-11-14
refactor: uses constants for otel.* attributes

Coolomina:
Takes care of #1250.

I did not find a better place for the Zipkin and Jaeger `otel.*` specific variables in the code. Do you think they should live under a separate file?, I guessed they belonged to the [common otel.* namespace](https://github.com/open-telemetry/opentelemetry-specification/blob/573c1c68f409cd6b5dc40a31f24b250a965425a3/specification/common/attribute-naming.md#otel-namespace)


Coolomina:
`CI / opentelemetry-exporter-otlp-common / windows-latest (pull_request)`  is failing due to a timeout when downlading a package. The endpoint seems to work properly now with `curl https://github.com/ruby/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/msys2.7z -L --output file`.

```
Downloading msys2 build tools
  https://github.com/ruby/setup-msys2-gcc/releases/download/msys2-gcc-pkgs/msys2.7z
  connect ETIMEDOUT 185.199.110.133:443
  Waiting 10 seconds before trying again
  connect ETIMEDOUT 185.199.110.133:443
  Waiting 12 seconds before trying again
  ```

Any of the reviewers to give it a retry maybe ?

File: semantic_conventions/lib/opentelemetry/semantic_conventions/common.rb [hunk 1]
```diff
@@ -0,0 +1,79 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SemanticConventions
+    module Common
+      # A comparison between the actual size of the Span and the BatchSpanProcessor @max_queue_size
+      OTEL_BSP_BUFFER_UTILIZATION = 'otel.bsp.buffer_utilization'
```
ericmustin, line 11: broadly this looks good, thank you for this work. i re-ran the windows job. a question that i'm not sure about is whether a few of these constants, like `buffer_utilization`, should live in the semantic conventions gem or whether it should be in the gem where the metrics reporter is being used, or maybe a different module like `Experimental` within this semconv gem, as `buffer_utilization` is not in the specification. don't think this is blocking tho, and i think this applies to only a few of the attributes in this pr. and i don't really mind if they end up under `Common` but just trying to give this a review, otherwise lgtm

ericmustin (commented):
Not sure whether a few of these constants should live in `Experimental`, `Common`, or some other option, but other than that lgtm, thank you for the work here!

semantic_conventions/lib/opentelemetry/semantic_conventions/common.rb
plantfansam: I'd vote for putting off-spec constants in a separate module within `OpenTelemetry::SemanticConventions`. `OpenTelemetry::SemanticConventions::OffSpec` or whatever, but this is a weak opinion, weakly held.

ericmustin: sorry for the delay, yes i think we're on the same page, the consensus from the tuesday SIG(see https://docs.google.com/document/d/1D15bO8o340sQm2CVZiukEJuCO_XMMHKPuTznoEhyFqE/edit#heading=h.951oi3x8ptux)  was that these off spec constants should live in their gems, rather than common, so this should apply to the following metric families in this pr.

- `otel.jaeger_exporter.*`
- `otel.otlp_exporter.*`
- `otel.zipkin_exporter.*`
- `otel.bsp.*`

ahayworth: Since I missed the SIG (again 😢) I presume that means we won't be taking this PR as-is @ericmustin?

(@Coolomina I appreciate your willingness to contribute and patience with us as we figure out where to put stuff like this. ❤️)

plantfansam: @ahayworth yeah, the SIG and @Coolomina are in agreement on this plan:

> Now that I have a bit more context, and looking at the angles @plantfansam and @ericmustin provided, maybe it makes sense for off-spec attributes to live in their respective libraries.

ahayworth: 👍 Sounds fine with me.

ericmustin: i mean we can take it and then just do a quick followup pr to move the stuff where we want it to be. i'm just a bit low on free cycles rn

Coolomina: Hello! 

No need @ericmustin, I'm preparing a commit with the changes. It's not a problem at all @ahayworth :D, just trying to learn here.

Coolomina:
For `OpenTelemetry::Exporters::*` I created the consensus of having an `OpenTelemetry::Exporter::#{name}::Attributes` because it felt wrong to put the constants in the main required file (ej: `exporter/jaeger/lib/opentelemetry/exporter/jaeger.rb`) and reference them without context in another file. 

Do you think referencing them with `Attributes::ATTR_NAME` makes sense?

For the BatchSpanProcessor I just put them in the class as non-private.

ahayworth:
@Coolomina This looks good to me, but I think there might be two more small changes:
1. I believe you need to update the gemspecs for everything where you added the new dependency on the `opentelemetry-semantic_conventions` gem. The tests pass because we install lots of things locally in CI, but that could break in the actual released versions. (I think it needs to be added, anyways; I could be wrong).
2. In the `./semantic_conventions` directory, can you run `bundle exec rake generate` and make sure your changes aren't overwritten? Most of that gem is actually generated from upstream specifications, so I just want to ensure things don't break. I _think_ it will be okay though.

Coolomina:
@ahayworth Thanks for the patience :D. 

Having added the dependency in `exporter/otlp-http/opentelemetry-exporter-otlp-http.gemspec`, the results do not seem to overwrite anything.

```
opentelemetry-ruby/semantic_conventions [otel-constants●] » bundle exec rake generate
...
...
Using default SPAN type for semantic convention 'webengine_resource' @ line 5

opentelemetry-ruby/semantic_conventions [otel-constants] » echo $?; git status      
0
On branch otel-constants
nothing to commit, working tree clean
```

ericmustin (approved):
i think this lgtm, ty for the work here

exporter/otlp-http/opentelemetry-exporter-otlp-http.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'opentelemetry-common', '~> 0.19.6'
   spec.add_dependency 'opentelemetry-exporter-otlp-common'
   spec.add_dependency 'opentelemetry-sdk', '~> 1.2'
+  spec.add_dependency 'opentelemetry-semantic_conventions'
```
ericmustin, line 32: as @ahayworth pointed out to me, i _think_ this needs to be version constrained ala 

https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/7cc8a21f80eb77519de0612c718d7a2702692604/instrumentation/trilogy/opentelemetry-instrumentation-trilogy.gemspec#L30

But im not really sure what version? i guess whatever we plan to release the semconv gem as (a patch release? minor? not sure)

ericmustin (commented):
actually one small open question wrt versioning, apols for the back and forth here

exporter/otlp-http/opentelemetry-exporter-otlp-http.gemspec
Coolomina: No problem!, to give you a bit of context I based my commit on the [examples of this repo](https://github.com/open-telemetry/opentelemetry-ruby/search?q=opentelemetry-semantic_conventions+extension%3Agemspec), which are not pinned.

Just tried to stay consistent, that's all :)

ahayworth, line 32: I think maybe we need a version constraint here, yeah?

exporter/jaeger/lib/opentelemetry/exporter/jaeger/attributes.rb
```diff
@@ -0,0 +1,19 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Exporter
+    module Jaeger
+      module Attributes
+        # Identifies Jaeger failed trace exports
+        OTEL_JAEGER_EXPORTER_FAILURE = 'otel.jaeger_exporter.failure'
```
fbogsany, line 12: IMO the prefixes of these constants can be removed. We're already nested within `OpenTelemetry::Exporter::Jaeger` here, so the context is clear.

These are also not attributes - they're metric names.

I think it'd be clearer to structure this like:
```ruby
module OpenTelemetry
  module Exporter
    module Jaeger
      # Metric counting Jaeger failed trace exports
      EXPORTER_FAILURE = 'otel.jaeger_exporter.failure'

      # Metric recording the time it took to flush Spans to Jaeger
      EXPORTER_REQUEST_DURATION = 'otel.jaeger_exporter.request_duration'
    end
  end
end
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -26,6 +26,24 @@ module Export
         # worker thread that exports the spans to wake up and start a new
         # export cycle.
         class BatchSpanProcessor # rubocop:disable Metrics/ClassLength
+          # A comparison between the actual size of the Span and the BatchSpanProcessor @max_queue_size
+          OTEL_BSP_BUFFER_UTILIZATION = 'otel.bsp.buffer_utilization'
```
fbogsany, line 30: We don't need the `OTEL_BSP_` prefix on these constants.

exporter/zipkin/lib/opentelemetry/exporter/zipkin/attributes.rb
```diff
@@ -0,0 +1,19 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Exporter
+    module Zipkin
+      module Attributes
+        # A count of failures when exporting Spans to Zipkin
+        OTEL_ZIPKIN_EXPORTER_FAILURE = 'otel.zipkin_exporter.failure'
```
fbogsany, line 12: These are also metrics rather than attributes, and don't need the `OTEL_ZIPKIN_` prefix.

exporter/otlp/lib/opentelemetry/exporter/otlp/attributes.rb
```diff
@@ -0,0 +1,25 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Exporter
+    module OTLP
+      module Attributes
+        # A count of failed exports of spans over HTTP as Protobuf encoded OTLP ExportTraceServiceRequests
+        OTEL_OTLP_EXPORTER_FAILURE = 'otel.otlp_exporter.failure'
```
fbogsany, line 12: Same comment here about the `OTEL_OTLP_` prefix and metrics, not attributes.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -230,14 +230,14 @@ def measure_request_duration
             ensure
               stop = Process.clock_gettime(Process::CLOCK_MONOTONIC)
               duration_ms = 1000.0 * (stop - start)
-              @metrics_reporter.record_value('otel.otlp_exporter.request_duration',
+              @metrics_reporter.record_value(SemanticConventions::Common::OTEL_OTLP_EXPORTER_REQUEST_DURATION,
```
fbogsany, line 233: This constant is defined in `...::Attributes` rather than `SemanticConventions::Common::`.

exporter/otlp-http/opentelemetry-exporter-otlp-http.gemspec
ahayworth: I think the only real reason is to prevent someone with a very old installation of `opentelemetry-semantic_conventions` from satisfying this constraint, since such an old version wouldn't have the constants needed.

PR #1404, opened 2022-11-16
fix: Adds User-Agent header in OTLP exporter

### LINKED ISSUE #1385: OTLP exporters should emit standard user agent string
As per the following change in the specification (1.14 releasing Oct 4):

> OpenTelemetry protocol exporters SHOULD emit a User-Agent header to at a minimum identify the exporter, the language of its implementation, and the version of the exporter. For example, the Python OTLP exporter version 1.2.3 would report the following:
> ```OTel OTLP Exporter Python/1.2.3```

See https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#user-agent

Coolomina:
Fixes #1385.

In the context of tests, I think it would be more readable to have the raw string in the expectation (Ej: `"OTel OTLP Exporter Ruby/0.24.1"`), rather than have it interpolated.

[Unlike rspec](https://relishapp.com/rspec/rspec-mocks/v/2-11/docs/stubbing-constants/stub-defined-constant#stub-nested-constant), I'm reading that `minitest` does not provide out of the box mechanisms for stubbing constants and other external libraries may be required, so even though I don't personally like it, I'm leaving it interpolated for the sake of not adding more dependencies in the codebase.

Also, I bumped the patch because #1385 is flagged as `bug`, even though it feels more like a `feature` to me.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -35,6 +35,8 @@ module OpenTelemetry
         ERROR_MESSAGE_INVALID_HEADERS = 'headers must be a String with comma-separated URL Encoded UTF-8 k=v pairs or a Hash'
         private_constant(:ERROR_MESSAGE_INVALID_HEADERS)
 
+        DEFAULT_USER_AGENT = "OTel-OTLP-Exporter-Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION} Ruby/#{RUBY_VERSION} (#{RUBY_PLATFORM}; #{RUBY_ENGINE}/#{RUBY_ENGINE_VERSION})"
+
         def self.ssl_verify_mode
           if ENV.key?('OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_PEER')
             OpenSSL::SSL::VERIFY_PEER
@@ -64,13 +66,7 @@ module OpenTelemetry
           @http = http_connection(@uri, ssl_verify_mode, certificate_file)
 
           @path = @uri.path
-          @headers = case headers
-                     when String then parse_headers(headers)
-                     when Hash then headers
-                     else
-                       raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS
-                     end
-          @headers['User-Agent'] = "OTel OTLP Exporter Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION}"
+          @headers = prepare_headers(headers)
           @timeout = timeout.to_f
           @compression = compression
           @metrics_reporter = metrics_reporter || OpenTelemetry::SDK::Trace::Export::MetricsReporter
@@ -388,6 +384,19 @@ module OpenTelemetry
             result.array_value = Opentelemetry::Proto::Common::V1::ArrayValue.new(values: values)
           end
           result
+        end
+
+        def prepare_headers(config_headers)
+          headers = case config_headers
+                    when String then parse_headers(config_headers)
+                    when Hash then config_headers.dup
+                    else
+                      raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS
+                    end
+
+          headers['User-Agent'] = (headers.fetch('User-Agent','') + ' ' + DEFAULT_USER_AGENT).strip
+
+          headers
         end
 
         def parse_headers(raw)
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -70,6 +70,7 @@ module OpenTelemetry
                      else
                        raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS
                      end
+          @headers['User-Agent'] = "OTel OTLP Exporter Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION}"
           @timeout = timeout.to_f
           @compression = compression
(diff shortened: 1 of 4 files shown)


ahayworth (approved):
Looks good to me! Thank you! 😄

exporter/otlp/lib/opentelemetry/exporter/otlp/version.rb
```diff
@@ -8,7 +8,7 @@ module OpenTelemetry
   module Exporter
     module OTLP
       ## Current OpenTelemetry OTLP exporter version
-      VERSION = '0.24.0'
+      VERSION = '0.24.1'
```
plantfansam, line 11: I think our release machinery will handle bumping this

plantfansam (changes requested):
Seems good to me! IDK about reaching into the exporter for instance variables during tests, but that pattern is already established. Don't need to bump the gem version in this PR, tho.

exporter/otlp/lib/opentelemetry/exporter/otlp/version.rb
Coolomina: Done!

Coolomina:
If necessary, I can squash those two commits together so we don't have that last crappy commit 😂

robertlaurin (approved):

robertlaurin:
@plantfansam any remaining blockers from you? I can't merge until your changes requested gets cleared.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -70,6 +70,7 @@ def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPOR
                      else
                        raise ArgumentError, ERROR_MESSAGE_INVALID_HEADERS
                      end
+          @headers['User-Agent'] = "OTel OTLP Exporter Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION}"
```
fbogsany, line 73: This is the first time we're mutating the `headers` Hash that's passed in. In general, we assume the caller maintains ownership and doesn't want us stomping on their hash. Arguably, we should have `dup`ed the hash in the `when Hash` branch above, in case the caller subsequently mutated it. I think we should do that anyway, then we own the hash and this line is 👍 .

robbkidd:
```suggestion
          @headers['User-Agent'] = "OTel-OTLP-Exporter-Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION}"
```

Product identifiers in a User-Agent come in [the format of a token](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.6) and therefore should not include spaces, despite [the current spaced-string example provided in the OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/ff19b9e8212fae4296f63d307591d781d4613304/specification/protocol/exporter.md#user-agent). Spaces in this value are going to mess up a looooot of parsers.

robbkidd (changes requested):
In addition to the request to dup headers before mutation, I'll add that the user-agent string we ship must _not_ include spaces in the "product identifier." That was totally what the OTel spec example provided, but the spec's example is ... not wholly correct.

Update: I've opened [a PR](https://github.com/open-telemetry/opentelemetry-specification/pull/3052) to the OTel spec to propose a no-spaces fix to the example there, too.

robbkidd:
Ooof. I'm sorry I've slept on this PR for so long.

@Coolomina, if you're still in on bring this PR home, I'd like to help! I've got an idea for how to implement @fbogsany's recommendation that we `dup` the incoming headers Hash to protect the exporter's config from later mutation.

I'd also like to propose

1. _Augmenting_ whatever User-Agent may appear in the incoming headers by concatenating ours to any value present.

2. Including more Ruby runtime information to the user-agent, something like:

```ruby
"OTel-OTLP-Exporter-Ruby//#{OpenTelemetry::Exporter::OTLP::VERSION} Ruby/#{RUBY_VERSION} (#{RUBY_PLATFORM})"
# -> OTel-OTLP-Exporter-Ruby/0.24.0 Ruby/3.2.1 (arm64-darwin22)
```

Or even including the engine name and version:

```ruby
"OTel-OTLP-Exporter-Ruby/#{OpenTelemetry::Exporter::OTLP::VERSION} Ruby/#{RUBY_VERSION} (#{RUBY_PLATFORM}; #{RUBY_ENGINE}/#{RUBY_VERSION})"
# -> OTel-OTLP-Exporter-Ruby/0.24.0 Ruby/3.2.1 (arm64-darwin22; ruby/3.2.1)
# … or … 
# -> OTel-OTLP-Exporter-Ruby/0.24.0 Ruby/3.2.0 (opal; opal/3.2.0)
# (I couldn't get JRuby running quickly to make this example.
```

robbkidd:
@Coolomina, are you still interested and available to continue work on this PR? If not, I'd like to take on the work to address PR comments and get it merged.

robbkidd:
As a followup to the removal of the version bump, I pushed a small removal of the changelog entry for that version bump, too.

robbkidd:
Another followup this week to remove spaces from the user-agent header, add more runtime info, and dupe the incoming headers from config to protect against future mutation.

robbkidd (approved):

robbkidd:
@plantfansam, would you be up for a re-review of this PR to address your requested changes (dupe headers to protect from outside mutation)?

robbkidd:
Bitten by new cops!

ericmustin (approved):
lgtm assuming cops r addressed

plantfansam (approved):
👍

fbogsany (approved):

PR #1406, opened 2022-11-27
Make use_all config friendlier

cwjenkins:
Add new flag, `suppress_not_found`, to silence instrumentation that isn't found.
Separate out 'failed to install' from 'not found' in logs during install_all method

### DIFF
--- a/registry/lib/opentelemetry/instrumentation/registry.rb
+++ b/registry/lib/opentelemetry/instrumentation/registry.rb
@@ -75,7 +75,9 @@ module OpenTelemetry
       end
 
       def install_instrumentation(instrumentation, config)
-        if instrumentation.install(config)
+        if !instrumentation.present?
+          OpenTelemetry.logger.debug "Instrumentation: #{instrumentation.name} skipping install given corresponding dependency not found"
+        elsif instrumentation.install(config)
           OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was successfully installed with the following options #{instrumentation.config}"
         else
           OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install"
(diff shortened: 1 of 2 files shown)


ahayworth (commented):
Things I like: a more detailed error message! 🎉 
Things I'm not sure about: should this be an option we ask users to pass in? Or is there a better way to do it?

I think I'm wondering specifically if we shouldn't do:
- logger.info "installed successfully!"
- logger.debug "failed to install... #{reason}" kind of thing

Basically I don't know that anyone _ever_ has the full set of dependencies for `instrumentation-all` available. I presume that everyone would want to use this option (so maybe we should default this to `true` instead of `false`).

Thoughts?

ahayworth:
The only thought that immediately comes to mind is "warn" is better than "error" I think, but ... I'm not sure otherwise. @open-telemetry/ruby-maintainers @open-telemetry/ruby-approvers any thoughts?

ahayworth (approved):
LGTM, thank you! Apologies for the delay, my work responsibilities have shifted so I haven't been super active around otel-ruby recently. I'm but a lowly approver on this repo, so we will need someone from @open-telemetry/ruby-maintainers to take a look.

cheempz:
Upvote! Running into lots of warning messages on startup due to missing dependency and this change would be really useful :)

fbogsany (approved):
This LGTM. I'd like to get a 👍 from @robertlaurin before I merge - he's the expert on the Registry & instrumentation config code.

robbkidd (approved):
👍🏻 Big fan of more nuanced console messaging.

simi (approved):
:pray: LGTM

cheempz:
Helloooo a plea/nag... could this be looked again by the maintainers please?

robertlaurin (approved):

simi (approved):

cheempz:
New year, new ping on this ;).  Could this one be merged and released soon?  Our customer highlighted again as a friction point.

kaylareopelle (approved):

mwear (approved):

ykitamura-mdsol:
is a new version of the opentelemetry-instrumentation-registry gem going to be released anytime soon?

PR #1412, opened 2022-12-09
fix: untraced only works with parent-based sampler

fbogsany:
The `untraced` helper currently only works when a parent-based sampler is installed. As an example, it will not suppress spans if the `ALWAYS_ON` sampler is installed.

This fixes `untraced` to work regardless of the sampler. It does this by communicating via a private `Context` key. An additional helper `untraced?` is introduced to query the state of the key from the `Context`. The SDK `Tracer#start_span` implementation is updated to check for `untraced?` before delegating either to the API `Tracer#start_span` method or to `TracerProvider#internal_start_span`.

The OpenTelemetry JS implementation offers the same functionality implemented in the same way, and serves as inspiration for the implementation in this PR.

### DIFF
--- a/api/lib/opentelemetry/trace/span.rb
+++ b/api/lib/opentelemetry/trace/span.rb
@@ -115,7 +115,7 @@ module OpenTelemetry
 
       # Sets the Status to the Span
       #
-      # If used, this will override the default Span status. Default is OK.
+      # If used, this will override the default Span status. Default status is unset.
       #
       # Only the value of the last call will be recorded, and implementations
       # are free to ignore previous calls.
--- a/api/test/test_helper.rb
+++ b/api/test/test_helper.rb
@@ -4,9 +4,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'simplecov'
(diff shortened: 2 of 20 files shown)


arielvalentin (approved):
Assuming linter errors are addressed.

fbogsany:
> Assuming linter errors are addressed.

I blame Copilot.

fbogsany:
We'll have to be careful with the version bumps when releasing this. We've added a method to the `-common` gem that the `-sdk` now depends on, but we haven't changed the `-sdk` interface.

fbogsany:
> Shouldn't we add a dependency on `-common` to the sdk gemspec?

It's already there.

ahayworth:
> > Shouldn't we add a dependency on -common to the sdk gemspec?
> It's already there.

🤦 Wow - it is indeed, I must have skipped over it when I was hunting for it. 🤦

ahayworth (approved):

arielvalentin (approved):

robertlaurin (approved):

robbkidd (approved):

Issue #1419, opened 2023-01-04
TruffleRuby builds failing on `main`


arielvalentin:
https://github.com/open-telemetry/opentelemetry-ruby/actions/runs/3838561094

nirvdrum:
Hello! I've just been alerted to this issue and I've begun looking into it. If you don't mind, please let us know about things like this on the [TruffleRuby issue tracker](https://github.com/oracle/truffleruby/issues). Alternatively, just ping me and I'll take care of opening the issue. We take compatibility pretty seriously.

All of the test failures look to be due to _simplecov_ 0.22. It's possible there are other failures that are masked, but I think they're all working for me locally. We're caught up in an a bit of an an unfortunate sequence of events. _simplecov_ requires some coordination from the VM in order to record coverage. The _simplecov_ 0.22 release at the end of 2022 inadvertently broke on TruffleRuby because it started calling a method our adapter didn't support. Since TruffleRuby isn't in _simplecov_'s CI, the issue wasn't caught before release. We've since added the missing method to our adapter and _simplecov_ 0.22 works on TruffleRuby 23.0.0-dev. But, that obviously won't fix previous TruffleRuby releases, such as the 22.3.0 build that this project's CI is using.

The issue can be resolved in _simplecov_ with a `respond_to?` call, which is [the fix](https://github.com/simplecov-ruby/simplecov/pull/1043) a community member submitted a PR a couple of days after the 0.22 release. But, the _simplecov_ team hasn't been active since the release. We're hopeful they'll merge that and cut a new release. Likewise, we're hopeful they'll add TruffleRuby to their CI to help avoid these sorts of breakages in the future.

In the meanwhile, these tests should pass running with `truffleruby-head` in the _setup-ruby_ action. I can appreciate not wanting to run CI with a dev build of a language, but it could suffice as a temporary measure. I'll redouble efforts to see if we can get a compatible _simplecov_ released. Alternatively, I think a small monkeypatch that should get things going with TruffleRuby 22.3.0 as well.

eregon:
What I have seen multiple gems do is stick to a previous release of simplecov, like 0.21.2, that's another possibility. E.g. https://github.com/notEthan/jsi/commit/35e9b5fbe3f485f58fffebd7ab6d395fabac9846

ahayworth:
[...]

PR #1428, opened 2023-02-17
ci: re-enable support for truffleruby on all but the GRPC exporter

ahayworth:
We still cannot install the `grpc` gem with truffleruby, but the rest of the otlp exporters build and pass tests. And, the GRPC version of the exporter is currently experimental, unfinished, and unreleased anyways.

cc #1419


ahayworth:
Edit: I also disabled the `Gemspec/DevelopmentDependencies` rubocop rule, which comes in on newer versions.

arielvalentin (approved):

plantfansam (approved):

PR #1438, opened 2023-04-19
chore: Add .tool-versions to .gitignore

elias19r:
Hi! I'm using [`asdf`] and it uses the `.tool-versions` file to keep
track of the Ruby version, similar to `.ruby-version`.

Is it okay to add it to `.gitignore` as well?

[`asdf`]: https://asdf-vm.com/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,6 +22,9 @@ instrumentation/active_record/db
 # IDE Settings
 /.idea/
 
+# asdf configuration
+.tool-versions
+
 # rbenv configuration
 .ruby-version
 


simi:
:thinking: It is common practice to keep your environmental specific files in "global ignore file" local to your environment.

https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer

elias19r:
I've updated my global gitignore file to include `.tool-versions`

PR #1440, opened 2023-04-19
test: Add coverage for other create instrument methods

elias19r:
Also,
- Make instrument name case-insensitive
- Avoid creating constants inside describe blocks because they are
  global to the whole test suite

### DIFF
--- a/metrics_api/lib/opentelemetry/metrics/meter.rb
+++ b/metrics_api/lib/opentelemetry/metrics/meter.rb
@@ -56,17 +56,33 @@ module OpenTelemetry
       private
 
       def create_instrument(kind, name, unit, description, callback)
-        raise InstrumentNameError if name.nil?
-        raise InstrumentNameError if name.empty?
-        raise InstrumentNameError unless NAME_REGEX.match?(name)
-        raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
-        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
+        raise InstrumentNameError if invalid_name?(name)
+        raise InstrumentUnitError if invalid_unit?(unit)
+        raise InstrumentDescriptionError if invalid_description?(description)
+
+        name = normalize_name(name)
 
         @mutex.synchronize do
           OpenTelemetry.logger.warn("duplicate instrument registration occurred for instrument #{name}") if @instrument_registry.include? name
 
           @instrument_registry[name] = yield
         end
+      end
+
+      def invalid_name?(name)
+        !NAME_REGEX.match?(name)
+      end
+
+      def normalize_name(name)
+        name.downcase
+      end
+
+      def invalid_unit?(unit)
+        unit && (!unit.ascii_only? || unit.size > 63)
+      end
+
+      def invalid_description?(description)
+        description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
       end
 
       def utf8mb3_encoding?(string)
--- a/metrics_api/test/opentelemetry/metrics/meter_test.rb
+++ b/metrics_api/test/opentelemetry/metrics/meter_test.rb
@@ -7,15 +7,15 @@
 require 'test_helper'
 
 describe OpenTelemetry::Metrics::Meter do
-  INSTRUMENT_NAME_ERROR = OpenTelemetry::Metrics::Meter::InstrumentNameError
-  INSTRUMENT_UNIT_ERROR = OpenTelemetry::Metrics::Meter::InstrumentUnitError
-  INSTRUMENT_DESCRIPTION_ERROR = OpenTelemetry::Metrics::Meter::InstrumentDescriptionError
-  DUPLICATE_INSTRUMENT_ERROR = OpenTelemetry::Metrics::Meter::DuplicateInstrumentError
+  let(:instrument_name_error)        { OpenTelemetry::Metrics::Meter::InstrumentNameError }
+  let(:instrument_unit_error)        { OpenTelemetry::Metrics::Meter::InstrumentUnitError }
+  let(:instrument_description_error) { OpenTelemetry::Metrics::Meter::InstrumentDescriptionError }
+  let(:duplicate_instrument_error)   { OpenTelemetry::Metrics::Meter::DuplicateInstrumentError }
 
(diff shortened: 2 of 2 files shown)


metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -69,6 +69,22 @@ def create_instrument(kind, name, unit, description, callback)
         end
       end
 
+      def invalid_name?(name)
+        !NAME_REGEX.match?(name)
+      end
```
elias19r, line 74: I noticed that `!NAME_REGEX.match?(name)` already accounts for `name.nil?` and `name.empty?` so I've removed them and kept only the regex match.

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -56,11 +56,11 @@ def create_observable_up_down_counter(name, unit: nil, description: nil, callbac
...
-        raise InstrumentNameError if name.nil?
-        raise InstrumentNameError if name.empty?
-        raise InstrumentNameError unless NAME_REGEX.match?(name)
-        raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
-        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
+        raise InstrumentNameError if invalid_name?(name)
+        raise InstrumentUnitError if invalid_unit?(unit)
+        raise InstrumentDescriptionError if invalid_description?(description)
+
+        name = normalize_name(name)
```
elias19r, line 63: This is new. From the Metrics' Meter API specification for the [Instrument Name Syntax](https://opentelemetry.io/docs/reference/specification/metrics/api/#instrument-name-syntax) it says:

> They are case-insensitive, ASCII strings.

So here I'm normalizing the name with `downcase` before using it in `@instrument_registry[name]`

metrics_api/test/opentelemetry/metrics/meter_test.rb
```diff
@@ -7,15 +7,15 @@
...
 
 describe OpenTelemetry::Metrics::Meter do
-  INSTRUMENT_NAME_ERROR = OpenTelemetry::Metrics::Meter::InstrumentNameError
-  INSTRUMENT_UNIT_ERROR = OpenTelemetry::Metrics::Meter::InstrumentUnitError
-  INSTRUMENT_DESCRIPTION_ERROR = OpenTelemetry::Metrics::Meter::InstrumentDescriptionError
-  DUPLICATE_INSTRUMENT_ERROR = OpenTelemetry::Metrics::Meter::DuplicateInstrumentError
+  let(:instrument_name_error)        { OpenTelemetry::Metrics::Meter::InstrumentNameError }
+  let(:instrument_unit_error)        { OpenTelemetry::Metrics::Meter::InstrumentUnitError }
+  let(:instrument_description_error) { OpenTelemetry::Metrics::Meter::InstrumentDescriptionError }
+  let(:duplicate_instrument_error)   { OpenTelemetry::Metrics::Meter::DuplicateInstrumentError }
```
elias19r, line 13: Constants defined in a `describe` block are global to the whole test suite which may not be the desirable effect, so I've replaced them with a `let` definition.

metrics_api/test/opentelemetry/metrics/meter_test.rb
```diff
@@ -25,49 +25,259 @@
...
+      ' ',
+      '1_meter',
+      '_meter',
+      '-meter',
+      '.meter',
+      'a' * 64,
+      'a@',
+      'a!'
+    ]
+  end
```
elias19r, line 267: I usually don't write DRY tests, but in this case I think having a singular place to define invalid names/units/descriptions helps because all of them use the same validation rules.

metrics_api/test/opentelemetry/metrics/meter_test.rb
```diff
@@ -25,49 +25,259 @@
...
+        meter.create_up_down_counter('AN_INSTRUMENT')
+        _(log_stream.string).must_match(/duplicate instrument registration occurred for instrument an_instrument/)
+      end
+    end
+  end
+
+  describe '#create_observable_counter' do
+    it 'raises an error when name does not match pattern' do
+      invalid_names.each do |invalid_name|
+        _(-> { meter.create_observable_counter(invalid_name, callback: -> {}) })
```
elias19r, line 149: **Not related to this PR, but I can include:**

I noticed the keyword arg `callback:` is required. So here I'm passing it even though I don't need it for the test case.

[From the specification](https://opentelemetry.io/docs/reference/specification/metrics/api/#asynchronous-instrument-api), it says:

> Users can provide `callback` functions, but it is up to their discretion. Therefore, this API MUST be structured to accept a variable number of `callback` functions, including none.

It **doesn't say** _"MUST NOT obligate a user to provide one"_ like for the other parameters, but it says _"including none"_.

I wonder if _"including none"_ would translate to optional in Ruby with a default empty `[]` array. What do you think?

In doc terms:

```rb
# @param callback [optional Proc, Array<Proc>]
```

metrics_api/test/opentelemetry/metrics/meter_test.rb
elias19r: After reading [Synchronous Instrument API](https://opentelemetry.io/docs/reference/specification/metrics/api/#synchronous-instrument-api) I noticed the following

> The API SHOULD be documented in a way to communicate to users that the
> `name` parameter needs to conform to the instrument name syntax.
> **The API SHOULD NOT validate the `name`; that is left to
> implementations of the API.**

and

> The `unit` parameter needs to support the instrument unit rule.
> Meaning, the API MUST accept a case-sensitive string that supports
> ASCII character encoding and can hold at least 63 characters.
> **The API SHOULD NOT validate the `unit`.**

Similar for `name` and `unit` in [Asynchronous Instrument API](https://opentelemetry.io/docs/reference/specification/metrics/api/#asynchronous-instrument-api)

So I'm thinking if we should actually remove all these validations from API.

elias19r: Because of this, I'm thinking of closing this PR and replacing it with https://github.com/open-telemetry/opentelemetry-ruby/pull/1442

elias19r:
I'm closing this PR in favor of https://github.com/open-telemetry/opentelemetry-ruby/pull/1442

PR #1442, opened 2023-04-27
feat: Rework Meter and MeterProvider API

elias19r:
- Remove argument validations from API and add doc comments:
  validation should be performed by the implementation of the API
- Add new parent classes `SynchronousInstrument` and `AsynchronousInstrument`
- Make `name` required for `Meter`
- Add prefix `NOOP_` to no-op constant instruments
- Normalize Instrument names with downcase before registering them
  because their name should be case-insensitive
- Add `advice` to Synchronous Instruments


metrics_api/.rubocop.yml
```diff
@@ -15,8 +16,14 @@ Metrics/CyclomaticComplexity:
   Max: 20
 Metrics/ParameterLists:
   Enabled: false
+
+Naming/VariableNumber:
+  Enabled: false
```
elias19r, line 21: I happy to revert any of these style changes, just let me know

nit: For this one, I think that naming variables with `something_1` and `something_2` in tests is easier to read

metrics_api/.rubocop.yml
```diff
@@ -15,8 +16,14 @@ Metrics/CyclomaticComplexity:
...
 Naming/FileName:
   Exclude:
     - "lib/opentelemetry-metrics-api.rb"
+
+Style/IfUnlessModifier:
+  Enabled: false
```
elias19r, line 27: nit: When this is enabled, we can be forced into writing long lines with `if/unless` at the end and I think may not be easy to read

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -8,71 +8,156 @@ module OpenTelemetry
...
       end
 
       private
 
       def create_instrument(kind, name, unit, description, callback)
-        raise InstrumentNameError if name.nil?
-        raise InstrumentNameError if name.empty?
-        raise InstrumentNameError unless NAME_REGEX.match?(name)
-        raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
-        raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
```
elias19r, line 63: After reading [Synchronous Instrument API](https://opentelemetry.io/docs/reference/specification/metrics/api/#synchronous-instrument-api) I noticed the following

> The API SHOULD be documented in a way to communicate to users that the
> `name` parameter needs to conform to the instrument name syntax.
> **The API SHOULD NOT validate the `name`; that is left to
> implementations of the API.**

and

> The `unit` parameter needs to support the instrument unit rule.
> Meaning, the API MUST accept a case-sensitive string that supports
> ASCII character encoding and can hold at least 63 characters.
> **The API SHOULD NOT validate the `unit`.**

Similar for `name` and `unit` in [Asynchronous Instrument API](https://opentelemetry.io/docs/reference/specification/metrics/api/#asynchronous-instrument-api)

So, I think we should remove these validations from here and add yard doc comments instead. What do you think?

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -8,71 +8,156 @@ module OpenTelemetry
...
+        name = name.downcase
```
elias19r, line 151: This is new. From the Metrics' Meter API specification for the [Instrument Name Syntax](https://opentelemetry.io/docs/reference/specification/metrics/api/#instrument-name-syntax) it says:

> They are case-insensitive, ASCII strings.

So here I'm normalizing the name with `downcase` before using it in `@instrument_registry[name]`

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,44 @@
...
+
+        # @param callback [Proc, Array<Proc>]
+        #   Callback functions should:
+        #   - be reentrant safe;
+        #   - not take an indefinite amount of time;
+        #   - not make duplicate observations (more than one Measurement with the same attributes)
+        #     across all registered callbacks;
+        def unregister_callback(callback)
+          @callback -= Array(callback)
+        end
```
elias19r, line 40: This is how I imagine register/unregister callback functions would look in Ruby. Feel free to suggest any changes

https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-instrument-api

metrics_api/lib/opentelemetry/metrics/instrument/synchronous_instrument.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module OpenTelemetry
+  module Metrics
+    module Instrument
+      # https://opentelemetry.io/docs/reference/specification/metrics/api/#synchronous-instrument-api
+      class SynchronousInstrument
+        attr_reader :name, :unit, :description
+
+        # @api private
+        def initialize(name, unit: nil, description: nil)
```
robertlaurin, line 15: > - advice for implementations.
>
>Users can provide advice, but its up to their discretion. Therefore, this API needs to be structured to accept advice, but MUST NOT obligate the user to provide it.
>
>advice needs to be structured as described in [instrument advice](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-advice), with parameters that are general and specific to a particular instrument kind. The API SHOULD NOT validate advice.

Did you want to work this in as a follow up PR? https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#synchronous-instrument-api

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -8,71 +8,156 @@ module OpenTelemetry
...
-      HISTOGRAM = Instrument::Histogram.new
-      OBSERVABLE_GAUGE = Instrument::ObservableGauge.new
-      UP_DOWN_COUNTER = Instrument::UpDownCounter.new
-      OBSERVABLE_UP_DOWN_COUNTER = Instrument::ObservableUpDownCounter.new
+      NOOP_COUNTER                    = Instrument::Counter.new('np-op')
+      NOOP_HISTOGRAM                  = Instrument::Histogram.new('np-op')
+      NOOP_UP_DOWN_COUNTER            = Instrument::UpDownCounter.new('np-op')
+      NOOP_OBSERVABLE_COUNTER         = Instrument::ObservableCounter.new('np-op')
+      NOOP_OBSERVABLE_GAUGE           = Instrument::ObservableGauge.new('np-op')
+      NOOP_OBSERVABLE_UP_DOWN_COUNTER = Instrument::ObservableUpDownCounter.new('np-op')
```
robertlaurin, line 16:
```suggestion
      NOOP_COUNTER                    = Instrument::Counter.new('no-op')
      NOOP_HISTOGRAM                  = Instrument::Histogram.new('no-op')
      NOOP_UP_DOWN_COUNTER            = Instrument::UpDownCounter.new('no-op')
      NOOP_OBSERVABLE_COUNTER         = Instrument::ObservableCounter.new('no-op')
      NOOP_OBSERVABLE_GAUGE           = Instrument::ObservableGauge.new('no-op')
      NOOP_OBSERVABLE_UP_DOWN_COUNTER = Instrument::ObservableUpDownCounter.new('no-op')
```

metrics_api/lib/opentelemetry/metrics/meter.rb
robertlaurin: >So, I think we should remove these validations from here and add yard doc comments instead. What do you think?

I think that's reasonable, the API spec is explicit in that it should perform any validations against the name field, so that will be the responsibility of the SDK implementation. Yard docs seem like a good solution for this.

metrics_api/lib/opentelemetry/metrics/meter.rb
elias19r: Oops, I didn't notice that. Thanks!

metrics_api/lib/opentelemetry/metrics/instrument/synchronous_instrument.rb
elias19r: Updated in https://github.com/open-telemetry/opentelemetry-ruby/pull/1442/commits/7cce415076ab44665b2ead3e4836de42796a21f5

I'm not sure how `advice` will be exactly used, but I think of it as an "options" Hash with recommended values. For example in docs, [the recommended bucket boundaries for a histogram](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#instrument-advice). I imagine they will add standard names to be used in `advice`

elias19r:
If you squash&merge, feel free to use the PR's description in the commit message (and add/edit anything). Otherwise, let me know and I can re-create commits with better messages :D

metrics_api/lib/opentelemetry/metrics/meter.rb
elias19r: Also in the "Metrics No-Op API Implementation" docs it makes clear No-Op Meters must not validate. For example:

> The Meter MUST accept these parameters. However, the Meter MUST NOT validate any argument it receives.

https://opentelemetry.io/docs/specs/otel/metrics/noop/#counter-creation

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
elias19r: I just realized that, with this, people are able to register callbacks to a No-Op instances of Asynchronous Instruments. I think I should delete these methods' bodies and leave them to be implemented by the SDK

robertlaurin:
@elias19r I'm about ready to merge, there's some branch conflicts that need to be resolved first though.

robertlaurin (approved):

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module Metrics
+    module Instrument
+      # https://opentelemetry.io/docs/reference/specification/metrics/api/#asynchronous-instrument-api
```
fbogsany, line 10: This link didn't work for me. I found the doc at:
```suggestion
      # https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-instrument-api
```

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,40 @@
...
+      class AsynchronousInstrument
+        attr_reader :name, :unit, :description, :callback
+
+        # @api private
+        def initialize(name, unit: nil, description: nil, callback: nil)
+          @name = name
+          @unit = unit || ''
+          @description = description || ''
+          @callback = callback ? Array(callback) : []
```
fbogsany, line 19: nit: since this is an array of functions, can we pluralize the name as `callbacks`?

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,40 @@
...
+        end
+
+        # @param callback [Proc, Array<Proc>]
+        #   Callback functions should:
+        #   - be reentrant safe;
+        #   - not take an indefinite amount of time;
+        #   - not make duplicate observations (more than one Measurement with the same attributes)
+        #     across all registered callbacks;
+        def register_callback(callback); end
```
fbogsany, line 28: Similarly, the plural form is preferable. I'd also like to use the "splat" form: 
```suggestion
        def register_callbacks(*callback); end
```
This allows calling as:
```ruby
instr.register_callbacks(cb1)
instr.register_callbacks(cb2, cb3)
```
instead of
```ruby
instr.register_callbacks(cb1)
instr.register_callbacks([cb2, cb3])
```

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
elias19r: Good catch! I think they changed it. https://opentelemetry.io/docs/reference/specification/metrics/api/#asynchronous-instrument-api is redirecting to https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-instrument-api with status 301. I'll update them!

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
elias19r: Sure, I can pluralize them!

I think the reason I initially didn't do that was to keep the same "symbol same" from the specification although this `@callback` instance variable is internal.

But in `Meter` API, in `metrics_api/lib/opentelemetry/metrics/meter.rb`, should I pluralize the keyword arg `callback:` as well in the `create_observable_counter`, `create_observable_gauge`, and `create_observable_up_down_counter` methods?

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
elias19r: I like the splat form, but I've been thinking if should actually support all those cases later in the implementation:

```rb
instr.register_callback(cb1)
instr.register_callback(cb2, cb3)
instr.register_callback([cb4, cb5]) # And support this as well?
# Just so it is consistent with `callbacks:` from the `Meter#create_observable_counter` API
```

In the implementation, I think we could support that with

```rb
def register_callbacks(*callbacks)
  @callbacks.concat(callbacks.flatten)
end
```

Because otherwise it would feel odd to be able to pass an array of procs when creating an instrument, like

```rb
meter = MeterProvider.get_meter('test')
instr = meter.create_observable_counter('test', callbacks: [cb1, cb2])
```

but not be able to pass an array of procs when registering callbacks later

```rb
meter = MeterProvider.get_meter('test')
instr = meter.create_observable_counter('test')

instr.register_callbacks([cb1, cb2]) # ?
```

metrics_api/test/opentelemetry/metrics/instrument/asynchronous_instrument_test.rb
```diff
@@ -0,0 +1,77 @@
...
+    end
+  end
+
+  describe '#register_callbacks' do
+    it 'responds without errors and returns nil' do
+      instrument = build_asynchronous_instrument('test-instrument')
+
+      assert(instrument.register_callbacks(-> {}).nil?)
+      assert(instrument.register_callbacks(-> {}, -> {}).nil?)
+      assert(instrument.register_callbacks([-> {}, -> {}]).nil?)
```
fbogsany, line 60:
```suggestion
      assert(instrument.register_callbacks(-> {}, -> {}).nil?)
```

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,40 @@
...
+
+        # @api private
+        def initialize(name, unit: nil, description: nil, callbacks: nil)
+          @name = name
+          @unit = unit || ''
+          @description = description || ''
+          @callbacks = callbacks ? Array(callbacks) : []
+        end
+
+        # @param callbacks [Proc, Array<Proc>]
```
fbogsany, line 22: Technically, `callbacks` is an `Array<Proc>` now, always - I'm not sure how to document a "splat" param.

metrics_api/lib/opentelemetry/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,40 @@
...
+        #   Callback functions should:
+        #   - be reentrant safe;
+        #   - not take an indefinite amount of time;
+        #   - not make duplicate observations (more than one Measurement with the same attributes)
+        #     across all registered callbacks;
+        def register_callbacks(*callbacks); end
+
+        # @param callbacks [Proc, Array<Proc>]
```
fbogsany, line 30: Same here.

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -8,71 +8,162 @@ module OpenTelemetry
...
+      #   Must conform to the instrument name syntax:
+      #   not nil or empty, case-insensitive ASCII string that matches `/\A[a-z][a-z0-9_.-]{0,62}\z/i`
+      # @param unit [optional String]
+      #   Must conform to the instrument unit rule:
+      #   case-sensitive ASCII string with maximum length of 63 characters
+      # @param description [optional String]
+      #   Must conform to the instrument description rule:
+      #   UTF-8 string but up to 3 bytes per charater with maximum length of 1023 characters
+      # @param advice [optional Hash] Set of recommendations aimed at assisting
+      #   implementations in providing useful output with minimal configuration
```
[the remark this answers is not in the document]
elias19r: It seems `advice` has been added to the specification recently. I'm not sure how it will be exactly used. For now, I'm thinking of it as an "options" Hash with recommended values. For example in docs, [the recommended bucket boundaries for a histogram](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advice). I _imagine_ they will define standard names to be used in `advice`, maybe:

```rb
advice: {
  histogram: {
    explicit_bucket_boundaries: [0.1, 0.5, 1.0, 5.0, 10.0, 25.0, Float::INFINITY]
  }
}
```

metrics_api/test/opentelemetry/metrics/instrument/asynchronous_instrument_test.rb
elias19r: What do you think of allowing all these use cases:

```rb
instr.register_callback(cb1)
instr.register_callback(cb2, cb3)
instr.register_callback([cb4, cb5]) # And support this as well?
# Just so it is consistent with `callbacks:` from the `Meter#create_observable_counter` API
```

See https://github.com/open-telemetry/opentelemetry-ruby/pull/1442#discussion_r1214852101

metrics_api/lib/opentelemetry/metrics/meter.rb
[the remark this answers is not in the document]
elias19r: Reading the SDK page, I found another reference to `advice` in [Default Aggregation](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#default-aggregation):

> The Default Aggregation informs the SDK to use the Instrument `kind` to select an aggregation and `advice` to influence aggregation configuration parameters (as noted in the "Selected Aggregation" column).

> Instrument Kind | Selected Aggregation
>  -- | --
>  [...] | [...]
>  Histogram | Explicit Bucket Histogram Aggregation, with `ExplicitBucketBoundaries` from `advice` if provided

elias19r:
@robertlaurin @fbogsany Hi, I haven't had a lot of time to work on this, but I could start a follow-up PR adjusting the existing SDK code to the changes I'm making to the API here.

https://github.com/thoughtbot/opentelemetry-ruby/pull/1
(I've opened it in another repo just for visibility. I'll move to here if this PR gets merged)

It is mainly about making the Proxy Instruments and SDK Instruments inherit from the API Instrument classes. If you get a chance to review it, let me know if that makes sense :)

After that, I wonder if we could complete the implementation of one instrument (maybe Counter). I don't quite see the big picture yet, but I'm starting to understand the concepts of the data model for metrics.

Hronom:
We interested in getting suport for metrics at least no auto instrumentation for ruby, it's shame that it still not have is maintainers can implement it?

metrics_api/lib/opentelemetry/metrics/meter.rb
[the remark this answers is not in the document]
kaylareopelle: I don't think this should block the current PR from being merged, but it seems that [`advice` has been renamed to `advisory parameters`](https://github.com/open-telemetry/opentelemetry-specification/pull/3662). 

* API: https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advisory-parameters 
* SDK: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-advisory-parameters 

It has a mixed status in the API and is experimental in the SDK.

elias19r:
I think I won't be able to keep up with those PRs, it's been a while since I read the docs. I'll close them but keep the branches in case the changes are useful.

PR #1443, opened 2023-05-01
Update wrong links in various gemspecs.

simi:


robertlaurin (approved):
Thanks for fixing this up. These gems are not published yet, and won't be for a little while yet so I'm glad you caught this before the are released.

PR #1447, opened 2023-05-15
fix: Consistent probability sampler init

fbogsany:
In the consistent probability sampler, we cache the exponent of the next highest and next lowest negative power-of-two values, and the probability of the highest. E.g. for a probability of 0.1, the next highest negative power-of-two is 2**-3 (0.125) and the next lowest is 2**-4 (0.0625), and the probability of the higher value is the fraction of difference between the 0.0625 and 0.125 that is 0.1 - i.e. 0.1 - 0.0625 / 0.0625 = 0.6.

Unfortunately, we calculated the ceiling incorrectly in the current implementation. Because these are negative power-of-two exponents, the _smaller_ value is the _upper_ bound and the _larger_ value is the _lower_ bound. This corrects that error.

Relevant song: https://www.youtube.com/watch?v=Esdnbq4hxHo

### DIFF
--- a/sdk_experimental/test/opentelemetry/sdk/trace/samplers/consistent_probability_based_test.rb
+++ b/sdk_experimental/test/opentelemetry/sdk/trace/samplers/consistent_probability_based_test.rb
@@ -62,4 +62,37 @@ describe OpenTelemetry::SDK::Trace::Samplers::ConsistentProbabilityBased do
 
     # TODO: statistical tests
   end
+
+  describe '#initialize' do
+    # Check that the internal state is initialized correctly. We cache the
+    # floor and ceil of the negative power-of-two of the provided probability,
+    # and the probability of the ceil. It should be possible to calculate the
+    # original probability from these.
+    #
+    # Note that these are negative power-of-two exponent values, so the ceiling
+    # is the smaller value.
+
+    it 'initializes correctly for power-of-two probabilities' do
+      p_values = 62.downto(0)
+      probabilities = p_values.map { |i| 2**-i }
+      p_values.zip(probabilities).each do |p, probability|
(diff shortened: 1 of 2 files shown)


sdk_experimental/test/opentelemetry/sdk/trace/samplers/consistent_probability_based_test.rb
```diff
@@ -62,4 +62,37 @@
...
+    it 'initializes correctly for 0.1' do
+      sampler = OpenTelemetry::SDK::Trace::Samplers::ConsistentProbabilityBased.new(0.1)
+      _(sampler.instance_variable_get(:@p_floor)).must_equal(4) # 2**-4 = 0.0625
+      _(sampler.instance_variable_get(:@p_ceil)).must_equal(3) # 2**-3 = 0.125
+      _(sampler.instance_variable_get(:@p_ceil_probability)).must_be_within_epsilon(0.6) # 0.0375 / 0.0625
+      ceil_prob = sampler.instance_variable_get(:@p_ceil_probability)
+      floor_prob = 1 - ceil_prob
+      p_floor = sampler.instance_variable_get(:@p_floor)
+      p_ceil = sampler.instance_variable_get(:@p_ceil)
+      _((2**-p_ceil * ceil_prob) + (2**-p_floor * floor_prob)).must_be_within_epsilon(0.1)
```
plantfansam, line 95:
```suggestion
      # Verify that the sampling probability encoded in each `p` adds up to the probability
      # in the the initializer argument  
      _((2**-p_ceil * ceil_prob) + (2**-p_floor * floor_prob)).must_be_within_epsilon(0.1)
```

PR #1500, opened 2023-07-27
chore(deps): Allow google-protobuf `~> 3.14`

arielvalentin:
GitHub still uses legacy versions of the protobuf gem. This makes it challenging to
upgrade to newer versions of the OTLP Exporter which only supports versions `~> 3.19`.

This change loosens the restrictions to allow GitHub to adopt newer versions of the protobuf
definitions using an older verson of the library.

~~This change explicitly skips 3.15 due to bugs like https://github.com/protocolbuffers/protobuf/issues/8337~~

### DIFF
--- a/exporter/otlp/Appraisals
+++ b/exporter/otlp/Appraisals
@@ -4,11 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-appraise "google-protobuf-3.14" do
-  gem 'google-protobuf', "3.14"
-end
-
-(16..23).each do |i|
+(14..23).each do |i|
   version = "3.#{i}"
   appraise "google-protobuf-#{version}" do
     gem 'google-protobuf', "~> #{version}"
--- a/exporter/otlp/Appraisals
+++ b/exporter/otlp/Appraisals
@@ -4,7 +4,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-(19..23).each do |i|
+appraise "google-protobuf-3.14" do
+  gem 'google-protobuf', "3.14"
+end
+
+(16..23).each do |i|
   version = "3.#{i}"
   appraise "google-protobuf-#{version}" do
     gem 'google-protobuf', "~> #{version}"
--- a/exporter/otlp/opentelemetry-exporter-otlp.gemspec
+++ b/exporter/otlp/opentelemetry-exporter-otlp.gemspec
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.0'
 
   spec.add_dependency 'googleapis-common-protos-types', '~> 1.3'
-  spec.add_dependency 'google-protobuf', '~> 3.19'
+  spec.add_dependency 'google-protobuf', '~> 3.14', '!= 3.15'
   spec.add_dependency 'opentelemetry-api', '~> 1.1'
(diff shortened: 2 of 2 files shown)


simi:
I'm not sure it is open-telemetry responsibility to prevent dependencies with bugs. If anyone runs `3.15` version of protobuf and is happy about, there is no need to block it. If 3.15 doesn't work at all (AFAIK it works, just with minor problems at some point), it should be considered yanked.

arielvalentin:
@simi I did this out of an abundance of caution based on our experience with 3.15 at GitHub and Shopify. 

I think the use case we are mostly concerned with is  where the application isn't using protobuf and for some reason the buggy 3.15 is pulled in then the application suffers from performance problem and runs into bugs as a result of the Exporter bringing in a bad dependency.

That being said, I do see your point about how 3.15 had additional bug fix releases and it will prevent users that are happy with it from using the gem. For those reasons I am going to change the constraints and allow 3.15 installs.

exporter/otlp/Appraisals
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+appraise "google-protobuf-3.14" do
+  gem 'google-protobuf', "3.14"
+end
+
```
arielvalentin, line 10:
```suggestion
```

exporter/otlp/Appraisals
```diff
@@ -0,0 +1,16 @@
...
+(16..23).each do |i|
```
arielvalentin, line 11:
```suggestion
(14..23).each do |i|
```

exporter/otlp/opentelemetry-exporter-otlp.gemspec
```diff
@@ -26,12 +26,13 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.0'
 
   spec.add_dependency 'googleapis-common-protos-types', '~> 1.3'
-  spec.add_dependency 'google-protobuf', '~> 3.19'
+  spec.add_dependency 'google-protobuf', '~> 3.14', '!= 3.15'
```
arielvalentin, line 29:
```suggestion
  spec.add_dependency 'google-protobuf', '~> 3.14'
```

robertlaurin:
>I'm not sure it is open-telemetry responsibility to prevent dependencies with bugs. If anyone runs 3.15 version of protobuf and is happy about, there is no need to block it. If 3.15 doesn't work at all (AFAIK it works, just with minor problems at some point), it should be considered yanked.

I agree with that point of view. We (Shopify) saw a host of issues with memory pressure as a result of that release. While we don't want this gem to be positioned as responsible for making sure no one pulls in potentially problematic releases of its dependencies, we saw a large amounts of reports coming in internally around "OpenTelemetry has a memory leak", when in fact it did not. 

We're really just trying to do the right thing in terms of potentially protecting downstream users of this package (ourselves included) and preventing support burden from misidentified memory leaks.

arielvalentin:
@robertlaurin @simi et.al. I rebased main onto this branch so it should be a clean review now.

robertlaurin (approved):
Just waiting on CI now

arielvalentin:
@robertlaurin thank you!!! This website where I opened this PR also thanks you!!!

PR #1506, opened 2023-08-01
fix(exporter/otlp): align endpoint environment varaible handling with spec

### LINKED ISSUE #1505: OTLP exporter env var handling is not spec compliant
**Description of the bug**

Hi everyone :slightly_smiling_face: It looks like the handling of the `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_ENDPOINT` environment variables is not in [compliance with the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp), there are a few examples of this that we discovered in some internal testing:

- `endpoint` and `OTEL_EXPORTER_OTLP_ENDPOINT` are both set to the same value, `
  - expected behavior: the URL from `endpoint` is used and not modified
  - actual behavior: `v1/traces` is appended to the `endpoint`
- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` and `OTEL_EXPORTER_OTLP_ENDPOINT` are set to the same value
  - expected behavior: the URL from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` is used and not modified
  - actual behavior: when v1/traces` is appended to the value from `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT`
- if `OTEL_EXPORTER_OTLP_ENDPOINT` does not have a trailing slash (example: `https://localhost:1234/api/v2/otlp`)
  - expected behavior: the URL from `OTEL_EXPORTER_OTLP_ENDPOINT` is appended with `/v1/traces` (example: `https://localhost:1234/api/v2/otlp` becomes `https://localhost:1234/api/v2/otlp/v1/traces`)
  - actual behavior: the URL from `OTEL_EXPORTER_OTLP_ENDPOINT` is stripped of the last part and appended with `/v1/traces` (example: `https://localhost:1234/api/v2/otlp` becomes `https://localhost:1234/api/v2/v1/otlp`)

Operating system details: Linux, Ubuntu 22.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.2.0"
RUBY_DESCRIPTION: "ruby 3.2.0 (2022-12-25 revision a528908271) [x86_64-linux]"

Tests intended for `exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb', all of them should fail in the current implementation

```rb
    it 'does not join endpoint with v1/traces if endpoint is set and is equal to OTEL_EXPORTER_OTLP_ENDPOINT' do
      exp = OpenTelemetry::TestHelpers.with_env(
        'OTEL_EXPORTER_OTLP_ENDPOINT' => 'https://localhost:1234/custom/path'
      ) do
        OpenTelemetry::Exporter::OTLP::Exporter.new(endpoint: 'https://localhost:1234/custom/path')
      end
      _(exp.instance_variable_get(:@path)).must_equal '/custom/path'
    end

    it 'appends the correct path if OTEL_EXPORTER_OTLP_ENDPOINT does have a path without a trailing slash' do
      exp = OpenTelemetry::TestHelpers.with_env(
        # simulate OTLP endpoints built on top of an exiting API
        'OTEL_EXPORTER_OTLP_ENDPOINT' => 'https://localhost:1234/api/v2/otlp'
      ) do
        OpenTelemetry::Exporter::OTLP::Exporter.new
      end
      _(exp.instance_variable_get(:@path)).must_equal '/api/v2/otlp/v1/traces'
    end

    it 'appends the correct path if OTEL_EXPORTER_OTLP_ENDPOINT does have a path without a trailing slash' do
      exp = OpenTelemetry::TestHelpers.with_env(
        # simulate OTLP endpoints built on top of an exiting API
        'OTEL_EXPORTER_OTLP_ENDPOINT' => 'https://localhost:1234/api/v2/otlp'
      ) do
        OpenTelemetry::Exporter::OTLP::Exporter.new
      end
      _(exp.instance_variable_get(:@path)).must_equal '/api/v2/otlp/v1/traces'
    end

```

pichlermarc:
This PR fixes the non spec compliant behavior outlined in #1505 :slightly_smiling_face: 

Fixes #1505

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -395,6 +395,8 @@ module OpenTelemetry
           elsif !ENV['OTEL_EXPORTER_OTLP_ENDPOINT'].nil?
             # Append v1/traces only if the non-signal specific env var is set.
             non_signal_specific_endpoint = ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+            # append '/' if not present
+            non_signal_specific_endpoint += '/' unless non_signal_specific_endpoint.end_with?('/')
             begin
               URI.join(non_signal_specific_endpoint, 'v1/traces')
             rescue URI::InvalidURIError
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -47,21 +47,15 @@ module OpenTelemetry
           end
         end
 
-        def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: 'http://localhost:4318/v1/traces'),
+        def initialize(endpoint: nil,
                        certificate_file: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE', 'OTEL_EXPORTER_OTLP_CERTIFICATE'),
                        ssl_verify_mode: Exporter.ssl_verify_mode,
                        headers: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS', default: {}),
                        compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: 'gzip'),
                        timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: 10),
                        metrics_reporter: nil)
-          raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" unless OpenTelemetry::Common::Utilities.valid_url?(endpoint)
+          @uri = determine_url(endpoint)
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
-
-          @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
-                   URI.join(endpoint, 'v1/traces')
-                 else
-                   URI(endpoint)
-                 end
 
           @http = http_connection(@uri, ssl_verify_mode, certificate_file)
 
@@ -384,6 +378,40 @@ module OpenTelemetry
             result.array_value = Opentelemetry::Proto::Common::V1::ArrayValue.new(values: values)
           end
           result
+        end
+
+        # Returns the url used by the exporter based on the precedence defined in the specification:
+        # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp
+        #
+        # @param [String, nil] endpoint
+        # @return [URI::Generic]
+        def determine_url(endpoint)
+          if !endpoint.nil?
+            # Use endpoint provided via argument as-is
+            parse_url_or_raise(endpoint)
+          elsif !ENV['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT'].nil?
+            # Use signal-specific endpoint as-is
+            parse_url_or_raise(ENV['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT'])
+          elsif !ENV['OTEL_EXPORTER_OTLP_ENDPOINT'].nil?
+            # Append v1/traces only if the non-signal specific env var is set.
+            non_signal_specific_endpoint = ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+            begin
+              URI.join(non_signal_specific_endpoint, 'v1/traces')
+            rescue URI::InvalidURIError
+              raise ArgumentError, "invalid url for OTLP::Exporter #{non_signal_specific_endpoint}"
+            end
+          else
+            # use default in all other cases
+            parse_url_or_raise('http://localhost:4318/v1/traces')
+          end
+        end
+
+        def parse_url_or_raise(endpoint)
+          raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" if endpoint.nil? || endpoint.strip.empty?
+
+          URI(endpoint)
+        rescue URI::InvalidURIError
+          raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}"
         end
 
         def prepare_headers(config_headers)
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -54,11 +54,7 @@ module OpenTelemetry
                        compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: 'gzip'),
                        timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: 10),
                        metrics_reporter: nil)
-          @uri = begin
-            prepare_endpoint(endpoint)
-          rescue ArgumentError => e
-            raise ArgumentError, e.message
-          end
+          @uri = prepare_endpoint(endpoint)
 
           raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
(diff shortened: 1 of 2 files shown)


mwear (approved):
LGTM. I have minor concerns that some people could be relying on the broken behavior, but this definitely a bug that should be fixed. Thanks @pichlermarc!

pichlermarc:
> LGTM. I have minor concerns that some people could be relying on the broken behavior, but this definitely a bug that should be fixed. Thanks @pichlermarc!

Fully agree, thank you for the review :slightly_smiling_face:

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -386,6 +380,42 @@ def as_otlp_any_value(value)
...
+            # Use endpoint provided via argument as-is
+            parse_url_or_raise(endpoint)
+          elsif !ENV['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT'].nil?
+            # Use signal-specific endpoint as-is
+            parse_url_or_raise(ENV['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT'])
+          elsif !ENV['OTEL_EXPORTER_OTLP_ENDPOINT'].nil?
+            # Append v1/traces only if the non-signal specific env var is set.
+            non_signal_specific_endpoint = ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+            # append '/' if not present
+            non_signal_specific_endpoint += '/' unless non_signal_specific_endpoint.end_with?('/')
```
fbogsany, line 399: This is not necessary. The `URI.join` call below will insert the `/` if required.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -386,6 +380,42 @@ def as_otlp_any_value(value)
...
+            begin
+              URI.join(non_signal_specific_endpoint, 'v1/traces')
+            rescue URI::InvalidURIError
+              raise ArgumentError, "invalid url for OTLP::Exporter #{non_signal_specific_endpoint}"
```
fbogsany, line 403: I strongly prefer to `raise ArgumentError` from the public method rather than a private one. It is more informative for the caller. Since this case is so specific, I think it would also be useful to call out the environment variable, in case the user wonders where the value came from.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -47,22 +47,16 @@ def self.ssl_verify_mode
...
-        def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_ENDPOINT', 'OTEL_EXPORTER_OTLP_ENDPOINT', default: 'http://localhost:4318/v1/traces'),
+        def initialize(endpoint: nil,
                        certificate_file: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_CERTIFICATE', 'OTEL_EXPORTER_OTLP_CERTIFICATE'),
                        ssl_verify_mode: Exporter.ssl_verify_mode,
                        headers: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS', default: {}),
                        compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: 'gzip'),
                        timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: 10),
                        metrics_reporter: nil)
-          raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" unless OpenTelemetry::Common::Utilities.valid_url?(endpoint)
+          @uri = determine_url(endpoint)
```
fbogsany, line 57:
```suggestion
          endpoint ||= ENV['OTEL_EXPORTER_OTLP_TRACES_ENDPOINT']
          @uri = begin
                   if endpoint.nil?
                     endpoint = ENV['OTEL_EXPORTER_OTLP_ENDPOINT'] || 'http://localhost:4318'
                     URI.join(endpoint, 'v1/traces')
                   elsif endpoint.strip.empty?
                     raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}"
                   else
                     URI(endpoint)
                   end
                 rescue URI::Error
                   raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}"
                 end
```
I think this captures the essence of the rules while raising as close to the public interface as possible.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
pichlermarc: Hmm, I could be wrong (this is the first time I'm using ruby). I also initially assumed it does, but trying it out in `irb`:

```
$ irb
3.2.0 :001 > require('uri')
 => true 
3.2.0 :002 > URI.join('http://test.com/api','v2/url')
 => #<URI::HTTP http://test.com/v2/url> 
```

drops `api` from the path. I think it may be debatable if it is directly intended by the [spec here](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/protocol/exporter.md#endpoint-urls-for-otlphttp) to append it. However, I think many language implementations do append '/' if it's not there (examples: [JavaScript](https://github.com/open-telemetry/opentelemetry-js/blob/abfe059d72d2a50d993ee7fb5ea50a1c29851a31/experimental/packages/otlp-exporter-base/src/util.ts#L52), [Python](https://github.com/open-telemetry/opentelemetry-python/blob/c9277ffb137fad999ee14bee4f99bad2d00b8b03/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py#L247), [Java](https://github.com/open-telemetry/opentelemetry-java/blob/ea7b761757caae434825a45ee6e7d9e559fe76cd/exporters/otlp/all/src/main/java/io/opentelemetry/exporter/otlp/internal/OtlpConfigUtil.java#L78))

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
pichlermarc: Yes, that's better (using some language features that I didn't know yet, nice :smile:). Thanks for the suggestion :raised_hands: 

Adapted the suggesting in [40204f9](https://github.com/open-telemetry/opentelemetry-ruby/pull/1506/commits/40204f970c45710608f2f8fbae95d0fbe4e557dc), but included appending `/` (see https://github.com/open-telemetry/opentelemetry-ruby/pull/1506#discussion_r1289966425)

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
pichlermarc: I'm re-raising in the constructor now to hide the private method [f2748b7](https://github.com/open-telemetry/opentelemetry-ruby/pull/1506/commits/f2748b7f89d29c61f6d255d051ddcbee208692fc), couldn't pull everything in the constructor unfortunately as complexity ended up being too high and rubocop failed.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -47,21 +47,20 @@ def self.ssl_verify_mode
...
                        ssl_verify_mode: Exporter.ssl_verify_mode,
                        headers: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS', default: {}),
                        compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: 'gzip'),
                        timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: 10),
                        metrics_reporter: nil)
-          raise ArgumentError, "invalid url for OTLP::Exporter #{endpoint}" unless OpenTelemetry::Common::Utilities.valid_url?(endpoint)
-          raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
+          @uri = begin
+            prepare_endpoint(endpoint)
+          rescue ArgumentError => e
```
arielvalentin, line 59: what am I missing here? Why is this rescuing an error and then copying the message to a new instance and raising that instead?

arielvalentin: Ok I see. Did Rubocop fail and tell you that you needed to handle the error raised by `prepare_endpoint`?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -47,21 +47,20 @@ def self.ssl_verify_mode
...
+            raise ArgumentError, e.message
+          end
```
arielvalentin, line 61:
```suggestion
          @uri = prepare_endpoint(endpoint)
```

Since `prepare_endpoint` is raising an error, there isn't really a need to `rescue` the error and re-raise it by copying the message.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
pichlermarc: There's https://github.com/open-telemetry/opentelemetry-ruby/pull/1506#discussion_r1289141806 suggesting to move the code from `prepare_endpoint()` (private method) to the constructor (public) and then throwing the `ArgumentError` from the constructor one as suggested in https://github.com/open-telemetry/opentelemetry-ruby/pull/1506#discussion_r1289028612. However, while pulling in the code from `prepare_endpoint()` to the constructor allows me to throw from the public interface, it also causes cognitive complexity to go up, and then robocop fails.

I'm also quite unhappy with re-throwing the exception and copying the message there, so if you agree, I'd remove it as per your suggestion https://github.com/open-telemetry/opentelemetry-ruby/pull/1506/files#r1331886287. :slightly_smiling_face:

pichlermarc: Applied [e4d31d6](https://github.com/open-telemetry/opentelemetry-ruby/pull/1506/commits/e4d31d6b58d35ad1f30201df74a08387915f483f) to get rid of it. :slightly_smiling_face:

mwear:
@arielvalentin @fbogsany can you confirm that your suggestions have been properly addressed?

PR #1510, opened 2023-08-03
chore: Raise test_gem ruby/setup-ruby version

kaylareopelle:
I'm trying to use SimpleCov in #1509, but TruffleRuby raises an error about `Coverage.running?` not being implemented. There's a fix for this in TruffleRuby 22. 

TruffleRuby 22 isn't accessible with the `test_gem` action's current version of `ruby/setup-ruby`. However, TruffleRuby 23 is available on the latest `ruby/setup-ruby` version. Using `v1.152.0` fixed the problem.

Example action failure: https://github.com/open-telemetry/opentelemetry-ruby/actions/runs/5744430524/job/15570701339?pr=1509#step:7:150

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -58,7 +58,7 @@ runs:
     # ...but not for appraisals, sadly.
     - name: Install Ruby ${{ inputs.ruby }} with dependencies
       if: "${{ steps.setup.outputs.appraisals == 'false' }}"
-      uses: ruby/setup-ruby@v1.144.1
+      uses: ruby/setup-ruby@v1.152.0
       with:
         ruby-version: "${{ inputs.ruby }}"
         working-directory: "${{ steps.setup.outputs.gem_dir }}"
@@ -69,7 +69,7 @@ runs:
     # If we're using appraisals, do it all manually.
     - name: Install Ruby ${{ inputs.ruby }} without dependencies
       if: "${{ steps.setup.outputs.appraisals == 'true' }}"
-      uses: ruby/setup-ruby@v1.144.1
+      uses: ruby/setup-ruby@v1.152.0
       with:
         ruby-version: "${{ inputs.ruby }}"
         bundler:  "latest"


simi (approved):
:information_source:  it is possible to setup dependabot for actions, I can try to setup that if welcomed in separate PR

arielvalentin (approved):

robbkidd:
The [setup-ruby maintainers recommend riding the lightning and tracking the v1 branch](https://github.com/ruby/setup-ruby#versioning) instead of specific pinned versions. Have we considered and rejected that idea already at some time in the past?

robbkidd (approved):
👍🏻 Newer is Better!

I approve newer ... and while we're on the subject, the [setup-ruby maintainers recommend riding the lightning and tracking the v1 branch](https://github.com/ruby/setup-ruby#versioning) instead of specific pinned versions. Have we considered and rejected that idea already at some time in the past?

mwear (approved):

simi:
> 👍🏻 Newer is Better!
> 
> I approve newer ... and while we're on the subject, the [setup-ruby maintainers recommend riding the lightning and tracking the v1 branch](https://github.com/ruby/setup-ruby#versioning) instead of specific pinned versions. Have we considered and rejected that idea already at some time in the past?

We had similar discussion in rubygems.org. Per https://securityscorecards.dev/, GitHub Actions should be pinned to actual commit, not to version, to minimise the risk of pulling in unexpected changes (potentially with security concerns). That works well combined with dependabot. That way you can scan the changelog at least (or changes), but get updated often.

I can try to configure both if welcomed - https://github.com/ossf/scorecard-action and monthly/weekly dependabot for actions.

kaylareopelle:
Thanks, all! I've added a note to chat about how we want to set the `ruby/setup-ruby` action versions to next week's SIG agenda.

https://docs.google.com/document/d/1D15bO8o340sQm2CVZiukEJuCO_XMMHKPuTznoEhyFqE/edit

PR #1517, opened 2023-08-25
feat: Create Logs SDK LoggerProvider

### LINKED ISSUE #1483: Logs SDK - Logger Provider
Implement the Logs SDK Logger Provider

https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerprovider 

- [x] Allow a [Resource](https://opentelemetry.io/docs/specs/otel/resource/sdk/) to be specified. If a Resource is specified, it SHOULD be associated with all the LogRecords produced by any Logger from the LoggerProvider.
- [x] [Create Logger Provider](https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerprovider-creation) - when? where?
- [x] Create Logger via Logger Provider
- [x] Shutdown
- [x] ForceFlush

kaylareopelle:
Fulfills the [Logs SDK specs for Logger Provider](https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerprovider)

The Logger Provider implementation is drawn mainly from the Tracer Provider implementation, as they have very similar specs.

* [Tracer Provider specs](https://opentelemetry.io/docs/specs/otel/trace/sdk/#tracer-provider)
* [TracerProvider class](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb) 

If merged, this PR would fulfill the following elements on the [spec-compliance-matrix for Logs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/spec-compliance-matrix.md#logs)

- [ ] LoggerProvider.GetLogger
- [ ] LoggerProvider.GetLogger accepts attributes
- [ ] LoggerProvider.Shutdown
- [ ] LoggerProvider.ForceFlush

Here are some other implementations of LoggerProvider for comparison:
* [Java (stable)](https://github.com/open-telemetry/opentelemetry-java/blob/90ab3665dc0fb56598fcf9b9c15941eedb3822f8/sdk/logs/src/main/java/io/opentelemetry/sdk/logs/SdkLoggerProvider.java)
* [JS (experimental)](https://github.com/open-telemetry/opentelemetry-js/blob/main/experimental/packages/sdk-logs/src/LoggerProvider.ts)
* [PHP (stable)](https://github.com/open-telemetry/opentelemetry-php/blob/main/src/SDK/Logs/LoggerProvider.php)
* [Python (experimental)](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/_logs/_internal/__init__.py#L581)

Resolves #1483

### DIFF
--- a/logs_sdk/lib/opentelemetry-logs-sdk.rb
+++ b/logs_sdk/lib/opentelemetry-logs-sdk.rb
@@ -4,6 +4,5 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'opentelemetry'
 require 'opentelemetry/sdk'
 require 'opentelemetry/sdk/logs'
--- a/logs_sdk/lib/opentelemetry-logs-sdk.rb
+++ b/logs_sdk/lib/opentelemetry-logs-sdk.rb
@@ -4,5 +4,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry'
+require 'opentelemetry/sdk'
 require 'opentelemetry/sdk/logs'
-require 'opentelemetry/sdk/logs/version'
--- a/logs_sdk/lib/opentelemetry/sdk/logs/export.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/export.rb
@@ -1,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The export module contains result codes for LoggerProvider#force_flush
+      # and LoggerProvider#shutdown
+      module Export
+        # The operation finished successfully.
+        SUCCESS = 0
+
+        # The operation finished with an error.
+        FAILURE = 1
+
+        # The operation timed out.
+        TIMEOUT = 2
+      end
+    end
+  end
+end
--- a/logs_sdk/lib/opentelemetry/sdk/logs/export.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/export.rb
@@ -7,8 +7,7 @@
 module OpenTelemetry
   module SDK
     module Logs
-      # The export module contains result codes for LoggerProvider#force_flush
-      # and LoggerProvider#shutdown
+      # The Export module contains result codes for exporters
       module Export
         # The operation finished successfully.
         SUCCESS = 0
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
@@ -26,7 +26,7 @@ module OpenTelemetry
         # the process after an invocation, but before the `Processor` exports
         # the completed spans.
         #
-        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @param [Numeric] timeout An optional timeout in seconds.
         # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
         #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
         def force_flush(timeout: nil)
@@ -35,7 +35,7 @@ module OpenTelemetry
 
         # Called when {LoggerProvider#shutdown} is called.
         #
-        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @param [Numeric] timeout An optional timeout in seconds.
         # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
         #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
         def shutdown(timeout: nil)
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
@@ -7,15 +7,15 @@
 module OpenTelemetry
   module SDK
     module Logs
-      # LogRecordProcessor describes a duck type and provides synchronous no-op hooks for when a
-      # {LogRecord} is started or when a {LogRecord} is ended. It is not required to subclass this
+      # LogRecordProcessor describes a duck type and provides a synchronous no-op hook for when a
+      # {LogRecord} is emitted. It is not required to subclass this
       # class to provide an implementation of LogRecordProcessor, provided the interface is
       # satisfied.
       class LogRecordProcessor
         # Called when a {LogRecord} is emitted. Subsequent calls are not
         # permitted after shutdown is called.
         # @param [LogRecord] log_record The emitted {LogRecord}
-        # @param [Context] context The resolved Context (the explicitly passed Context or the current Context)
+        # @param [Context] context The {Context}
         def on_emit(log_record, context); end
 
         # Export all log records to the configured `Exporter` that have not yet
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
@@ -15,7 +15,7 @@ module OpenTelemetry
         # Called when a {LogRecord} is emitted. Subsequent calls are not
         # permitted after shutdown is called.
         # @param [LogRecord] log_record The emitted {LogRecord}
-        # @param [Context] context The resolved Context
+        # @param [Context] context The resolved Context (the explicitly passed Context or the current Context)
         def on_emit(log_record, context); end
 
         # Export all log records to the configured `Exporter` that have not yet
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # LogRecordProcessor describes a duck type and provides synchronous no-op hooks for when a
+      # {LogRecord} is started or when a {LogRecord} is ended. It is not required to subclass this
+      # class to provide an implementation of LogRecordProcessor, provided the interface is
+      # satisfied.
+      class LogRecordProcessor
+        # Called when a {LogRecord} is emitted. Subsequent calls are not
+        # permitted after shutdown is called.
+        # @param [LogRecord] log_record The emitted {LogRecord}
+        # @param [Context] context The resolved Context
+        def on_emit(log_record, context); end
+
+        # Export all log records to the configured `Exporter` that have not yet
+        # been exported.
+        #
+        # This method should only be called in cases where it is absolutely
+        # necessary, such as when using some FaaS providers that may suspend
+        # the process after an invocation, but before the `Processor` exports
+        # the completed spans.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def force_flush(timeout: nil)
+          Export::SUCCESS
+        end
+
+        # Called when {LoggerProvider#shutdown} is called.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def shutdown(timeout: nil)
+          Export::SUCCESS
+        end
+      end
+    end
+  end
+end
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
@@ -1,0 +1,33 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The SDK implementation of OpenTelemetry::Logs::Logger
+      class Logger < OpenTelemetry::Logs::Logger
+        attr_reader :instrumentation_scope, :logger_provider
+
+        # @api private
+        #
+        # Returns a new {OpenTelemetry::SDK::Logs::Logger} instance. This should
+        # not be called directly. New loggers should be created using
+        # {LoggerProvider#logger}.
+        #
+        # @param [String] name Instrumentation package name
+        # @param [String] version Instrumentation package version
+        # @param [LoggerProvider] logger_provider The {LoggerProvider} that
+        #   initialized the logger
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def initialize(name, version, logger_provider)
+          @instrumentation_scope = InstrumentationScope.new(name, version)
+          @logger_provider = logger_provider
+        end
+      end
+    end
+  end
+end
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
@@ -9,8 +9,6 @@ module OpenTelemetry
     module Logs
       # The SDK implementation of OpenTelemetry::Logs::Logger
       class Logger < OpenTelemetry::Logs::Logger
-        attr_reader :instrumentation_scope, :logger_provider
-
         # @api private
         #
         # Returns a new {OpenTelemetry::SDK::Logs::Logger} instance. This should
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
@@ -29,7 +29,7 @@ module OpenTelemetry
           @stopped = false
         end
 
-        # Creates an {OpenTelemetry::SDK::Logs::Logger} instance.
+        # Returns an {OpenTelemetry::SDK::Logs::Logger} instance.
         #
         # @param [String] name Instrumentation package name
         # @param [String] version Optional instrumentation package version
@@ -43,7 +43,7 @@ module OpenTelemetry
               "invalid name. Name provided: #{name.inspect}")
           end
 
-          OpenTelemetry::SDK::Logs::Logger.new(name, version, self)
+         Logger.new(name, version, self)
         end
 
         # Adds a new log record processor to this LoggerProvider's
@@ -78,19 +78,19 @@ module OpenTelemetry
           @mutex.synchronize do
             if @stopped
               OpenTelemetry.logger.warn('LoggerProvider#shutdown called multiple times.')
-              return OpenTelemetry::SDK::Logs::Export::FAILURE
+              return Export::FAILURE
             end
 
             start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
             results = @log_record_processors.map do |processor|
               remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
-              break [OpenTelemetry::SDK::Logs::Export::TIMEOUT] if remaining_timeout&.zero?
+              break [Export::TIMEOUT] if remaining_timeout&.zero?
 
               processor.shutdown(timeout: remaining_timeout)
             end
 
             @stopped = true
-            results.max || OpenTelemetry::SDK::Logs::Export::SUCCESS
+            results.max || Export::SUCCESS
           end
         end
 
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
@@ -1,0 +1,127 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The SDK implementation of OpenTelemetry::Logs::LoggerProvider.
+      class LoggerProvider < OpenTelemetry::Logs::LoggerProvider
+        attr_reader :resource, :log_record_processors
+
+        UNEXPECTED_ERROR_MESSAGE = 'unexpected error in ' \
+          'OpenTelemetry::SDK::Logs::LoggerProvider#%s'
+        # Returns a new LoggerProvider instance.
+        #
+        # @param [optional Resource] resource The resource to associate with
+        #   new LogRecords created by {Logger}s created by this LoggerProvider.
+        # @param [optional Array] log_record_processors The
+        #   {LogRecordProcessor}s to associate with this LoggerProvider.
+        #
+        # @return [OpenTelemetry::SDK::Logs::LoggerProvider]
+        def initialize(
+          resource: OpenTelemetry::SDK::Resources::Resource.create,
+          log_record_processors: []
+        )
+          @log_record_processors = log_record_processors
+          @mutex = Mutex.new
+          @resource = resource
+          @stopped = false
+        end
+
+        # Creates an {OpenTelemetry::SDK::Logs::Logger} instance.
+        #
+        # @param [optional String] name Instrumentation package name
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def logger(name = nil, version = nil)
+          name ||= ''
+          version ||= ''
+
+          OpenTelemetry.logger.warn('LoggerProvider#logger called without providing a logger name.') if name.empty?
+
+          OpenTelemetry::SDK::Logs::Logger.new(name, version, self)
+        end
+
+        # Adds a new log record processor to this LoggerProvider's
+        # log_record_processors.
+        #
+        # @param [LogRecordProcessor] log_record_processor The
+        #   {LogRecordProcessor} to add to this LoggerProvider.
+        def add_log_record_processor(log_record_processor)
+          @mutex.synchronize do
+            @log_record_processors = log_record_processors.dup.push(log_record_processor)
+          end
+        end
+
+        # Attempts to stop all the activity for this LoggerProvider. Calls
+        # {LogRecordProcessor#shutdown} for all registered {LogRecordProcessor}s.
+        #
+        # This operation may block until all log records are processed. Must
+        # be called before turning off the main application to ensure all data
+        # are processed and exported.
+        #
+        # After this is called all newly created {LogRecord}s will be no-op.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def shutdown(timeout: nil)
+          @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('LoggerProvider#shutdown called multiple times.')
+              return OpenTelemetry::SDK::Logs::Export::FAILURE
+            end
+
+            start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
+            results = log_record_processors.map do |processor|
+              remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+              break [OpenTelemetry::SDK::Logs::Export::TIMEOUT] if remaining_timeout&.zero?
+
+              processor.shutdown(timeout: remaining_timeout)
+            end
+
+            @stopped = true
+            results.max || OpenTelemetry::SDK::Logs::Export::SUCCESS
+          end
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: UNEXPECTED_ERROR_MESSAGE % __method__)
+          Export::FAILURE
+        end
+
+        # Immediately export all {LogRecord}s that have not yet been exported
+        # for all the registered {LogRecordProcessor}s.
+        #
+        # This method should only be called in cases where it is absolutely
+        # necessary, such as when using some FaaS providers that may suspend
+        # the process after an invocation, but before the {LogRecordProcessor}
+        # exports the completed {LogRecord}s.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def force_flush(timeout: nil)
+          @mutex.synchronize do
+            return Export::SUCCESS if @stopped
+
+            start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
+            results = log_record_processors.map do |processor|
+              remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+              return Export::TIMEOUT if remaining_timeout&.zero?
+
+              processor.force_flush(timeout: remaining_timeout)
+            end
+
+            results.max || Export::SUCCESS
+          end
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: UNEXPECTED_ERROR_MESSAGE % __method__)
+          Export::FAILURE
+        end
+      end
+    end
+  end
+end
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
@@ -53,6 +53,11 @@ module OpenTelemetry
         #   {LogRecordProcessor} to add to this LoggerProvider.
         def add_log_record_processor(log_record_processor)
           @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('calling LoggerProvider#' \
+                'add_log_record_processor after shutdown.')
+              return
+            end
             @log_record_processors = @log_record_processors.dup.push(log_record_processor)
           end
         end
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
@@ -18,7 +18,7 @@ module OpenTelemetry
 
         # Returns a new LoggerProvider instance.
         #
-        # @param [Resource] resource An optional resource to associate with
+        # @param [optional Resource] resource The resource to associate with
         #   new LogRecords created by {Logger}s created by this LoggerProvider.
         #
         # @return [OpenTelemetry::SDK::Logs::LoggerProvider]
@@ -32,7 +32,7 @@ module OpenTelemetry
         # Returns an {OpenTelemetry::SDK::Logs::Logger} instance.
         #
         # @param [String] name Instrumentation package name
-        # @param [String] version Optional instrumentation package version
+        # @param [optional String] version Instrumentation package version
         #
         # @return [OpenTelemetry::SDK::Logs::Logger]
         def logger(name:, version: nil)
@@ -71,7 +71,7 @@ module OpenTelemetry
         #
         # After this is called all newly created {LogRecord}s will be no-op.
         #
-        # @param [Numeric] timeout An optional timeout in seconds.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
         #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
         def shutdown(timeout: nil)
@@ -102,7 +102,7 @@ module OpenTelemetry
         # the process after an invocation, but before the {LogRecordProcessor}
         # exports the completed {LogRecord}s.
         #
-        # @param [Numeric] timeout An optional timeout in seconds.
+        # @param [optional Numeric] timeout An optional timeout in seconds.
         # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
         #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
         def force_flush(timeout: nil)
--- a/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
@@ -9,7 +9,7 @@ module OpenTelemetry
     module Logs
       # The SDK implementation of OpenTelemetry::Logs::LoggerProvider.
       class LoggerProvider < OpenTelemetry::Logs::LoggerProvider
-        attr_reader :resource, :log_record_processors
+        attr_reader :resource
 
         UNEXPECTED_ERROR_MESSAGE = 'unexpected error in ' \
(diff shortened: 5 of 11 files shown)


logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+        # After this is called all newly created {LogRecord}s will be no-op.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def shutdown(timeout: nil)
+          @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('LoggerProvider#shutdown called multiple times.')
+              return OpenTelemetry::SDK::Logs::Export::FAILURE
```
kaylareopelle, line 81: The code for `LoggerProvider#shutdown` and `LoggerProvider#force_flush` are drawn mainly from `TracerProvider` methods of the same names. 

Why does `#shutdown` return a failure status code if called when `@stopped` is true, but `#force_flush` returns a success status code?

JS, [for example](https://github.com/open-telemetry/opentelemetry-js/blob/b400c2e5d9729c3528482781a93393602dc6dc9f/experimental/packages/sdk-logs/src/LoggerProvider.ts#L124-L145 ), does similar things for both `#force_flush` and `#shutdown`.

The specs on `#force_flush` (very similar for [Logs](https://opentelemetry.io/docs/specs/otel/logs/sdk/#forceflush) and [Traces](https://opentelemetry.io/docs/specs/otel/trace/sdk/#forceflush)), don't describe the desired behavior in cases where the method is called, but the provider has stopped/shutdown.

kaylareopelle, line 81: The spec states: 
>After the call to Shutdown, subsequent attempts to get a Logger are not allowed. SDKs SHOULD return a valid no-op Logger for these calls, if possible.
[Source](https://opentelemetry.io/docs/specs/otel/logs/sdk/#shutdown)

This is also true for `TracerProvider#shutdown` [(source)](https://opentelemetry.io/docs/specs/otel/trace/sdk/#shutdown). This leads me to believe the shutdown method should return  `OpenTelemetry::Logs::LoggerProvider::NOOP_LOGGER` when it is stopped. If so, should this be applied to the Trace SDK too?

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+
+              processor.force_flush(timeout: remaining_timeout)
+            end
+
+            results.max || Export::SUCCESS
+          end
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: FORCE_FLUSH_ERROR)
+          Export::FAILURE
+        end
```
kaylareopelle, line 125: There is not a `rescue` in the [TracerProvider#force_flush method](https://github.com/open-telemetry/opentelemetry-ruby/blob/4347e0f47bb0d6a4870f1e9311f20d3d344fb34a/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L89-L113), but the method documentation states "`Export::FAILURE` should be returned if a non-specific failure occurred." I'm assuming a non-specific failure would be any `StandardError` instance. 

* Is there a condition to return `Export::FAILURE` on non-specific errors in `TracerProvider#force_flush` that I'm missing? 
* Similar documentation exists for the `LoggerProvider`/`TracerProvider#shutdown` methods. If we choose to add a `rescue` here, should the `#shutdown` methods also have a similar `rescue`?

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+
+        # Adds a new log record processor to this LoggerProvider's
+        # log_record_processors.
+        #
+        # @param [LogRecordProcessor] log_record_processor The
+        #   {LogRecordProcessor} to add to this LoggerProvider.
+        def add_log_record_processor(log_record_processor)
+          @mutex.synchronize do
+            @log_record_processors = log_record_processors.dup.push(log_record_processor)
```
kaylareopelle, line 61: This is copied from `TracerProvider#add_span_processor`. @fallwith and I were chatting about this LOC. He was asking about the use of `.dup.push` here.

Does `dup.push` provide atomicity in a way that the `push` method alone couldn't deliver? Or is there a different reason `dup.push` was used?

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
jack-berg: The goal is that after shutdown, all existing and new loggers are effectively noop meaning that subsequent calls to emit a log record do not make it to any registered log record processor. Languages have flexibility in how to implement this behavior.

jack-berg: I think there's some flexibility in the spec language around the behavior of shutdown. It says:

> After the call to Shutdown, subsequent attempts to get a Logger are not allowed. 

In java, we log an error message if when shutdown is called multiple times, but don't return an error. I think the intent of "not allowed" is to avoid performing the behavioral side effects of calling shutdown. Specifically, called shutdown on the provider is support to be implemented by calling shutdown on all registered processors, which we don't want to do multiple times.

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,22 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # Presently no-op LogRecordProcessor
+      class LogRecordProcessor
```
jack-berg, line 11: No 'onEmit' method?

kaylareopelle: The LoggerProvider specs don't refer to the `onEmit` method, so I thought I'd introduce that functionality in a separate PR. I'm open to adding it if you think it'd clarify things.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: Thanks for clarifying! It sounds like Ruby is in alignment as-is.

kaylareopelle: Ah, ok. That makes a lot of sense. The `@stopped` instance variable is taking care of that for us in Ruby.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+        #
+        # @param [optional String] name Instrumentation package name
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def logger(name = nil, version = nil)
+          name ||= ''
+          version ||= ''
+
+          OpenTelemetry.logger.warn(EMPTY_NAME_ERROR) if name.empty?
```
kaylareopelle, line 47: > In the case where an invalid name (null or empty string) is specified, a working Logger MUST be returned as a fallback rather than returning null or throwing an exception, its name SHOULD keep the original invalid value, and a message reporting that the specified value is invalid SHOULD be logged.
[source](https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerprovider)

This logic is drawn from [TracerProvider#tracer](https://github.com/open-telemetry/opentelemetry-ruby/blob/4347e0f47bb0d6a4870f1e9311f20d3d344fb34a/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L51-L56). The null or empty string will not stop InstrumentationScope from being generated during Logger initialization.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+
+          @mutex.synchronize do
+            OpenTelemetry::SDK::Logs::Logger.new(name, version, self)
```
kaylareopelle, line 50: > The input provided by the user MUST be used to create an [InstrumentationScope](https://opentelemetry.io/docs/specs/otel/glossary/#instrumentation-scope) instance which is stored on the created Logger.

[source](https://opentelemetry.io/docs/specs/otel/logs/sdk/#loggerprovider)

The name and version here are passed to the Logger to create the InstrumentationScope.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+          end
+        end
+
+        # Adds a new log record processor to this LoggerProvider's
+        # log_record_processors.
+        #
+        # @param [LogRecordProcessor] log_record_processor The
+        #   {LogRecordProcessor} to add to this LoggerProvider.
+        def add_log_record_processor(log_record_processor)
```
kaylareopelle, line 59: > The LoggerProvider MAY provide methods to update the configuration. If configuration is updated (e.g., adding a LogRecordProcessor), the updated configuration MUST also apply to all already returned Loggers (i.e. it MUST NOT matter whether a Logger was obtained from the LoggerProvider before or after the configuration change). Note: Implementation-wise, this could mean that Logger instances have a reference to their LoggerProvider and access configuration only via this reference.

[source](https://opentelemetry.io/docs/specs/otel/logs/sdk/#configuration)

Every logger instance has a reference to the logger provider that created it. [(see L50)](https://github.com/open-telemetry/opentelemetry-ruby/pull/1517/files#diff-f9a2c98dbf5b32983909ec3963c7cf01517197a75482ba1f27ed1151ab5be862R50) This gives loggers access to the log record processors.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,129 @@
...
+        # This operation may block until all log records are processed. Must
+        # be called before turning off the main application to ensure all data
+        # are processed and exported.
+        #
+        # After this is called all newly created {LogRecord}s will be no-op.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def shutdown(timeout: nil)
```
kaylareopelle, line 77: See: https://opentelemetry.io/docs/specs/otel/logs/sdk/#shutdown 

* Only called once by setting `@stopped` to `true` L92
* Provides a way to let the caller know whether it succeeded, failed, or timed out by returning  `OpenTelemetry::SDK::Logs::EXPORT` constants (L81, L87, L93)
* Implements a timeout through calls to `OpenTelemetry::Common::Utilities.timeout_timestamp` and `OpenTelemetry::Common::Utilities.maybe_timeout` (L84, L86)
* Invokes shutdown on all the processors (L85 - 89)

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle, line 125: Spec reference: https://opentelemetry.io/docs/specs/otel/logs/sdk/#forceflush 

Out of time atm, but I'll circle back with more 1:1 details.

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,22 @@
...
+        def shutdown(timeout: nil)
+          # TODO: implement
```
mwear, line 13:
```suggestion
          Export::SUCCESS
```

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,22 @@
...
+        end
+
+        def force_flush(timeout: nil)
+          # TODO: implement
```
mwear, line 17:
```suggestion
          Export::SUCCESS
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
mwear: I'm uncertain, but @fbogsany probably knows. I think it has to do with mutating the `processors` array. Looking at `TracerProvider` a span gets a reference to `processors` on start. If a span processor is added after a span is started, and the `processors` array is mutated, a span will see a processor that wasn't there when it was started. I'm not sure if this would happen in the real world, or if this is what we're protecting against, but that is a reason why we might prefer `dup.push` over `push`.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
mwear: We might be assuming that processors will handle their own errors, but that might be optimistic. I would be fine with rescuing from `StandardError` in `force_flush` and `shutdown` across the board. I'll let others chime in if they have opinions.

mwear (commented):
Looks good. I have a few minor comments.

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
kaylareopelle: @mwear - Sounds great. Updates made in: 26ef80d0fd93561f1283ca32f50460db760fd4cc

kaylareopelle:
@mwear - Re: [this comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/1517#discussion_r1343275519)

I've updated #shutdown to include error handling to match #force flush in: 3c0dd5769f3d3dac1a7e5e1c86ffee33cdbbb38e

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def logger(name = nil, version = nil)
+          name ||= ''
+          version ||= ''
+
+          OpenTelemetry.logger.warn('LoggerProvider#logger called without providing a logger name.') if name.empty?
+
+          OpenTelemetry::SDK::Logs::Logger.new(name, version, self)
```
kaylareopelle, line 46: @arielvalentin, @robbkidd - Per our discussion during the SIG, I've removed the `@mutex.synchronize` wrapper.

mwear (approved):
LGTM, thanks @kaylareopelle!

kaylareopelle:
Thanks, @mwear! I've rebased the branch into one commit.

arielvalentin (approved):
_Assuming we gain a better understanding of duping the log processors..._

robertlaurin (approved):

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
fbogsany: Yes, that's exactly the reason - each SDK `Span` instance has a reference to the list of span processors, and that list should not change between the span calling `on_start` and `on_finish`. IDK if that applies in your case - I need to read the rest of the PR.

fbogsany (changes requested):
No changes yet - I just want to block so I have a chance to read it closely.

logs_sdk/lib/opentelemetry-logs-sdk.rb
```diff
@@ -4,5 +4,6 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+require 'opentelemetry'
```
fbogsany, line 7: This shouldn't be necessary.
```suggestion
```

logs_sdk/lib/opentelemetry/sdk/logs/export.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The export module contains result codes for LoggerProvider#force_flush
```
fbogsany, line 10: These _should_ also be result codes for exporters - their use shouldn't just relate to these two methods.

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,47 @@
...
+    module Logs
+      # LogRecordProcessor describes a duck type and provides synchronous no-op hooks for when a
+      # {LogRecord} is started or when a {LogRecord} is ended. It is not required to subclass this
+      # class to provide an implementation of LogRecordProcessor, provided the interface is
+      # satisfied.
+      class LogRecordProcessor
+        # Called when a {LogRecord} is emitted. Subsequent calls are not
+        # permitted after shutdown is called.
+        # @param [LogRecord] log_record The emitted {LogRecord}
+        # @param [Context] context The resolved Context
```
fbogsany, line 18: What is the "resolved Context" here? Who does the resolution? (Note that I haven't read this part of the spec or all of this PR, or the logs API, so maybe this is answered clearly somewhere else - just asking as a naive reviewer.)

logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
```diff
@@ -0,0 +1,33 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The SDK implementation of OpenTelemetry::Logs::Logger
+      class Logger < OpenTelemetry::Logs::Logger
+        attr_reader :instrumentation_scope, :logger_provider
```
fbogsany, line 12: I'm curious why we expose these attributes. In the `Tracer` and `Meter`, they're private, so the inconsistency is a little weird.

logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
```diff
@@ -0,0 +1,33 @@
...
+        # @param [String] name Instrumentation package name
+        # @param [String] version Instrumentation package version
+        # @param [LoggerProvider] logger_provider The {LoggerProvider} that
+        #   initialized the logger
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def initialize(name, version, logger_provider)
+          @instrumentation_scope = InstrumentationScope.new(name, version)
+          @logger_provider = logger_provider
+        end
```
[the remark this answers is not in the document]
fbogsany: The `@api private` comment is our standard way of marking "internal" methods. There is no practical way to enforce it.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The SDK implementation of OpenTelemetry::Logs::LoggerProvider.
+      class LoggerProvider < OpenTelemetry::Logs::LoggerProvider
+        attr_reader :resource, :log_record_processors
```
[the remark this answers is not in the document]
fbogsany: It is also not exposed in the spec.
```suggestion
        attr_reader :resource
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        UNEXPECTED_ERROR_MESSAGE = 'unexpected error in ' \
+          'OpenTelemetry::SDK::Logs::LoggerProvider#%s'
```
fbogsany, line 15: I'd expect this constant to be `private`
```suggestion
        UNEXPECTED_ERROR_MESSAGE = 'unexpected error in ' \
          'OpenTelemetry::SDK::Logs::LoggerProvider#%s'

        private_constant :UNEXPECTED_ERROR_MESSAGE

```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+        #
+        # @param [optional Resource] resource The resource to associate with
+        #   new LogRecords created by {Logger}s created by this LoggerProvider.
+        # @param [optional Array] log_record_processors The
+        #   {LogRecordProcessor}s to associate with this LoggerProvider.
+        #
+        # @return [OpenTelemetry::SDK::Logs::LoggerProvider]
+        def initialize(
+          resource: OpenTelemetry::SDK::Resources::Resource.create,
+          log_record_processors: []
```
fbogsany, line 26: Providing the `log_record_processors` at creation time is _MAY_ in the spec. We don't do that in the `MeterProvider` (for `metric_readers`) or the `TracerProvider` (for `span_processors`). Instead, we rely on the `add_...` methods to add them individually. I would start with _not_ exposing it during creation. We can always add it later if that's a common pattern (and then consider adding to `TracerProvider` and `MeterProvider` as well).

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+          @stopped = false
+        end
+
+        # Creates an {OpenTelemetry::SDK::Logs::Logger} instance.
+        #
+        # @param [optional String] name Instrumentation package name
+        # @param [optional String] version Instrumentation package version
+        #
+        # @return [OpenTelemetry::SDK::Logs::Logger]
+        def logger(name = nil, version = nil)
```
fbogsany, line 40: This signature is inconsistent between `TracerProvider` and `MeterProvider`. In the latter, we used:
```ruby
def meter(name, version: nil)
```
so `name` is required and `version` is a keyword arg. IIRC we're going to have to add one or two more args at some point 🔜 , so keywords for optional args are vastly preferred.

We should go back and fix this for `TracerProvider`. I _think_ that can be done in a backwards-compatible fashion.

For `MeterProvider`, we opted to require `name`. If a user explicitly passes `nil` or `""`, we can handle it, but those are both considered invalid - I'd rather have an explicitly invalid required arg than an invalid default arg.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
fbogsany: We use a "registry" in the `MeterProvider` and `TracerProvider` to ensure the same instance is returned for the identical `name` and `version` pairs. I don't see that requirement in the spec for `LoggerProvider`, and maybe it doesn't matter (e.g. if `Logger` has no state). I'm mildly surprised and need to 🤔 a bit more about it.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
fbogsany: It's worth considering whether it should be valid to add log record processors after `@stopped`. In the `TracerProvider`, we warn and don't allow that.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+
+              processor.shutdown(timeout: remaining_timeout)
+            end
+
+            @stopped = true
+            results.max || OpenTelemetry::SDK::Logs::Export::SUCCESS
+          end
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: UNEXPECTED_ERROR_MESSAGE % __method__)
+          Export::FAILURE
```
fbogsany, line 92: In the `TracerProvider` and `MeterProvider` we expect the processors/readers to _not_ `raise` from `#shutdown` or `#force_flush` and instead only return one of the 3 supported result codes.

fbogsany (changes requested):

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: Sounds great. I prefer the keyword args, I just defaulted to what was in TracerProvider since that was already stable. 

Updated in: ca7c15f5bbb2834e339fb723bfe24667fe2035a3

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: I'm happy to remove the arg from the initialize method. I'll double-check about its commonality in other languages. I've also removed the public reader. See: 37befe6f0a5b0d460e368f8fcbf85f31284e384c

logs_sdk/lib/opentelemetry/sdk/logs/export.rb
kaylareopelle: Absolutely. I was being overly specific in the class description with the intent of making it more general when there were more places the constants were used. 

Updated in: 5528277401c8354cfab8ade8a17f4d65b20f7dd9

kaylareopelle: Great point. Updated to be more general in: d63a4e9dc1ce01b6d63914dc7e68a235027048d3

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
kaylareopelle: The "resolved Context" here is coming from the Logs SDK spec description for the method arguments. 

https://opentelemetry.io/docs/specs/otel/logs/sdk/#onemit

The description has a bit more detail to include that might clear things up. I've updated the description to include the rest of the text in: 5528277401c8354cfab8ade8a17f4d65b20f7dd9

kaylareopelle: I'm not sure who does the resolution. I'm also not quite sure how we'll use this argument. 

I'll consult some of the implementations in other languages and come back with more info.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: I have a registry that follows the `MeterProvider` and `TracerProvider` patterns in some code I have queued up for future PRs. I'm happy to bring it into this PR if you'd like to see it now. It became quite necessary once instrumentation was in the mix.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: Understood. Fixed in: a619d36d9bda299e196a7af037595055e8b0d75c

kaylareopelle:
@fbogsany - I'm not able to respond directly to [this comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/1517#discussion_r1362457528), but I've updated `#add_log_record_processor` to not permit adding new processors if `@stopped` is true and emit a warning.

See: 336ca7dd1b17ed4c298b601e96481531adda04c7

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
kaylareopelle: @fbogsany  - I can't find any examples where the `context` argument is used in other languages. 
* Python doesn't include it anywhere. 
* PHP and Java have it as an argument but do not use it in the `onEmit` functions. 
* Javascript doesn't include it as an argument in their Simple or Batch processors, but the duck type includes it as an argument.

So I don't have a great answer about who does the resolution. I didn't end up needing this argument in the Simple or Batch processors because I used the SpanContext to pull `trace_id`, `span_id`, and `trace_flags`. Those are the only context-related attributes in the [Logs Data Model](https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition).

Do you have a sense of what approach you'd prefer Ruby take?

logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
kaylareopelle: This `attr_reader` call has been removed. I'm not sure why I initially broke from `Tracer` and `Meter`.

kaylareopelle:
@fbogsany  - Thank you for your thorough feedback. I believe I've addressed all your comments. Please re-review when you can.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+          @log_record_processors = []
+          @mutex = Mutex.new
+          @resource = resource
+          @stopped = false
+        end
+
+        # Creates an {OpenTelemetry::SDK::Logs::Logger} instance.
+        #
+        # @param [String] name Instrumentation package name
+        # @param [optional String] version Instrumentation package version
```
olleolleolle, line 35: Minor: [YARD documentation's type parser](https://yardoc.org/types.html) does not support the "optional" marker in the type description. Perhaps the word can be moved to the descriptive string?

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+        # Attempts to stop all the activity for this LoggerProvider. Calls
+        # {LogRecordProcessor#shutdown} for all registered {LogRecordProcessor}s.
+        #
+        # This operation may block until all log records are processed. Must
+        # be called before turning off the main application to ensure all data
+        # are processed and exported.
+        #
+        # After this is called all newly created {LogRecord}s will be no-op.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
```
olleolleolle, line 74:
```suggestion
        # @param [Numeric] timeout An optional timeout in seconds.
```

(Minor, same as earlier note on YARD tags.)

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        # Immediately export all {LogRecord}s that have not yet been exported
+        # for all the registered {LogRecordProcessor}s.
+        #
+        # This method should only be called in cases where it is absolutely
+        # necessary, such as when using some FaaS providers that may suspend
+        # the process after an invocation, but before the {LogRecordProcessor}
+        # exports the completed {LogRecord}s.
+        #
+        # @param [optional Numeric] timeout An optional timeout in seconds.
```
olleolleolle, line 105:
```suggestion
        # @param [Numeric] timeout An optional timeout in seconds.
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle:
```suggestion
        # @param [String] version Optional instrumentation package version
```

kaylareopelle: Thanks for the catch, @olleolleolle! This type parser is new to me. Out of curiosity, what happens if the type parser fails? I was able to generate docs that display the type with "optional" present.

For example: Screenshot 2023-11-07 at 11 33 44 AM

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # LogRecordProcessor describes a duck type and provides synchronous no-op hooks for when a
+      # {LogRecord} is started or when a {LogRecord} is ended. It is not required to subclass this
```
fbogsany, line 11:
```suggestion
      # LogRecordProcessor describes a duck type and provides a synchronous no-op hook for when a
      # {LogRecord} is emitted. It is not required to subclass this
```

logs_sdk/lib/opentelemetry/sdk/logs/log_record_processor.rb
```diff
@@ -0,0 +1,47 @@
...
+      # class to provide an implementation of LogRecordProcessor, provided the interface is
+      # satisfied.
+      class LogRecordProcessor
+        # Called when a {LogRecord} is emitted. Subsequent calls are not
+        # permitted after shutdown is called.
+        # @param [LogRecord] log_record The emitted {LogRecord}
+        # @param [Context] context The resolved Context (the explicitly passed Context or the current Context)
```
fbogsany, line 18:
```suggestion
        # @param [Context] context The {Context}
```
The additional text for this parameter doesn't provide additional clarity (I don't know what it means in the context of this API).

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+        #
+        # @return [OpenTelemetry::SDK::Logs::LoggerProvider]
+        def initialize(resource: OpenTelemetry::SDK::Resources::Resource.create)
+          @log_record_processors = []
+          @mutex = Mutex.new
+          @resource = resource
+          @stopped = false
+        end
+
+        # Creates an {OpenTelemetry::SDK::Logs::Logger} instance.
```
fbogsany, line 32:
```suggestion
        # Returns an {OpenTelemetry::SDK::Logs::Logger} instance.
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
fbogsany: Spec says:
> When more than one Logger of the same `name`, `version`, and `schema_url` is created, it is unspecified whether or under which conditions the same or different Logger instances are returned.

So there's some flexibility here. We'll need to implement equality in `Logger` if we choose not to implement a registry.

fbogsany: Do we need to fully qualify the class here?
```suggestion
         Logger.new(name, version, self)
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+        # Attempts to stop all the activity for this LoggerProvider. Calls
+        # {LogRecordProcessor#shutdown} for all registered {LogRecordProcessor}s.
+        #
+        # This operation may block until all log records are processed. Must
+        # be called before turning off the main application to ensure all data
+        # are processed and exported.
+        #
+        # After this is called all newly created {LogRecord}s will be no-op.
+        #
+        # @param [Numeric] timeout An optional timeout in seconds.
```
fbogsany, line 74:
```suggestion
        # @param [optional Numeric] timeout An optional timeout in seconds.
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+        # @return [Integer] Export::SUCCESS if no error occurred, Export::FAILURE if
+        #   a non-specific failure occurred, Export::TIMEOUT if a timeout occurred.
+        def shutdown(timeout: nil)
+          @mutex.synchronize do
+            if @stopped
+              OpenTelemetry.logger.warn('LoggerProvider#shutdown called multiple times.')
+              return OpenTelemetry::SDK::Logs::Export::FAILURE
```
fbogsany, line 81: Does this need to be fully qualified?
```suggestion
              return Export::FAILURE
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+            end
+
+            start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
+            results = @log_record_processors.map do |processor|
+              remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+              break [OpenTelemetry::SDK::Logs::Export::TIMEOUT] if remaining_timeout&.zero?
```
fbogsany, line 87:
```suggestion
              break [Export::TIMEOUT] if remaining_timeout&.zero?
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+
+              processor.shutdown(timeout: remaining_timeout)
+            end
+
+            @stopped = true
+            results.max || OpenTelemetry::SDK::Logs::Export::SUCCESS
```
fbogsany, line 93:
```suggestion
            results.max || Export::SUCCESS
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+
+        # Immediately export all {LogRecord}s that have not yet been exported
+        # for all the registered {LogRecordProcessor}s.
+        #
+        # This method should only be called in cases where it is absolutely
+        # necessary, such as when using some FaaS providers that may suspend
+        # the process after an invocation, but before the {LogRecordProcessor}
+        # exports the completed {LogRecord}s.
+        #
+        # @param [Numeric] timeout An optional timeout in seconds.
```
fbogsany, line 105:
```suggestion
        # @param [optional Numeric] timeout An optional timeout in seconds.
```

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: Awaiting discussion on [L21](https://github.com/open-telemetry/opentelemetry-ruby/pull/1517/files#diff-f9a2c98dbf5b32983909ec3963c7cf01517197a75482ba1f27ed1151ab5be862R21)

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -0,0 +1,126 @@
...
+        attr_reader :resource
+
+        UNEXPECTED_ERROR_MESSAGE = 'unexpected error in ' \
+          'OpenTelemetry::SDK::Logs::LoggerProvider#%s'
+
+        private_constant :UNEXPECTED_ERROR_MESSAGE
+
+        # Returns a new LoggerProvider instance.
+        #
+        # @param [Resource] resource An optional resource to associate with
```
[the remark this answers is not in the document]
kaylareopelle: I'm going to commit these suggestions for now and talk about the YARD documentation type parser in a separate issue.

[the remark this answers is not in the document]
kaylareopelle: New issue: https://github.com/open-telemetry/opentelemetry-ruby/issues/1553

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: Good point!

I received requests from one of opentelemetry-ruby's maintainers to revert the changes that move `optional` into the description and out of the type declaration. 

This project has a consistent pattern of using `[optional Type]` in the argument documentation. Rather than make the change in this PR, we should consider updating the standard for the entire project in one go. I've created https://github.com/open-telemetry/opentelemetry-ruby/issues/1553 for this discussion.

kaylareopelle:
@fbogsany, checking back in here! Have your concerns been addressed?

fbogsany (approved):

PR #1522, opened 2023-09-13
fix: Small perf improvement to generate_r

fbogsany:
This is a small performance improvement to the `generate_r` method for the consistent-probability sampler. This method generates a `r` value from the `trace_id`. The current implementation unnecessarily allocates a slice of the `trace_id` string. Using the `offset:` argument to `unpack1`, we can avoid this allocation.

Benchmark results:
```
          generate_r      3.920M (± 0.4%) i/s -     19.654M
 generate_r_in_place      4.596M (± 0.5%) i/s -     23.085M

Comparison:
 generate_r_in_place:  4595613.7 i/s
          generate_r:  3920215.1 i/s - 1.17x slower
```

### DIFF
--- a/sdk_experimental/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
+++ b/sdk_experimental/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
@@ -106,7 +106,7 @@ module OpenTelemetry
           end
 
           def generate_r(trace_id)
-            x = trace_id[8, 8].unpack1('Q>') | 0x3
+            x = trace_id.unpack1('Q>', offset: 8) | 0x3
             64 - x.bit_length
           end
         end
--- a/api/benchmarks/id_generation_bench.rb
+++ b/api/benchmarks/id_generation_bench.rb
@@ -35,6 +35,16 @@ def generate_span_id_while
   id
 end
 
+def generate_r(trace_id)
+  x = trace_id[8, 8].unpack1('Q>') | 0x3
+  64 - x.bit_length
+end
+
+def generate_r_in_place(trace_id)
+  x = trace_id.unpack1('Q>', offset: 8) | 0x3
+  64 - x.bit_length
+end
+
 Benchmark.ipsa do |x|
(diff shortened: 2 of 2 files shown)


robertlaurin (approved):

sdk_experimental/lib/opentelemetry/sdk/trace/samplers/consistent_probability_tracestate.rb
```diff
@@ -106,7 +106,7 @@ def decimal(str)
           end
 
           def generate_r(trace_id)
-            x = trace_id[8, 8].unpack1('Q>') | 0x3
+            x = trace_id.unpack1('Q>', offset: 8) | 0x3
```
casperisfine, line 109: Not sure if you are aware but `offset` was added in 3.1, and IIRC otel is supposed to be compatible with Ruby 3.0?

But since the offset is static here, you can use the `@` syntax:

```ruby
>> "abcedfghsdfdsfhdsfjsdkjfdsjfkdlsfjdsfjksdfk".unpack1('Q>', offset: 8)
=> 8314883393651632228
>> "abcedfghsdfdsfhdsfjsdkjfdsjfkdlsfjdsfjksdfk".unpack1('@8Q>')
=> 8314883393651632228
```

`offset` was added to avoid allocating a new pattern string every time when the offset is dynamic.

casperisfine, line 109: Not sure if you are aware but `offset` was added in 3.1, and IIRC otel is supposed to be compatible with Ruby 3.0?

But since the offset is static here, you can use the `@` syntax:

```ruby
>> "abcedfghsdfdsfhdsfjsdkjfdsjfkdlsfjdsfjksdfk".unpack1('Q>', offset: 8)
=> 8314883393651632228
>> "abcedfghsdfdsfhdsfjsdkjfdsjfkdlsfjdsfjksdfk".unpack1('@8Q>')
=> 8314883393651632228
```

`offset` was added to avoid allocating a new pattern string every time when the offset is dynamic.

PR #1535, opened 2023-10-27
Assert correct elements are removed from hashes larger than limit.

ptolts:
Adds missing test coverage for Tracestates larger than 32.

### DIFF
--- a/api/test/opentelemetry/trace/tracestate_test.rb
+++ b/api/test/opentelemetry/trace/tracestate_test.rb
@@ -74,9 +74,12 @@ describe OpenTelemetry::Trace::Tracestate do
       tracestate = OpenTelemetry::Trace::Tracestate.from_hash(h)
       _(tracestate.to_h).must_equal('the-good-key' => '  the-good-value', '0x_the-vendor/key@my*v3nd0r' => 'another-great-value')
     end
-    it 'will not exceed 32 members' do
-      tracestate = OpenTelemetry::Trace::Tracestate.from_hash(33.times.collect { |n| [n.to_s, n.to_s] }.to_h)
-      _(tracestate.to_h.size).must_be :<=, 32
+    it 'will not exceed 32 members and drops excess from the beginning' do
+      input_hash = 33.times.collect { [('a'..'z').to_a.sample(5).join, ('a'..'z').to_a.sample(5).join] }.to_h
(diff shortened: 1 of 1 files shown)


fbogsany (changes requested):
This is technically incorrect. From https://github.com/open-telemetry/opentelemetry-ruby/pull/1533#issuecomment-1783535817

> Followup: the [W3C `trace-context` spec](https://www.w3.org/TR/trace-context/#tracestate-limits) states:
> 
> > In a situation where `tracestate` needs to be truncated due to size limitations, the vendor _MUST_ truncate whole entries. Entries larger than 128 characters long _SHOULD_ be removed first. Then entries _SHOULD_ be removed starting from the end of `tracestate`. Note that other truncation strategies like safe list entries, blocked list entries, or size-based truncation _MAY_ be used, but are highly discouraged. Those strategies decrease the interoperability of various tracing vendors.

Our implementation is wrong and we should remove entries from the end rather than the start (and, of course, we should remove entries > 128 characters first).

ptolts:
I'll close, thanks Bogs.

fbogsany:
@ptolts could you please click the link in https://github.com/open-telemetry/opentelemetry-ruby/pull/1535#issuecomment-1783522860 to authorize the CLA?

api/test/opentelemetry/trace/tracestate_test.rb
```diff
@@ -74,9 +74,12 @@
...
     end
-    it 'will not exceed 32 members' do
-      tracestate = OpenTelemetry::Trace::Tracestate.from_hash(33.times.collect { |n| [n.to_s, n.to_s] }.to_h)
-      _(tracestate.to_h.size).must_be :<=, 32
+    it 'will not exceed 32 members and drops excess from the beginning' do
+      input_hash = 33.times.collect { [('a'..'z').to_a.sample(5).join, ('a'..'z').to_a.sample(5).join] }.to_h
+      tracestate = OpenTelemetry::Trace::Tracestate.from_hash(input_hash)
+      output_hash = tracestate.to_h
+      _(output_hash.size).must_be :<=, 32
+      _(output_hash).must_equal(input_hash.drop(1).to_h)
```
fbogsany, line 82:
```suggestion
      _(output_hash).must_equal(input_hash.first(32).to_h)
```

kaylareopelle:
Hi, @ptolts! Checking in on this PR. Would you be open to signing the CLA?

PR #1536, opened 2023-10-31
fix: spec compliance for span attribute limit

### LINKED ISSUE #1534: Attribute limits should discard from the end
**Description of the bug**

See https://github.com/open-telemetry/opentelemetry-ruby/pull/1533#issuecomment-1783501206 for background. https://github.com/open-telemetry/opentelemetry-specification/pull/1130 added span limits to the spec. A key requirement is:
> - set a limit of unique attribute keys such that:
>  - for each unique attribute key, addition of which would result in exceeding
>    the limit, SDK MUST discard that key/value pair.

Our implementation of span limits instead drops attributes added earlier when the attribute count limit is exceeded: https://github.com/open-telemetry/opentelemetry-ruby/blob/afc7bc56b39adae925ebd465948a722bd64e1d2e/sdk/lib/opentelemetry/sdk/trace/span.rb#L347-L354

E.g. if `ENV['OTEL_ATTRIBUTE_COUNT_LIMIT'] == '2'` then the following code should produce a span with attributes `{"a" => 1, "b" => 2}` but instead it contains `{"b" => 2, "c" => 3}`:
```ruby
tracer.in_span('bad limit', attributes: {"a" => 1, "b" => 2}) do |span|
  span["c"] = 3
end
```

```rb
ENV['OTEL_ATTRIBUTE_COUNT_LIMIT'] = '2'

require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'

span_processor = OpenTelemetry::SDK::Trace::Export::SimpleSpanProcessor.new(
  OpenTelemetry::SDK::Trace::Export::ConsoleSpanExporter.new
)

  c.add_span_processor(span_processor)
end

SimpleTracer = OpenTelemetry.tracer_provider.tracer('Bug Report')

SimpleTracer.in_span('bad limit', attributes: {"a" => 1, "b" => 2}) do |span|
  span["c"] = 3
end

OpenTelemetry.tracer_provider.shutdown
```

fbogsany:
Fixes #1534

Note: this does not address Event or Link attribute limits (and, indeed, Event and Link limits), at least some of which are similarly non-compliant. I think we should refactor `trim_links` and `append_event` to use the `trim_span_attributes` method, but that needs to be changed to accept the limit as a parameter. I'd prefer to do that in a subsequent PR.

As a side-effect, trimming span attributes is faster. The old `set_attribute` is about 14% slower than the new code in simple benchmarks, and the old `trim_span_attributes` is about 28% slower than the new implementation.

### DIFF
--- a/sdk/test/opentelemetry/sdk/trace/span_test.rb
+++ b/sdk/test/opentelemetry/sdk/trace/span_test.rb
@@ -60,10 +60,10 @@ describe OpenTelemetry::SDK::Trace::Span do
       _(span.attributes).must_equal('foo' => 'bar')
     end
 
-    it 'trims the oldest attribute' do
+    it 'trims the newest attribute' do
       span.set_attribute('old', 'oldbar')
       span.set_attribute('foo', 'bar')
-      _(span.attributes).must_equal('foo' => 'bar')
+      _(span.attributes).must_equal('old' => 'oldbar')
     end
 
     it 'truncates attribute value length based if configured' do
@@ -138,10 +138,10 @@ describe OpenTelemetry::SDK::Trace::Span do
       _(span.attributes).must_equal('foo' => 'bar')
     end
 
-    it 'trims the oldest attributes' do
+    it 'trims the newest attributes' do
       span.add_attributes('old' => 'oldbar')
       span.add_attributes('foo' => 'bar', 'bar' => 'baz')
-      _(span.attributes).must_equal('bar' => 'baz')
+      _(span.attributes).must_equal('old' => 'oldbar')
     end
 
     it 'truncates attribute value length based if configured' do
--- a/sdk/test/opentelemetry/sdk/trace/span_test.rb
+++ b/sdk/test/opentelemetry/sdk/trace/span_test.rb
@@ -499,22 +499,23 @@ describe OpenTelemetry::SDK::Trace::Span do
     end
 
     it 'trims excess attributes' do
-      attributes = { 'foo': 'bar', 'other': 'attr' }
+      attributes = { 'foo' => 'bar', 'other' => 'attr' }
       span = Span.new(context, Context.empty, OpenTelemetry::Trace::Span::INVALID, 'name', SpanKind::INTERNAL, nil, span_limits,
                       [], attributes, nil, Time.now, nil, nil)
       _(span.to_span_data.total_recorded_attributes).must_equal(2)
       _(span.attributes.length).must_equal(1)
+      _(span.attributes).must_equal('foo' => 'bar')
     end
 
     it 'truncates attributes if configured' do
-      attributes = { 'foo': 'oldbaroldbaroldbaroldbaroldbaroldbar' }
+      attributes = { 'foo' => 'oldbaroldbaroldbaroldbaroldbaroldbar' }
       span = Span.new(context, Context.empty, OpenTelemetry::Trace::Span::INVALID, 'name', SpanKind::INTERNAL, nil, span_limits,
                       [], attributes, nil, Time.now, nil, nil)
-      _(span.attributes[:foo]).must_equal('oldbaroldbaroldbaroldbaroldba...')
+      _(span.attributes["foo"]).must_equal('oldbaroldbaroldbaroldbaroldba...')
     end
 
     it 'counts attributes' do
-      attributes = { 'foo': 'bar', 'other': 'attr' }
+      attributes = { 'foo' => 'bar', 'other' => 'attr' }
       span = Span.new(context, Context.empty, OpenTelemetry::Trace::Span::INVALID, 'name', SpanKind::INTERNAL, nil, span_limits,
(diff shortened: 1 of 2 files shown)


sdk/test/opentelemetry/sdk/trace/span_test.rb
```diff
@@ -60,10 +60,10 @@
       _(span.attributes).must_equal('foo' => 'bar')
     end
 
-    it 'trims the oldest attribute' do
+    it 'trims the newest attribute' do
```
arielvalentin, line 63: I believe this also happens during construction. Bonus points for adding test coverage there.

arielvalentin (approved):
Since its late and has been a long day I am having trouble groking this statement:

> set a limit of unique attribute keys such that:
> for each unique attribute key, addition of which would result in exceeding the limit, SDK MUST discard that key/value pair. https://opentelemetry.io/docs/specs/otel/common/#attribute-limits

If the unique combination of key value pairs is present, then this seems to imply the implementation is not a Hash or Dictionary? How would this end up exceeding the limit? Is it assuming the k/v is an Ordered List and then reduced into a Dictionary at export time? Seems odd to me. 

Anyway, a curious behavior for hashes is when merged, the hash preserves the insertion order of the receiver:

```console
irb(main):001> h = { a: 1, b: 2}
=> {:a=>1, :b=>2}
irb(main):002> h.merge!(a: 5)
=> {:a=>5, :b=>2}
irb(main):003> h
=> {:a=>5, :b=>2}
irb(main):004> h.merge!(b:8, a: 5)
=> {:a=>5, :b=>8}
irb(main):005> y = { b: 1,  a: 3 }
=> {:b=>1, :a=>3}
irb(main):007> y.merge!(h)
=> {:b=>8, :a=>5}
```

Does the spec imply that adding a new entry to the hash means its replacement should be treated as an entry at the end?

Then that would mean an atomic update where the hash should have the duplicate entries deleted first and then the replacements added after in order to keep them younger?

fbogsany:
> Since its late and has been a long day I am having trouble groking this statement:
> 
> > set a limit of unique attribute keys such that:
> > for each unique attribute key, addition of which would result in exceeding the limit, SDK MUST discard that key/value pair. https://opentelemetry.io/docs/specs/otel/common/#attribute-limits
> 
> If the unique combination of key value pairs is present, then this seems to imply the implementation is not a Hash or Dictionary? How would this end up exceeding the limit? Is it assuming the k/v is an Ordered List and then reduced into a Dictionary at export time? Seems odd to me.

We only care about uniqueness of keys, not key-value pairs. If a key already exists as a span attribute, then it is within the limit - overwriting it is permitted without changing ordering. If a key does not already exist as a span attribute and adding it would exceed the limit, we must drop the key (and therefore the value as well).

> 
> Anyway, a curious behavior for hashes is when merged, the hash preserves the insertion order of the receiver:
> 
> ```
> irb(main):001> h = { a: 1, b: 2}
> => {:a=>1, :b=>2}
> irb(main):002> h.merge!(a: 5)
> => {:a=>5, :b=>2}
> irb(main):003> h
> => {:a=>5, :b=>2}
> irb(main):004> h.merge!(b:8, a: 5)
> => {:a=>5, :b=>8}
> irb(main):005> y = { b: 1,  a: 3 }
> => {:b=>1, :a=>3}
> irb(main):007> y.merge!(h)
> => {:b=>8, :a=>5}
> ```
> 
> Does the spec imply that adding a new entry to the hash means its replacement should be treated as an entry at the end?

No. _New_ keys should be added at the end (which they are). Existing keys can have their values overwritten in place.

> 
> Then that would mean an atomic update where the hash should have the duplicate entries deleted first and then the replacements added after in order to keep them younger?

No, we don't need to do that (although updates are atomic regardless).

fbogsany:
For historical interest, when implementing limits and validity checking, we chose to apply limits eagerly (so as to not retain memory unnecessarily) and check validity lazily (so as to not check validity of something we might overwrite or drop). This choice means we _might_ drop something due to limits that would have been retained if something else was dropped due to invalidity.

sdk/test/opentelemetry/sdk/trace/span_test.rb
fbogsany: We have a test, but it only asserts the length rather than the contents: https://github.com/open-telemetry/opentelemetry-ruby/blob/dcda280c9e919d66038a452b29c268434a4138c6/sdk/test/opentelemetry/sdk/trace/span_test.rb#L501-L508
I'll add an assertion for the contents.

fbogsany: https://github.com/open-telemetry/opentelemetry-ruby/pull/1536/commits/ba0eacca7512f2b3cd424a6e2dcb092fcd1672e2

PR #1537, opened 2023-10-31
BatchSpanProcessor#force_flush: purge inherited spans even on shutdown

casperisfine:
Ref: https://github.com/open-telemetry/opentelemetry-ruby/pull/363
Ref: https://github.com/Shopify/pitchfork/pull/74

At Shopify we're using the Pitchfork web server, and one thing it does is that it sometimes do a "double fork", similar to the classic daemonization dance.

Recently we noticed that sometimes the "middle process" of that daemonization dance would get "stuck". After adding some debug we figured it's because we have:

```ruby
at_exit do
  # Triggers flushing of buffer on shutdown
  OpenTelemetry.tracer_provider.shutdown(timeout: 60)
end
```

So the child process inherits lots of spans from its parents, and normally OpenTelemetry deal gracefully with that by droping the inherited spans.

But here because we called `shutdown`, `@keep_running` is set to `false` hence `reset_on_fork` is not called.

I don't understand why we shouldn't call it during shutdown though, aside from avoiding to respawn the thread.

With this change we expect our "middle process" to drop all the inherited spans and to have nothing to flush.

@fbogsany @robertlaurin
cc @DazWorrall @bmansoob

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/span.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/span.rb
@@ -347,8 +347,14 @@ module OpenTelemetry
         def trim_span_attributes(attrs)
           return if attrs.nil?
 
-          excess = attrs.size - @span_limits.attribute_count_limit
-          excess.times { attrs.shift } if excess.positive?
+          if attrs.size > @span_limits.attribute_count_limit
+            n = @span_limits.attribute_count_limit
+            attrs.delete_if do |_key, _value|
+              n -= 1
+              n.negative?
+            end
+          end
+
           truncate_attribute_values(attrs, @span_limits.attribute_length_limit)
           nil
         end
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -103,7 +103,7 @@ def on_finish(span)
           def force_flush(timeout: nil) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity, Metrics/MethodLength
```
fbogsany, line 103:
```suggestion
          def force_flush(timeout: nil) # rubocop:disable Metrics/MethodLength
```

fbogsany (approved):

PR #1548, opened 2023-12-05
feat: Update metrics configuration patch

kaylareopelle:
This PR updates the metrics configuration patch to match the functionality for traces in the `OpenTelemetry::SDK::Configurator`. 

Once this PR is merged, the metric exporter can be automatically configured using the `OTLP_METRICS_EXPORTER` environment variable. 

The environment variable has options for:
* otlp (default)
* console
* in-memory (pull exporter)
* none

Like the Traces exporter, more than one exporter can be configured.

This draws a lot from the structure in: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/configurator.rb

It intends to follow the specification: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md?plain=1#L256


kaylareopelle:
Blocked by #1551

metrics_sdk/lib/opentelemetry/sdk/metrics/configuration_patch.rb
```diff
@@ -10,12 +10,53 @@ module Metrics
...
+          exporters.split(',').map do |exporter|
+            case exporter.strip
+            when 'none' then nil
+            when 'console'
+              OpenTelemetry.meter_provider.add_metric_reader(Metrics::Export::ConsoleMetricPullExporter.new)
+            when 'in-memory'
+              OpenTelemetry.meter_provider.add_metric_reader(Metrics::Export::InMemoryMetricPullExporter.new)
+            when 'otlp'
+              begin
+                OpenTelemetry.meter_provider.add_metric_reader(OpenTelemetry::Exporter::OTLP::MetricsExporter.new)
```
kaylareopelle, line 48: Check to make sure this is still the right constant name

mwear (approved):

metrics_sdk/lib/opentelemetry/sdk/metrics/configuration_patch.rb
```diff
@@ -10,12 +10,57 @@ module Metrics
...
+          exporters.split(',').map do |exporter|
+            case exporter.strip
+            when 'none' then nil
+            when 'console'
+              OpenTelemetry.meter_provider.add_metric_reader(
+                Metrics::Export::PeriodicMetricReader.new(
+                  exporter: Metrics::Export::ConsoleMetricPullExporter.new
+                )
+              )
+            when 'otlp'
```
kaylareopelle, line 46: @mwear @xuan-cao-swi - [The specs state that OTLP should be the default exporter or `OTEL_METRICS_EXPORTER`](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#exporter-selection), so I updated the tests to reflect this and installed the OTLP exporter in the test group of the Gemile.

However, the OTLP exporter brings in the protobuf dependencies, making the entire metrics_sdk test suite incompatible with JRuby. 

It seems like the tracing test suite gets around this by avoiding any mention of the OTLP exporter in the tests, even though the spec also expects OTLP to be the default exporter for that signal (the tests use zipkin instead).

I don't want to skip JRuby for all tests in the metrics_sdk. 

My current thoughts for solutions are: 
1. conditionally install the OTLP exporter in the test environment if JRuby is present and skip the related tests if JRuby is present 
2. omit the OTLP exporter from the tests entirely, and bring in an alternate exporter for the tests, following the pattern for traces (in-memory metrics exporter seems like the best candidate, even though it's not part of the spec)
3. something else...? 

What do y'all think would be the best next step?

kaylareopelle: We talked about this during the SIG today. 

For testing, use the in-memory exporter when you can.
For this particular case, where it's testing this default, just skip that on JRuby.

mwear (approved):

PR #1556, opened 2023-12-21
feat: Add support for mutual TLS.

megabus-tobin:
Add support for specifying client certificate and key for mutual TLS when exporting.

Also log more information about why an SSL error occured as this can be very difficult to diagnose. This is logged in the same way as other probably fatal issues are, e.g. HTTP 404s.


arielvalentin:
@megabus-tobin Thank you for your contribution!

Please consider separating the error handling changes in a separate PR. We will be able to review and release that change independent of verifying the correctness of the changes required for mTLS support.

Happy Holidays!

megabus-tobin:
@arielvalentin I've split the small logging change off into #1557

kaylareopelle (commented):
Thank you, @megabus-tobin! 

For anyone who's interested, the specs for these env vars can be found here: https://opentelemetry.io/docs/specs/otel/protocol/exporter/ 

Could you incorporate the new environment variables into the exporter tests? [(here's the file for the OTLP library)](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb)

Anywhere the `certificate_file` is mentioned may be a good place to add content for the client certificate and key.

megabus-tobin:
> Could you incorporate the new environment variables into the exporter tests? [(here's the file for the OTLP library)](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb)
> 
> Anywhere the `certificate_file` is mentioned may be a good place to add content for the client certificate and key.

The `Net:HTTP` class makes this a bit harder than it should be. Unlike `ca_cert`, the `cert` and `key` members aren't strings, they're `OpenSSL::X509::Certificate` and `OpenSSL::PKey::RSA` objects.

I'm not sure what the best approach would be. Should I add a test certificate and key to the repository or refactor `OpenTelemetry::Exporter::OTLP::Exporter.initialize` to optionally take the object directly and fallback to trying to load them from the environment?

The latter would be very similar to what `OpenTelemetry::Exporter::OTLP::Exporter.prepare_endpoint` already does. It would also have the small advantage of allowing for more exotic ways of obtaining the certificate and key but the utility of that is questionable since you need a file for the CA cert anyway.

kaylareopelle:
> The Net:HTTP class makes this a bit harder than it should be. Unlike ca_cert, the cert and key members aren't strings, they're OpenSSL::X509::Certificate and OpenSSL::PKey::RSA objects.
>
> I'm not sure what the best approach would be. Should I add a test certificate and key to the repository or refactor OpenTelemetry::Exporter::OTLP::Exporter.initialize to optionally take the object directly and fallback to trying to load them from the environment?
>
> The latter would be very similar to what OpenTelemetry::Exporter::OTLP::Exporter.prepare_endpoint already does. It would also have the small advantage of allowing for more exotic ways of obtaining the certificate and key but the utility of that is questionable since you need a file for the CA cert anyway.

Sorry for my delated response. Great questions! I think adding the test certificate and key seems the least intrusive to what you've already built if that's not too much effort. If other approvers/maintainers disagree, please chime in!

I like your idea about following the pattern in `OpenTelemetry::Exporter::OTLP::Exporter.prepare_endpoint`, but like you mentioned, I'm not sure about the utility of that approach.

megabus-tobin:
@kaylareopelle I have updated the tests to check that the client certificate and key are handled correctly.

Some small notes:
1. To test that explicit TLS settings are preferred to environment variables I needed a second cert/key that would be loaded in the unlikely event the test was going to fail.
2. Initially I generated the cert/key on the fly but since a path is needed this would have made it necessary to ensure a writeable filesystem when running tests. Instead I have put the test certs/keys into the same directory as the tests and referenced that via `__FILE__`.
3. The certs/keys I generated are obviously meaningless. E.g. `CN = a.example.tld`, `O = Internet Widgits Pty Ltd`. By the time this gets reviewed they will probably also be expired, which the tests don't care about.
4. Rather than reading the certs/keys at the point of comparison, I've done it once near the top of the file. Apart from avoiding a little repetition this makes it much easier to spot if the files got removed, etc. rather than an individual test appearing to fail.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -107,11 +109,13 @@ def fetch_ssl_verify_mode
             end
           end
 
-          def http_connection(uri, ssl_verify_mode, certificate_file)
+          def http_connection(uri, ssl_verify_mode, certificate_file, client_certificate_file, client_key_file)
             http = Net::HTTP.new(uri.host, uri.port)
             http.use_ssl = uri.scheme == 'https'
             http.verify_mode = ssl_verify_mode
             http.ca_file = certificate_file unless certificate_file.nil?
+            http.cert = OpenSSL::X509::Certificate.new(File.read(client_certificate_file)) unless client_certificate_file.nil?
```
szechyjs, line 117: Especially when running in containers, and using environment variables, it would be nice to be able to load the PEM encoded cert/key directly from environment variables, instead of having to find a way to write it to a file first.

szechyjs:
Would be nice to update the gRPC exporter as well.

```rb
creds = if mtls_config_provided
  GRPC::Core::ChannelCredentials.new(ca, key, cert)
else
  :this_channel_is_insecure
end

@client = Opentelemetry::Proto::Collector::Trace::V1::TraceService::Stub.new(
  "#{uri.host}:#{uri.port}",
  creds
)
```

https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-grpc/lib/opentelemetry/exporter/otlp/grpc/trace_exporter.rb#L34

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
megabus-tobin: I agree but at least one file will still be needed for many/most users because it's assigned to `Net::HTTP#ca_file` which can only be a path. Should I make the change to support either for the client cert and key despite that?

kaylareopelle:
> @kaylareopelle I have updated the tests to check that the client certificate and key are handled correctly.

@megabus-tobin, thank you for updating the tests! I think this is a solid approach.

Are you planning to make the [GRPC changes mentioned here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1556#issuecomment-2096908236) as part of this PR?

megabus-tobin:
@kaylareopelle 
> Are you planning to make the [GRPC changes mentioned here](https://github.com/open-telemetry/opentelemetry-ruby/pull/1556#issuecomment-2096908236) as part of this PR?

I have made these changes now. I've only performed minimal testing on this as it's not something I've needed to use previously and am not familiar with it. Unlike the HTTP exporter, there are almost no existing tests included in the library.

exporter/otlp-http/README.md
```diff
@@ -68,15 +68,17 @@ For additional examples, see the [examples on github][examples-github].
...
+| Parameter                 | Environment variable                         | Default                             |
+|---------------------------|----------------------------------------------|-------------------------------------|
+| `endpoint:`               | `OTEL_EXPORTER_OTLP_ENDPOINT`                | `"http://localhost:4318/v1/traces"` |
+| `certificate_file:`       | `OTEL_EXPORTER_OTLP_CERTIFICATE`             |                                     |
+| `client_certificate_file` | `OTEL_EXPORTER_OTLP_CLIENT_CERTIFICATE`      |                                     |
+| `client_key_file`         | `OTEL_EXPORTER_OTLP_CLIENT_KEY`              |                                     |
+| `headers:`                | `OTEL_EXPORTER_OTLP_HEADERS`                 |                                     |
+| `compression:`            | `OTEL_EXPORTER_OTLP_COMPRESSION`             | `"gzip"`                            |
+| `timeout:`                | `OTEL_EXPORTER_OTLP_TIMEOUT`                 | `10`                                |
+| `ssl_verify_mode:`        | `OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_PEER` or | `OpenSSL::SSL:VERIFY_PEER`          |
```
arielvalentin, line 80: is there a constraint that we would like to put on disabling peer verification when using mTLS?

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
arielvalentin: I think that would make it inconsistent with other SDK implementations which all use file paths. 

I think that if we have a request to do this we may want to steer folks in a direction to generate a file via a configMap or equivalent.

exporter/otlp-http/README.md
megabus-tobin: I've found it useful to still be able to disable host verification when testing as the certs don't always match the servers. As they're often just localhost or were determined by an overly dynamic DNS.

In our internal use case the certificate verification is _sometimes_ only actually useful from the server's perspective, which is actually our internal use case: a publicly exposed collector behind a reverse proxy that verifies the client using a certificate we know it must already have. Admittedly in that case we do know hostname matches but in our earlier iterations of the setup it didn't and caused problem with some of the OTEL SDKs for other languages.

kaylareopelle:
We chatted about this PR in the SIG last week. A question was raised about whether other languages also have this feature. 

The environment variables are referenced in [the protocol exporter specification](https://opentelemetry.io/docs/specs/otel/protocol/exporter/#:~:text=Certificate%20File%3A%20The,OTEL_EXPORTER_OTLP_TRACES_CLIENT_CERTIFICATE%20OTEL_EXPORTER_OTLP_METRICS_CLIENT_CERTIFICATE%20OTEL_EXPORTER_OTLP_LOGS_CLIENT_CERTIFICATE). 

They were added in this PR: https://github.com/open-telemetry/opentelemetry-specification/pull/2370 

The `OTEL_EXPORTER_OTLP_CLIENT_KEY` has been added to Go, PHP, and JS [(search)](https://github.com/search?q=org%3Aopen-telemetry+OTEL_EXPORTER_OTLP_CLIENT_KEY&type=code&p=2)

cc @mwear

megabus-tobin:
> The `OTEL_EXPORTER_OTLP_CLIENT_KEY` has been added to Go, PHP, and JS [(search)](https://github.com/search?q=org%3Aopen-telemetry+OTEL_EXPORTER_OTLP_CLIENT_KEY&type=code&p=2)

It's available in the Java SDK too but is configured differently (because Java). See https://opentelemetry.io/docs/languages/java/configuration/ and in particular the `otel.exporter.otlp.client.key` system property.

kaylareopelle (approved):
Thanks for your work on this!

mwear (approved):
LGTM. @robertlaurin would you mind taking a look at this when you get a chance?

PR #1557, opened 2023-12-23
feat: Improve SSL error logging.

megabus-tobin:
Log more information about why an SSL error occured as this can be very difficult to diagnose. This is logged in the same way as other probably fatal issues are, e.g. HTTP 404s.


arielvalentin (approved):

arielvalentin (approved):

kaylareopelle (approved):

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -180,7 +180,8 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/MethodLength
             rescue Net::OpenTimeout, Net::ReadTimeout
               retry if backoff?(retry_count: retry_count += 1, reason: 'timeout')
               return FAILURE
-            rescue OpenSSL::SSL::SSLError
+            rescue OpenSSL::SSL::SSLError => e
+              OpenTelemetry.handle_error(exception: e, message: 'SSL error in OTLP::Exporter#send_bytes')
```
arielvalentin, line 184: @open-telemetry/ruby-approvers I think that this is a helpful error message but I'm wondering if it would be even more helpful to record a different message if it is backing off vs when the exporter gives up ?

xuan-cao-swi: I like the idea of different msg between `backoff/retry` vs `give up/return FAILTURE`. Maybe refactoring the `backoff?` to log the retry count (i.e. now it's only the metric records the retry count)?

I am also wondering why we don't log the error msg for other kind of error (e.g. EOFError, SocketError, etc.)?

```ruby
rescue OpenSSL::SSL::SSLError => e
  retry if backoff?(retry_count: retry_count += 1, reason: 'openssl_error')
  # I think adding the debug msg here can avoid repeat error message
  OpenTelemetry.handle_error(exception: e, message: 'SSL error in OTLP::Exporter#send_bytes')
  return FAILURE
```

megabus-tobin: Would it make sense to not retry on SSL related errors at all? They aren't something you'd expect to correct themselves. The handling of them should probably be more like `Net::HTTPNotFound`.

megabus-tobin: I've swapped the error logging to after the `retry`.

I haven't added better error logging to all the other types of failure since I was focusing only on SSL errors as part of getting #1556 done. Should I expand the scope and do it here or keep this narrowly focussed?

kaylareopelle: Sorry for the delay here, @megabus-tobin. Thanks for the update! 

I'd prefer keeping things scoped to SSL errors to accomplish #1556, and address the other changes in a future PR.

megabus-tobin:
I missed that this had been closed automatically. It wold still be good to get this in if possible to make debugging SSL related errors easier.

kaylareopelle:
Hi @megabus-tobin , I missed that too. Thanks for bringing this to our attention. 

I'm not able to reopen the PR, the button is null with the message: `The ssl-error-logging branch was force pushed or recreated.` Are you able to reopen in a new PR?

megabus-tobin:
@kaylareopelle ah, I didn't realise Github couldn't handle that happening. I've been using the branch locally and rebasing it periodically using the official repo's main branch.

New MR: #1644

PR #1560, opened 2024-01-08
fix: require csv for ruby-3.4 compatibility

romuloceccon:
Loading opentelemetry-exporter-otlp on ruby-3.3 now prints a warning:

    .../lib/opentelemetry/exporter/otlp/exporter.rb:10: warning: csv
    was loaded from the standard library, but will no longer be part
    of the default gems since Ruby 3.4.0. Add csv to your Gemfile or
    gemspec. Also contact author of opentelemetry-exporter-otlp-0.26.1
    to add csv into its gemspec.

This commit fixes that by adding an explicit dependency from opentelemetry-exporter-otlp to the csv gem at version '~> 3.1' (ruby-3.0 bundles csv at 3.1.9).

### DIFF
--- a/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
+++ b/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
@@ -8,7 +8,6 @@ require 'opentelemetry/common'
 require 'opentelemetry/exporter/otlp/common'
 require 'opentelemetry/sdk'
 require 'net/http'
-require 'csv'
 require 'zlib'
 
 require 'google/rpc/status_pb'
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -7,7 +7,6 @@
 require 'opentelemetry/common'
 require 'opentelemetry/sdk'
 require 'net/http'
-require 'csv'
 require 'zlib'
 
 require 'google/rpc/status_pb'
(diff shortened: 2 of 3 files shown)


kaylareopelle:
Hi @romuloceccon! Thanks for submitting this fix! Getting rid of the deprecation warning sounds great to me. 

I was looking at the uses of CSV in this library, and I think it may not be used outside the require statement in exporter/ot|p/lib/opentelemetry/exporter/ot|p/ exporter.rb line 10. 

Could you try removing the require statement and see if the tests pass and the deprecation warning goes away?

arielvalentin:
> Hi @romuloceccon! Thanks for submitting this fix! Getting rid of the deprecation warning sounds great to me.
> 
> I was looking at the uses of CSV in this library, and I think it may not be used outside the require statement in exporter/ot|p/lib/opentelemetry/exporter/ot|p/ exporter.rb line 10.
> 
> Could you try removing the require statement and see if the tests pass and the deprecation warning goes away?

@kaylareopelle we use the CSV library to parse configuration values from environment variables in at least the zipkin exporter: https://github.com/open-telemetry/opentelemetry-ruby/blob/ff9a7d3af31ad83ed6e07445a9e1609bec53d779/exporter/zipkin/lib/opentelemetry/exporter/zipkin/exporter.rb#L50

I could have sworn it was used in the OTLP Exporters but looks like it now uses custom parsing? https://github.com/open-telemetry/opentelemetry-ruby/blob/ff9a7d3af31ad83ed6e07445a9e1609bec53d779/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb#L269

romuloceccon:
Thanks for the heads-up, @kaylareopelle and @arielvalentin! I have adjusted the commit accordingly, and also fixed otlp/http exporter.

kaylareopelle (approved):
Thank you, @romuloceccon! This looks good to me. I think the job failures are due to an unrelated problem on `main` causing TruffleRuby installs to fail, that will hopefully be fixed with #1559 🤞

mwear (approved):

Issue #1160, opened 2022-03-25
Log more details for "Unable to export # spans"


francisdb:
**Description of the bug**

I'm trying to send traces to our otlp compliant tempo service
Our jvm services can send traces without issues with only configuring the trace endpoint

Now I'm trying to do the same for our ruby code but have the following issue:

OTEL_EXPORTER_OTLP_ENDPOINT="https://internal.url"
OTEL_EXPORTER_OTLP_PROTOCOL=grpc
OTEL_LOG_LEVEL=DEBUG

```
D, [2022-03-25T20:23:01.357950 #71922] DEBUG -- : Upgrading default proxy tracer provider to OpenTelemetry::SDK::Trace::TracerProvider
I, [2022-03-25T20:23:01.361189 #71922]  INFO -- : Instrumentation: OpenTelemetry::Instrumentation::Faraday was successfully installed

perform some http calls here

E, [2022-03-25T20:23:32.033916 #71922] ERROR -- : OpenTelemetry error: Unable to export 30 spans
E, [2022-03-25T20:23:42.271592 #71922] ERROR -- : OpenTelemetry error: Unable to export 72 spans
```

The main issue is that we don't have more output than this so it is very hard to know what fails.

The http calls correctly get the `traceparent` header with the surrounding traceId  as we can look up the spans from the services this code talks to.

**Share details about your runtime**

Operating system details: macOS 12.3 / Ubuntu Linux
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.1.1"
RUBY_DESCRIPTION: "ruby 3.1.1p18 (2022-02-18 revision 53f5fc4236) [arm64-darwin21]"

fbogsany:
This problem occurs because we don't support gRPC in the OTLP exporter. We only support HTTP. Unfortunately, we don't support the `OTEL_EXPORTER_OTLP_PROTOCOL` env var either, so we're configuring a OTLP/HTTP exporter regardless of that variable.

We don't anticipate having a OTLP/gRPC exporter 🔜  (tho PRs are welcome), but we should interpret the `OTEL_EXPORTER_OTLP_PROTOCOL` env var and log a warning if `grpc` is requested, and not configure the OTLP/HTTP exporter.

francisdb:
That would already help if the code would log some warning that think are not supported and falling back to other protocols.

[...]

PR #1565, opened 2024-01-16
Consider OTLP export failures handleable errors

plantfansam:
This PR attempts to address https://github.com/open-telemetry/opentelemetry-ruby/issues/1160.

The general idea is that:

- we can consider OTLP exporter failures to be errors worthy of handling once retries have been exhausted
- we should let the `handle_error` method decide whether or not these particular failures are important
- the uri is useful information in these failures

Not everyone has a metrics reporter installed, so some log-level visibility might be desirable.

Alternatively, we could `OpenTelemetry::Logger.warn` (or `.debug` or `info` whatever) in these scenarios. I'm happy with that outcome too. Or whatever other great ideas folks may have 😄.

### DIFF
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -223,7 +223,7 @@ module OpenTelemetry
             klass_or_nil = ::Google::Protobuf::DescriptorPool.generated_pool.lookup(detail.type_name).msgclass
             detail.unpack(klass_or_nil) if klass_or_nil
           end.compact
-          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}} from #{@uri}")
+          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}} for uri=#{@uri}")
         rescue StandardError => e
           OpenTelemetry.handle_error(exception: e, message: 'unexpected error decoding rpc.Status in OTLP::Exporter#log_status')
         end
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -179,7 +179,7 @@ module OpenTelemetry
               redo if backoff?(retry_after: 0, retry_count: retry_count += 1, reason: response.code)
             else
               @http.finish
-              OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response_code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
+              OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response.code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
               @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => 'unexpected_response', 'response_code' => response.code })
               FAILURE
             end
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -228,9 +228,9 @@ module OpenTelemetry
           OpenTelemetry.handle_error(exception: e, message: 'unexpected error decoding rpc.Status in OTLP::Exporter#log_status')
         end
 
-        def log_request_failure(response_code, labels = {})
+        def log_request_failure(response_code)
           OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response_code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
-          @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response_code }.merge(labels))
+          @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response_code })
         end
 
         def measure_request_duration
--- a/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
+++ b/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
@@ -179,8 +179,7 @@ module OpenTelemetry
               redo if backoff?(retry_after: 0, retry_count: retry_count += 1, reason: response.code)
             else
               @http.finish
-              OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response.code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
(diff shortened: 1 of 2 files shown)


exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -61,6 +61,7 @@ def initialize(endpoint: nil,
           @http = http_connection(@uri, ssl_verify_mode, certificate_file)
 
           @path = @uri.path
+          @uri_string = @uri.to_s
```
fbogsany, line 64: Doesn't feel worthwhile caching this just for use in the error message.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -161,14 +162,15 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/CyclomaticComplexity,
             when Net::HTTPServiceUnavailable, Net::HTTPTooManyRequests
               response.body # Read and discard body
               redo if backoff?(retry_after: response['Retry-After'], retry_count: retry_count += 1, reason: response.code)
+              log_request_failure(response.code)
               FAILURE
             when Net::HTTPRequestTimeOut, Net::HTTPGatewayTimeOut, Net::HTTPBadGateway
               response.body # Read and discard body
               redo if backoff?(retry_count: retry_count += 1, reason: response.code)
+              log_request_failure(response.code)
```
fbogsany, line 170: These two calls will double count if `backoff?` returns `false` because `backoff?` also does `@metrics_reporter.add_to_counter('otel.otlp_exporter.failure' ...)`.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -222,11 +225,16 @@ def log_status(body)
             klass_or_nil = ::Google::Protobuf::DescriptorPool.generated_pool.lookup(detail.type_name).msgclass
             detail.unpack(klass_or_nil) if klass_or_nil
           end.compact
-          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}}")
+          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}}, uri #{@uri_string}")
```
fbogsany, line 228:
```suggestion
          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}} from #{@uri}")
```
reads better IMO

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -222,11 +225,16 @@ def log_status(body)
...
         rescue StandardError => e
           OpenTelemetry.handle_error(exception: e, message: 'unexpected error decoding rpc.Status in OTLP::Exporter#log_status')
         end
 
+        def log_request_failure(response_code, labels = {})
+          OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response_code} for uri: '#{@uri_string}' in OTLP::Exporter#send_bytes")
```
fbogsany, line 234:
```suggestion
          OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response_code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
```
use of `=` feels more consistent within the message.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
plantfansam: Nice catch 😅  - will 🤔  and 💻

plantfansam: 86f99e80fa1ecbb4cddfec6ffdfad1a8aa52cf5b

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -222,11 +223,16 @@ def log_status(body)
...
           end.compact
-          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}}")
+          OpenTelemetry.handle_error(message: "OTLP exporter received rpc.Status{message=#{status.message}, details=#{details}} for uri=#{@uri}")
         rescue StandardError => e
           OpenTelemetry.handle_error(exception: e, message: 'unexpected error decoding rpc.Status in OTLP::Exporter#log_status')
         end
 
+        def log_request_failure(response_code)
+          OpenTelemetry.handle_error(message: "OTLP exporter received http.code=#{response_code} for uri='#{@uri}' in OTLP::Exporter#send_bytes")
+          @metrics_reporter.add_to_counter('otel.otlp_exporter.failure', labels: { 'reason' => response_code })
```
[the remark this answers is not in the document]
plantfansam: I don't think we need the extra label; I think we can do `log_request_failure(response.code)`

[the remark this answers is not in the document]
plantfansam: https://github.com/open-telemetry/opentelemetry-ruby/pull/1565/commits/3751534fd40bc324ab5703b0f1d2bfafa0da5d58

kaylareopelle (approved):
I'm so glad you submitted this PR!

I've manually added logging like this around these errors so many times, and it'll be great to have to have that baked in. 🍪

fbogsany (approved):

muripic:
Hey! 👋 
After this change we're getting a lot of error logs due to retries that eventually succeed (specifically, due to [this line](https://github.com/open-telemetry/opentelemetry-ruby/pull/1565/files#diff-34fb5aed920b31dca334d76d897e2cd337ee9ac0380b550ee3aef6f5eab60eb5R249).

> Alternatively, we could OpenTelemetry::Logger.warn (or .debug or info whatever) in these scenarios. I'm happy with that outcome too. Or whatever other great ideas folks may have 😄.

Going back a bit to what you said there, we appreciate the visibility 💯  but at least for us it would make more sense if these failures coming from retries were logged at debug, info or warning level, since it's retried anyway (a different situation would be if it's still failing after all retry attempts). What do you think? Are you all still open to discussing it? 😄

fbogsany:
😬 yeah, that’s bad. We shouldn’t call `log_request_failure` in `backoff?`, both because we’re going to retry (maybe) and because there are many non-http-response-code `reason`s, so the message can be incorrect.

plantfansam:
Sorry for the trouble @muripic, we will address this.

muripic:
Thank you! ❤️

plantfansam:
> Thank you! ❤️

Of course! Should be addressed by https://github.com/open-telemetry/opentelemetry-ruby/pull/1589

PR #1566, opened 2024-01-16
Add context to metrics reporting of buffer-full events

plantfansam:
We report `buffer-full` dropped spans in two contexts: `on_finish` and `force_flush`. Since `force_flush` is used in specific contexts, I thought it would be useful to supply a label so that users can have visibility into that.

We could alternatively use the `reason` to disambiguate the two scenarios, rather than introducing a new tag.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -82,7 +82,7 @@ module OpenTelemetry
               n = spans.size + 1 - max_queue_size
               if n.positive?
                 spans.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full', function: 'on_finish')
+                report_dropped_spans(n, reason: 'buffer-full', function: __method__)
               end
               spans << span
               @condition.signal if spans.size > batch_size
@@ -205,7 +205,7 @@ module OpenTelemetry
           end
 
           def report_dropped_spans(count, reason:, function: nil)
-            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: count, labels: { 'reason' => reason, OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => function }.compact)
+            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: count, labels: { 'reason' => reason, OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => __method__ }.compact)
           end
 
           def fetch_batch
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -147,7 +147,7 @@ module OpenTelemetry
             thread&.join(timeout)
             force_flush(timeout: OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time))
             dropped_spans = lock { spans.shift(spans.length) }
-            report_dropped_spans(dropped_spans, reason: 'terminating') if dropped_spans.positive?
+            report_dropped_spans(dropped_spans, reason: 'terminating') if dropped_spans.any?
             @exporter.shutdown(timeout: OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time))
           end
 
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -123,7 +123,7 @@ module OpenTelemetry
               n = spans.size + snapshot.size - max_queue_size
               if n.positive?
                 snapshot.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full', function: 'force_flush')
+                report_dropped_spans(n, reason: 'buffer-full', function: __method__)
               end
               spans.unshift(snapshot) unless snapshot.empty?
               @condition.signal if spans.size > max_queue_size / 2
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -123,7 +123,7 @@ def force_flush(timeout: nil) # rubocop:disable Metrics/MethodLength
               n = spans.size + snapshot.size - max_queue_size
               if n.positive?
                 snapshot.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full')
+                report_dropped_spans(n, reason: 'buffer-full', context: "force_flush")
```
fbogsany, line 126:
```suggestion
                report_dropped_spans(n, reason: 'buffer-full', context: 'force_flush')
```
single quotes for consistency

fbogsany (approved):
Small suggestion. Otherwise LGTM.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -82,7 +82,7 @@ def on_finish(span)
               n = spans.size + 1 - max_queue_size
               if n.positive?
                 spans.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full')
+                report_dropped_spans(n, reason => 'buffer-full', OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => 'on_finish')
```
plantfansam, line 85: Need to use hash rocket syntax to have constant as key (afaik)

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -204,8 +204,8 @@ def report_result(result_code, batch)
             end
           end
 
-          def report_dropped_spans(count, reason:)
-            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: count, labels: { 'reason' => reason })
+          def report_dropped_spans(count, labels = {})
```
plantfansam, line 207: Although I changed the method signature, we do not need to change other call sites for `report_dropped_spans`  since they were using the kwarg `reason: 'foo'` which we can also interpret as a hash.

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -82,7 +82,7 @@ def on_finish(span)
               n = spans.size + 1 - max_queue_size
               if n.positive?
                 spans.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full')
+                report_dropped_spans(n, reason: 'buffer-full', context: 'on_finish')
```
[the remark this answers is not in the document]
plantfansam: Stupendous idea! 03fb66cd841d488d2d80c24e954d3aa30e6d6280

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
fbogsany: 🤔 I'd rather we used named arguments consistently, and Strings for keys and values in the labels. I.e. I'd prefer:
```ruby
def report_dropped_spans(count, labels: nil) ... end

report_dropped_spans(n, labels: { 'reason' => 'buffer-full', 'code.function' => 'force_flush' })
```
or:
```ruby
def report_dropped_spans(count, reason:, function: nil)
  @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: count, labels: { 'reason' => reason, 'code.function' => function }.compact)
end

report_dropped_spans(n, reason: 'buffer-full', function: 'force_flush')
```

I prefer the latter, since it is cleaner for callers.

plantfansam: Yes, let's do the latter

plantfansam: dfab14e06e961a925be24be69d4ae91d0d463814

ericmustin:
Seems completely arbitrary and unrelated to the specification. Also lgtm

ericmustin (approved):

kaylareopelle (approved):

plantfansam:
Woo hoo, I just need an adult to merge it 😄

fbogsany:
> Seems completely arbitrary and unrelated to the specification.

Unrelated, yes. Not completely arbitrary. For transparency, the goal is to _also_ enable logging of the `trace_id`s of the spans that are dropped. It isn't completely clear how to do that in a way that'll make everyone happy, e.g. with levelled logging, so we're going to experiment with this in Shopify by monkey-patching `report_dropped_spans`. The existing code only passes the count of dropped spans to this method, but we actually need the dropped spans themselves. Hence this change.

fbogsany:
Also, to be clear, we are dropping spans in production due to `buffer-full` and we don't know which of the callsites is responsible, hence adding a label to disambiguate. Changing the meaning of the existing label might break some consumers of the metric.

sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
```diff
@@ -222,7 +222,7 @@ def to_span_data
       _(test_exporter.failed_batches.size).must_equal(0)
       _(test_exporter.batches.size).must_equal(0)
 
-      _(bsp.instance_variable_get(:@spans).size).must_equal(1)
+      _(bsp.instance_variable_get(:@spans).size).must_equal(0)
```
plantfansam, line 225: Since we are now properly dropping spans during `shutdown` we change the test expectation

fbogsany (approved):

PR #1572, opened 2024-01-19
log spanz

plantfansam:
- add context to metrics reporting of buffer-full events
- Update sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
- Use semantic convention for code function label
- Kwargs are better; use kwargs
- stray paren
- maybe this

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -81,8 +81,8 @@ module OpenTelemetry
               reset_on_fork
               n = spans.size + 1 - max_queue_size
               if n.positive?
-                spans.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full', function: 'on_finish')
+                dropped_spans = spans.shift(n)
+                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: 'on_finish')
               end
               spans << span
               @condition.signal if spans.size > batch_size
@@ -122,8 +122,8 @@ module OpenTelemetry
             lock do
               n = spans.size + snapshot.size - max_queue_size
               if n.positive?
-                snapshot.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full', function: 'force_flush')
+                dropped_spans = snapshot.shift(n)
+                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: 'force_flush')
               end
               spans.unshift(snapshot) unless snapshot.empty?
               @condition.signal if spans.size > max_queue_size / 2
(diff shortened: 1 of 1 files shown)


sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -81,8 +81,8 @@ def on_finish(span)
               reset_on_fork
               n = spans.size + 1 - max_queue_size
               if n.positive?
-                spans.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full')
+                dropped_spans = spans.shift(n)
+                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: 'on_finish')
```
fbogsany, line 85:
```suggestion
                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: __method__.to_s)
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -122,8 +122,8 @@ def force_flush(timeout: nil) # rubocop:disable Metrics/MethodLength
             lock do
               n = spans.size + snapshot.size - max_queue_size
               if n.positive?
-                snapshot.shift(n)
-                report_dropped_spans(n, reason: 'buffer-full')
+                dropped_spans = snapshot.shift(n)
+                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: 'force_flush')
```
fbogsany, line 126:
```suggestion
                report_dropped_spans(dropped_spans, reason: 'buffer-full', function: __method__.to_s)
```

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
```diff
@@ -146,7 +146,7 @@ def shutdown(timeout: nil)
 
             thread&.join(timeout)
             force_flush(timeout: OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time))
-            dropped_spans = lock { spans.size }
+            dropped_spans = lock { spans}
```
fbogsany, line 149: This is problematic if multiple threads are running. We know the exporter thread has finished, and we _should_ be the only thread executing `shutdown`, but there is some possibility that other application threads are running and enqueuing finished spans.

Technically, we haven't "dropped spans" because we're just assuming that the process is shutting down and the BSP will go away anyway, but I think we should not make that assumption. Instead, we should _actually_ drop the spans by shifting them out of the `spans` array. That will also satisfy mutual exclusion concerns.

So, something like:
```suggestion
            dropped_spans = lock { spans.shift(spans.length) }
```

plantfansam: > Technically, we haven't "dropped spans" because we're just assuming that the process is shutting down and the BSP will go away anyway, but I think we should not make that assumption. Instead, we should _actually_ drop the spans by shifting them out of the `spans` array. That will also satisfy mutual exclusion concerns.

I think this is a good idea, but also, I'm not clear on why we shouldn't assume the BSP will go away anyway. I was under the impression that there was one BSP worker thread per forked process 🤔

plantfansam:
closing in favor of https://github.com/open-telemetry/opentelemetry-ruby/pull/1566

sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
plantfansam: closed in favor of https://github.com/open-telemetry/opentelemetry-ruby/pull/1566; we can continue this discussion here or there :)

PR #1582, opened 2024-01-25
feat: add spans to Trace::ExportError

fbogsany:
This adds a `spans` attribute to `SDK::Trace::ExportError`, containing the array of spans that could not be exported. It refactors the BSP a little to ensure an array of `Span` rather than (sometimes) an array of `SpanData` is wrapped by the exception.

The intent is to allow a pluggable error handler to recognize the error type, retrieve the array of spans, and (for example) report more detailed information (such as the trace IDs of the spans that are dropped).

Example:
```ruby
OpenTelemetry.error_handler = ->(exception: nil, message: nil) do
  case exception
  when OpenTelemetry::SDK::Trace::ExportError
    logger.error("OpenTelemetry error: #{exception.message} from trace IDs: #{exception.spans.collect {|span| span.context.trace_id }.uniq.join(', ')}")
  else
    logger.error("OpenTelemetry error: #{[message, exception&.message].compact.join(' - ')}")
  end
end
```

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export.rb
@@ -12,9 +12,12 @@ module OpenTelemetry
       module Export
         # Raised when an export fails; spans are available via :spans accessor
         class ExportError < OpenTelemetry::Error
+          # Returns the {Span} array for this exception
+          #
+          # @return [Array<OpenTelemetry::SDK::Trace::Span>]
           attr_reader :spans
 
-          # @param [Array<OpenTelemetry::Trace::Span>] spans the array of spans that failed to export
+          # @param [Array<OpenTelemetry::SDK::Trace::Span>] spans the array of spans that failed to export
           def initialize(spans)
             super("Unable to export #{spans.size} spans")
             @spans = spans


plantfansam:
Proving to myself that you can do instance vars + attr_reader on an exception

```
class FooError < StandardError
  attr_reader :str
  def initialize(str)
    super("str")
    @str = str
  end
end

begin
  raise FooError.new("hi")
rescue FooError => e
  pp "in rescue block"
  pp e.str
end
```

>  ruby foo.rb
> "in rescue block"
> "hi"

sdk/lib/opentelemetry/sdk/trace/export.rb
```diff
@@ -10,7 +10,16 @@ module Trace
       # The Export module contains the built-in exporters and span processors for the OpenTelemetry
       # reference implementation.
       module Export
-        ExportError = Class.new(OpenTelemetry::Error)
+        # Raised when an export fails; spans are available via :spans accessor
+        class ExportError < OpenTelemetry::Error
+          attr_reader :spans
+
+          # @param [Array<OpenTelemetry::Trace::Span>] spans the array of spans that failed to export
```
fbogsany, line 17:
```suggestion
          # @param [Array<OpenTelemetry::SDK::Trace::Span>] spans the array of spans that failed to export
```

sdk/lib/opentelemetry/sdk/trace/export.rb
```diff
@@ -10,7 +10,16 @@ module Trace
       # The Export module contains the built-in exporters and span processors for the OpenTelemetry
       # reference implementation.
       module Export
-        ExportError = Class.new(OpenTelemetry::Error)
+        # Raised when an export fails; spans are available via :spans accessor
+        class ExportError < OpenTelemetry::Error
+          attr_reader :spans
```
fbogsany, line 15:
```suggestion
          # Returns the {Span} array for this exception
          #
          # @return [Array<OpenTelemetry::SDK::Trace::Span>]
          attr_reader :spans
```

PR #1604, opened 2024-02-22
feat: add basic metrics view

xuan-cao-swi:
## Description

I'd like to contribute on [metrics view](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view). I think view will be helpful for customized aggregation over default aggregation on instrumenting metrics.

1. Changed the location of `data_points` to `metric_stream`, because I think aggregation should act as aggregator that provide transformative (on `data_points`) feature rather than storage feature.
2. Didn't implement the logic for [conflicting metric identities](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/data-model.md#opentelemetry-protocol-data-model-producer-recommendations). For current implementation, it will export all the metrics with different aggregation if the metric_stream has multiple views applied.
3. Need schema_url as part of instrumentation selection. https://github.com/open-telemetry/opentelemetry-ruby/pull/1237


kaylareopelle:
As a continuation from our discussion during the SIG on July 9, 2024: 

I was experimenting with this branch to create an explicit bucket histogram with custom buckets. It worked great! Thank you for building this! 

The only error I've encountered was while exporting the metrics via OTLP. The structure of the metrics object passed as an argument to [the export method](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics_exporter.rb#L86-L91) has changed.

Before: 
```ruby
Array[MetricData, MetricData]
```

In this branch: 
```ruby
Array[[MetricData],[MetricData]]
```

We could handle this in the exporter, but it may be preferable to go back to the original structure. As far as I can [see in the spec](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#exportbatch:~:text=batch%20%2D%20a%20batch%20of%20Metric%20Points.%20The%20exact%20data%20type%20of%20the%20batch%20is%20language%20specific%2C%20typically%20it%20is%20some%20kind%20of%20list.%20The%20exact%20type%20of%20Metric%20Point%20is%20language%20specific%2C%20and%20is%20typically%20optimized%20for%20high%20performance.), it just needs to be some kind of list, so either way seems compliant.

xuan-cao-swi:
Thanks for catching it. I use `flatten!` to ensure it return `[MetricData]`. I remembered that I made this mistake while I was thinking about the case for multiple view for one metric stream.

kaylareopelle:
> Thanks for catching it. I use `flatten!` to ensure it return `[MetricData]`. I remembered that I made this mistake while I was thinking about the case for multiple view for one metric stream.

Thanks for the fix! What else did you want to accomplish before moving the PR out of draft mode?

xuan-cao-swi:
Not in my current mind. I will open it for review

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -0,0 +1,36 @@
...
+          attr_reader :name, :aggregation, :attribute_keys
+
+          def initialize(name, **options)
+            @name = name
+            @options = options
+            @aggregation = options[:aggregation]
+            @attribute_keys = options[:attribute_keys] || {}
+          end
+
+          def match_instrument(metric_stream)
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated!

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -125,13 +126,12 @@ def register_synchronous_instrument(instrument)
...
         end
 
-        # The type of the Instrument(s) (optional).
         # The name of the Instrument(s). OpenTelemetry SDK authors MAY choose to support wildcard characters, with the question mark (?) matching exactly one character and the asterisk character (*) matching zero or more characters.
-        # The name of the Meter (optional).
-        # The version of the Meter (optional).
-        # The schema_url of the Meter (optional).
-        def add_view
-          # TODO: For each meter add this view to all applicable instruments
+        # The options of the Instrument(s). Useful key include: aggregation, type, unit, meter_name, meter_version, attribute_keys
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated the doc. Let me know which part needs further improve

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -30,30 +30,60 @@ def initialize(
             @instrument_kind = instrument_kind
```
kaylareopelle, line 30: I assume that initializing the MetricStream falls into this part of the spec: https://opentelemetry.io/docs/specs/otel/metrics/sdk/#stream-configuration 

If so, we might want to consider updating the defaults of the MetricStream to more closely match the spec. I don't think we need to do this as part of this PR. If we don't, we may want to create an issue to tackle this later on.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -30,30 +30,60 @@ def initialize(
...
+              start_time,
+              end_time
+            )
+          end
+
+          def find_registered_view
+            registered_views = []
+            @meter_provider.registered_views.each { |view| registered_views << view if view.match_instrument?(self) }
+            registered_views
           end
```
kaylareopelle, line 87: I'm concerned about running `find_registered_view` every time we run the `collect` or `update` methods.

It seems like we may be duplicating work that could be preserved across runs.

Do we expect the registered views to change over time? Could we save the registered views we know match with a particular metrics stream? Or maybe just check to see if there are new views and add those to an existing list each time?

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -125,13 +126,33 @@ def register_synchronous_instrument(instrument)
...
 
-        # The type of the Instrument(s) (optional).
-        # The name of the Instrument(s). OpenTelemetry SDK authors MAY choose to support wildcard characters, with the question mark (?) matching exactly one character and the asterisk character (*) matching zero or more characters.
-        # The name of the Meter (optional).
-        # The version of the Meter (optional).
-        # The schema_url of the Meter (optional).
-        def add_view
-          # TODO: For each meter add this view to all applicable instruments
+        # Register a view.
+        #
```
kaylareopelle, line 130:
```suggestion
        #
        # A View provides SDK users with the flexibility to customize the metrics that are output by the SDK.
        #
```

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -0,0 +1,36 @@
...
+            return false if @aggregation.nil?
+            return false if @name && @name != metric_stream.name
```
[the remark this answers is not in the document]
xuan-cao-swi: Thanks, I have include a basic wildcard matching for name in `name_match`

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -0,0 +1,36 @@
...
+
+          def initialize(name, **options)
+            @name = name
+            @options = options
+            @aggregation = options[:aggregation]
+            @attribute_keys = options[:attribute_keys] || {}
+          end
+
+          def match_instrument(metric_stream)
+            return false if @aggregation.nil?
```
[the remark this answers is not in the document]
xuan-cao-swi: I think it was intended to ignore the view without any (valid) aggregation, but it seems like not appropriate as it's kind confusing with spec. I have removed it and use `valid_aggregation?` to check if the aggregation is valid one

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
xuan-cao-swi: Yeah, it's probably best to address it in different PR dedicates for metric_stream.

View seems doesn't need to have `attribute_keys`, but I was confused by this [spec's example](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#view)
```
meter_provider
    .add_view("X", aggregation=SumAggregation())
    .add_view("*", attribute_keys=["a", "b"]) # wildcard view matches everything, including X
    ...
```

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
xuan-cao-swi: I didn't see any place indicate user can or should add_view after create the instrument. Can we make that assumption? However, if that's the case, the instrument initialization order has to be
```
OpenTelemetry.meter_provider.add_view(...)
OpenTelemetry.meter_provider.add_view(...)
... # add more views

counter = meter.create_counter('counter') # instrument initialization
```
I updated the logic based on the assumption that registered views doesn't change for initialized instrument

xuan-cao-swi:
Thanks @kaylareopelle, I updated the suggestions. I didn't do commit the suggestions on the github page to avoid github freeze on my browser and create commit issue (I had bad experience last time click each `Commit suggestion`).

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
kaylareopelle: Hmm... that is odd. I'm on board for leaving this as-is for this PR. New issue to track the work: https://github.com/open-telemetry/opentelemetry-ruby/issues/1656

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
kaylareopelle: The initialization order is a good point. I'm not sure that's quite what we want. I think we should defer this point to another PR. I'll open a new issue and we can continue the conversation there.

kaylareopelle: @xuan-cao-swi  - Here's the new issue to think about this further: https://github.com/open-telemetry/opentelemetry-ruby/issues/1672

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -125,13 +126,32 @@ def register_synchronous_instrument(instrument)
...
-        # The name of the Instrument(s). OpenTelemetry SDK authors MAY choose to support wildcard characters, with the question mark (?) matching exactly one character and the asterisk character (*) matching zero or more characters.
-        # The name of the Meter (optional).
-        # The version of the Meter (optional).
-        # The schema_url of the Meter (optional).
-        def add_view
-          # TODO: For each meter add this view to all applicable instruments
+        # A View provides SDK users with the flexibility to customize the metrics that are output by the SDK.
+        #
+        # A View provides SDK users with the flexibility to customize the metrics that are output by the SDK.
+        #
```
kaylareopelle, line 132: This line is duplicated
```suggestion
```

metrics_sdk/lib/opentelemetry/sdk/metrics/view.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      # View
```
kaylareopelle, line 10:
```suggestion
      # A View provides SDK users with the flexibility to customize the metrics that are output by the SDK. 
```

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -0,0 +1,49 @@
...
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module View
+        # RegisteredView
```
kaylareopelle, line 11:
```suggestion
        # RegisteredView is an internal class used to match Views with a given {MetricStream}
```

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -0,0 +1,49 @@
...
+          attr_reader :name, :aggregation, :attribute_keys
+
+          def initialize(name, **options)
+            @name = name
+            @options = options
+            @aggregation = options[:aggregation]
+            @attribute_keys = options[:attribute_keys] || {}
+          end
+
+          def match_instrument?(metric_stream)
```
kaylareopelle, line 22: There's one more pattern we need to consider for the spec. [If the name is `*`, then it should always match.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md?plain=1#L275).

kaylareopelle (changes requested):

metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
xuan-cao-swi: For current implementation, the wildcard `*` should match everything.

kaylareopelle (approved):
Thank you for adding this and for bearing with me through all the back and forth! 😄

mwear (commented):
Just a couple of questions, nothing blocking.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_stream.rb
```diff
@@ -30,30 +30,59 @@ def initialize(
...
-                @unit,
-                @instrument_kind,
-                @meter_provider.resource,
-                @instrumentation_scope,
-                @aggregation.collect(start_time, end_time),
-                @aggregation.aggregation_temporality,
-                start_time,
-                end_time
-              )
+            metric_data = []
```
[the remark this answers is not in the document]
xuan-cao-swi: Added mutex back. Collect should be thread-safe since it will update the data_points depends on the aggregation_temporality.

metrics_sdk/lib/opentelemetry/sdk/metrics/state/metric_store.rb
```diff
@@ -25,7 +25,7 @@ def collect
               @epoch_end_time = now_in_nano
               snapshot = @metric_streams.map { |ms| ms.collect(@epoch_start_time, @epoch_end_time) }
               @epoch_start_time = @epoch_end_time
-              snapshot
+              snapshot.flatten!
```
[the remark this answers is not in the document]
xuan-cao-swi: Updated!

metrics_sdk/lib/opentelemetry/sdk/metrics/export/in_memory_metric_pull_exporter.rb
```diff
@@ -25,7 +25,7 @@ def pull
 
           def export(metrics, timeout: nil)
             @mutex.synchronize do
-              @metric_snapshots << metrics
+              metrics.instance_of?(Array) ? @metric_snapshots.concat(metrics) : @metric_snapshots << metrics
```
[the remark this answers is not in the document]
xuan-cao-swi: Actually, for current implementation, there will be no non-array options metrics based on metric_stream.collect. I can just use `@metric_snapshots.concat(Array(metrics))` only.

mwear (approved):

Issue #1386, opened 2022-10-04
Asynchronous Counter


ahayworth:
https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-counter

PR #1610, opened 2024-03-05
feat: add basic implementation of asynchronous metrics

xuan-cao-swi:
## Description

I'd like to make contribution on basic implementation on asynchronous metrics.

Related: https://github.com/open-telemetry/opentelemetry-ruby/issues/1386, [Asynchronous Up Down Counter card](https://github.com/open-telemetry/opentelemetry-ruby/projects/2#card-84597629), [Asynchronous Gauge](https://github.com/open-telemetry/opentelemetry-ruby/projects/2#card-84597614).
Spec: [asynchronous-instrument-api](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md#asynchronous-instrument-api)

WIP:
1. when multiple callbacks exist in single instrument, they may all impose (accumulative) modification on data points, which need further consideration of use case.

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/asynchronous_instrument.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/asynchronous_instrument.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Metrics
+      module Instrument
+        # {AsynchronousInstrument} contains the common functionality shared across
+        # the asynchronous instruments SDK instruments.
+        class AsynchronousInstrument
+          def initialize(name, unit, description, callback, instrumentation_scope, meter_provider)
+            @name = name
+            @unit = unit
+            @description = description
+            @instrumentation_scope = instrumentation_scope
+            @meter_provider = meter_provider
+            @metric_streams = []
+            @callbacks = []
+
+            register_callback(callback)
+            meter_provider.register_asynchronous_instrument(self)
+          end
+
+          # @api private
+          def register_with_new_metric_store(metric_store, aggregation: default_aggregation)
+            ms = OpenTelemetry::SDK::Metrics::State::AsynchronousMetricStream.new(
+              @name,
+              @description,
+              @unit,
+              instrument_kind,
+              @meter_provider,
+              @instrumentation_scope,
+              aggregation,
+              @callbacks
+            )
+            @metric_streams << ms
+            metric_store.add_metric_stream(ms)
+          end
+
+          def register_callback(callback)
+            callbacks = [callback] unless callback.instance_of? Array
+
+            callbacks.each do |cb|
+              if cb.instance_of? Proc
+                @callbacks << cb
+              else
+                OpenTelemetry.logger.warn "The callback registeration failed for instrument #{@name}"
+              end
+            end
+            @meter_provider.register_callback(self, @callbacks) # meter_provider should register list of callback
+          end
+
+          def remove_callback(callback)
+            orig_callback_size = @callbacks.size
+            callbacks = [callback] unless callback.instance_of? Array
+
+            callbacks.each { |cb| @callback.delete(cb) }
+            @meter_provider.register_callback(self, @callbacks) if @callback.size != orig_callback_size
+          end
+
+          private
+
+          def update(timeout, attributes)
+            @metric_streams.each { |ms| ms.invoke_callback(timeout, attributes) }
+          end
+        end
+      end
+    end
+  end
+end
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/asynchronous_instrument.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/asynchronous_instrument.rb
@@ -22,7 +22,7 @@ module OpenTelemetry
             @timeout   = nil
             @attributes = {}
 
-            register_callback(callback)
+            init_callback(callback)
             meter_provider.register_asynchronous_instrument(self)
           end
 
@@ -44,16 +44,29 @@ module OpenTelemetry
             metric_store.add_metric_stream(ms)
           end
 
-          # For multiple callbacks in single instrument
+          # The API MUST support creation of asynchronous instruments by passing zero or more callback functions
+          # to be permanently registered to the newly created instrument.
+          def init_callback(callback)
+            if callback.instance_of?(Proc)
+              @callbacks << callback
+            elsif callback.instance_of?(Array)
+              callback.each { |cb| @callbacks << cb if cb.instance_of?(Proc) }
+            else
+              OpenTelemetry.logger.warn "Only accept single Proc or Array of Proc for initialization with callback (given callback #{callback.class}"
+            end
+          end
+
+          # Where the API supports registration of callback functions after asynchronous instrumentation creation,
+          # the user MUST be able to undo registration of the specific callback after its registration by some means.
           def register_callback(callback)
             if callback.instance_of?(Proc)
               @callbacks << callback
+              callback
             else
               OpenTelemetry.logger.warn "Only accept single Proc for registering callback (given callback #{callback.class}"
             end
           end
 
-          # For callback functions registered after an asynchronous instrument is created,
           def unregister(callback)
             @callbacks.delete(callback)
           end
@@ -77,4 +90,10 @@ module OpenTelemetry
       end
     end
   end
+end
+
+def create_callback(callbacks)
+  return callbacks if callbacks.instance_of?(Proc)
+
+  puts 'a'
 end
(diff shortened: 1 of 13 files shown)


metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -125,6 +125,10 @@ def register_synchronous_instrument(instrument)
           end
         end
 
+        def register_asynchronous_instrument(instrument)
```
dan-corneanu, line 128: Do we need both `register_asynchronous_instrument` and `register_synchronous_instrument`?
Could we not have just one function
```
# @api private
def register_instrument(instrument)
...
end
```

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/asynchronous_instrument.rb
```diff
@@ -0,0 +1,99 @@
...
+          def update(timeout, attributes)
+            @metric_streams.each { |ms| ms.invoke_callback(timeout, attributes) }
+          end
+        end
+      end
+    end
+  end
+end
+
+def create_callback(callbacks)
```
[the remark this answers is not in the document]
xuan-cao-swi: Yes, thanks!

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
xuan-cao-swi: Updated!

cdan-youdo (approved):

dan-corneanu:
Any chance one of the code owners could take a look at this PR?

zvkemp:
I think there are a few other things to consider here; top of my list would be the ability to return multiple values from a callback, meaning the callback should also be able to return arbitrary attributes (perhaps in addition to the attributes declared when the instrument is created). An example might be reporting variable-length statistics about the current state of the system, like activity-per-cpu or RSS for each pid:

```ruby
callback = lambda do |state|
  state.observe(2000, { 'pid' => 12, 'process_name' => 'puma' })
  state.observe(2500, { 'pid' => 14, 'process_name' => 'sidekiq' })
end
```

(In this example, the state object is some data container that exists for this callback invocation; the next operation would be to loop over each observed result and call `update` with the value and callback attributes, probably merged with the instrument's attributes).

The spec has a few other examples [similar to this](https://opentelemetry.io/docs/specs/otel/metrics/api/#asynchronous-gauge).

xuan-cao-swi:
Hi @zvkemp, thank you for your feedback. Initially, I considered only a single value. However, we can address this by modifying the operation [here](https://github.com/open-telemetry/opentelemetry-ruby/blob/5566e01ecd51661c093d5a459330e87f14e00375/metrics_sdk/lib/opentelemetry/sdk/metrics/state/asynchronous_metric_stream.rb#L72) to check the type of value returned from the user's code block.

zvkemp:
As far as I can tell, the expected modes of the callback are:

1. it returns a single value
2. it returns a value along with some attributes (e.g. `[1, { 'cpu' => '1' }]`)
3. it returns multiple values
4. it returns multiple values, each with their own attributes

Since the actual API of the callback isn't defined by the spec, it should be valid to specify that in order to return multiple values, a helper object should be used. Otherwise, for simple cases it should be ok to just return a value.

This was the thing I prototyped (before noticing this PR):

```ruby
class `
  attr_reader :data
  def initialize
    @data = []
  end

  def observe(value, attributes = nil)
    @data << [value, attributes]
    self
  end

  def self.enum_for(value)
    if value.is_a?(self)
      value.data.to_enum
    else
      # assume we have a single value
      [[value, nil]].to_enum
    end
  end
end
```

invoke_callback would be:
```ruby
          def invoke_callback(timeout, attributes)
            @mutex.synchronize do
              Timeout.timeout(timeout || 30) do
                @callback.each do |cb|
                   observations = CallbackObservations.new
                  CallbackObservations.enum_for(cb.call(observations)).each do |value, attributes|
                    @aggregation.update(value, attributes || {}, @data_points)
                  end
                end
              end
            end
          end
```

... and then callbacks could be any of the following:

```ruby
-> (obs) do
  obs.observe(1)
end

-> (obs) do
  obs.observe(1, cpu: 1)
end

-> () { 1 }
```

This is sort of maximally flexible without needing to do nested type-checking (but doesn't yet include any validation on the return values of the callbacks — they are assumed to be the observations object or a single value). A _safer_ version would be to enforce that all observations are recorded via `CallbackObservations#observe`, and ignore the return values from the procs.

xuan-cao-swi:
Thanks @zvkemp , I think your solution may work. I am thinking this as an additional improvement (single value export -> multi-value export) on asy instrument, and you can open a pr once this one get merged (not sure if you can direct open a pr based on this branch; but if that works, please do so).

zvkemp:
I have a fork of your branch here: https://github.com/zvkemp/opentelemetry-ruby/tree/metrics-asynchronous (but no PR yet, but it supports this proposal: https://github.com/zvkemp/opentelemetry-ruby-contrib/pull/1). This implementation uses the more restrictive version, where all observations need to go on the recorder argument.

zvkemp:
I've been doing some additional thinking on the 'multi-value export' issue, and I would suggest that the simple case may be too simple if the eventual goal is to support multiple values — i.e., if the callback is allowed to return a value, an array of values, a 'tuple' (array) of a value plus attributes, an array of tuples, or nothing, it seems like there's too much room for error or ambiguity.

The other SDKs I've looked at (python, rust, javascript, and go) require observations to be forwarded to the callback argument, or have defined an `Observation` type, and declare that the callback must return an iterable of observations. The second option isn't very idiomatic in Ruby, so I think `-> (result) { result.observe(1, { ... }) }` is probably the best option here, ignoring the return value altogether.

kaylareopelle:
> I think -> (result) { result.observe(1, { ... }) } is probably the best option here, ignoring the return value altogether.

I agree, @zvkemp. I think we should support multi-value exports from the start. I like where you're headed in your fork of this branch.

kaylareopelle (approved):
This looks good to me. Thank you, @xuan-cao-swi! (and thank you everyone else for your patience)

In the near future, I'd like to see someone: 
* Update the API to include `#observe` methods for the Observable instruments
* Update the OTLP exporter to support the asynchronous instruments

metrics_api/lib/opentelemetry/metrics/instrument/observable_gauge.rb
```diff
@@ -9,7 +9,14 @@ module Metrics
     module Instrument
       # No-op implementation of ObservableGauge.
       class ObservableGauge
-        # TODO
+        # Observe the ObservableCounter with fixed timeout duartion.
```
kaylareopelle:
```suggestion
        # Observe the ObservableGauge with fixed timeout duration.
```

metrics_api/lib/opentelemetry/metrics/instrument/observable_counter.rb
```diff
@@ -9,7 +9,14 @@ module Metrics
     module Instrument
       # No-op implementation of ObservableCounter.
       class ObservableCounter
-        # TODO
+        # Observe the ObservableCounter with fixed timeout duartion.
```
kaylareopelle:
```suggestion
        # Observe the ObservableCounter with fixed timeout duration.
```

metrics_api/lib/opentelemetry/metrics/instrument/observable_up_down_counter.rb
```diff
@@ -9,7 +9,15 @@ module Metrics
     module Instrument
       # No-op implementation of ObservableUpDownCounter.
       class ObservableUpDownCounter
-        # TODO
+        # Observe the ObservableCounter with fixed timeout duartion.
```
kaylareopelle:
```suggestion
        # Observe the ObservableUpDownCounter with fixed timeout duration.
```

metrics_api/lib/opentelemetry/metrics/instrument/observable_up_down_counter.rb
```diff
@@ -9,7 +9,15 @@ module Metrics
...
+        # Everytime observe, the value should be sent to backend through exporter
```
kaylareopelle: This doesn't match with how I thought the asynchronous instruments worked. I thought you'd call `observe` once, perhaps in an initializer, and then it would be invoked on every export cycle by default or on intervals based on the provided timeout. Kind of like a heartbeat? 

Is it more like `record`, where you have to call it every time you want a measurement?

kaylareopelle (commented):
Thanks for adding this documentation! A few small edits and one question.

metrics_api/lib/opentelemetry/metrics/instrument/observable_up_down_counter.rb
xuan-cao-swi: Yes, you can record data by calling observe; and yes, the observe will be invoke every export cycle, not based on time interval. I think export based on time interval, just use periodical reader with async instrument (e.g. record cpu utilization).

This comment maybe misleading; I will update it.

xuan-cao-swi: User don't need to call observe to have it be invoked on every export cycle.
User probably can call observe if they really want to.

kaylareopelle: What do you think about taking out the second line for now, and we can add more documentation later? 

```suggestion
```

Issue #1484, opened 2023-06-28
Logs SDK - LogRecord 


kaylareopelle:
Implement the LogRecord specifications within the Logs SDK.

- [ ] Log Record Definition: https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition 
- [ ] ReadableLogRecord: https://opentelemetry.io/docs/specs/otel/logs/sdk/#readablelogrecord
- [ ] ReadWriteLogRecord: https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord

PR #1611, opened 2024-03-05
feat: Update Logger#on_emit to create LogRecords

kaylareopelle:
This PR creates a `LogRecord` class that is read/writeable to start with, and becomes read-only when it transitions to `LogRecordData`. 

Additionally, it provides the SDK implementation of the Logs Bridge API `Logger#on_emit` method to emit `LogRecord`s to the processing pipeline.

Relates to the following specs:
* [Log Record Definition](https://opentelemetry.io/docs/specs/otel/logs/data-model/#log-and-event-record-definition)
* [Readable Log Record](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readablelogrecord) 
* [Read Write Log Record](https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord)
* [Logger Operations: Emit a log record](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#emit-a-logrecord)

Resolves: https://github.com/open-telemetry/opentelemetry-ruby/issues/1484


logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
```diff
@@ -25,6 +27,56 @@ def initialize(name, version, logger_provider)
...
+        #   Boolean>, Hash{String => String, Numeric, Boolean, Array<String,
+        #   Numeric, Boolean>}] A value containing the body of the log record.
+        # @param attributes [optional Hash{String => String, Numeric, Boolean,
+        #   Array<String, Numeric, Boolean>}] Additional information about the
+        #   event.
+        #
+        # @api public
+        def on_emit(timestamp: nil,
+                 observed_timestamp: nil,
+                 span_context: nil, # or should this just be context? like in the API?
```
kaylareopelle, line 62: Discussion here: https://github.com/open-telemetry/opentelemetry-ruby/pull/1611/files#r1513443549

logs_sdk/lib/opentelemetry/sdk/logs/logger.rb
```diff
@@ -25,6 +27,56 @@ def initialize(name, version, logger_provider)
...
+                                     span_context: span_context ||= OpenTelemetry::Trace.current_span.context,
+                                     severity_text: severity_text,
+                                     severity_number: severity_number,
+                                     body: body,
+                                     attributes: attributes,
+                                     logger: self)
+
+          logger_provider.instance_variable_get(:@log_record_processors).each do |processor|
+            processor.on_emit(log_record, span_context)
+          end
```
kaylareopelle, line 78: I think this might not be the best implementation approach. I'd like some feedback here.

This made more sense when we were exposing `log_record_processors` with an `attr_reader`. However, we decided not to do that because [it might not be safe for concurrent execution.](https://github.com/open-telemetry/opentelemetry-ruby/pull/1517/commits/de7f36ce8fa8eab285dea579afe69e2501bfb635#r1346607101)

The [Span gets its processors defined as an attribute during creation](https://github.com/open-telemetry/opentelemetry-ruby/blob/98c629edbc6445305bed65c4f4306431d683b474/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L145), but it's also created by the TracerProvider rather than the Tracer. The Tracer is equivalent to the Logger in this scenario.

Also unlike Span, there isn't a `finish` operation separate from `start`, so [`emit` both creates the log record and sends it to the processing pipeline.](https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#emit-a-logrecord)

kaylareopelle: That's a great idea, @xuan-cao-swi. Incorporated in 1f9f9027219888422d638fc4c24b6be15f52dd78 (along with other context-related updates).

logs_api/lib/opentelemetry/logs/logger.rb
```diff
@@ -12,36 +12,45 @@ class Logger
...
+      #   with the current context.
+      # @param [optional Hash{String => String, Numeric, Boolean,
+      #   Array<String, Numeric, Boolean>}] attributes Additional information
+      #   about the event.
+      # @param [optional Context] context The Context to associate with the
+      #   LogRecord. Intended default: OpenTelemetry::Context.current
       #
       # @api public
-      def emit(
+      def on_emit(
```
kaylareopelle, line 43: When we renamed the method in https://github.com/open-telemetry/opentelemetry-ruby/pull/1517, I forgot to go back and change the references in the API.

robertlaurin (approved):

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -0,0 +1,99 @@
...
+          body: nil,
+          attributes: nil,
+          trace_id: nil,
+          span_id: nil,
+          trace_flags: nil,
+          logger: nil
+        )
+          @timestamp = timestamp
+          @observed_timestamp = observed_timestamp || timestamp || Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
+          @span_context = span_context
```
robertlaurin, line 67: Where is `span_context` coming from here?

kaylareopelle: Fixed in: 647e0c930921fec0b14282ce4ed7ca920cc083f7

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -0,0 +1,97 @@
...
+          @observed_timestamp = observed_timestamp || timestamp || Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
+          @severity_text = severity_text
+          @severity_number = severity_number
+          @body = body
+          @attributes = attributes.nil? ? nil : Hash[attributes] # We need a mutable copy of attributes
+          @trace_id = trace_id
+          @span_id = span_id
+          @trace_flags = trace_flags
+          @resource = logger&.resource
+          @instrumentation_scope = logger&.instrumentation_scope
```
robertlaurin, line 74: Is it possible for a log record to get created from an api logger? Otherwise I would expect instrumentation scope to always be present (do we need the safe operator here?).

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -0,0 +1,97 @@
...
+          @timestamp = timestamp
+          @observed_timestamp = observed_timestamp || timestamp || Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)
+          @severity_text = severity_text
+          @severity_number = severity_number
+          @body = body
+          @attributes = attributes.nil? ? nil : Hash[attributes] # We need a mutable copy of attributes
+          @trace_id = trace_id
+          @span_id = span_id
+          @trace_flags = trace_flags
+          @resource = logger&.resource
```
robertlaurin, line 73: I know there's a method added to the sdk logger here to support this but it still reads like `logger.logger_provider.resource` to me. 

I wonder if there's any merit in following the same pattern we did with the tracer https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/tracer.rb#L35

logs_sdk/lib/opentelemetry/sdk/logs/log_record_data.rb
```diff
@@ -0,0 +1,43 @@
...
+                                 #   Boolean>, Hash{String => String, Numeric, Boolean,
+                                 #   Array<String, Numeric, Boolean>}
+                                 :attributes, # optional Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}
+                                 :trace_id,              # optional String (16-byte binary)
+                                 :span_id,               # optional String (8-byte binary)
+                                 :trace_flags,           # optional Integer (8-bit byte of bit flags)
+                                 :resource,              # optional OpenTelemetry::SDK::Resources::Resource
+                                 :instrumentation_scope, # OpenTelemetry::SDK::InstrumentationScope
+                                 :total_recorded_attributes) do # Integer
+        def unix_nano_timestamp
```
xuan-cao-swi, line 25: nit: is that possible to do `alias_method :unix_nano_observed_timestamp, :unix_nano_timestamp`?

logs_sdk/lib/opentelemetry/sdk/logs/log_record_data.rb
```diff
@@ -0,0 +1,43 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # LogRecordData is a Struct containing {LogRecord} data for export.
+      LogRecordData = Struct.new(:timestamp,             # optional Integer nanoseconds since Epoch
+                                 :observed_timestamp,    # Integer nanoseconds since Epoch
```
fbogsany, line 12: The `timestamp` and `observed_timestamp` are documented here to be "Integer nanoseconds since Epoch", so there should be no need for the `unix_nano_timestamp` and `unix_nano_observed_timestamp` methods below.

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -0,0 +1,97 @@
...
+                      :resource,
+                      :instrumentation_scope
+
+        # Creates a new {LogRecord}.
+        #
+        # @param [optional Float, Time] timestamp Time when the event occurred.
+        # @param [optional Float, Time] observed_timestamp Time when the event
+        #   was observed by the collection system. If nil, will first attempt
+        #   to set to `timestamp`. If `timestamp` is nil, will set to
+        #   `Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)`.
```
fbogsany, line 30: Why do these accept `Float`? They should only accept `Time`, like corresponding methods in the tracing API. This makes it easier to ensure the correctness of conversion to "Integer nanoseconds since Epoch" internally, and reduces confusion about the units.

logs_api/lib/opentelemetry/logs/logger.rb
```diff
@@ -12,36 +12,45 @@ class Logger
...
       # Emit a {LogRecord} to the processing pipeline.
       #
-      # @param timestamp [optional Float, Time] Time in nanoseconds since Unix
+      # @param [optional Float, Time] timestamp Time in nanoseconds since Unix
       #   epoch when the event occurred measured by the origin clock, i.e. the
       #   time at the source.
-      # @param observed_timestamp [optional Float, Time] Time in nanoseconds
+      # @param [optional Float, Time] observed_timestamp Time in nanoseconds
       #   since Unix epoch when the event was observed by the collection system.
       #   Intended default: Process.clock_gettime(Process::CLOCK_REALTIME)
```
fbogsany, line 20: The units are unclear here. If we accept nanoseconds, I'd expect that to be Integer rather than Float, and if we accept `Time`, then the units are not relevant. IMO we should only accept `Time`, and we should convert that to "Integer nanoseconds since Epoch" internally.

Related: documenting the `Process.clock_gettime(Process::CLOCK_REALTIME)` as the default increases confusion about the units (and precision).

logs_sdk/lib/opentelemetry/sdk/logs/log_record_data.rb
kaylareopelle: Good question! The current setup [follows the pattern for export helpers in the SpanData class](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/span_data.rb#L34-L50).

They're not exactly aliasable because the `unix_nano_timestamp` method references the `timestamp` attribute on the LogRecordData Struct, whereas the other method uses the `observed_timestamp` attribute. I could refactor it to accept an argument, but then we'd be breaking from the pattern in SpanData. 

I'd be curious to hear what approach the maintainers prefer we take.

kaylareopelle: Whoops, commented on this before I saw Francis' comment. These methods will be removed.

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: @robertlaurin - From my reading, it seems like it's in-spec to allow an API Logger to emit a log record. 

Here's the part of the spec I'm working from: https://opentelemetry.io/docs/specs/otel/logs/bridge-api/#emit-a-logrecord 

What are your thoughts?

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: That's a great point, following that approach cleans up the code nicely. Fixed in ab570713d783b1dac3edea013c2d6f929cc03549

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: This isn't really relevant anymore after the refactor in ab570713d783b1dac3edea013c2d6f929cc03549

logs_sdk/lib/opentelemetry/sdk/logs/log_record_data.rb
kaylareopelle: Great point. Fixed in 0fb4606e20c8d71d4f2c47a71ac59898f4deb2e1

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: I got a little too overzealous when trying to adhere to the spec. The datatype in the Logs Data Model was listed as "Type: Timestamp, uint64 nanoseconds since Unix epoch." Since the timestamp could be efficiently accessed using `Process.clock_gettime(Process::CLOCK_REALTIME, :nanosecond)`, I decided to add it and make it the default.

Based on these questions, I decided to remove the `Float` option, go all in on the `Time` object and transform it into a nanosecond `Integer` in `LogRecord#to_log_record_data`. 

Fixes across two commits: 
* d3607592e8127ec0e6f0ca719a7d52649e7b5a5e 
* 0fb4606e20c8d71d4f2c47a71ac59898f4deb2e1

logs_api/lib/opentelemetry/logs/logger.rb
kaylareopelle: Great points, I agree. Addressed here: https://github.com/open-telemetry/opentelemetry-ruby/pull/1611#discussion_r1678505898

kaylareopelle:
@robertlaurin, @fbogsany - Thanks for your feedback! I've been able to return to this project and addressed your comments. This PR is ready for another review.

mwear (approved):
LGTM. @fbogsany can you confirm that your changes have been addressed?

mwear:
I'll plan to merge this EOD Thursday if there aren't any objections.

PR #1622, opened 2024-04-01
feat: Add severity number to logs-api

khushijain21:
This adds severity constant to log-api - this is required for logging bridges to map to OTel severity levels.


logs_api/lib/opentelemetry/logs/severity_number.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+module Logs
```
kaylareopelle, line 7: @khushijain21 - Could you nest the `Logs` module under the `OpenTelemetry` module?

```ruby
module OpenTelemetry
  module Logs
    class SeverityNumber
      # ...
    end
  end
end
```

I think I did a poor job of explaining the nesting earlier today. I'll try to find a blog post that can describe things better.

kaylareopelle (changes requested):
Thank you, @khushijain21! 

In addition to my comment in the PR, we need to require the new `severity_number` file in `logs_api/lib/opentelemetry/logs.rb`, like we do for other files in the `logs` directory:

https://github.com/open-telemetry/opentelemetry-ruby/blob/c0595b7cb80313bc72e194f4e969ef6c8b3f8471/logs_api/lib/opentelemetry/logs.rb#L7-L11 

Then I think this should be good to go.

kaylareopelle (approved):
I think this looks great, thank you @khushijain21!

mwear (approved):

PR #1623, opened 2024-04-02
feat: Add add_link to span api/sdk

dmathieu:
The specification has recently been updated to add the ability to add span links after span creation.

See:

* https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#add-link
* https://github.com/open-telemetry/opentelemetry-specification/pull/3678
* https://github.com/open-telemetry/opentelemetry-specification/pull/3887

This updates the API and SDK to add this new method.

Note: I am not fully sure how you all are handling additive changes to the API. With Ruby being very permissive with method definitions, I assume shipping everything together is fine.
If you think that could cause an issue (like the API being used with a lower version of the SDK), I'm happy to split this PR into two, so the SDK is shipped first, and the API in another release.


api/lib/opentelemetry/trace/span.rb
```diff
@@ -78,6 +78,28 @@ def add_attributes(attributes)
         self
       end
 
+      # Add a link to a {Span}.
+      #
+      # Adding links at span creation using the `links` option is preferred
+      # to calling add_link later, because head sampling decisions can only
+      # consider information present during span creation.
+      #
+      # Example>
```
kaylareopelle, line 87: Using a colon matches the documentation style for `Span#add_event`:
https://github.com/open-telemetry/opentelemetry-ruby/blob/6623a3e41534622fb30bd30ec7892ea2e1b4a6b8/api/lib/opentelemetry/trace/span.rb#L81-L85

```suggestion
      # Example:
```

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -118,6 +118,37 @@ def add_attributes(attributes)
           self
         end
 
+        # Add a link to a {Span}.
+        #
+        # Adding links at span creation using the `links` option is preferred
+        # to calling add_link later, because head sampling decisions can only
+        # consider information present during span creation.
+        #
+        # Example>
```
kaylareopelle, line 127: Same suggestion as above:
```suggestion
        # Example:
```

kaylareopelle (commented):
Thank you for bringing the Ruby project into compliance with this spec update!

I'll defer to the @open-telemetry/ruby-maintainers to your questions about API and SDK releases. 

I have a few small documentation comments. 

I'm also curious about this line in the spec: 
> Links added after Span creation may not be considered by [Samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampler).

Can you tell me how links added using the `add_link` method are ignored by Samplers?

dmathieu:
> Can you tell me how links added using the add_link method are ignored by Samplers?

[SDK samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampler), which perform head sampling (not collector samplers, or any other tail sampler) run on span creation. So when `add_link` is called, the head sampling decision has already been made, and is not reconsidered with the new link.

kaylareopelle:
> [SDK samplers](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampler), which perform head sampling (not collector samplers, or any other tail sampler) run on span creation. ...

Thank you, @dmathieu. This connects the dots!

kaylareopelle (approved):

sdk/lib/opentelemetry/sdk/trace/span.rb
```diff
@@ -118,6 +118,37 @@ def add_attributes(attributes)
...
+        # @param [OpenTelemetry::Trace::Link] the link object to add on the {Span}.
+        #
+        # @return [self] returns itself
+        def add_link(link)
+          @mutex.synchronize do
+            if @ended
+              OpenTelemetry.logger.warn('Calling add_link on an ended Span.')
+            else
+              @links ||= []
+              @links = trim_links(@links + [link], @span_limits.link_count_limit, @span_limits.link_attribute_count_limit)
```
fbogsany, line 145: `@links + [link]` will allocate 2 additional arrays (one to wrap `link` and one for the `+` result). It should be more efficient to `@links << link`, which will append to the `@links` array in place. The potentially inconsistent intermediate state should be invisible to observers due to the mutex.

dmathieu: The links array is frozen in `trim_links`. So in order to do that, I'd have to stop freezing it.

fbogsany: yes - stop freezing it because it is now mutable (it wasn't before). It should be frozen in `finish` instead.

fbogsany (approved):

PR #1624, opened 2024-04-03
Fix/change some messaging attributes to match conventions

s22su:
I used `opentelemetry-semantic_conventions` in one of my projects and noticed that some attributes are different from the docs on following pages:
- https://opentelemetry.io/docs/specs/semconv/messaging/messaging-spans/#messaging-attributes
- https://opentelemetry.io/docs/specs/semconv/messaging/kafka/

All kafka attributes described in docs are present in `opentelemetry-semantic_conventions` gem now.
Messaging batch count is added because I wat to use it in my code and it was missing before.

### DIFF
--- a/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb
+++ b/semantic_conventions/lib/opentelemetry/semantic_conventions/trace.rb
@@ -421,6 +421,9 @@ module OpenTelemetry
       # @note If the key type is not string, it's string representation has to be supplied for the attribute. If the key has no unambiguous, canonical string form, don't include its value
       MESSAGING_KAFKA_MESSAGE_KEY = 'messaging.kafka.message.key'
 
+      # The offset of a record in the corresponding Kafka partition
+      MESSAGING_KAFKA_MESSAGE_OFFSET = 'messaging.kafka.message.offset'
+
       # Name of the Kafka Consumer Group that is handling the message. Only applies to consumers, not producers
       MESSAGING_KAFKA_CONSUMER_GROUP = 'messaging.kafka.consumer.group'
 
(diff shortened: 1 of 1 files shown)


kaylareopelle:
👋 Hi @s22su! Thank you for your contribution! 

I'm not sure about the exact status of the semantic conventions gem right now. The approvers/maintainers spoke in January about wanting to rework it. I'll bring your PR up during the SIG meeting tomorrow and try to get some next steps.

kaylareopelle:
Hi @s22su, I'm sorry for taking so long to get back to you on this.

Our approach to semantic conventions is in flux. Our current approach is brittle for a number of reasons, including we don't have a great way to handle removals of constants without potentially breaking things for users upstream. 

We're grateful that you pointed out the changes in the messaging conventions and opened this PR. Your PR reinvigorated our discussion about #1567, opened by @robbkidd. This PR presents an alternate approach for the semantic conventions gem.

As an alternative to your PR, we'd like to update the messaging conventions portion of #1567 to include the latest content.

Would you be willing to test out that PR, once we've updated it, to see if it works in your environment?

s22su:
@kaylareopelle it seems to me like the intended usage of the gem after linked PR is merged, will be to update `opentelemetry-semantic_conventions` gem version (I believe it will be a major version bump) and use `OpenTelemetry::SemanticConventions_1_20_0::Trace::MESSAGING_KAFKA_MESSAGE_KEY` instead of `OpenTelemetry::SemanticConventions::Trace::MESSAGING_KAFKA_MESSAGE_KEY` so not sure what is there is to test.

When next version will be released I'll happily use it in our internal library. Feel free to close this PR, I'll keep an eye on #1567 to know when I can update.

kaylareopelle:
Thanks for your understanding, @s22su.

PR #1638, opened 2024-05-21
docs: Add ruby compatibility documentation

dmathieu:
This is a proposal to document the supported ruby versions in the README, so folks know what to expect.


kaylareopelle (approved):
I think this is a great addition. Thank you!

fbogsany (approved):

PR #1648, opened 2024-06-21
fix: Add csv as a dependency for Ruby 3.4

michaelherold:
csv has been a default gem for a long while. In Ruby 3.4, it will stop being a default gem and will need to be explicitly required as a dependency.

> warning: /opt/rubies/3.3.1/lib/ruby/3.3.0/csv.rb was loaded from the standard library, but will no longer be part of the default gems since Ruby 3.4.0. Add csv to your Gemfile or gemspec. Also contact author of opentelemetry-exporter-otlp-0.26.1 to add csv into its gemspec.

This change adds it as a dependency for the OTLP exporter since it makes use of the library. It uses the same version constraint as the Zipkin exporter for consistency.

See eb89cd62f5df71886651f87f38824cbcb6e47251 for a similar change.

### DIFF
--- a/exporter/otlp/opentelemetry-exporter-otlp.gemspec
+++ b/exporter/otlp/opentelemetry-exporter-otlp.gemspec
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
   spec.require_paths = ['lib']
   spec.required_ruby_version = '>= 3.0'
 
+  spec.add_dependency 'csv', '~> 3.1'
   spec.add_dependency 'googleapis-common-protos-types', '~> 1.3'
   spec.add_dependency 'google-protobuf', '>= 3.18'
   spec.add_dependency 'opentelemetry-api', '~> 1.1'


kaylareopelle:
Thanks for submitting this PR, @michaelherold! CSV is no longer a dependency of `opentelemetry-exporter-otlp`. If you update the gem to version 0.26.2 or above, the warning should go away.

michaelherold:
Hi @kaylareopelle! I double checked when I submitted and [this](https://github.com/open-telemetry/opentelemetry-ruby/blob/8a16f1fc18dd0b4365ecc58e3517caee5e274a71/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics_exporter.rb#L10) is where the require happens. I haven't vetted when or where it is used, though; perhaps it's not used anywhere but the require was left behind?

kaylareopelle:
Ah, thank you! That link was very helpful. That file is for the opentelemetry-exporter-otlp-**metrics** gem, not the opentelemetry-exporter-otlp gem.

It looks like we're not using the CSV library in gem anymore, so I think the require was left behind. 

Could you update this PR to remove the `require 'csv'` line from [exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics_exporter.rb](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics_exporter.rb)?

kaylareopelle:
Hi, @michaelherold  - The `require 'csv'` line will be removed by https://github.com/open-telemetry/opentelemetry-ruby/pull/1670 

The warning will still appear on `opentelemetry-exporter-otlp-0.26.1`. The warning should no longer be reported on  versions 0.26.2+ versions of that gem, as well as `opentelemetry-exporter-otlp-metrics` (when it is released).

kaylareopelle:
Closing, as #1670 has been merged.

PR #1651, opened 2024-07-10
feat(semconv): update to v1.25 in a non-breaking way

robbkidd:
Now that semconv won't have deprecated names removed without a major version bump, here's a go at updating semantic conventions to 1.25.

* no `const_missing` hijinks! (like in #1567)
* pulls from the new semconv repo

* updates jinja2 template for the new structure of convention data

* stable semconv names will appear in OpenTelemetry::SemanticConventions
  while all semconv names (experimental, deprecated, stable, etc) appear
  in OpenTelemetry::SemanticCandidates
    - This maps to what other langs (Java, Python, JS) are putting into
      "Incubating" modules. We opted for "candidates" because
      OpenTelemetry::SemanticConventions::Incubating::... was getting
      too long and OpenTelemetry::SemanticIncubating didn't make much
      sense.

* generates a module per root semconv namespace to make navigating docs
  more focused and to decrease the number of constants loaded when
  using more precise requires in downstream code

* generates only attribute name constants and to an attributes subfolder
  to prepare for metric name constants to be generated to metrics
  subfolders, also for decreasing the scope of what gets loaded when
  downstream instrumentation requires these constants

### Some TODOs

A few of these appear in the code in this PR, too.

- [ ] usage examples. How weird is it to explicitly require the nested modules?
- ~test(s) to make sure the trace and resource constants are present in SemanticCandidates~
  - See dc42bf59e86b8ef767ab435a92d19646a04ceab6. I did the best I could with a test to find which existing v1.10 constants had no match in the entirety of 1.25. I used that list of missing consts to update their deprecation notices accordingly.
- [x] test(s) to make sure the `SemanticConventions` (stable) constants are all still present in the `SemanticCandidates` constants
- [x] add deprecation notices to all the `SemanticConventions::(Resource|Trace)` constants with pointers to their replacements
- [ ] figure out how to `Firstcap` or `PascalCase` the module names in the jinja2 template with exceptions for things like AspNetCore, HTTP, etc?
  - The module name will become a part of the constants interface, so the fancier we make the code generation logic, the more likely something breaks and we release a version where module names have a new backwards-breaking shape.
  - 👎🏻 I vote we don't do this and have non-idiomatic `SCREAMING` module names for simplicity of up-keep.
- [x] Do something about JRuby
  - ~Turn off testing constants exist on JRuby?~
  - Find a markdown parser/renderer (1) supported by Yard, (2) supports inline-code-blocks-inside-links in markdown content, and (3) works on JRuby?
    - Found kramdown. Using it instead of redcarpet.


semantic_conventions/lib/opentelemetry/semantic_candidates/attributes/aws.rb
```diff
@@ -0,0 +1,293 @@
...
+    
+      # The value of the `TotalSegments` request parameter
+      #
+      # @note StabilityLevel.EXPERIMENTAL
+      TOTAL_SEGMENTS = 'aws.dynamodb.total_segments'
+    
+      # The ARN of an [ECS cluster](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/clusters.html)
+      #
+      # @note StabilityLevel.EXPERIMENTAL
+      CLUSTER_ARN = 'aws.ecs.cluster.arn'
```
robbkidd, line 139: Found one flaw (repeated in several places): using attribute.attr_id in the j2 template is not enough. "ecs" is omitted from the `attr_id`, it's [present in the prefix for the attribute group](https://github.com/open-telemetry/semantic-conventions/blob/v1.26.0/model/registry/aws.yaml#L257-L268). 😭

robbkidd:
Probably going to dial this back from semconv v1.26 to v1.25 because there's some name hijinks going on in `db.*` in 1.26. At least for this initial cut at catching Ruby's semconv library up to recent.

robbkidd:
Guh. [New markdown processor](https://github.com/open-telemetry/opentelemetry-ruby/pull/1651/commits/c3c116fb6ec8d210d4448a96db41f3ea07d16a14) (redcarpet) for yard to use in doc generation doesn't support JRuby [because native extensions](https://github.com/open-telemetry/opentelemetry-ruby/actions/runs/9911194737/job/27383322016?pr=1651#step:7:99).

```
An error occurred while installing redcarpet (3.6.0), and Bundler cannot
continue.

In Gemfile:
  redcarpet
```

robbkidd:
Taking this out of draft. It may not be ready to merge, but it's ready for people to review and form opinions about.

robbkidd:
Taking this out of draft. It may not be ready to merge, but it's ready for people to review and form opinions about.

There's a lot going on here, so [walking the commits](https://github.com/open-telemetry/opentelemetry-ruby/pull/1651/commits) might help. I left notes about the whats and whys of each change in the commit messages.

JamieDanielson:
> * stable semconv names will appear in OpenTelemetry::SemanticConventions
>   while all semconv names (experimental, deprecated, stable, etc) appear
>   in OpenTelemetry::SemanticCandidates
>   
>   * This maps to what other langs (Java, Python, JS) are putting into
>     "Incubating" modules. We opted for "candidates" because
>     OpenTelemetry::SemanticConventions::Incubating::... was getting
>     too long and OpenTelemetry::SemanticIncubating didn't make much
>     sense.
> *

Wanted to note while I'm looking at it - there's a chance we may go with Incubating in JS. If we do, it may be worth reconsidering "candidates" to help with consistency, and also to follow what seems to be [guidance provided from semconv on artifact structure](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/non-normative/code-generation.md#semantic-conventions-artifact-structure)

> * Artifact name:
>    * opentelemetry-semconv - stable conventions
>    * opentelemetry-semconv-incubating - (if applicable) the preview artifact containing all (stable and experimental) conventions
> * Namespace: opentelemetry.semconv and opentelemetry.semconv.incubating

So maybe OpenTelemetry::SemConvIncubating:: as an option?

kaylareopelle:
@JamieDanielson  - I like the idea of mirroring the package name more closely and reducing `SemanticConventions` to `SemConv`. That'd give our fingers a bit of a break. 

So to match the artifact namespace more closely, it could also be: 
`OpenTelemetry::SemConv` - stable constants 
`OpenTelemetry::SemConv::Incubating` - others

robbkidd:
Yea. "Candidates" was probably never going to fly. This whole thing is about trying to be _conventional_.

I don't hate shortening `SemanticConventions` to `SemConv`.

kaylareopelle (changes requested):
This looks good! 

As we discussed in the comments, the plan is to update the constant prefixes to: 
`OpenTelemetry::SemConv` - stable constants
`OpenTelemetry::SemConv::Incubating` - others

Once that's done, I think this will be ready to go. Adding the "request changes" label to make it easier to see.

kaylareopelle:
Pairing recap! 

@robb - Next steps: 
* Add some post-processing using Rake/Ruby to generate a rollup file for each namespace to require the `attributes.rb` and `metrics.rb` files
  * Use Ruby file tools to look at all the new directories that were created
  * Produce new files with those names at the same level as the directory (ex. `semconv/http` is the existing directory and `semconv/http.rb` will be the generated the rollup file)
  * Within the rollup file, require the `attributes.rb` and `metrics.rb` files 
    * (maybe an `ls` on the root namespace directory may help verify the presence of an `attributes.rb` and `metrics.rb` within each namespace?)
  * _Maybe_ add the empty module definition for the namespace in the rollup file too:
```ruby
module OpenTelemetry
  module SemConv
    module HTTP
    end
  end
end
``` 

Questions: 
* Do we want some sort of identifier in the name of the constant to distinguish whether the constant is an attribute or a metric?
  * We have a label for templates (suffix `*_LAMBDA`)
  * The constant would look something like: `OpenTelemetry::SemConv::HTTP::HTTP_SERVER_REQUEST_DURATION` for a metric and `OpenTelemetry::SemConv::HTTP::HTTP_STATUS_CODE` for an attribute, 
  * To match the file path for module name nesting: `OpenTelemetry::SemConv::HTTP::Attribute::HTTP_STATUS_CODE`
  * To match the augment the constant name: `OpenTelemetry::SemConv::HTTP::HTTP_STATUS_CODE_ATTR`

Other stuff:
* Check to see if "Now available in the stable namespace..." comment in the incubating code is hyperlinkable to the stable constant 
* Can you put comments within a Heredoc? (docker command)

trask:
hi @kaylareopelle @robbkidd! just checking what's the latest status of this PR? we are hoping to remove/archive the old semantic convention build tooling in the near-ish future, thanks!

lmolkova:
Heads up on the semconv change https://github.com/open-telemetry/semantic-conventions/pull/2047.

If merged, semantic conventions (starting with next release) will become incompatible with the build-tools. Weaver will become the only tool that can generate them.

kaylareopelle:
Thanks for the updates, @lmolkova and @trask!

kaylareopelle:
Following up [from Slack](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1755717448458099?thread_ts=1755540447.721939&cid=C01NWKKMKMY), here's a PR with my tweaks that bump the version to 1.36.0, which was the latest semconv version last week. 

https://github.com/robbkidd/opentelemetry-ruby/pull/1 

The script worked great! I made a few small tweaks to account for formatting (rubocop) and the tests. Otherwise, this seems very easy to maintain as long as weaver remains stable. I'm comfortable approving this PR with those changes. I think we should also update the README to reflect this new structure, which I would be happy to do.

What else do we need to consider before we merge this? What am I missing?

robbkidd:
Guh. _Linters_, amirite?

semantic_conventions/Gemfile
```diff
@@ -13,4 +13,5 @@ eval_gemfile '../contrib/Gemfile.shared'
 group :development, :test do
   gem 'opentelemetry-api', path: '../api'
   gem 'pry'
+  gem 'mutex_m' if RUBY_VERSION >= '3.4'
```
kaylareopelle: 🤖 🙏 

```suggestion
  gem 'mutex_m' if RUBY_VERSION >= '3.4'
  gem 'opentelemetry-api', path: '../api'
  gem 'pry'
```

semantic_conventions/test/opentelemetry/semantic_conventions/attributes_test.rb
```diff
@@ -0,0 +1,26 @@
...
+    .reject { |const| const == :Incubating }
+    .each do |root_namespace|
+      describe "#{root_namespace} stable constants still exist in SemConv::Incubating" do
+        OpenTelemetry::SemConv
+          .const_get(root_namespace)
+          .constants
+          .each do |stable_const|
+            it stable_const.to_s do
+              candidate_namespace = OpenTelemetry::SemConv::Incubating.const_get(root_namespace)
+              assert_includes candidate_namespace.constants, stable_const, "Missing stable constant in incubating: #{stable_const}"
```
kaylareopelle, line 22: 🤖 🙏 
```suggestion

              assert_includes candidate_namespace.constants, stable_const, "Missing stable constant in incubating: #{stable_const}"
```

kaylareopelle (approved):
🎊

kaylareopelle:
Whenever we want to look at README improvements, here's a draft: https://github.com/robbkidd/opentelemetry-ruby/pull/2

Issue #14, opened 2019-07-11
API: decide what to do for invalid arguments


fbogsany:
I'm not sure what we should do for invalid arguments in the minimal implementation, and am somewhat torn on the issue.

I just wanted to point out that there is a proposal currently in PR form here: https://github.com/open-telemetry/opentelemetry-specification/pull/153.

There seem to be various opinions as you read through the comments. This [comment](https://github.com/open-telemetry/opentelemetry-specification/pull/153#discussion_r296447000) proposed a somewhat pragmatic approach of falling back to a default when an argument is invalid. Quote (from the comment)

> Specific case: I do think using a null name for Span could fallback to using a default name, instead of throwing. Documenting it is nice, but providing defaults for these kind of things is a good alternative ;)

It might be too early to make a final call on how to deal exceptions in the minimal implementation, but we should keep an eye on that PR and the specification repo for guidance.

_Originally posted by @mwear in https://github.com/open-telemetry/opentelemetry-ruby/diffs_

fbogsany:
We settled on removing most runtime checks #115 and adding a "strict mode" #126

PR #1657, opened 2024-07-30
feat: coerce OpenTelemetry::Trace::Status.description to String

flux-ricky:
Accepting non string values results in obscure errors when exporting traces.

----

The motivation for this PR is due to me spending a chunk of hours trying to track down an error I was seeing in our logs.

The error was: `OpenTelemetry error: unexpected error in OTLP::Exporter#encode - Invalid argument for string field 'message' (given Class).`

This was pretty obscure to me it took a while to get to the bottom of it. As it turned out it was because I was passing a class to Status like so `OpenTelemetry::Trace::Status.error(e.class)`. I dunno how I didn't see this issue before in testing or if behaviour has changes in recent versions, but in any case it was painful to debug especially as someone not familiar with the OTEL API/SDK.

Unsure what the maintainers will think of this but I feel the DX could be improved with more leniency in what is accepted or improved logging.

### DIFF
--- a/api/lib/opentelemetry/trace/status.rb
+++ b/api/lib/opentelemetry/trace/status.rb
@@ -59,7 +59,7 @@ module OpenTelemetry
       # @param [String] description
       def initialize(code, description: '')
         @code = code
-        @description = description
+        @description = description.to_s
       end
 
       # Returns false if this {Status} represents an error, else returns true.
--- a/api/test/opentelemetry/trace/status_test.rb
+++ b/api/test/opentelemetry/trace/status_test.rb
@@ -26,6 +26,11 @@ describe OpenTelemetry::Trace::Status do
       status = OpenTelemetry::Trace::Status.ok('ok')
       _(status.description).must_equal('ok')
     end
+
+    it 'coerces the value passed in' do
+      status = OpenTelemetry::Trace::Status.ok(true)
+      _(status.description).must_equal('true')
+    end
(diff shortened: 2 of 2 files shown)


kaylareopelle:
Thanks for your submission, @flux-ricky! I'm sorry this was such an ordeal to track down. I've run into issues debugging the `OTLP::Exporter#encode` error messages too, and think we should do a better job describing what's going on when one of those errors are raised. I'll open an issue to work on this further.

We log when span attribute key/value pairs are the wrong class [(example)](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/internal.rb#L51-L60). Perhaps that's the right move here too.  

I think this is a good workaround for now. I'd like to hear what the other approvers/maintiners think.

kaylareopelle:
Issue to track improved error messaging: https://github.com/open-telemetry/opentelemetry-ruby/issues/1669

mwear (changes requested):
We decided early on in this project to avoid excessive type checking for arguments see (#14, #203). The same philosophy generally holds for coercion as well. While this change is minimal, it's not something we want to add everywhere, and would recommend fixing callers of the API to pass in a string directly. We should focus on making this more easily debuggable. Hopefully #1674 is a step in that direction. I'm going to suggest that we close this PR if there aren't any strong objections.

PR #1661, opened 2024-07-31
Add code spell task, fix issues and run the task in CI

dmathieu:
This adds [codespell](https://github.com/codespell-project/codespell) to the project, which allows finding spelling issues within code.
It also fixes every spelling issue found, and runs the task within CI.


dmathieu:
The failing test appears unrelated.

kaylareopelle (approved):
Thanks for your contribution, @dmathieu! 

codespell is new to me. [CSpell was added to contrib](https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1061) to mimic what [OpenTelemetry.io uses](https://opentelemetry.io/docs/contributing/style-guide/#spell-checking), but it looks like the OTel Go and Python projects both use this tool.

I don't think we need to use the same tool in contrib and core. CSpell is still pretty new to contrib, so maybe this is a good opportunity to see which one we prefer and align the two later.

That the `.codespellignore` is so short leans me toward this tool.

dmathieu:
cspell is different. It checks for spelling errors in written text (documentation, markdown documents). That makes sense for opentelemetry.io, but not so much for a codebase.

mwear (approved):

Issue #1487, opened 2023-06-28
Logs SDK - LogRecordExporter


kaylareopelle:
Implement the LogRecordExporter for the Logs SDK.

https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordexporter

- [ ] Export - https://opentelemetry.io/docs/specs/otel/logs/sdk/#export  
- [ ] ForceFlush - https://opentelemetry.io/docs/specs/otel/logs/sdk/#forceflush-2 
- [ ] Shutdown - https://opentelemetry.io/docs/specs/otel/logs/sdk/#shutdown-2

kaylareopelle:
WIP, unreviewed code for this issue, along with other parts of the logs SDK, can be found in https://github.com/kaylareopelle/opentelemetry-ruby/pull/6

PR #1666, opened 2024-07-31
feat: Add log record exporter interface

kaylareopelle:
This PR adds the duck type for the log record exporter.

Spec: https://opentelemetry.io/docs/specs/otel/logs/sdk/#logrecordexporter 

Closes: #1487


logs_sdk/lib/opentelemetry/sdk/logs/export.rb
```diff
@@ -7,6 +7,9 @@
 module OpenTelemetry
   module SDK
     module Logs
+      ExportError = Class.new(OpenTelemetry::Error)
```
mwear, line 10: In order for this to match the convention in traces, this should be a `Logs::Export::ExportError` (needs to be nested under the Export module).

see: https://github.com/open-telemetry/opentelemetry-ruby/blob/92748f2b74c510d55c50b226380f9f1c51db4a4a/sdk/lib/opentelemetry/sdk/trace/export.rb#L9-L14

kaylareopelle: Thank you! Fixed in 425ab1c6ebf4287f43f26a87caaff20c6f603f02

logs_sdk/lib/opentelemetry/sdk/logs/export.rb
```diff
@@ -7,8 +7,10 @@
 module OpenTelemetry
   module SDK
     module Logs
-      # The Export module contains result codes for exporters
+      # The Export module contains the built-in exporters and log record
+      # processors for the OpenTelemetry reference implementation.
```
kaylareopelle, line 11: This more closely matches the description of the Trace Export module.

mwear (approved):

Issue #1669, opened 2024-08-01
Improve error reporting for `OTLP::Exporter#encode`


kaylareopelle:
Currently, we get error messages like, `OpenTelemetry error: unexpected error in OTLP::Exporter#encode - Invalid argument for string field 'message' (given Class)` when there's an error in the encoding step for traces/spans. 

We've had multiple reports that this error message does not contain enough information to easily debug the problem (see: #1657 as one example). 

Improve the error messages, or provide an easy option to get enhanced messages for users who aren't satisfied with the current messages.

For example, when I'm debugging locally, I often add the backtrace to the output to help locate the issue.

kaylareopelle:
Closed by #1674, but feel free to reopen if this doesn't seem like a sufficient default.

PR #1674, opened 2024-08-09
feat: include backtrace first line for better debug info

xuan-cao-swi:
Propose a solution/improvement for https://github.com/open-telemetry/opentelemetry-ruby/issues/1669.

Since we are passing the exception class into `handle_error`, I think it's beneficial to include the last line of the error in the log message for developers to track down the cause.

I believe a full backtrace is too much information to show; however, the end user can always customize their `@error_handler` to include it.

I didn't add error handling in the encoding function like `as_otlp_span` because:
1. After rescuing, we need to provide a replacement value for `spans: sds.map { |sd| as_otlp_span(sd) }`; otherwise, the `encode` function will raise another type of error, which I think is redundant information.
2. If we provide an empty `Opentelemetry::Proto::Trace::V1::Span.new` like `as_otlp_key_value` does, it will raise a single error, but the encode will not throw an error and will export incorrect data (e.g., an empty span), which I think is misleading.


kaylareopelle (dismissed):
I like this! Thank you! 

As an example, here is the output when I reproduce a known error related to the `examples/metrics_sdk/metrics_collect_otlp.rb` file when no attributes are given to `histogram.record`:
```
E, [2024-08-09T15:03:07.514396 #44506] ERROR -- : OpenTelemetry error: unexpected error in OTLP::MetricsExporter#encode - undefined method `map' for nil - /Users/kreopelle/dev/opentelemetry-ruby/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics_exporter.rb:274:in `histogram_data_point'
```

xuan-cao-swi:
Thanks,
> If we provide an empty Opentelemetry::Proto::Trace::V1::Span.new like as_otlp_key_value does, it will raise a single error, but the encode will not throw an error and will export incorrect data (e.g., an empty span), which I think is misleading.

I took second thought, and I think it's ok to pass empty Opentelemetry::Proto::Trace::V1::Span or nil, and add check like [L126](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-exporter-otlp/v0.28.1/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L126)

mwear (approved):

PR #1698, opened 2024-08-28
chore: Loosen metrics gems versions on otlp exporter

kaylareopelle:
The OTLP metrics exporter fails to install for tests because the latest metrics SDK is newer than the previous version constraint. The rest of the gems are installed based on minor versions. Now metrics_sdk/metrics_api are in sync.


kaylareopelle:
Example log message from GHA: 

```
Because opentelemetry-metrics-sdk >= 0.1.0, < 0.2.A could not be found in source
at `../../metrics_sdk`
and every version of opentelemetry-exporter-otlp-metrics depends on
opentelemetry-metrics-sdk ~> 0.1.0,
  opentelemetry-exporter-otlp-metrics cannot be used.
So, because Gemfile depends on opentelemetry-exporter-otlp-metrics >= 0,
  version solving has failed.
```

mwear (approved):

PR #1703, opened 2024-08-28
feat: Explicit bucket boundary advisory parameter

kaylareopelle:
The advisory parameter `:explicit_bucket_boundaries` allows instrumentation authors to set bucket boundaries for histograms that differ from the default bucket set. Since this only applies to histograms, advice is not explicitly assigned in other instruments, so shared methods have an optional argument that defaults to nil. 

Any key passed to advice will be accepted, but only the `:explicit_bucket_boundaries` symbol will have an effect.

* [API Spec](https://opentelemetry.io/docs/specs/otel/metrics/api/#instrument-advisory-parameters)
* [SDK Spec](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#instrument-advisory-parameters)


metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
```diff
@@ -8,19 +8,22 @@ module OpenTelemetry
   module Internal
     # @api private
     class ProxyInstrument
-      def initialize(kind, name, unit, desc, callable)
+      def initialize(kind, name, unit, desc, callable, advice = nil)
```
kaylareopelle, line 11: I'm not sure what the ideal default argument is. We expect advice to be a hash when it's present, but that might cause a lot of unnecessary object allocations. 

Right now, the API defaults to nil and the SDK defaults to an empty hash.

kaylareopelle, line 11: Also, `advice` is shorter than `advisory_parameters`, so that's why I chose that word. However, it's not the latest terminology used in the spec. I'm happy to change to `advisory_parameters` if that's preferred. Perhaps some documentation would clarify things.

metrics_sdk/test/opentelemetry/sdk/metrics/instrument/histogram_test.rb
```diff
@@ -35,4 +35,29 @@
...
+    _(last_snapshot[0].name).must_equal('histogram')
+    _(last_snapshot[0].unit).must_equal('s')
+    _(last_snapshot[0].description).must_equal('test')
+    _(last_snapshot[0].instrumentation_scope.name).must_equal('test')
+    _(last_snapshot[0].data_points[0].count).must_equal(5)
+    _(last_snapshot[0].data_points[0].sum).must_equal(40)
+    _(last_snapshot[0].data_points[0].min).must_equal(0)
+    _(last_snapshot[0].data_points[0].max).must_equal(20)
+    _(last_snapshot[0].data_points[0].attributes).must_be_nil
+    _(last_snapshot[0].aggregation_temporality).must_equal(:delta)
```
kaylareopelle, line 61: idk if we need all these assertions, when really we just want to make sure the bucket was assigned properly. it felt safer JIC adding the new bucket boundaries somehow interfered with the bucket-related calculations (ex. `sum`)

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
mwear: Would a constant, empty, frozen hash be an option?

kaylareopelle: Good idea! Updated in d2318b0da04dfea617ff52911ddc6deb760bfe00

Issue #1695, opened 2024-08-28
Add `OnEnding` callback to allow processors to mutate spans on the end


kaylareopelle:
> `OnEnding`
Status: [Development](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/document-status.md)
>
> `OnEnding` is called during the span End() operation, after the end timestamp has been set. The Span object is still mutable (i.e., `SetAttribute`, `AddLink`, `AddEvent` can be called) while `OnEnding` is called. This method MUST be called synchronously within the [`Span.End()` API](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#end), therefore it should not block or throw an exception. If multiple `SpanProcessors` are registered, their `OnEnding` callbacks are invoked in the order they have been registered. The SDK MUST guarantee that the span can no longer be modified by any other thread before invoking `OnEnding` of the first `SpanProcessor`. From that point on, modifications are only allowed synchronously from within the invoked `OnEnding` callbacks. All registered `SpanProcessor` `OnEnding` callbacks are executed before any `SpanProcessor`'s `OnEnd` callback is invoked.
>
> Parameters:
>
> `span` - a [read/write span object](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#additional-span-interfaces) for the span which is about to be ended.
> Returns: `Void`

Spec: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onending
PR describing the Spec: https://github.com/open-telemetry/opentelemetry-specification/pull/4024

pantuza:
Hi, @kaylareopelle . I would love to try adding this feature.

kaylareopelle:
Thanks @pantuza! You're assigned 🎉

PR #1713, opened 2024-09-01
Adds on_ending callback to allow processors to mutate spans before End operation

pantuza:
This Pull Request aims to solve the issue #1695 by adding the `on_ending` callback to `SpanProcessor` class.
It also make sure to call that method during the span End operation (`finish()` method), right after the end
timestamp has been set but while the span is still mutable.

Based on [the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#onending), this implementation complies with:
- The `on_ending` callback is optional within ANY `SpanProcessor`. Only called when exists.
- called during the span End() operation
- Span object is still mutable
- MUST be called synchronously within the Span End operation
-  it should not block or throw an exception
- MUST guarantee that the span can no longer be modified by any other thread before invoking `on_ending`
- modifications are only allowed synchronously from within the invoked `on_ending`
- SpanProcessor `on_ending` callbacks are executed before any SpanProcessor's `on_end` callback is invoked


sdk/lib/opentelemetry/sdk/trace/span_processor.rb
```diff
@@ -23,6 +23,20 @@ class SpanProcessor
...
+        #
+        # This method is called synchronously and should not blocks the current
+        # thread nor throw exceptions.
+        #
+        # This method is optional on the Span Processor interface. It will only
+        # get called if it exists within the processor.
+        #
+        # @param [Span] span the {Span} that just is ending (still mutable).
+        # @return [void]
+        def on_ending(span); end
```
dmathieu, line 38: This is an experimental feature and may have breaking changes. How about specifying that here ?

pantuza: Thank you for the suggestion. I will update it.

sdk/lib/opentelemetry/sdk/trace/span_processor.rb
```diff
@@ -23,6 +23,22 @@ class SpanProcessor
         #  started span.
         def on_start(span, parent_context); end
 
+        # The on_ending is an experimental feature and may have breaking changes.
```
kaylareopelle, line 26:
```suggestion
        # The on_ending method is an experimental feature and may have breaking changes.
```

sdk/lib/opentelemetry/sdk/trace/span_processor.rb
```diff
@@ -23,6 +23,22 @@ class SpanProcessor
...
+        #
+        # Called when a {Span} is ending, after the end timestamp has been set
+        # but before span becomes immutable. This allows for updating the span
+        # by setting attributes or adding links and events.
+        #
+        # This method is called synchronously and should not blocks the current
```
kaylareopelle, line 32:
```suggestion
        # This method is called synchronously and should not block the current
```

kaylareopelle (changes requested):
Thanks for your PR, @pantuza! I want to think about this a little more, but I have some feedback on the documentation for now.

pantuza:
> Thanks for your PR, @pantuza! I want to think about this a little more, but I have some feedback on the documentation for now.

Thank you for your feedback and documentation suggestions. I really appreciate it. 
Take the necessary time to think about the PR and please, let me know in case you think there is anything left to be done in this PR. 

To give you a little background context, have started this work because of the following discussion in this Pull Request #1713

robbkidd:
> … have started this work because of the following discussion in this Pull Request https://github.com/open-telemetry/opentelemetry-ruby/pull/1713 [👀]

1713 is _this_ pull request. Is there a different discussion you meant to link to?

pantuza:
> > … have started this work because of the following discussion in this Pull Request #1713 [👀]
> 
> 1713 is _this_ pull request. Is there a different discussion you meant to link to?

You are right! Thank you for pointing it out. The correct discussion is in #1690

mwear (approved):
Overall, this looks good to me. For reference, Java already has this: https://github.com/open-telemetry/opentelemetry-java/pull/6367. We chose to rename the `end` method to `finish` since `end` is a Ruby keyword. One thought I had was whether or not we should considering calling this `on_finishing`, but I am ok with this as is. Curious if anyone else has any thoughts.

kaylareopelle (approved):
Thank you!

kaylareopelle:
> One thought I had was whether or not we should considering calling this `on_finishing`, but I am ok with this as is. Curious if anyone else has any thoughts.

I do like the idea of using `on_finishing` here to keep the same term around, but don't feel strongly one way or the other. If we use `on_finishing` instead, we may want to add a link to the SDK documentation or "On Ending" to clear up any confusion that might exist connecting the `on_finishing` method to its specification.

pantuza:
I also like the idea of using `on_finishing` to keep names concise with the project decisions. 
If you don't mind I will update the code that way and make sure to document it properly.

PR #1725, opened 2024-09-10
feat: OpenTelemetry.logger_provider API, ProxyLoggers, Configuration, and Instrument Registry

### LINKED ISSUE #1699: Add Logs Configuration Patch
Similar to the Metrics Configuration Patch, add a patch for the Logs SDK to load if the logs_sdk gem is present.

### LINKED ISSUE #1702: Logs SDK - Add Instrumentation Registry
Add an instrumentation registry for loggers to the logs SDK. This can follow the patterns established in the tracers.

kaylareopelle:
This PR has a few different features that are all somewhat related:
 * Adds the APIs to set a global logger provider using `OpenTelemetry.logger_provider`
   * Related to this, introduces `OpenTelemetry::Internal::ProxyLoggerProvider` and `OpenTelemetry::Internal::ProxyLogger` classes for defaults. 
   * Like in the main SDK with traces, these proxies are upgraded to real LoggerProviders/Loggers when set.
 * Adds the ability to configure the logs exporter using `OTEL_LOGS_EXPORTER`. The options available are `console`, `none`, and `otlp`. The OTLP exporter option will fail gracefully unless the yet to be released `opentelemetry-exporter-otlp-logs` gem is available. This takes the ConfigurationPatch approach, similar to https://github.com/open-telemetry/opentelemetry-ruby/pull/1548
 * Adds a thread-safe registry for loggers to access the same logger based on name and version
 
 Closes #1699
 Closes #1702


logs_api/lib/opentelemetry/internal/proxy_logger.rb
```diff
@@ -0,0 +1,70 @@
...
+    # the ProxyLoggerProvider until a delegate logger provider is installed. After the delegate
+    # logger provider is installed, the ProxyLogger will delegate to the corresponding "real"
+    # logger.
+    class ProxyLogger < Logs::Logger
+      # Returns a new {ProxyLogger} instance.
+      #
+      # @return [ProxyLogger]
+      def initialize
+        super
+        @mutex = Mutex.new
```
mwear, line 21: Do we need a mutex here? We don't have it for the proxy tracer. I believe we can remove this because the delegate is set from proxy logger provider, which synchronizes access using its mutex.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
```diff
@@ -34,14 +39,21 @@ def initialize(resource: OpenTelemetry::SDK::Resources::Resource.create)
         #
         # @return [OpenTelemetry::SDK::Logs::Logger]
         def logger(name:, version: nil)
-          version ||= ''
+          if @stopped
```
mwear, line 42: Is this case necessary? I ask because it differs from what we're doing in the `TracerProvider`.

logs_api/lib/opentelemetry/internal/proxy_logger.rb
kaylareopelle: Great question! I don't think we need it here. I was copying the setup in the ProxyMeter and didn't cross-reference what was going on with the ProxyTracer. 

Should we remove the mutex from the ProxyMeter too?
https://github.com/open-telemetry/opentelemetry-ruby/blob/672739120a605c5c14afc48e5aa07d1320443355/metrics_api/lib/opentelemetry/internal/proxy_meter.rb#L28-L37

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
kaylareopelle: It's been a minute since I wrote this, but I believe I did it to adhere to this part of the shutdown protocol: 

> `Shutdown` MUST be called only once for each `LoggerProvider` instance. After
the call to `Shutdown`, subsequent attempts to get a `Logger` are not allowed.
SDKs SHOULD return a valid no-op `Logger` for these calls, if possible. 

[(source)](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/sdk.md?plain=1#L131-L133)

There's similar verbiage for [the Trace SDK](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md?plain=1#L132-L134). 

I'm not sure if we adhere to this part of the Trace spec elsewhere. Maybe this part does the job well enough? But I'm not sure if that would sufficiently cover an attempt to get a tracer after shutdown. 

https://github.com/open-telemetry/opentelemetry-ruby/blob/a1f1629abb7c87ae316cf0f4541111e48e433d67/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L72-L75

What do you think?

logs_api/lib/opentelemetry/internal/proxy_logger.rb
mwear: I think it makes sense to align the three. For the purposes of this PR we can remove it from the proxy logger.

logs_sdk/lib/opentelemetry/sdk/logs/logger_provider.rb
mwear: I think the other piece of the puzzle (for traces) is here: https://github.com/open-telemetry/opentelemetry-ruby/blob/a1f1629abb7c87ae316cf0f4541111e48e433d67/sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb#L146-L167

My interpretation of the spec is that after calling shutdown on a tracer provider all tracers issued by it, past or future, must be non-operational. That is, they will create no-op spans from there on out. They do not necessarily need to be an instance of a no-op tracer, if that makes sense.

I think we need to ensure that all loggers return early from `on_emit` after `shutdown` is called, then it doesn't really matter what kind of logger we return from the logger provider. For consistency, we can follow what we're doing for tracing. I'm open to other considerations though.

kaylareopelle: I hadn't thought about it that way! Thanks for breaking that down. I've updated the code to use this approach. See: 06e4b05744729af78047a22dcbd1fa18b26a064d

logs_api/lib/opentelemetry/internal/proxy_logger.rb
kaylareopelle: Great! Updated in 9e2b8eeb87d3ca900bb687ddedd5421e527079a7

mwear (approved):
LGTM

logs_sdk/lib/opentelemetry/sdk/logs/configuration_patch.rb
```diff
@@ -0,0 +1,71 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+require 'opentelemetry/sdk/configurator'
+
+module OpenTelemetry
+  module SDK
+    module Logs
+      # The ConfiguratorPatch implements a hook to configure the logs portion
+      # of the SDK.
+      module ConfiguratorPatch
```
wsmoak, line 14: I am curious, is it intentional the the module name is ConfiguratorPatch (with an r) but the filename is configuration_patch.rb (with an n) ?

PR #1728, opened 2024-09-12
feat: sem_conv stability mode

hannahramadan:
Lays the foundation for accepting a new environment variable, `OTEL_SEMCONV_STABILITY_OPT_IN`, which gives users the ability to determine which semantic conventions to use. Individual instrumentation are expected to add metrics/attributes by calling the methods present in `OpenTelemetry::SemanticConventions::StabilityMode`. This borrows from Pythons [implementation](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/opentelemetry-instrumentation/src/opentelemetry/instrumentation/_semconv.py).

Only HTTP is supported for now ([spec](https://opentelemetry.io/docs/specs/semconv/http/)) and the values defined so far are:
- `http` - emit the new, stable HTTP and networking conventions only
- `http/dup` - emit both the old and the stable HTTP and networking conventions
- `default` - continue emitting whatever version of the old experimental HTTP and networking conventions the instrumentation was emitting previously.
- Note: `http/dup` has higher precedence than `http` in case both values are present

### DIFF
--- a/semantic_conventions/lib/opentelemetry/semantic_conventions/stability_mode.rb
+++ b/semantic_conventions/lib/opentelemetry/semantic_conventions/stability_mode.rb
@@ -21,13 +21,18 @@ module OpenTelemetry
     #   - `default`: Emit old conventions ONLY
     #
     # If no `OTEL_SEMCONV_STABILITY_OPT_IN` is set, the `default` behavior is used. `http/dup` has higher precedence than `http` in case both values are present.
+    #
+    # For instrumentation authors of libraries that emit HTTP semantic conventions, it is recommended to use the helper methods provided by this class to assign 
+    # attributes on span to support the opt-in preference of users. This class provides one method per convention and each method accepts at least two arguments. The 
+    # first is a hash to add the attribute(s) to. Additional arguments make up the value. The return value is nil, but the hash that was passed in will be updated.
     # 
     # Example usage:
     # ```ruby
+    # result = {} # Hash to add attributes to
     # stability_mode = OpenTelemetry::SemanticConventions::StabilityMode.new
-    # stability_mode.set_http_method(result, 'GET')
     # stability_mode.set_http_status_code(result, 200)
-    # stability_mode.set_http_url(result, 'https://example.com')
+    # 
+    # span.add_attributes(result) # Add the attributes to the span
     # ```
     class StabilityMode
       OTEL_SEMCONV_STABILITY_OPT_IN = 'OTEL_SEMCONV_STABILITY_OPT_IN'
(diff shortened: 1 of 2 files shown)


semantic_conventions/lib/opentelemetry/semantic_conventions/stability_mode.rb
```diff
@@ -0,0 +1,159 @@
...
+    #   - `default`: Emit old conventions ONLY
+    #
+    # If no `OTEL_SEMCONV_STABILITY_OPT_IN` is set, the `default` behavior is used. `http/dup` has higher precedence than `http`` in case both values are present.
+    # 
+    # Example usage:
+    # ```ruby
+    # stability_mode = OpenTelemetry::SemanticConventions::StabilityMode.new
+    # stability_mode.set_http_method(result, 'GET')
+    # stability_mode.set_http_status_code(result, 200)
+    # stability_mode.set_http_url(result, 'https://example.com')
```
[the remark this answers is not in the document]
hannahramadan: Made some updates! https://github.com/open-telemetry/opentelemetry-ruby/pull/1728/commits/36cfd9a3b49bb48520e6e0f60c588e7f37b1fa9e

kaylareopelle:
Link to instrumentation that uses this PR: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1166

semantic_conventions/lib/opentelemetry/semantic_conventions/stability_mode.rb
```diff
@@ -0,0 +1,164 @@
...
+      attr_accessor :initialized, :lock, :otel_semconv_stability_signal_mapping
+
+      def initialize
+        @lock = Mutex.new
+        @otel_semconv_stability_signal_mapping = {}
+
+        @lock.synchronize do
+          opt_in = ENV.fetch('OTEL_SEMCONV_STABILITY_OPT_IN', nil)
+          opt_in_list = opt_in.split(',').map(&:strip) if opt_in
+          http_set_sability_mode(opt_in_list) if opt_in_list
```
kaylareopelle, line 64:
```suggestion
          http_set_stability_mode(opt_in_list) if opt_in_list
```

semantic_conventions/lib/opentelemetry/semantic_conventions/stability_mode.rb
```diff
@@ -0,0 +1,164 @@
...
+        end
+      end
+
+      def http_set_sability_mode(opt_in_list)
```
kaylareopelle, line 68:
```suggestion
      def http_set_stability_mode(opt_in_list)
```

kaylareopelle (changes requested):

PR #1731, opened 2024-09-20
fix: update references to logging exporter

codeboten:
This exporter has been replaced by the debug exporter and will be removed soon

Related to https://github.com/open-telemetry/opentelemetry-collector/pull/11037

### DIFF
--- a/examples/metrics_sdk/README.md
+++ b/examples/metrics_sdk/README.md
@@ -36,8 +36,8 @@ receivers:
       # Default endpoints: 0.0.0.0:4317 for gRPC and 0.0.0.0:4318 for HTTP
 
 exporters:
-  logging:
-    loglevel: debug
+  debug:
+    verbosity: detailed
 
 processors:
   batch:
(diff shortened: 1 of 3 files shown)


kaylareopelle (changes requested):
Thank you for this PR, @codeboten! 

I think the metrics example updates are great. The `otel-collector` example updates don't work quite yet.

The version of the collector in `examples/otel-collector/.env` predates the debug exporter. It currently uses 0.35.0. However, when we update to the earliest version with the debug exporter, [0.86.0](https://github.com/open-telemetry/opentelemetry-collector/blob/main/CHANGELOG.md#v100-rcv0015v0860), we hit problems with the Jaeger exporter no longer being present in the collector. 

I've tried a few things to fix this, including following [the documentation to update Jaeger](https://opentelemetry.io/blog/2023/jaeger-exporter-collector-migration/), but couldn't get it to work. It feels serendipitous that you're the author of that doc and of this PR.

I totally get it if this was just a quick PR and you don't have more time to spend on this problem. If you do have more time to spend, I would appreciate your help getting this example to work on a newer version of the collector.

Happy to send along more details on my exploits if you have the bandwidth!

codeboten:
@kaylareopelle i've updated the example to use v0.109.0 of the collector, take a look and see if it works on your end. I tested it locally and was able to see a trace in jaeger

kaylareopelle (approved):
Thank you, @codeboten! This works for me too 😄

mwear (approved):

Issue #1710, opened 2024-08-29
Fallback Meter when an invalid name is provided


kaylareopelle:
When an invalid name is specified during a call to `MeterProvider#meter`, a working Meter implementation is returned as a fallback.

Furthermore (though optional), the fallback Meter name property keeps its original invalid value.

API: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/api.md?plain=1#L123-L132  
SDK: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md?plain=1#L126-L129

As of now, the behavior is that `name` is a required attribute on a Meter. There are no warnings about invalid names (empty strings or nil values).

PR #1734, opened 2024-09-27
fix: add warning if invalid meter name given

xuan-cao-swi:
Invalid meter name will return working meter with original name. It's missing the warning msg so added here.
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/1710


kaylareopelle (approved):
Thank you!

PR #1741, opened 2024-10-07
fix: Coerce aggregation_temporality value to a symbol

kaylareopelle:
Fetching the `OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` environment variable when setting the `:aggregation_temporality` on some aggregation classes (sum, explicit bucket histogram) sets the temporality as a string. However, checks throughout the metrics SDK and the OTLP exporter expect the temporality to be a symbol. 

This would cause applications that set the environment variable and used the OTLP exporter to have the aggregation temporality set to `Unspecified`.

Now, when aggregations that use this environment variable are initialized, the temporality will be coerced into a symbol.


metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -15,7 +15,7 @@ class Sum
 
           def initialize(aggregation_temporality: ENV.fetch('OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE', :delta))
             # TODO: the default should be :cumulative, see issue #1555
-            @aggregation_temporality = aggregation_temporality
+            @aggregation_temporality = aggregation_temporality.to_sym
```
arielvalentin, line 18: Can this be user specified or is there a constraint on the temporality preference?

I.e. should this be open ended or should it be a sort of enum object?

arielvalentin (approved):

mwear (approved):

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
kaylareopelle: Great point! I think an enum would be a good addition here. 

[According to the spec](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#temporality), the only valid values here are `:delta` or `:cumulative`. The OTLP exporter will fall back to `unspecified` otherwise. 

We don't have support for `:cumulative` aggregation yet, so really, the only true option should just be `:delta`.

kaylareopelle: I opened an issue to fix this as a follow-up: https://github.com/open-telemetry/opentelemetry-ruby/issues/1742

PR #1747, opened 2024-10-12
fix: Handle IPv6 addresses in OTLP HTTP connection

### LINKED ISSUE #1721: opentelemetry-exporter-otlp: http_connection does not handle ipv6 host
The OT library uses `host` from URI which fails to unwrap ipv6 brackets properly.  Fix is to update this to use `hostname` instead, which will return a plain ipv6 address which is what Net::HTTP expects as the first argument.
https://github.com/open-telemetry/opentelemetry-ruby/blob/75a48853b3525231a06aa17b0f76a247f4782060/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L107

Operating system details: FROM ruby:3.3.5-bookworm
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.5"
RUBY_DESCRIPTION: "ruby 3.3.5 (2024-09-03 revision ef084cc8f4) [aarch64-linux]"

failure.rb
```
require 'net/http'
uri = URI('http://[::1]:4318')
http.start
```
(hard error) getaddrinfo: Name or service not known (Socket::ResolutionError)

success.rb
```
require 'net/http'
uri = URI('http://[::1]:4318')
http = Net::HTTP.new(uri.hostname, uri.port)
http.start
```
(would be a success if something were on the port) Connection refused - connect(2) for "::1" port 4318 (Errno::ECONNREFUSED)

create2000:
Fixes #1721. This PR updates the Net::HTTP connection to handle IPv6 addresses correctly by using hostname instead of host.


exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -109,7 +109,7 @@ def fetch_ssl_verify_mode
           end
 
           def http_connection(uri, ssl_verify_mode, certificate_file, client_certificate_file, client_key_file)
-            http = Net::HTTP.new(uri.host, uri.port)
+            http = Net::HTTP.new(uri.hostname, uri.port)
```
arielvalentin, line 112: Are there any unit tests that you may also include with this change would have exposed this bug? 

e.g. setting the endpoint to `"http://[::1]/v1/traces"` and see the error goes away.

.rspec
```diff
@@ -0,0 +1 @@
+--require spec_helper
```
kaylareopelle, line 1: This repository uses [minitest](https://github.com/minitest/minitest), rather than [RSpec](https://rspec.info/) as the testing framework.

Would you mind deleting the `.rspec` file and the `spec/spec_helper.rb` file?

If you're having trouble running the tests locally, you can do so by:
1. Installing the bundle 
```bash
$ bundle install
```
2. Using Rake to run the tests 
```bash
$ bundle exec rake tests
```

If this doesn't work, let us know and we can try to troubleshoot.

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
create2000: Are you proposing i create some unit tests?

kaylareopelle: @create2000 - It would be great to have unit tests for this change, especially if you can write some sort of regression-style test that demonstrates the bug was fixed due to your code updates.

create2000: How would you suggest I go about the regression-style test? This is my first time contributing, and I am contributing as an Outreachy Intern applicant.

.rspec
create2000: Alright

exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
kaylareopelle: Thanks for the context, @create2000! 

What I mean is to write a test that would fail without your change. You could undo your change, write a test that uses an IPv6 address, and watch it fail. Then, add your change back and see it pass. This means your test will protect against regressions in the code that may re-introduce the bug.

There may be an existing test you can adjust from the OTLP exporter gem to make it easier. 
The file you could look at is `exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb`

PR #1753, opened 2024-10-22
chore: Update appraisals of google-protobuf

kaylareopelle:
The `google-protobuf` gem has a 4.x release. Also, our minimum supported version is now 3.18.0.

This PR updates the Appraisals file for the OTLP exporter and the OTLP metrics exporter to test different versions.


mwear (approved):

arielvalentin (commented):
Seems like protobuf as are backward compatible?

What is different in these major version changes?

Do we need to regenerate any source to target 4.x version of the gem?

PR #1767, opened 2024-11-22
fix: handle float value in NumberDataPoint

percygrunwald:
It's currently possible to have a `Counter` with a `Float` value, but this fails to export because the `NumberDataPoint` is always assumed to be an `Integer` for the purposes of ProtoBuf encoding. The `.proto` for a `NumberDataPoint` allows for a float value in the `as_double` key, so there should be no reason that we should not be able to support `Float` `Counter`s.

This commit adds support for `Float` values in `NumberDataPoint`, which in turn adds support for `Float` `Counter`s. I also updated the documentation in the structs for both `NumberDataPoint.value` and `HistogramDataPoint.sum` to match the expected type based on the `.proto`.

Adds a test to confirm that the metrics_exporter handles both `Integer` and `Float` values for a `NumberDataPoint`.

Noting the error message output without this change:

```
irb(main):004:0> E, [2024-11-22T11:29:40.853932 #1] ERROR -- : OpenTelemetry error: unexpected error in OTLP::MetricsExporter#encode - Non-integral floating point value assigned to integer field 'as_int' (given Float). - /home/percy/co/manage/vendor/bundle/ruby/3.1.0/gems/opentelemetry-exporter-otlp-metrics-0.2.0/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb:287:in `initialize'
```

### DIFF
--- a/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
+++ b/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
@@ -284,13 +284,20 @@ module OpenTelemetry
           end
 
           def number_data_point(ndp)
-            Opentelemetry::Proto::Metrics::V1::NumberDataPoint.new(
+            args = {
               attributes: ndp.attributes.map { |k, v| as_otlp_key_value(k, v) },
-              as_int: ndp.value,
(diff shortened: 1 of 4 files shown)


kaylareopelle (approved):
Thank you, @percygrunwald! This is a great catch.

PR #1773, opened 2024-12-04
feat: Add compatibility with env OTEL_SDK_DISABLED

pedrofurtado:
feat: Add compatibility with env OTEL_SDK_DISABLED

Currently, the ruby gem does not have compatibility with env `OTEL_SDK_DISABLED` defined in Open Telemetry specification.

Specification: https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#general-sdk-configuration

This PR adds this integration.


sdk/lib/opentelemetry/sdk.rb
```diff
@@ -61,6 +61,11 @@ module SDK
     #       c.use_all
     #     end
     def configure
+      if ENV['OTEL_SDK_DISABLED'] == 'true'
+        OpenTelemetry.logger.warn 'Environment variable OTEL_SDK_DISABLED is defined as true. SDK is disabled'
```
kaylareopelle, line 65:
```suggestion
        OpenTelemetry.logger.warn 'Environment variable OTEL_SDK_DISABLED is defined as true. SDK is disabled.'
```

sdk/lib/opentelemetry/sdk.rb
```diff
@@ -61,6 +61,11 @@ module SDK
     #       c.use_all
     #     end
     def configure
+      if ENV['OTEL_SDK_DISABLED'] == 'true'
```
arielvalentin, line 64: I see that you had added but then removed tests. 

Where the tests failing?

kaylareopelle (commented):
Thanks for this PR, @pedrofurtado! 

There is a test helper, `OpenTelemetry::TestHelpers.with_env` that can be used to help test behavior with environment variables.

For example:
https://github.com/open-telemetry/opentelemetry-ruby/blob/84d64caa1feb269542952fd766c0552bf805c78c/common/test/opentelemetry/common/utilities_test.rb#L102-L104

I started writing this comment before you removed the test. Could you add a test that leverages this helper?

sdk/lib/opentelemetry/sdk.rb
pedrofurtado: I created the test incorrectly

pedrofurtado:
@arielvalentin @kaylareopelle I've created a test for this. 
But the CI is not running, is pending approval. Can you approve the workflow to run? 👍

sdk/lib/opentelemetry/sdk.rb
pedrofurtado: Test recreated now

pedrofurtado:
@arielvalentin @kaylareopelle The CI finished, but a rubocop offense was notified. I fixed that, but I need approval again for CI please 😄

arielvalentin:
> @arielvalentin @kaylareopelle The CI finished, but a rubocop offense was notified. I fixed that, but I need approval again for CI please 😄 

Thanks for your patience. You'll require approvals to run CI until your first PR is merged.

sdk/lib/opentelemetry/sdk.rb
arielvalentin, line 64: @robertlaurin is this the best place for us to check this? 

Does it make sense to check that in the instance method instead? https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/configurator.rb#L144

Do we want to support being able to change the envar and re-initialize in a running program?

pedrofurtado: Here in our applications works like a charm in this place (using monkey patch).

The PR content was based on this comment: https://github.com/open-telemetry/opentelemetry-ruby/discussions/1605#discussioncomment-8568551

arielvalentin: Yes, that was my comment 😄 

I reached out to @robertlaurin here since he designed this component and wanted to solicit his input about where this check should happen. 

If a user is _not_ using the DSL, they can still create an instance of the `Configurator` and call the `configure` method bypassing the `envar`. 

They could also not use the configurator at all and assemble the tracer provider and resources themselves. 

I just want to make sure we are hooking this into the appropriate scope, so folks are not surprised that it does not work as anticipated in specific scenarios; or if we are OK with this minimal implementation.

kaylareopelle:
The spec for this environment variable feels a little vague to me. Do we know the scope of the `OTEL_SDK_DISABLED` env var? 

The current implementation would skip creating tracer providers through configuration, but someone could still manually create functional tracer providers and tracers. Is that acceptable? 

I took a look at [Python's implementation](https://github.com/open-telemetry/opentelemetry-python/pull/3648), and it seems like they always return no-ops when `OTEL_SDK_DISABLED` is detected.

However, [Java's implementation](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java#L449) seems closer to this approach. Configuration is skipped, but I'm not sure if that means you cannot create working tracers, etc.

sdk/lib/opentelemetry/sdk.rb
arielvalentin: I had similar questions @kaylareopelle to the ones you brought up

kaylareopelle: thank you! missed this! [(link to comment)](https://github.com/open-telemetry/opentelemetry-ruby/pull/1773#issuecomment-2521597058)

pedrofurtado:
Hi @arielvalentin @kaylareopelle! :wave: 
Is the PR ok?

arielvalentin:
I don't know. We don't quite understand how this environment variable is supposed to work per the specification. Right now this PR only works in a specific use case. 

If our users initialize the SDK in a different way, I.e. manual configuration, then the envar will not work as expected and we may get a bug report. 

That is why we are asking the question about whether or not this check is happening the the correct place.

kaylareopelle:
I opened an issue in the semantic conventions repo to see if we can get some help clarifying the functionality: https://github.com/open-telemetry/semantic-conventions/issues/1667 

Once we understand the spec better, we can move forward. Thanks for your patience, @pedrofurtado!
cc @arielvalentin

arielvalentin:
@kaylareopelle I think that should be in the spec repo thought right? Not semconv?

kaylareopelle:
@arielvalentin  - Thank you, you're right! 🤦 I'll get that fixed.

kaylareopelle:
New issue in the right repo! https://github.com/open-telemetry/opentelemetry-specification/issues/4332 

Thanks, @arielvalentin!

kaylareopelle (approved):
Hi @pedrofurtado! Thanks for your patience. We talked about this during the SIG on Tuesday. 

The conversation at https://github.com/open-telemetry/opentelemetry-specification/issues/4332 seemed to conclude that environment variables can have different scopes based on the implementation. Having `OTEL_SDK_DISABLED` scoped to just the configure method falls within the guidelines. Having it create no-ops outside of auto-configuration also falls within the guidelines. 

I think this is a great first step toward getting this functionality into the SDK, and it seems to be fully spec-compliant.

If, in the future, we prefer to have no-ops for all tracers/meters/loggers; we can implement that in a separate PR.

kaylareopelle:
>I think this is a great first step toward getting this functionality into the SDK, and it seems to be fully spec-compliant.
>
>If, in the future, we prefer to have no-ops for all tracers/meters/loggers; we can implement that in a separate PR.

@arielvalentin, are you comfortable with this approach?

arielvalentin:
> > I think this is a great first step toward getting this functionality into the SDK, and it seems to be fully spec-compliant.
> > If, in the future, we prefer to have no-ops for all tracers/meters/loggers; we can implement that in a separate PR.
> 
> @arielvalentin, are you comfortable with this approach?

I am not sure. If it is true that each SDK has to implement it themselves then maybe this is a good start, but if the API gem should _prevent_ any implementation of the SDK from being initialized, then maybe that is fine. 

As you all discussed in that spec issue, it seems like other language SIGs separate the responsiblities of managing envars and instead leave it to their configuration modules. 

I am probably overthinking this, but I do not want to end up with cases where folks are initializing multiple tracer providers and manually setting the Global Tracer Provider, which is part of the API gem and ignore the envar altogether. 

Lets go ahead with incremental steps. As "bug" reports flow in we can address them and discover all of the interesting ways our end users configure their SDKs.

arielvalentin (approved):

arielvalentin (approved):
> > I think this is a great first step toward getting this functionality into the SDK, and it seems to be fully spec-compliant.
> > If, in the future, we prefer to have no-ops for all tracers/meters/loggers; we can implement that in a separate PR.
> 
> @arielvalentin, are you comfortable with this approach?

I am not sure. If it is true that each SDK has to implement it themselves then maybe this is a good start, but if the API gem should _prevent_ any implementation of the SDK from being initialized, then maybe that is fine. 

As you all discussed in that spec issue, it seems like other language SIGs separate the responsiblities of managing envars and instead leave it to their configuration modules. 

I am probably overthinking this, but I do not want to end up with cases where folks are initializing multiple tracer providers and manually setting the Global Tracer Provider, which is part of the API gem and ignore the envar altogether. 

Lets go ahead with incremental steps. As "bug" reports flow in we can address them and discover all of the interesting ways our end users configure their SDKs.

Issue #1783, opened 2024-12-18
Support monotonic operation for counter/sum


xuan-cao-swi:
The counter should be adjustable for monotonic operations so that, when enabled, it prevents any negative values.

Related discussion: https://github.com/open-telemetry/opentelemetry-ruby/discussions/1778

Related docs: 
- https://opentelemetry.io/docs/specs/otel/metrics/data-model/#opentelemetry-protocol-data-model
- https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums
- https://opentelemetry.io/docs/specs/otel/metrics/data-model/#timeseries-model

garoazinha:
I also couldn't export metrics to prometheus and could fix it by changing the way the exporting is done. According to the [specs](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#sum-aggregation) non_monotonic type sum is based on the instrument. Doing this:

 ```ruby
case instrument_type      
  when :counter
   Opentelemetry::Proto::Metrics::V1::Metric.new(
     name: metrics.name,
     description: metrics.description,
     unit: metrics.unit,
     sum: Opentelemetry::Proto::Metrics::V1::Sum.new(
       aggregation_temporality: as_otlp_aggregation_temporality(metrics.aggregation_temporality),
       data_points: metrics.data_points.map do |ndp|
         number_data_point(ndp)
       end,
       is_monotonic: true
      )
    )
end
```

on the [metrics otlp exporter](https://github.com/open-telemetry/opentelemetry-ruby/blob/baeda39731d70f3f18ac8693168039fb8a1d2c8c/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb#L220) already solved the issue for me. The is_monotonic attribute should be tied to the Sum class though, as it is in other implementations. The parameter is usually set when initializing the sum aggregation object, in:
		[python impl](https://github.com/open-telemetry/opentelemetry-python/blob/f3b9e6ec96b7cc3717d5c8b529da7a8923810973/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/metric_reader_storage.py#L172)
		[js impl](https://github.com/open-telemetry/opentelemetry-js/blob/fc0edd82f6749bfe1ca733ed1f80ad4a6015285f/packages/sdk-metrics/src/view/Aggregation.ts#L55)
[...]

PR #1793, opened 2025-01-15
fix: adding is_monotonic flag to sum 

garoazinha:
The is_monotonic flag for sum aggregation otlp model is required by the [specs](https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums) 

 [The SDK should reject negative inputs to monotonic cumulative metrics, but it is not required to.](https://github.com/open-telemetry/opentelemetry-specification/blob/main/oteps/metrics/0003-measure-metric-type.md#cumulative-metric)

Implementations such as python and go appear to not enforce monotonicity, but the C++ and js implementations do. 
[js impl](https://github.com/open-telemetry/opentelemetry-js/blob/fc0edd82f6749bfe1ca733ed1f80ad4a6015285f/packages/sdk-metrics/src/aggregator/Sum.ts#L38)
[C++ impl](https://github.com/lalitb/opentelemetry-cpp/blob/62b71970e5ca0e1331a5d4d87980ee56f881f34c/sdk/src/metrics/aggregation/sum_aggregation.cc#L36)

Issue: #1783


metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -47,6 +48,8 @@ def update(increment, attributes, data_points)
               nil
             )
 
+            return if is_monotonic && increment < 0
```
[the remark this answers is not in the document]
garoazinha: I think checking if increment is negative fulfills the specs for cumulative sums. for delta sums it might be the case that we need to check if the sum (value + increment) is negative? only checking if increment is negative ensures the sum value is never negative but might reject some valid increments. is my logic correct?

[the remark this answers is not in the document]
xuan-cao-swi: Yes, I think your logic is correct (also checked other implementation as your referenced).
The spec is kind confusing, but `In this case of metrics, this means the sum is nominally increasing` should ease the confusion.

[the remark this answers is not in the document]
xuan-cao-swi: LGTM

xuan-cao-swi (approved):

kaylareopelle (approved):
Thanks, @garoazinha! This looks good to me. I appreciate your work on this!

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -11,11 +11,12 @@ module Aggregation
         # Contains the implementation of the Sum aggregation
         # https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#sum-aggregation
         class Sum
-          attr_reader :aggregation_temporality
+          attr_reader :aggregation_temporality, :is_monotonic
```
arielvalentin, line 14: @xuan-cao-swi typically I would think of these kinds of attributes hidden behind predicates. 

Would it not make sense to make the public interface of this attribute `monotonic?`

kaylareopelle: The value in the protobuf is `is_monotonic`, so I think the goal was to match. I also found it confusing with Ruby's predicates.

https://github.com/open-telemetry/opentelemetry-ruby/blob/289b4aaa35f8b13803ab6c6c8b34fddd41525376/exporter/otlp/lib/opentelemetry/proto/metrics/v1/metrics_pb.rb#L11 

But, I believe some agents (Js?) just reduce this to `monotonic` in the SDK, so there's precedent to go the other way.

metrics_sdk/.tool-versions
kaylareopelle, line 1: @garoazinha, would you mind removing this file? I think it snuck in on the last commit.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
arielvalentin: Thanks for explaining that, though isn't this an internal Ruby Class and not generated from protobuf source?

Are we leaking the protobuf source and idioms into the SDK?

kaylareopelle: Great point. I don't think we need to. I've used the protobuf as a guide before for implementation, but those are behind the scenes details that probably don't need to be propagated upstream. @garoazinha has updated the code to use the predicate style you suggested.

PR #1802, opened 2025-01-30
feat: Support 3.1 Min Version

arielvalentin:
Ruby 3.0
status: eol
release date: 2020-12-25
normal maintenance until: 2023-04-01
EOL: 2024-04-23

https://www.ruby-lang.org/en/downloads/branches/

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -58,21 +58,21 @@ runs:
     # ...but not for appraisals, sadly.
     - name: Install Ruby ${{ inputs.ruby }} with dependencies
       if: "${{ steps.setup.outputs.appraisals == 'false' }}"
-      uses: ruby/setup-ruby@v1.190.0
+      uses: ruby/setup-ruby@v1.215.0
       with:
         ruby-version: "${{ inputs.ruby }}"
         working-directory: "${{ steps.setup.outputs.gem_dir }}"
-        bundler:  "2.5.17"
+        bundler: "latest"
         bundler-cache: true
         cache-version: "v1-${{ steps.setup.outputs.cache_key }}"
 
     # If we're using appraisals, do it all manually.
     - name: Install Ruby ${{ inputs.ruby }} without dependencies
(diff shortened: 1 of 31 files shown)


kaylareopelle (approved):
Thank you!

kaylareopelle (commented):
Thinking about this a little more.... Unlike contrib, this repo has some gems with a 1.x release. I think we should mark this as a breaking change to avoid accidental upgrades for 3.0 users.

Unfortunately we don't have any post-install messages to warn folks, but I hope the contrib messages reached most users of these libraries. 

What do you think about using #otel-ruby to announce this for the core libraries? 

I want both the core and contrib libraries to be aligned with a min version, so I'd like to get this out relatively soon, while minimizing surprises.

arielvalentin:
@kaylareopelle I have changed the title and added a `!` to signal a "breaking" change but I am not sure that it matters to qualify it that way. 

Users of Ruby 3.0 or older are not really impacted by it. They just won't be able to upgrade and will be pinned at an old version. This should not cause any conflicts for them.

kaylareopelle:
> @kaylareopelle I have changed the title and added a `!` to signal a "breaking" change but I am not sure that it matters to qualify it that way.
> 
> Users of Ruby 3.0 or older are not really impacted by it. They just won't be able to upgrade and will be pinned at an old version. This should not cause any conflicts for them.

@arielvalentin, now that I look into it further, I agree. I've encountered complaints about increasing minimum Ruby versions before, but modern bundler versions should take `required_ruby_version` into account and prevent version negotiation problems. 

It seems like this was finessed in the 1.11.x series of Bundler, though there are some improvements in 2.x. Since we set a minimum bundler version of 1.17.0 in the core libraries, I believe most of our users are safe. Maybe we should chat about this in the SIG to make sure everyone's on the same page.

Breaking changes can be a big deal, so I don't want to force one unnecessarily.

kaylareopelle:
We talked about this in the SIG and decided that a minor version is appropriate given our minimum Bundler requirement has the required functionality to limit update problems for our users. I'll update the PR title.

mwear (approved):

PR #1760, opened 2024-11-14
fix(context): do not modify stack array directly when attach and detach

tientt-holistics:
As mentioned in [open-telemetry/opentelemetry-ruby/pull/1598](https://github.com/open-telemetry/opentelemetry-ruby/pull/1598)

> As identified in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/772, Rails copies all thread-local variables in ActionController::Live. Our use of an array to back the Context stack results in two threads sharing the same stack. Context.clear only clears the array, but continues to share it (in fact, its effect is visible to both threads), so there is no way to cleanly separate the thread's Context stacks using the public API.

This PR will replace the `stack.push` and `stack.pop` with Thread.current assignment

Issue #1766, opened 2024-11-21
Context concurrency and parallelism issues


arielvalentin:
The `OpenTelemetry::Context` uses a [fiber local Ruby Array](https://github.com/open-telemetry/opentelemetry-ruby/blob/08b1130ed265e82680c7845bc44ad9526d9a0a2e/api/lib/opentelemetry/context.rb#L130) to keep track of the active `Context` instance.  

There are at least 2 instances in gems that are breaking encapsulation and copying fiber local variables to child threads:

- [Rails ActionController::Metal::Live](https://github.com/rails/rails/blob/5287cfa17681527761c50fd8c6ae169f4500bd97/actionpack/lib/action_controller/metal/live.rb#L284)
- [GraphQL::Dataloader](https://github.com/rmosolgo/graphql-ruby/blob/8a21eb17d58902b20867f18b4c25937b75baa830/lib/graphql/dataloader.rb#L80)

This leads to concurrency and parallel processing issues because multi-threads are pushing and popping entries on the same Stack, resulting in [`OpenTelemetry::Context::DetachError`](https://github.com/open-telemetry/opentelemetry-ruby/blob/08b1130ed265e82680c7845bc44ad9526d9a0a2e/api/lib/opentelemetry/context.rb#L58) as well as non-deterministic behavior where the `OpenTelemetry::Context.current` method returns the wrong instance of the context object. 

Helpers that rely on the implicitly active Context, like `OpenTelemetry::Trace.current_span`, will end up receiving the wrong Span object instance.

There is a proposed fixed here that would end up using Immutable Arrays, however this will increase the number of object allocations per Context stack mutation: 

https://github.com/open-telemetry/opentelemetry-ruby/pull/1760

Another option I have considered is to propose using Thread Local variables instead of Fiber local variables, however that would result in concurrency issues when using Fibers where multiple fibers would have access to the same Array instance. We would need a mechanism to seed a spawned thread with a snapshot of the stack so that there could be some consistent context propagation for the gems listed above.  

I would like to get some feedback from @open-telemetry/ruby-maintainers on whether or not there are addition alternatives we can consider.

ioquatix:
Copying around thread locals like this is like copying around private instance variables between objects, it's a horrible idea and should never be done.

There is now a feature in Ruby, which does this and is expected to do this: `Fiber.storage`

```
irb(main):001> Fiber[:foo] = :bar
=> :bar
irb(main):002* Thread.new do
[...]

PR #1807, opened 2025-02-05
Use a Fiber attribute for Context

fbogsany:
Alternative to #1760

### DIFF
--- a/api/lib/opentelemetry/context.rb
+++ b/api/lib/opentelemetry/context.rb
@@ -8,11 +8,12 @@ require 'opentelemetry/context/key'
 require 'opentelemetry/context/propagation'
 
 module OpenTelemetry
+  Fiber.attr_accessor :opentelemetry_context
+
   # Manages context on a per-fiber basis
   class Context
     EMPTY_ENTRIES = {}.freeze
-    STACK_KEY = :__opentelemetry_context_storage__
-    private_constant :EMPTY_ENTRIES, :STACK_KEY
+    private_constant :EMPTY_ENTRIES
 
     DetachError = Class.new(OpenTelemetry::Error)
 
@@ -113,11 +114,9 @@ module OpenTelemetry
         current.value(key)
       end
 
-      # Clears the fiber-local Context stack. This allocates a new array for the
-      # stack, which is important in some use-cases to avoid sharing the backing
-      # array between fibers.
+      # Clears the fiber-local Context stack.
       def clear
-        Thread.current[STACK_KEY] = []
+        Fiber.current.opentelemetry_context = []
       end
 
       def empty
@@ -127,7 +126,7 @@ module OpenTelemetry
       private
 
       def stack
-        Thread.current[STACK_KEY] ||= []
+        Fiber.current.opentelemetry_context ||= []
       end
     end
 
--- a/api/test/opentelemetry/context_test.rb
+++ b/api/test/opentelemetry/context_test.rb
@@ -289,7 +289,73 @@ describe OpenTelemetry::Context do
     end
   end
 
+  describe 'fibers' do
+    it 'is isolated with respect to Fiber-local variable manipulation' do
+      OpenTelemetry::TestHelpers.with_test_logger do |log_stream|
+        ctx = new_context
+        Context.with_current(ctx) do
+          # This is based on code in ActionController::Live#process:
+          # https://github.com/rails/rails/blob/ad0105c13f61d145a659004efb928a643104e973/actionpack/lib/action_controller/metal/live.rb#L270
+          t1 = Thread.current
+          locals = t1.keys.map { |key| [key, t1[key]] }
+          Fiber.new do
+            t2 = Thread.current
+            locals.each { |k, v| t2[k] = v }
+            # Manipulate _this fiber's_ context stack.
+            Context.attach(new_context)
+          ensure
+            locals.each { |k, _| t2[k] = nil }
+          end.resume
+        end
+        _(log_stream.string).must_be_empty
+      end
+    end
+
+    it 'is isolated with respect to Fiber-local storage manipulation' do
+      OpenTelemetry::TestHelpers.with_test_logger do |log_stream|
+        ctx = new_context
+        Context.with_current(ctx) do
+          # This is based on code in ActionController::Live#process, modified to use Fiber-local storage:
+          # https://github.com/rails/rails/blob/ad0105c13f61d145a659004efb928a643104e973/actionpack/lib/action_controller/metal/live.rb#L270
+          f1 = Fiber.current
+          Fiber[:foo] = :bar
+          locals = f1.storage
+          Fiber.new do
+            f2 = Fiber.current
+            locals.each { |k, v| f2.storage[k] = v }
+            # Manipulate _this fiber's_ context stack.
+            Context.attach(new_context)
+          ensure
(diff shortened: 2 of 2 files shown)


ioquatix:
I haven't reviewed the code, but I agree with the approach (since I suggested it :).

ioquatix (approved):

ivoanjo:
Hey :wave:. I saw this after Ariel pinged on https://github.com/open-telemetry/opentelemetry-ruby/issues/1766 . Amazing work on enumerating and contrasting the alternatives :)

I work at the Datadog on our [Ruby profiler](https://github.com/DataDog/dd-trace-rb/tree/master/ext/datadog_profiling_native_extension). One of the things we currently support is correlating the profiling data we gather with OTEL traces. The intuition here is you can look at the "profiles for this span/trace".

Right now the way we do this with otel-ruby... [complicated](https://github.com/DataDog/dd-trace-rb/pull/3984) and [error-prone](https://github.com/DataDog/dd-trace-rb/pull/4240). But it does work :)

TL;DR:
* The profiler is implemented with native code because there are things that can only be gathered from the C side, for instance profiling allocations and the Global VM Lock.
* The profiler samples across threads. E.g. it reads this context from different threads, not just from the current thread.
* If we want to read the current span in the middle of an allocation, or GVL event, we can't call Ruby code at that point. Thus, we're "barging in" an directly reading `Thread.current[STACK_KEY]` using C code.

The semi-sad news is that I [prototyped changing our code to read from `Fiber.current`](https://github.com/DataDog/dd-trace-rb/compare/ivoanjo/otel-try-from-fiber?expand=1)... and discovered that Ruby doesn't allow us to get "the current fiber object for a given thread".

(We often slightly cheat by using internal Ruby headers to get access to things that we wouldn't otherwise have, but fibers actually are defined in [`cont.c`](https://github.com/ruby/ruby/blob/master/cont.c) so no headers involved there.)

So this would complicate our life quite a bit.

Obviously, an important question is: how much should you care about this? Having the otel tracing context being in an accessible location is something that is also relevant for the [OTEL profiling support](https://opentelemetry.io/blog/2024/state-profiling/) (Datadog is also participating in standardizing this effort).

Right now this whole stack and contexts structure per-fiber is very flexible at the Ruby level, but it makes it quite complicated to answer the query "what's active in this thread/fiber right now".

ioquatix:
@ivoanjo what's the correct way to fix this? `Thread#fiber`? It seems like it would be extremely racy.

ivoanjo:
>  @ivoanjo what's the correct way to fix this? Thread#fiber? It seems like it would be extremely racy.

Yeah, something like that is missing and could be used. In fact, it wouldn't even need to be exposed on the Ruby side -- e.g. could only exist as a C-level API.

And indeed it would be racy, but for profilers that would probably be ok, since they often interrupt Ruby so have ~some control over the GVL. And it probably wouldn't be more racy than other things involving threads?

---

On the other hand, "getting context for connecting to other things" could be somewhat decoupled from the internal management of otel-ruby state. E.g. if the library exposed the latest `span_id`/`trace_id` somewhere easily accessible (e.g. maybe keep that in `Thread#[]` but then use the Fiber as a source of truth) it would make it easier for external tools to get this info while still keeping the state consistent.

Alternatively alternatively, having some kind of very lightweight callback when a span was activated/deactivated on a fiber could allow 3rd parties to build their own tracking. That would probably well enough for profilers that work inside the Ruby process, but doesn't actually solve the issue for tools that work from outside the process, such as https://github.com/open-telemetry/opentelemetry-ebpf-profiler or even something like [rbspy](https://rbspy.github.io/).

robertlaurin (approved):

ioquatix:
That all makes sense.

I am okay with introducing a C API but it might be good to get feedback from @eregon about it as he has some strong opinions about these issues/design. He may have some other insights into the problem too.

Ideas for implementation:

```c
// Safe to call from any thread provided you have the GVL?
VALUE rb_thread_current_fiber(VALUE self) {
    rb_thread_t *thread = rb_thread_ptr(self);

    if (thread->ec->fiber_ptr) {
        return rb_fiberptr_self(thread->ec->fiber_ptr);
    }
    else {
        return Qnil;
    }
}
```

If I understand correctly, your use case would be given a list of threads, getting the current fiber, and extracting fiber-instance attributes.

eregon:
Thanks for the ping.
That does seem problematic for me, this method cannot work without a GVL, e.g. on TruffleRuby it would always return a stale Fiber which once upon a time was (and maybe still is or not) the current Fiber of that Thread.

I don't have time to read all the context, but assuming we want to associate profiling information to Fibers, the API to get the profiling information should tell us which Fiber was executing at the time of the sample. There is no way to recover the current Fiber from a Thread after the fact.

ioquatix:
> always return a stale Fiber which once upon a time was (and maybe still is or not) the current Fiber of that Thread.

For the sake of the profiler, it's probably good enough?

ivoanjo:
>> always return a stale Fiber which once upon a time was (and maybe still is or not) the current Fiber of that Thread.
>
> For the sake of the profiler, it's probably good enough?

Yeah, it would probably be fine. This is already a challenge that sampling profilers have -- they observe only periodically, so they may "blame" cpu-time or another resource on a tiny once-in-a-lifetime-method. The usual approach is that with enough samples, it evens out and thus these errors end up not impacting the overall picture (think lossy video/image/audio codecs).

Having said that...

> I don't have time to read all the context, but assuming we want to associate profiling information to Fibers, the API to get the profiling information should tell us which Fiber was executing at the time of the sample.

...this would be nice as well. In fact, the GO VM does have a built-in profiler, and has a mechanism where you can attach context (simple key/value pairs) to goroutines and these get captured by the profiler as well.

---

But I don't want to get too outside the scope from this PR! Apologies to @fbogsany for the noise. The TL;DR is that this change makes it a bit harder to observe otel Ruby apps, due to no fault of your amazing work. And at least for me there's not a clear suggestion that gives us no trade-offs :|

robertlaurin:
Hey all, appreciate the conversation that came up around this change. We're still in a position where the latest release of Rails breaks Otel instrumentation for LiveController instances. I'd like to move ahead with merging and releasing this change.

@ivoanjo I don't necessarily want us to commit to supporting what we intended to be a private interface, but I realize this has implication on work you have been doing with profiling. I think it would be beneficial if we saw some examples of what a more formalized solution might look like.

ivoanjo:
One good thing from this being an internal detail currently, is that evolving/changing minds seems like a low-cost thing. (The only cost is to external profilers that are reaching to internal APIs :wink: )

I've quickly prototyped an alternative to `ArrayContext` that uses `Thread.current` in [this branch](https://github.com/open-telemetry/opentelemetry-ruby/compare/fiber-attribute-for-context...DataDog:opentelemetry-ruby:ivoanjo/defensive-array-context?expand=1). The intuition is -- what if the array context was defensive and checked if it was copied:

```ruby
    def stack
      current_stack, owner = Thread.current[STACK_KEY]

      if owner != Fiber.current
        current_stack = []
        Thread.current[STACK_KEY] = [current_stack, Fiber.current]
      end

      current_stack
    end
```

Benchmarks show this is only slightly more costly than `ArrayContext`.

Having said that, I've also spent more time looking at the CRuby VM and while there's no public ways of getting the current Fiber for a thread, it's not impossible, so we can support the current PR as-is if needed with some acrobatics on our side.

PR #1603, opened 2024-02-22
feat: add basic periodic exporting metric_reader

xuan-cao-swi:
## Description

I'd like to contribute on [Periodic exporting MetricReader](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#periodic-exporting-metricreader). I think the periodic metric_reader is useful for asynchronous instrument that collect metrics (e.g. cpu, memory) within fixed duration.

Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/1627

PR #1811, opened 2025-02-18
docs: Mention support for periodic exporting in README

gjtorikian:
I think #1603 made this available.


kaylareopelle (approved):
Great catch, thank you!

Issue #1797, opened 2025-01-22
Metrics otlp exporter sample broken


KieranP:
**Description of the bug**

Using metrics otlp exporter sample (https://github.com/open-telemetry/opentelemetry-ruby/blob/main/examples/metrics_sdk/metrics_collect_otlp.rb#L33) fails with:

```
undefined method `pull' for an instance of OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader
```

**Share details about your runtime**

Operating system details: Mac
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.3.6"
RUBY_DESCRIPTION: "ruby 3.3.6 (2024-11-05 revision 75015d4c1f) [arm64-darwin24]"

    opentelemetry-api (1.4.0)
    opentelemetry-common (0.21.0)
    opentelemetry-exporter-otlp (0.29.1)
    opentelemetry-exporter-otlp-metrics (0.3.0)
    opentelemetry-metrics-api (0.2.0)
    opentelemetry-metrics-sdk (0.5.0)
    opentelemetry-registry (0.3.1)
    opentelemetry-sdk (1.6.0)
    opentelemetry-semantic_conventions (1.10.1)

PR #1815, opened 2025-02-19
fix: update metrics example

xuan-cao-swi:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/1797

The current metrics configuration patch will load the periodic reader by default. The example doesn't need the periodic reader to test if the exporter can send out the metrics to the collector.

The Otel collector config.yaml seems to need the endpoint defined; updated the README.


examples/metrics_sdk/metrics_collect.rb
```diff
@@ -15,6 +15,8 @@
 require 'opentelemetry/sdk'
 require 'opentelemetry-metrics-sdk'
 
+ENV['OTEL_METRICS_EXPORTER'] = 'none'
```
kaylareopelle, line 18: What do you think about adding a comment to describe why we're doing this?
```suggestion
# this example manually configures the exporter, turn off automatic configuration
ENV['OTEL_METRICS_EXPORTER'] = 'none'
```

examples/metrics_sdk/metrics_collect_otlp.rb
```diff
@@ -18,6 +18,8 @@
 require 'opentelemetry-metrics-sdk'
 require 'opentelemetry-exporter-otlp-metrics'
 
+ENV['OTEL_METRICS_EXPORTER'] = 'none'
```
kaylareopelle, line 21: If you like the suggestion above, here's another comment for some parallelism
```suggestion
# this example manually configures the exporter, turn off automatic configuration
ENV['OTEL_METRICS_EXPORTER'] = 'none'
```

kaylareopelle (approved):
Thanks, @xuan-cao-swi! One small comment, but feel free to dismiss the suggestion if you disagree.

Issue #1814, opened 2025-02-18
MetricsReporter: possible outdate protobuf datatype


xuan-cao-swi:
Running the metrics exporter and received this issue when `log_status` is called.
 
```
E, [2025-02-18T16:51:36.131790 #19] ERROR -- : OpenTelemetry error: unexpected error decoding rpc.Status in OTLP::MetricsExporter#log_status - undefined method `type_name' for <Google::Protobuf::Any: type_url: "type.googleapis.com/google.rpc.BadRequest", value: "
>
.resourceMetrics.scopeMetricsPath contained no usable data
1
.resourceMetricsPath contained no usable data">:Google::Protobuf::Any

              klass_or_nil = ::Google::Protobuf::DescriptorPool.generated_pool.lookup(detail.type_name).msgclass
                                                                                            ^^^^^^^^^^ - /usr/local/bundle/gems/opentelemetry-exporter-otlp-metrics-0.3.0/lib/opentelemetry/exporter/otlp/metrics/util.rb:120:in `method_missing'
```

Not sure if this is caused by protobuf any type removed the `type_name` (https://cloud.google.com/ruby/docs/reference/google-cloud-container_analysis-v1/latest/Google-Protobuf-Any)

Current protobuf any have two values: 
```
<Google::Protobuf::Any: type_url: "type.googleapis.com/google.rpc.BadRequest", value: ">
```
Currently possible workaround
```ruby
require 'google/rpc/error_details_pb'
type_name = detail.type_url.split('/').last # => google.rpc.BadRequest
klass_or_nil=::Google::Protobuf::DescriptorPool.generated_pool.lookup(type_name)&.msgclass => Google::Rpc::BadRequest
```

**Share details about your runtime**

Operating system details: Debian GNU/Linux 11 (bullseye)
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.1.0"
RUBY_DESCRIPTION: "ruby 3.1.0p0 (2021-12-25 revision fb4df44d16) [aarch64-linux]"
OpenTelemetry::Exporter::OTLP::Metrics::VERSION: "0.3.0"

PR #1816, opened 2025-02-19
fix: update out-of-date google-protobuf-any

xuan-cao-swi:
Closes: https://github.com/open-telemetry/opentelemetry-ruby/issues/1814

This issue won't cause any major breakdown since it's only for logging the error message.

If the change is valid, then it needs to be updated for all exporters.

### DIFF
--- a/exporter/otlp-metrics/README.md
+++ b/exporter/otlp-metrics/README.md
@@ -70,6 +70,8 @@ Or, if you use [bundler][bundler-home], include `opentelemetry-sdk`, `openteleme
 Then, configure the SDK to use the OTLP metrics exporter
 
 ```ruby
+ENV['OTEL_METRICS_EXPORTER'] = 'none'
+
 require 'opentelemetry/sdk'
 require 'opentelemetry-metrics-sdk'
 require 'opentelemetry/exporter/otlp_metrics'
--- a/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/util.rb
+++ b/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/util.rb
@@ -118,7 +118,7 @@ module OpenTelemetry
             status = Google::Rpc::Status.decode(body)
             details = status.details.map do |detail|
               type_name = detail.type_url.to_s.split('/').last.to_s
-              klass_or_nil = ::Google::Protobuf::DescriptorPool.generated_pool.lookup(detail.type_name)&.msgclass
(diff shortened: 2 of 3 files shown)


kaylareopelle:
Do we need to update the min version of the protobuf dependencies for the gems too?

xuan-cao-swi:
> Do we need to update the min version of the protobuf dependencies for the gems too?

I'm not very familiar with protobuf but I can share what I found:

I don't think we need to update protobuf dependencies (`well_known_types.rb` file's last commit is 2 years ago); I think that's just an edge case that was missed.

By looking at the [well_known_types.rb](https://github.com/protocolbuffers/protobuf/blob/main/ruby/lib/google/protobuf/well_known_types.rb#L42), it seems like we can access `type_name` ? But strangely it returns `no method` when access it. 

The `type_name` seems also derived from type_url (e.g. type_url split), so I think it's safe to use type_url split in our implementation.

kaylareopelle (commented):
Hi @xuan-cao-swi, coming back to this one. I think this is a good change. Would you mind adding a regression test?

xuan-cao-swi:
Hi @kaylareopelle, I think this [test](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-metrics/test/opentelemetry/exporter/otlp/metrics/metrics_exporter_test.rb#L483) cover the case; but I still add one test case that parse the request from raw byte.

I still couldn't figure out why this type_name is missing, so I asked on [protobuf repo](https://github.com/protocolbuffers/protobuf/issues/21204).

FYI, our current test case use protobuf 3.25, but latest one is 4.30.

kaylareopelle:
> Hi @kaylareopelle, I think this [test](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp-metrics/test/opentelemetry/exporter/otlp/metrics/metrics_exporter_test.rb#L483) cover the case; but I still add one test case that parse the request from raw byte.
> 
> I still couldn't figure out why this type_name is missing, so I asked on [protobuf repo](https://github.com/protocolbuffers/protobuf/issues/21204).
> 
> FYI, our current test case use protobuf 3.25, but latest one is 4.30.

Thank you and thanks for raising the issue! I think it would be helpful to start testing using newer versions of protobuf. [I created a PR to do this](https://github.com/open-telemetry/opentelemetry-ruby/pull/1753), but it fell down in priority and I couldn't see it through to the end. Feel free to open one if you have the bandwidth!

kaylareopelle (approved):

PR #1822, opened 2025-03-03
fix: OTLP exporter compression should be none by default. fixes #1798.

### LINKED ISSUE #1798: OTLP Exporter Compression should be `none` by default
> Compression: Compression key for supported compression types. Supported compression: gzip.
> Default: No value [3]
> Env vars: OTEL_EXPORTER_OTLP_COMPRESSION OTEL_EXPORTER_OTLP_TRACES_COMPRESSION OTEL_EXPORTER_OTLP_METRICS_COMPRESSION OTEL_EXPORTER_OTLP_LOGS_COMPRESSION

Type: [Enum](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#enum)

https://opentelemetry.io/docs/specs/otel/protocol/exporter/#configuration-options

The current default is set to `gzip`: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L54

patriciomacadden:


fbogsany:
The spec carries the important caveat:
> [3]: If no compression value is explicitly specified, SIGs can default to the value they deem most useful among the supported options.

We chose to make `gzip` the default in https://github.com/open-telemetry/opentelemetry-ruby/pull/934 as a result of this spec change: https://github.com/open-telemetry/opentelemetry-specification/pull/1923. Do we really want to reverse this decision? For users that rely on the default compression, this change in behaviour risks an erosion of trust in the project.

cc @arielvalentin 

Also note that there's another copy of this code here: https://github.com/open-telemetry/opentelemetry-ruby/blob/aacd8c8e264b110507c2a733dcc5309ca26aac66/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb#L39

patriciomacadden:
thanks for reviewing @fbogsany . I'll amend the trace exporter as well. 

Regarding your other comment, I'm not sure and I'll leave the decision to you. I'm an opentelemetry user, I was reading the issue list and I thought it was a good opportunity to start contributing to the project. So feel free to close this PR.

arielvalentin:
> The spec carries the important caveat:
> 
> > [3]: If no compression value is explicitly specified, SIGs can default to the value they deem most useful among the supported options.
> 
> We chose to make `gzip` the default in #934 as a result of this spec change: [open-telemetry/opentelemetry-specification#1923](https://github.com/open-telemetry/opentelemetry-specification/pull/1923). Do we really want to reverse this decision? For users that rely on the default compression, this change in behaviour risks an erosion of trust in the project.

What process have you all used in the past to make big decisions like this? I haven't earned my maintainers badge so all I can do is lay out a case for this. I did not get any feedback/input/concerns on the issue I opened so I assumed this was ok? https://github.com/open-telemetry/opentelemetry-ruby/issues/1798

I advocated for switching to none since is my experience so far has been inconsistent accross language SIGs. AFAICT Java, Go, and JavaScript default to no compression making Ruby the outlier and that has caught some of our end users by surprise.  

I do empathaize with how this _will_ impact our end users, however IMO it is also difficult to tell measure "erosion of trust" since we do not capture those metrics or have a sense outside of GitHub how important these kind of changes are. 

I do understand the sentiment of making a big change like this but the gem has not yet made 1.x status giving us some flexibility in making changes. It may be frustrating for users who are expecting default compression to `gzip` to change in the OTLP Exporters but if we accept this change we should probably communicate it as a breaking change. 

All that being said, I will leave the final decision of which `optional` default settings  in the SDK components to the maintainers.

kaylareopelle:
@arielvalentin, @fbogsany - checking in on this issue.

arielvalentin:
I don't think my plea was successful.

smoke (approved):

PR #1823, opened 2025-03-06
fix: Recover periodic metric readers after forking

chrisholmes:
This PR provides a fix to `PeriodicMetricReader` so that it continues working after a Ruby process is forked. This problem currently surfaces itself on applications that use forking webservers.

When forking a Ruby process only the thread creating the child process is coped across from the parent process, which means that background threads need to be created for monitoring tools to continue working. This means that, before this fix, `PeriodicMetricReader` would not work on the child process as the background thread is not copied.

The fix is implemented by overriding the `Process._fork` method provided in [Ruby versions 3.1+](https://ruby-doc.org/core-3.1.1/Process.html#method-c-_fork). Overriding this method allows the authors of monitoring tools to introduce callbacks on fork events. 

An alternative approach, required prior to 3.1, would require clients to call hooks directly from their forking webserver's configuration such as use `on_worker_boot` in `puma.rb`. Overriding `Process._fork` removes the need for this boilerplate.

When resetting the `PeriodicMetricReader`, the `after_fork` hook will first read and discard metrics that would have been published by the parent process. This is in order to avoid duplicate publishing of metrics that will be handled by the parent process.

The parent process' `PeriodicMetricReader` is left running after forking as it is assumed that clients will want to publish metrics from it still.


xuan-cao-swi:
Thanks @chrisholmes , just curious if it's possible to do something like [reset_on_fork](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb#L70) instead of prepend `Process`.

chrisholmes:
> Hi @chrisholmes , thanks for the explanation. My thinking is that as long as the thread (in child) keep running and export data for every period of time, then there is no need to recover it. Please correct me if I am misunderstand something or wrong.

hi @xuan-cao-swi , I'll try to explain my understanding:

Recovery is required because when we fork only the thread that performed the fork is copied from the parent onto the child process. This means that any background thread behaviour, such as PeriodicMetricReader, is stalled until the thread is re-created.

We have a few options for recovery:
1. Override `Process._fork`, which was introduced for this purpose
2. Ask developers to add a recovery hook into their codebase (such as in `on_worker_boot` in puma)
3. Use a trigger, such as the creation of a metric or the recording of a metric value, to recreate the thread

For options 1 & 2, there would be immediate recovery. For option 3, there would be a delay in recovery until the triggering event.

One scenario, where immediate recovery is desired, is if a developer creates gauges prior to forking. For option 3, these gauges will not be reported by the child process until the thread is recovered by the trigger, whose immediacy is not guaranteed. Whereas, a user of the metrics would probably expect metrics to be reported continuously without gaps.

Does this scenario help?

xuan-cao-swi:
Hi @chrisholmes, thanks for the explanation—you're right that there's no recovery point like BSP's `on_finish` for spans.

Option 3 would delay metric export, but could the recorded timestamps still help correlate metrics on the representation side? As you mentioned, recovery from the `metric_stream` record could work. That said, I’m not a fan of checking whether the forked periodic reader needs to start its thread on every record—though BSP does something similar per span finish.

The main concern with Option 1 is the monkey patching of the `Process` module. Option 2 seems more reasonable, but we’re aiming to minimize direct user interaction with the SDK. We’ll bring this up in the SIG meeting next week.

chrisholmes:
Hi, sorry for the bit of a delay here (work and holiday),

> Option 3 would delay metric export, but could the recorded timestamps still help correlate metrics on the representation side? As you mentioned, recovery from the `metric_stream` record could work. That said, I’m not a fan of checking whether the forked periodic reader needs to start its thread on every record—though BSP does something similar per span finish.

Honestly, I can't say I know fully. My understanding is that gauge style metrics will not report until there is a change in value so there will be a potential loss of data there.

> The main concern with Option 1 is the monkey patching of the `Process` module. Option 2 seems more reasonable, but we’re aiming to minimize direct user interaction with the SDK. We’ll bring this up in the SIG meeting next week.

I don't think this is a warranted concern as `_fork` was provided for this exact purpose. Here's a quick list of notable projects that use this feature:
- [ruby/debug](/blob/bead0984d241a91235e3d3bacd247f0363d2d530/lib/debug/session.rb#L2442)
- [mperham/connectionpool](https://github.com/mperham/connection_pool/blob/main/lib/connection_pool.rb#L52)
- [rails](https://github.com/rails/rails/blob/747f85f200e7bb2c1a31b4e26e5a5655e2dc0cdc/activesupport/lib/active_support/fork_tracker.rb#L6)
- [sqlite3-ruby](https://github.com/sparklemotion/sqlite3-ruby/blob/a0dd5c8e080891ee9f9ca9b5ec139e39c0efd8c0/lib/sqlite3/fork_safety.rb#L9)

On Option 2, I also saw this earlier that suggests this approach is no longer recommended for some projects (specifically sidekiq) as gem vendors should auto-detect for you: https://github.com/sidekiq/sidekiq/issues/5288#issuecomment-1100423464

xuan-cao-swi:
Thanks for the reference. I think this is a strong use case for the Ruby SDK to adopt prepending the `Process` module.
Regarding option 2, I meant to use the `after_fork` helper provided by other libraries.

kaylareopelle (approved):
I'm comfortable patching the Process module for this fix. I think the code footprint and communication to handle things in contrib is too complicated and difficult to maintain. This fix is my preference. 

Thanks for this contribution, @chrisholmes and for your patience waiting for a review.

Issue #1781, opened 2024-12-16
Unexpected behavior for log record attributes with symbols for keys


kaylareopelle:
When a log record attribute has a Symbol for a key, the following OpenTelemetry error log message is output:
`ERROR -- : OpenTelemetry error: invalid log record attribute key type Symbol on record: 'Thuja plicata with symbol attribute key'` 
Where `Thuja plicata` is the log body, and `:cedar` is the attribute key. 

The expectation is that all attribute keys are Strings. This is the same as the expectation for attributes on stable Ruby OTel signals, like Spans/SpanEvents for Traces. 

Despite this warning, the attribute is passed along to the exporter anyway, and it is successfully sent over OTLP. 

For example ([source](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1734228402085199)):

**OpenTelemetry Ruby logger SDK**
```ruby
# Use your Logger to  emit a log record
logger.on_emit(
  timestamp: Time.now,
  severity_text: 'INFO',
  body: 'Thuja plicata with symbol attribute key',
  attributes: { cedar: true },
)
```

**Shell output from running the code above**
```sh
[4:10](https://cloud-native.slack.com/archives/C01NWKKMKMY/p1734307839893059?thread_ts=1734228402.085199&cid=C01NWKKMKMY)
$ pwd
/Users/wsmoak/Projects/opentelemetry-ruby/examples/logs_sdk
wsmoak logs_sdk $ ruby logs_otlp.rb
Resolving dependencies...
Fetching gem metadata from https://rubygems.org/..
E, [2024-12-15T19:10:02.812695 #81798] ERROR -- : OpenTelemetry error: invalid log record attribute key type Symbol on record: 'Thuja plicata with symbol attribute key'
```

**Collector output from running the code above**
```sh
$ cat collector-output.txt | grep -3 Thuja
--
Timestamp: 2024-12-16 00:10:02.81267 +0000 UTC
SeverityText: INFO
SeverityNumber: Unspecified(0)
Body: Str(Thuja plicata with symbol attribute key)
Attributes:
     -> cedar: Bool(true)
Trace ID:
```

### Questions / Next Steps
1. Why are symbols forbidden as keys on trace-related attributes?
2. Should this reasoning extend to logs? 
3. If the reasoning should extend to logs, update the code to reject attributes with symbols as keys and update the error message to state that strings are required
4. If the reasoning should not extend to logs, remove the error message

[...]

PR #1825, opened 2025-03-11
feat: Update error message for invalid log record attribute

kaylareopelle:
Update the error message for invalid log record attribute keys and values. This keeps the error reporting strategy consistent with traces (using `handle_error`), and adds the attribute key and class to help with debugging.

Relates to #1781


logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -139,10 +139,10 @@ def validate_attributes(attrs)
           # Future refactor opportunity: https://github.com/open-telemetry/opentelemetry-ruby/issues/1739
           attrs.keep_if do |k, v|
             if !Internal.valid_key?(k)
-              OpenTelemetry.handle_error(message: "invalid log record attribute key type #{k.class} on record: '#{body}'")
+              OpenTelemetry.logger.warn("Invalid log record attribute key type #{k.class} for key #{k.inspect} on record: '#{body}'. Attribute keys must be Strings. Dropping attribute.")
```
arielvalentin, line 142: This is going to produce very noisy log output. 

Do we log warnings in the SDK for traces?

kaylareopelle: Great point. The Trace SDK does not use the logger to warn, instead, we handle the error like the current behavior in the logger. I wonder if updating the message for the `handle_error` call would be sufficient to let people know about the problem?

Here's the SDK logging code: https://github.com/open-telemetry/opentelemetry-ruby/blob/98c629edbc6445305bed65c4f4306431d683b474/sdk/lib/opentelemetry/sdk/internal.rb#L51-L63

kaylareopelle: I think the main issue with the previous message was that the `key` was missing from the output. That made it more difficult to identify what attribute was the problem. I'm fine with keeping it as a `handle_error` call or using this PR to move it to a `logger.warn`.

I might lean more toward the `handle_error` approach so that people can define their own error handlers to control the output behavior.

logs_sdk/Gemfile
```diff
@@ -11,6 +11,8 @@ gemspec
 eval_gemfile '../contrib/Gemfile.shared'
 
 group :test, :development do
+  gem 'mutex_m' if RUBY_VERSION >= '3.4'
```
ericmustin: I take it this is required for testing rails logging bits?

ericmustin (approved):

logs_sdk/Gemfile
kaylareopelle: Good question. This isn't needed anymore. Minitest was missing a mutex_m requirement on the version in my Gemfile.lock. However, that version was outdated. Minitest 5.21+ removed the dependency. I'll remove this.
```suggestion
```

logs_sdk/Gemfile
```diff
@@ -11,6 +11,7 @@ gemspec
 eval_gemfile '../contrib/Gemfile.shared'
 
 group :test, :development do
+
```
kaylareopelle:
```suggestion
```

Issue #1662, opened 2024-07-31
Metrics feedback


kaylareopelle:
We want your feedback on the `opentelemetry-metrics-sdk` and `opentelemetry-metrics-api` gems! Understanding how these libraries are used by the community will help us prioritize future work on Metrics.

chloe-meister:
Hi there 👋 

I've started some prototype implementation of metrics in a Rails app, and I could get some metrics exported from the SDK example. 

However, From the README it is not clear where and when I should pull the metrics and shutdown the provider?
These 2 lines:
```Ruby
OpenTelemetry.meter_provider.metric_readers.each(&:pull)
OpenTelemetry.meter_provider.shutdown
```

What we'd want to do is adding a counter metric each time a specific controller action is called. I can do this with a standard `before_action` filter, but from there I don't understand well when to pull the metrics from the reader? Any advice?

kaylareopelle:
Hi @chloe-meister! Thanks for reaching out! 

The README provides an example for a script that would run once, pull the metric, and exit the program.

If you're open to it, I think a periodic metric reader is a better fit for a Rails application. It'll collect the metrics and pass them to the exporter on a set interval. The process will live as long as the Rails application does, or until `OpenTelemetry.meter_provider.shutdown` is manually called.

Assuming you're using the OTLP exporter, rather than calling:
https://github.com/open-telemetry/opentelemetry-ruby/blob/2f87a1d1c90e51679838b36609e07cd3acdc9bcf/examples/metrics_sdk/metrics_collect_otlp.rb#L23-L25

You would call something like:
```ruby
# config/initializers/opentelemetry.rb
reader = OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader.new(
          exporter: OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporter.new,
          export_interval_millis: 3000,
          export_timeout_millis: 10000)

OpenTelemetry.meter_provider.add_metric_reader(reader)
```

You can create the meter and the counter the same way as in the example. 
If you prefer the console exporter, just swap `OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporter` with `OpenTelemetry::SDK::Metrics::Export::ConsoleMetricPullExporter`.

Using arguments above, your metrics would be exported every three seconds and would have a 10-second timeout. 

Keep in mind that if you're using the OTLP exporter, you also need to install the `opentelemetry-exporter-otlp-metrics` gem.

[...]

PR #1826, opened 2025-03-18
fix: use condition signal to replace sleep and remove timeout.timeout…

xuan-cao-swi:
… (use exporter timeout function - http.timeout)

PR #1827, opened 2025-03-19
Prevent slow shutdown caused by waiting full metrics export interval

akahn:
We noticed that when using the new metrics functionality in the library that our Rails application now took longer to shut down. We want to export metrics every 60 seconds, but we don't want to wait an additional 60 seconds for the app to stop. This was caused by the shutdown logic in `PeriodicMetricReader` which worked by setting `@continue` to false, indicating to the loop in `start` to no longer continue sleeping and writing. However, this meant that on shutdown, it takes a full `@export_interval` in order to "notice" that `@continue` is false an the program should exit. 

With this change, upon `shutdown`, the program performs a final `export` and then kills the writer thread. This is an approach suggested by @xuan-cao-swi in https://github.com/open-telemetry/opentelemetry-ruby/issues/1662#issuecomment-2714994848. 

Some questions:

1. Does the `kill` need to acquire the lock first? 
3. Is it still necessary to set `@continue` to false, if `@continue` is no longer used to control shutdown behavior?
4. Is `@continue` even needed any more? 
5. Would it be preferable, rather than the thread-killing approach, to use a `ConditionVariable` with `wait(@export_mutex, @export_interval)` in the writing loop, and `signal()` in `shutdown`? This would allow the writing loop "wake up" responsively on shutdown, while ordinarily waiting `@export_interval` before looping and writing again.

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
@@ -43,7 +43,11 @@ module OpenTelemetry
               @continue = false # force termination in next iteration
               @thread
             end
-            thread&.join(@export_interval)
+
+            # Trigger a final export before killing the writing thread
+            export(timeout: @export_timeout)
+            thread.kill
+
             @exporter.force_flush if @exporter.respond_to?(:force_flush)
             @exporter.shutdown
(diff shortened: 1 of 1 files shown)


kaylareopelle:
Hi @akahn, is this PR still needed now that #1826 has been merged?

akahn:
Fixed by #1826

PR #1836, opened 2025-04-15
chore: add Rubocop additional linter

xuan-cao-swi:
Related to https://github.com/open-telemetry/opentelemetry-ruby-contrib/issues/1392

I suggest use shared gamefile that contain all development dependencies (e.g. rubocop) since they are not part of gem bundle (i.e. only for internal use). Appraisal won't work with `eval_gemfile` but they have one active PR that will introduce the support for `eval_gemfile`. For gems (mostly exporter) that use appraisal, the development dependencies are in gemspec.


common/test/opentelemetry/common/utilities_test.rb
```diff
@@ -19,19 +19,19 @@ def shutdown(timeout: nil); end
 
   describe '#untraced?' do
     it 'returns true within an untraced block' do
-      assert_equal(true, common_utils.untraced { common_utils.untraced? })
+      assert(common_utils.untraced { common_utils.untraced? })
```
kaylareopelle, line 22: These suggestions from minitest can sometimes be false positives. I usually keep `assert_equal` when I want to validate the return value is `true` instead of any truthy value. Given the description of the test, this assertion may be a good candidate to keep using `assert_equal`. 

If you agree, the test below this one should also be reverted.

kaylareopelle (commented):
Thanks for putting this together, @xuan-cao-swi! I'm a little confused about your comment in the PR description about the Appraisals. It looks like they ran fine in the CI? I know we talked about this in the SIG, but I forgot -- do we need to consider that Appraisal issue a blocker before this is safe to merge?

kaylareopelle (approved):
Thank you! I'm going to change the conventional commit style to a `chore` for this PR so that it doesn't trigger a release, as the changes are largely style-based.

Issue #1841, opened 2025-04-22
Docs for start_timestamp inaccurately say they expect nanoseconds


cretz:
**Description of the bug**

In latest SDK/API, places like https://github.com/open-telemetry/opentelemetry-ruby/blob/b2cac915e2662f91e032097c4e0c6e9395128d64/api/lib/opentelemetry/trace/tracer.rb#L29 incorrectly state it's integer nanoseconds since the epoch when it's really `Time` or `Float` of seconds expected.

dmathieu:
There's a use case to passing nanoseconds here. So I'm not sure the doc should be changed.
Instead, I believe the SDK should be able to handle nanoseconds, as well as milliseconds like it currently does.

kaylareopelle:
Hi @cretz, thanks for reaching out.

@dmathieu is correct. The Tracing spec permits both nanoseconds and milliseconds for timestamps.

> A timestamp is the time elapsed since the Unix epoch.
>
>* The minimal precision is milliseconds.
>* The maximal precision is nanoseconds.

Source: https://opentelemetry.io/docs/specs/otel/trace/api/#timestamp 

Looking at the contrib code, when the `start_timestamp` is specified, it's a `Time` object. 

* [Example in httpx](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/fed9a0f71d45ee726baa7c6d20f7ebe55bc71935/instrumentation/httpx/lib/opentelemetry/instrumentation/httpx/plugin.rb#L69-L88)
* Example with rack: [link 1](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/5c647b56c2794c137c77a28c17b4bac1ec82a5bc/instrumentation/rack/lib/opentelemetry/instrumentation/rack/middlewares/tracer_middleware.rb#L105-L111), [link 2](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/bc4a937ed2a0d1898f0f19ae45a2b3a0ef9a067c/instrumentation/rack/lib/opentelemetry/instrumentation/rack/util/queue_time.rb#L19-L43)

[...]

PR #1846, opened 2025-05-15
fix: update docs for in_span method to list start_timestamp as a Time object

tannalynn:
Updates the docs line for start_timestamp so that it is listed as a Time object now, instead of an integer. 

resolves https://github.com/open-telemetry/opentelemetry-ruby/issues/1841


kaylareopelle (approved):
Thanks, @tannalynn!

PR #1851, opened 2025-06-04
feat: add exponential histogram in otlp metrics exporter

xuan-cao-swi:
Adding support to otlp metrics exporter when recording histogram that use exponential histogram bucket.

When determining the instrument kind (e.g., via `as_otlp_metrics(metrics)`), OpenTelemetry Ruby typically uses `metrics.instrument_kind`. However, this approach doesn't work for exponential_histogram, so it needs to inspect the type of the data points instead. This is similar to how Python uses `isinstance(metric.data, ExponentialHistogramType)`. Refactoring the Ruby implementation to follow a similar pattern could be a potential improvement.


kaylareopelle (approved):
Thank you for making these updates, Xuan. I like the idea you shared in your description about the way Python's exporter works. Could you capture that in an issue so we can work on it later?

PR #1857, opened 2025-06-28
feat!: Set logs exporter default to OTLP

kaylareopelle:
The OTLP logs exporter is available on RubyGems. This changes the default to match the other signals.


ericmustin (approved):
🎉

PR #1859, opened 2025-07-01
fix: Stop exporting metrics with empty data points

kaylareopelle:
The OTLP exporter raised the following error when attempting to export a histogram with a view without data points:
"ERROR -- : OpenTelemetry error: unexpected error in OTLP::MetricsExporter#encode - nil message not allowed here."

### DIFF
--- a/metrics_sdk/CHANGELOG.md
+++ b/metrics_sdk/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Release History: opentelemetry-metrics-sdk
 
+### v0.7.2 / 2025-07-03
+
+* FIXED: Coerce aggregation temporality to be a symbol for exponential histograms
+
 ### v0.7.1 / 2025-05-28
 
 * FIXED: Recover periodic metric readers after forking
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/version.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/version.rb
@@ -8,7 +8,7 @@ module OpenTelemetry
   module SDK
     module Metrics
       # Current OpenTelemetry metrics sdk version
-      VERSION = '0.7.1'
+      VERSION = '0.7.2'
     end
   end
 end


kaylareopelle:
A user encountered a similar problem. They tested the branch associated with this PR and the problem was resolved.

https://github.com/open-telemetry/opentelemetry-ruby/issues/1662#issuecomment-3033139374

metrics_sdk/test/integration/periodic_metric_reader_test.rb
```diff
@@ -11,6 +11,51 @@
...
+
+      metric_exporter = OpenTelemetry::SDK::Metrics::Export::InMemoryMetricPullExporter.new
+      periodic_metric_reader = OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader.new(export_interval_millis: 5000, export_timeout_millis: 5000, exporter: metric_exporter)
+
+      OpenTelemetry.meter_provider.add_metric_reader(periodic_metric_reader)
+
+      meter = OpenTelemetry.meter_provider.meter('test')
+      meter.create_histogram('example', unit: 's', description: 'test')
+
+      sleep(1)
```
arielvalentin, line 26: The export interval is quite long for the test. Is the sleep necessary?

metrics_sdk/test/integration/periodic_metric_reader_test.rb
```diff
@@ -11,6 +11,51 @@
...
+      boundaries = [0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10]
+
+      OpenTelemetry.meter_provider.add_view('http.server.request.duration',
+                                            type: :histogram,
+                                            aggregation: OpenTelemetry::SDK::Metrics::Aggregation::ExplicitBucketHistogram.new(boundaries: boundaries))
+
+      meter = OpenTelemetry.meter_provider.meter('test')
+      meter.create_histogram('http.server.request.duration', unit: 's', description: 'Duration of HTTP server requests.')
+
+      sleep(8)
```
arielvalentin, line 51: Similarly, can you make the export intervals shorter than 5 seconds to speed up the test a bit?

ericmustin (approved):
lgtm

metrics_sdk/test/integration/periodic_metric_reader_test.rb
kaylareopelle: Great point! I'm not sure if the sleep is necessary. I'll take a look and see if we can make these intervals shorter. I created #1862 to track this and will follow up post-conference.

PR #1872, opened 2025-07-19
fix: support hash bodies and attribute values in OTLP logs exporter

jmalcic:
Currently the logs exporter ignores hashes when encoding log bodies and attributes. This seems to be unintended, because the spec explicitly supports arbitrary nesting.

- [Type definitions](https://opentelemetry.io/docs/specs/otel/logs/data-model/#definitions-used-in-this-document)


kaylareopelle (approved):
Thank you!

PR #1881, opened 2025-08-07
feat!: add noop for Trace::Span attributes and events

xuan-cao-swi:
### Description

Adds no-op `attributes` and `events` methods to `OpenTelemetry::Trace::Span`.

This change ensures that when `in_span` returns a non-recording or invalid span (i.e., a no-op span), calls like `span.attributes` or `span.events` won't raise method errors. This prevents crashes in cases like [AWS Lambda instrumentation](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/opentelemetry-instrumentation-aws_lambda/v0.3.0/instrumentation/aws_lambda/lib/opentelemetry/instrumentation/aws_lambda/wrap.rb#L46), where such calls may occur on no-op spans.

Operational function from sdk: [attributes](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/span.rb#L39), [events](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/sdk/lib/opentelemetry/sdk/trace/span.rb#L50)

### DIFF
--- a/api/lib/opentelemetry/trace/span.rb
+++ b/api/lib/opentelemetry/trace/span.rb
@@ -78,6 +78,18 @@ module OpenTelemetry
         self
       end
 
+      # Retrieve attributes
+      #
+      # Note that the OpenTelemetry project
+      # {https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-semantic-conventions.md
+      # documents} certain "standard attributes" that have prescribed semantic
+      # meanings.
+      #
+      # @return [hash] returns empty hash
+      def attributes
+        {}
+      end
+
       # Add a link to a {Span}.
       #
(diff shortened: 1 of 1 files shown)


ericmustin:
seems fine

kaylareopelle (changes requested):
This works for me too. Like we discussed in the SIG, I'm going to remove the `!` from the conventional commit to avoid triggering a major version release. Adding APIs isn't a breaking change. 

For consistency with the rest of the methods in this file, could you add some basic tests for the new methods?

kaylareopelle (approved):
Thanks!

kaylareopelle:
Thank you for catching this. We'll work on reverting the change.

Issue #1882, opened 2025-08-07
Make Gauge instruments compatible with Prometheus


kaylareopelle:
In a comment on #1877: 
> I switched to using a plain Gauge, however, it does not respect OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE like Counter does. I need cumulative metrics for Prometheus.

The Last Value aggregation is used for gauges. It currently has a keyword argument for aggregation_temporality, but it does not take the environment variable into account, likely because the data model for gauges doesn't reference aggregation temporality.

Instead, it says: 
> Gauges do not provide an aggregation semantic, instead “last sample value” is used when performing operations like temporal alignment or adjusting resolution. [^1]

Sums, in contrast, have this bullet point: 
> An Aggregation Temporality of delta or cumulative. [^2]

Since Prometheus needs a cumulative temporality, let's look into how best to support; consulting implementations from other languages. Whatever we use should also be tested with ObservableGauge.

[^1]: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#gauge
[^2]: https://opentelemetry.io/docs/specs/otel/metrics/data-model/#sums

wsmoak:
With the latest code on main 6168e9c8277867 (after #1855 merged) I am not seeing `AggregationTemporality` being sent for Gauge any more in the debug output.  And my Gauge value shows up in Prometheus.

Taking it out of the constructor for LastValue sounds good, it isn't being used / doesn't make sense there.

wsmoak:
Yes you can resolve it, I'm not having trouble with gauges and Prometheus using the latest code.

PR #1885, opened 2025-08-11
feat: drop aggregation_temporality method and instance variable for last_va…

xuan-cao-swi:
…lue and drop aggregation

### Description:

LastValue aggregation doesn't have temporality, so remove it for better clarity.

[Reference](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#last-value-aggregation)

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/state/asynchronous_metric_stream.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/state/asynchronous_metric_stream.rb
@@ -10,11 +10,9 @@ module OpenTelemetry
       module State
         # @api private
         #
-        # The MetricStream class provides SDK internal functionality that is not a part of the
-        # public API.
-        class AsynchronousMetricStream
-          attr_reader :name, :description, :unit, :instrument_kind, :instrumentation_scope, :data_points
-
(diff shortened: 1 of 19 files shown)


kaylareopelle:
Relates to #1882

metrics_sdk/lib/opentelemetry/sdk/metrics/state/asynchronous_metric_stream.rb
```diff
@@ -49,6 +49,9 @@ def initialize(
           def collect(start_time, end_time)
             invoke_callback(@timeout, @attributes)
 
+            is_monotonic = @aggregation.respond_to?(:monotonic?) ? @aggregation.monotonic? : nil
+            aggregation_temporality = @aggregation.respond_to?(:aggregation_temporality) ? @aggregation.aggregation_temporality : nil
```
kaylareopelle: This doesn't need to be addressed as part of this PR, but I just realized our tests for the metric streams are lacking. They seem to be hit through other tests with SimpleCov. I think we should consider adding class-specific tests in addition.

xuan-cao-swi: Sure, I created the [issue](https://github.com/open-telemetry/opentelemetry-ruby/issues/1890)

Issue #1880, opened 2025-08-06
Add support for "LowMemory" temporality preference


kaylareopelle:
`OTEL_EXPORTER_OTLP_METRICS_TEMPORALITY_PREFERENCE` has three valid values, `Cumulative`, `Delta`, and `LowMemory`. 

We updated the default value to be `cumulative` to match the spec in #1555, but have not added conditions for low memory.

With low memory, the behavior is: 
> * `LowMemory`: This configuration uses Delta aggregation temporality for Synchronous Counter and Histogram and uses Cumulative aggregation temporality for Synchronous UpDownCounter, Asynchronous Counter, and Asynchronous UpDownCounter instrument kinds.

[Source](https://github.com/open-telemetry/opentelemetry-specification/blob/10acd40725898fff74fb455c788060ce06725540/specification/metrics/sdk_exporters/otlp.md?plain=1#L50-L55)

PR #1886, opened 2025-08-11
feat: add low memory to TEMPORALITY_PREFERENCE

xuan-cao-swi:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/1880

Based on [source](https://github.com/open-telemetry/opentelemetry-specification/blob/10acd40725898fff74fb455c788060ce06725540/specification/metrics/sdk_exporters/otlp.md?plain=1#L50-L55), it should follow the diagram:

| Preference Value | Counter    | Async Counter    | Histogram  | UpDownCounter | Async UpDownCounter |
|------------------|------------|------------------|----------- |---------------|-------------------- |
| **Cumulative**   | Cumulative | Cumulative       | Cumulative | Cumulative    | Cumulative          |
| **Delta**        | Delta      | Delta            | Delta      | Cumulative    | Cumulative          |
| **LowMemory**    | Delta      | Cumulative       | Delta      | Cumulative    | Cumulative          |

Updown counter should always be cumulative, unless user specifically define it to delta through view.

### DIFF
--- a/metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/aggregation_temporality_test.rb
+++ b/metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/aggregation_temporality_test.rb
@@ -1,0 +1,295 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+require 'test_helper'
+
+describe OpenTelemetry::SDK::Metrics::Aggregation::AggregationTemporality do
+  describe '.determine_temporality' do
+    describe 'with CUMULATIVE preference' do
+      it 'returns cumulative for counter instrument' do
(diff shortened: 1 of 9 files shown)


metrics_sdk/test/opentelemetry/sdk/metrics/aggregation/aggregation_temporality_test.rb
```diff
@@ -0,0 +1,295 @@
...
+        result = OpenTelemetry::SDK::Metrics::Aggregation::AggregationTemporality.determine_temporality(
+          aggregation_temporality: 'lowmemory',
+          instrument_kind: :counter,
+          default: :cumulative
+        )
+        _(result.temporality).must_equal :delta
+      end
+    end
+
+    describe 'with unknown string values' do
```
kaylareopelle, line 229: What do you think about adding a log message to let people know when the string value doesn't match? I could see myself writing `low_memory` instead of `lowmemory` or making some other typo that would be hard to track down.

xuan-cao-swi: Added the `low_memory` as one extra condition.

PR #1899, opened 2025-08-19
fix!: Remove Span APIs for attributes and events

kaylareopelle:
These APIs are not spec compliant. The SDK description makes this clear. This reverts #1881.

We need to find another solution for AWS Lambda instrumentation.

cc @xuan-cao-swi @fbogsany


robertlaurin (approved):
Thanks for the quick fix @kaylareopelle

PR #1894, opened 2025-08-15
tests: add test case for metric_store and metric_view

xuan-cao-swi:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/1890

### Description

Found a couple of issues while developing the test case:
1. The current callback timeout applies to the entire set of callbacks of the instrument instead of each callback individually. This PR fixes that.
2. A faulty callback that raises an error can crash the entire application. This update adds a safeguard to log the error while allowing the application to continue running. Also we shouldn't update to instrument and merge attributes if callback return non-numeric data or nil. ([Reference](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.48.0/specification/metrics/sdk.md#observations-inside-asynchronous-callbacks))
3. Currently, each instrument supports only one view. Assigning multiple views to the same instrument can lead to inconsistent data_point values, as each view may modify the underlying data. This PR does not address the issue, but it will be resolved in the future.

Each metric_stream should be bound to its own instrument. While most functionality is tested through instruments, here we focus on fundamental correctness, which will be important when views and exemplars are updated.

PR #1908, opened 2025-08-26
refactor: Refactor view that allow multiple view for single instrument 

xuan-cao-swi:
### Description

Currently, a view modifies the `data_points` directly from the `metric_stream`. As a result, when multiple views attempt to modify the same `data_points`, they are no longer working with the original data.

This PR moves `data_points` back into the aggregation, ensuring each view is bound to its own independent data storage.

PS. This branch is derived from https://github.com/open-telemetry/opentelemetry-ruby/pull/1894


xuan-cao-swi:
Close it as it's not a good solution

PR #1909, opened 2025-08-26
feat(metrics-sdk)!: Metrics cardinality limit

xuan-cao-swi:
### Description

This PR introduces **cardinality limits** for metric attributes, as defined in the [metrics SDK specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/metrics/sdk.md#cardinality-limits).

Users can configure the cardinality limit in two ways:

1. **Through a view**:

```ruby
OpenTelemetry.meter_provider.add_view(
  'test_histogram',
  aggregation: OpenTelemetry::SDK::Metrics::Aggregation::ExplicitBucketHistogram.new,
  aggregation_cardinality_limit: 1
)
```

2. **Through an exporter**:

```ruby
exporter_with_limit = ConsoleMetricPullExporter.new(aggregation_cardinality_limit: 100)
```

If no limit is provided, the default value of **2000** is used (also negative value will turn to 0 for the limit).

The cardinality limit is enforced during the aggregation update process. When the number of unique attribute combinations exceeds the limit, the extra data points are grouped into a separate data point with the attribute:

```ruby
{ 'otel.metric.overflow' => true }
```

This overflow data point is treated like any other data point.

**Future work**: There is currently duplicated logic across different aggregation implementations. A shared base aggregation class should be introduced to reduce repetition.


metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/explicit_bucket_histogram.rb
```diff
@@ -59,60 +60,76 @@ def collect(start_time, end_time, data_points)
...
-          def update(amount, attributes, data_points, exemplar_offer: false)
-            hdp = data_points.fetch(attributes) do
-              if @record_min_max
-                min = Float::INFINITY
-                max = -Float::INFINITY
-              end
+          def update(amount, attributes, data_points, cardinality_limit, exemplar_offer: false)
+            hdp = if data_points.key?(attributes)
+                    data_points[attributes]
+                  elsif data_points.size >= cardinality_limit
```
mwear: We have an off by 1 error. As written we will end up with cardinality_limit + 1 data points, since the overflow is added to an already full map. FWIW, this change would align with how the Go and JS SDKs handle this:

JS:
https://github.com/open-telemetry/opentelemetry-js/blob/3f1a493ac6a2d1eceee6ea8a179602826cf27c3c/packages/sdk-metrics/src/state/DeltaMetricProcessor.ts#L31

https://github.com/open-telemetry/opentelemetry-js/blob/3f1a493ac6a2d1eceee6ea8a179602826cf27c3c/packages/sdk-metrics/src/state/DeltaMetricProcessor.ts#L44

Go:
https://github.com/open-telemetry/opentelemetry-go/blob/a51ddb125a449109e6a3d2be88989995e3fb89ae/sdk/metric/internal/aggregate/limit.go#L36
  

```suggestion
                  elsif data_points.size >= cardinality_limit - 1
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
```diff
@@ -221,49 +223,53 @@ def collect(start_time, end_time, data_points)
...
-            # fetch or initialize the ExponentialHistogramDataPoint
-            hdp = data_points.fetch(attributes) do
-              if @record_min_max
-                min = Float::INFINITY
-                max = -Float::INFINITY
-              end
+          def update(amount, attributes, data_points, cardinality_limit, exemplar_offer: false)
+            hdp = if data_points.key?(attributes)
+                    data_points[attributes]
+                  elsif data_points.size >= cardinality_limit
```
mwear:
```suggestion
                  elsif data_points.size >= cardinality_limit - 1
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/last_value.rb
```diff
@@ -33,29 +34,51 @@ def collect(start_time, end_time, data_points)
...
-            unless reservoir
-              reservoir = @exemplar_reservoir.dup
-              reservoir.reset
-              @exemplar_reservoir_storage[attributes] = reservoir
-            end
+          def update(increment, attributes, data_points, cardinality_limit, exemplar_offer: false)
+            # Check if we already have this attribute set
+            ndp = if data_points.key?(attributes)
+                    data_points[attributes]
+                  elsif data_points.size >= cardinality_limit
```
mwear:
```suggestion
                  elsif data_points.size >= cardinality_limit - 1
```

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/sum.rb
```diff
@@ -51,41 +51,61 @@ def collect(start_time, end_time, data_points)
             end
           end
 
+          def update(increment, attributes, data_points, cardinality_limit, exemplar_offer: false)
+            return if @monotonic && increment < 0
+
+            # Check if we already have this attribute set
+            ndp = if data_points.key?(attributes)
+                    data_points[attributes]
+                  elsif data_points.size >= cardinality_limit
```
mwear:
```suggestion
                  elsif data_points.size >= cardinality_limit - 1
```

mwear (commented):
Mostly LGTM. The current calculations are off by one, however. I added suggestions and some links to the Go and JS implementations. The tests will also need to be updated since they are also off by one.

mwear (approved):

kaylareopelle (approved):

PR #1910, opened 2025-08-29
Issue 1889

bschrag620:


test_helpers/lib/opentelemetry/test_helpers.rb
```diff
@@ -42,7 +42,7 @@ def with_test_logger
     end
 
     def exportable_timestamp(time = Time.now)
-      (time.to_r * 1_000_000_000).to_i
+      OpenTelemetry::Common::Utilities.time_in_nanoseconds(time))
```
kaylareopelle: There's a syntax error here causing some CI failures:
```suggestion
      OpenTelemetry::Common::Utilities.time_in_nanoseconds(time)
```

common/lib/opentelemetry/common/utilities.rb
```diff
@@ -42,6 +42,21 @@ def timeout_timestamp
...
+        time_in_nanoseconds(Time.now)
+      end
+
+      # Converts the provided timestamp to nanosecond integer
+      #
+      # @param timestamp [Time] the timestamp to convert
+      # @return [Integer]
+      def time_in_nanoseconds(timestamp)
+        (timestamp.to_r * 1_000_000_000).to_i
+      end
```
kaylareopelle: These methods are pretty similar. What do you think about taking the approach from the test_helpers.rb file, and setting `Time.now` as the default arg for a nanosecond conversion method to reduce to one method? 

Also, could you add some unit tests?

bschrag620: Done and done!

PR #1911, opened 2025-08-29
feat: add record_exception option for in_span

robertlaurin:
Non breaking change, default behaviour is preserved. Allows for the `in_span` convenience method to be configured to not capture exception events. Maintains recording of the span status as is.

Related context: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1653

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
@@ -140,12 +140,13 @@ module OpenTelemetry
               scale_change = get_scale_change(low, high)
               downscale(scale_change, hdp.positive, hdp.negative)
               new_scale = @mapping.scale - scale_change
-              hdp.scale = new_scale
               @mapping = new_mapping(new_scale)
               bucket_index = @mapping.map_to_index(amount)
 
               OpenTelemetry.logger.debug "Rescaled with new scale #{new_scale} from #{low} and #{high}; bucket_index is updated to #{bucket_index}"
             end
+
+            hdp.scale = @mapping.scale
 
             # adjust buckets based on the bucket_index
             if bucket_index < buckets.index_start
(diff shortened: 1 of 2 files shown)


fbogsany (approved):

api/lib/opentelemetry/trace/tracer.rb
```diff
@@ -31,12 +31,12 @@ class Tracer
...
       #   span to the block.
-      def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil)
+      def in_span(name, attributes: nil, links: nil, start_timestamp: nil, kind: nil, record_exception: true)
         span = nil
         span = start_span(name, attributes: attributes, links: links, start_timestamp: start_timestamp, kind: kind)
         Trace.with_span(span) { |s, c| yield s, c }
       rescue Exception => e # rubocop:disable Lint/RescueException
-        span&.record_exception(e)
+        span&.record_exception(e) if record_exception
         span&.status = Status.error("Unhandled exception of type: #{e.class}")
```
kirs, line 40: I find that `Unhandled` might be misleading here. Maybe just `Exception of type:`?

fbogsany: `Unhandled` was included in the message to indicate that the exception was unhandled at the time the span was finished, based on the [language in the spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md#recording-an-exception):

> An exception SHOULD be recorded as an `Event` on the span during which it occurred if and only if it remains unhandled when the span ends and causes the span status to be set to ERROR.

It may be handled later, but it was not handled within the lifetime of the span.

The spec [recommends](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.48.0/specification/trace/api.md#set-status):

> When the status is set to `Error` by Instrumentation Libraries, the `Description` SHOULD be documented and predictable.

Sadly, we're violating [this other recommendation](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/general/recording-errors.md#recording-errors-on-spans), but I think changing the description would be more problematic at this point:

> It's NOT RECOMMENDED to duplicate status code or `error.type` in span status description.

PR #1929, opened 2025-09-26
Expose a setter to configure service namespace resource.

dhruvCW:
Add `Configurator#service_namespace` so that we can configure the service namespace when setting up the sdk

example
```ruby
  c.service_name = 'demo OTEL app'
  c.service_version = '2.0'
  c.service_namespace = 'OTEL'
    
  c.use_all
end
```


dhruvCW:
Ah my bad didn't realize this is still under development 😅

PR #1932, opened 2025-10-08
docs: upgrade http example

bellvat:
While following the example, I tried visiting the links and found them broken. 
This commit fixes the broken links and updates the example to use the latest span attributes.


examples/http/client.rb
```diff
@@ -26,14 +26,14 @@
...
-# https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/http.md
+# https://github.com/open-telemetry/semantic-conventions/blob/main/docs/http/http-spans.md#http-client-span
 
 # Span name should be set to URI path value:
 tracer.in_span(
   url,
   attributes: {
-    'component' => 'http',
-    'http.method' => 'GET',
+    'url.scheme' => 'http',
```
dmathieu: How about using the semantic conventions package here, and for all the other attributes?


```suggestion
    OpenTelemetry::Semconv::URL::URL_SCHEME => 'http',
```

bdory: Hello @dmathieu, thank you for your feedback. I have updated the files accordingly.

PR #1935, opened 2025-10-12
fix: issue with sending traces to IPv6 endpoints

chen-anders:
Resolves: https://github.com/open-telemetry/opentelemetry-ruby/issues/1721

This PR resolves the issue around using the opentelemetry-ruby SDK for apps that send traces to an IPv6 endpoint using the suggested fix in the original issue. We also add some tests for verifying the expected value of the collector hostname based on a passed-in value.

*Reviewer's Note*: Tests were written with significant AI assistance.


exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -109,7 +109,7 @@ def fetch_ssl_verify_mode
           end
 
           def http_connection(uri, ssl_verify_mode, certificate_file, client_certificate_file, client_key_file)
-            http = Net::HTTP.new(uri.host, uri.port)
+            http = Net::HTTP.new(uri.hostname, uri.port)
```
kaylareopelle, line 112: Thank you for opening this PR, @chen-anders! I appreciate the Reviewer's note. The tests look good to me, but I think we should add them to the other libraries getting the IPv6 compatible hostname too. Could you update the tests in `otlp-http`, `otlp-logs` and `otlp-metrics` too?

chen-anders: I went ahead and added the tests requested for the other exporters.

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -357,6 +357,148 @@
...
+      _(http.address).must_equal '::1'
+      _(http.port).must_equal 4318
+      _(exp.instance_variable_get(:@path)).must_equal '/v1/traces'
+    end
+
+    it 'exports span data with IPv6 address' do
+      stub_request(:post, 'http://[::1]:4318/v1/traces').to_return(status: 200)
+      exp = OpenTelemetry::Exporter::OTLP::Exporter.new(endpoint: 'http://[::1]:4318/v1/traces')
+      span_data = OpenTelemetry::TestHelpers.create_span_data
+      result = exp.export([span_data])
```
acarella1: Hi @chen-anders thank you for pushing this along.
Overall looks good, but might need to consider the performance of the numerous test additions which use `exp.export(...)`, which could lead to test pipeline performance slowdown.

Also I believe that the of actual exporting of span data is covered by other tests, so these new tests should only be concerned with parser validation such as:
```
      http = exp.instance_variable_get(:@http)
      _(http.address).must_equal '::1'
      _(http.port).must_equal 4318
```

Do you agree @kaylareopelle ?

chen-anders: I went ahead and removed the tests in https://github.com/open-telemetry/opentelemetry-ruby/pull/1935/commits/0d18d617a496469a544efe77f735c332ec50f19b

kaylareopelle:
Hi @chen-anders. I'm sorry this has been open and approved for so long without a merge. We have some merge conflicts and other changes in this PR now that I don't quite understand. Can you take a look? I can try to figure out what's going on with the CLA.

Issue #1888, opened 2025-08-11
Add logging about export status to Metrics


kaylareopelle:
As a follow-up to https://github.com/open-telemetry/opentelemetry-ruby/issues/1662#issuecomment-3170642537:

> Please add some debug output for the metrics exporter. Working with logs, if I set
> 
> OpenTelemetry.logger.level = Logger::DEBUG
> 
> Then I get
> 
> D, [2025-08-09T07:27:50.918650 #25729] DEBUG -- : Successfully exported 1 log records
> 
> But for metrics there is no output.

Let's add some information about successful export to Metrics. I think the Periodic Metric Reader's `export` method may be a good place to start: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/bd01ee9abda7568bcc8a5e65848b910107dec7d4/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb#L117-L128

wsmoak:
This is what I'm using locally.  I copied the idea from the debug output for logging.  
Looks like I'm in a different place than you suggested however.

```
diff --git a/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb b/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
index eacf0b3d..2a4f0297 100644
--- a/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
+++ b/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
@@ -82,10 +82,21 @@ module OpenTelemetry
             export(collect)
           end
 
+          def report_result(result_code, metrics)
+            if result_code == SUCCESS
+              OpenTelemetry.logger.debug("Successfully exported #{metrics.size} metrics")
+            else
+              OpenTelemetry.handle_error(exception: ExportError.new("Unable to export #{metrics.size} metrics"))
+              OpenTelemetry.logger.error("Result code: #{result_code}")
+            end
+          end
+
           # metrics Array[MetricData]
           def export(metrics, timeout: nil)
             @mutex.synchronize do
-              send_bytes(encode(metrics), timeout: timeout)
+              result = send_bytes(encode(metrics), timeout: timeout)
+              report_result(result, metrics)
+              result
             end
           end
```

PR #1947, opened 2025-10-18
feat: Add logging about export status to Metrics SDK

wsmoak:
Adds some debug log output for successfully exporting metrics, and additional error logging.  Implementation is borrowed from the Batch Log Record Processor.

Does not include a count.  Eventually that can be handled with the `otel.sdk.exporter.metric_data_point.exported` internal metric.  See https://opentelemetry.io/docs/specs/semconv/otel/sdk-metrics/

Fixes: #1888

### DIFF
--- a/metrics_sdk/test/opentelemetry/sdk/metrics/export/periodic_metric_reader_test.rb
+++ b/metrics_sdk/test/opentelemetry/sdk/metrics/export/periodic_metric_reader_test.rb
@@ -21,7 +21,6 @@ describe OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader do
     attr_reader :exported_metrics
 
     def export(metrics, timeout: nil)
-      puts "TestExporter export called"
       s = @status_codes.shift
       if s.nil? || s == SUCCESS
         @exported_metrics.concat(metrics)


wsmoak:
I need advice on how to test this, if it will require tests.  Everything I try ends up with an awful mess of mocks.

wsmoak:
Okay, I started over by copying the test for the BatchLogRecordProcessor, and that worked out better.

That test has `skip if Gem.win_platform?` -- let's see if it is needed here too.

xuan-cao-swi (approved):
LGTM

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
```diff
@@ -123,7 +123,18 @@ def start
           def export(timeout: nil)
             @export_mutex.synchronize do
               collected_metrics = collect
-              @exporter.export(collected_metrics, timeout: timeout || @export_timeout) unless collected_metrics.empty?
+              result_code = @exporter.export(collected_metrics, timeout: timeout || @export_timeout) unless collected_metrics.empty?
+              report_result(result_code)
```
utay, line 127: I think there's a bug: when `collected_metrics` is an empty array, `report_result` will log an error because `result_code` is `nil`

wsmoak: Sorry about that!  I opened this to fix it, please comment if you have any feedback: https://github.com/open-telemetry/opentelemetry-ruby/pull/1953

utay: Thanks! LGTM

PR #1952, opened 2025-10-30
feat: Bump semantic conventions to version 1.37.0

kaylareopelle:
Semantic conventions update for version 1.37.0

Also includes an adjustment to the attributes jinja template to account for multi-line comments in response to https://github.com/open-telemetry/semantic-conventions/issues/3019

---
> ### Outdated
> This was pretty smooth overall. The only problem was with the Gen AI examples. They needed to be manually commented out. After the opening `[` the rest of the example wasn't commented out. This problem also exists in the latest release, 1.38.0.
> 
> It might be related to a fix in the actual semconv repo. Things look pretty good from the weaver perspective to me, but maybe you have an idea, @robbkidd? 
> 
> Where the sample values are in the weaver template:
> https://github.com/open-telemetry/opentelemetry-ruby/blob/1f4f85ef19914fdb31cbd314c59f3f247b61df07/semantic_conventions/templates/registry/ruby/attributes.j2#L44-L51

### DIFF
--- a/semantic_conventions/lib/opentelemetry/semconv/incubating/aws/attributes.rb
+++ b/semantic_conventions/lib/opentelemetry/semconv/incubating/aws/attributes.rb
@@ -23,7 +23,7 @@ module OpenTelemetry
     module Incubating
       module AWS
         # @!group Attribute Names
-      
+
         # The unique identifier of the AWS Bedrock Guardrail. A [guardrail](https://docs.aws.amazon.com/bedrock/latest/userguide/guardrails.html) helps safeguard and prevent unwanted behavior from model responses or user messages.
         #
         # @note Stability Level: development
@@ -31,7 +31,7 @@ module OpenTelemetry
         # @example Sample Values
         # sgi5gkybzqak
         AWS_BEDROCK_GUARDRAIL_ID = 'aws.bedrock.guardrail.id'
-    
+
         # The unique identifier of the AWS Bedrock Knowledge base. A [knowledge base](https://docs.aws.amazon.com/bedrock/latest/userguide/knowledge-base.html) is a bank of information that can be queried by models to generate more relevant responses and augment prompts.
         #
         # @note Stability Level: development
@@ -39,7 +39,7 @@ module OpenTelemetry
         # @example Sample Values
         # XFWUPB9PAW
         AWS_BEDROCK_KNOWLEDGE_BASE_ID = 'aws.bedrock.knowledge_base.id'
-    
+
         # The JSON-serialized value of each item in the `AttributeDefinitions` request field.
         #
         # @note Stability Level: development
@@ -47,7 +47,7 @@ module OpenTelemetry
         # @example Sample Values
         # { "AttributeName": "string", "AttributeType": "string" }
         AWS_DYNAMODB_ATTRIBUTE_DEFINITIONS = 'aws.dynamodb.attribute_definitions'
-    
+
         # The value of the `AttributesToGet` request parameter.
         #
         # @note Stability Level: development
@@ -56,12 +56,12 @@ module OpenTelemetry
         # lives
         # id
         AWS_DYNAMODB_ATTRIBUTES_TO_GET = 'aws.dynamodb.attributes_to_get'
-    
+
         # The value of the `ConsistentRead` request parameter.
         #
         # @note Stability Level: development
         AWS_DYNAMODB_CONSISTENT_READ = 'aws.dynamodb.consistent_read'
-    
+
         # The JSON-serialized value of each item in the `ConsumedCapacity` response field.
         #
         # @note Stability Level: development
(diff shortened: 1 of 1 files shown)


semantic_conventions/lib/opentelemetry/semconv/incubating/client/attributes.rb
```diff
@@ -31,9 +31,9 @@ module CLIENT
         # @note Stability Level: stable
         #
         # @example Sample Values
-        #   client.example.com
-        #   10.1.2.80
-        #   /tmp/my.sock
+        #   # client.example.com
```
arielvalentin: where are these extra octothorps `#` coming from? The jinja template?

semantic_conventions/templates/registry/ruby/attributes.j2
```diff
@@ -44,7 +44,7 @@ module OpenTelemetry
       # @example Sample Values
         {%- if attribute.examples is sequence %}
           {%- for example in attribute.examples %}
-      #   {{ example }}
+      #   {{ example | comment }}
```
arielvalentin: Is that because the example or comment may _already_ have an octothorp in it?

semantic_conventions/lib/opentelemetry/semconv/incubating/client/attributes.rb
kaylareopelle: Good catch, I'm not sure. I can take a look next week.

semantic_conventions/templates/registry/ruby/attributes.j2
kaylareopelle: I was directed to use this for multi line strings. 

However, looking at the comment, I may have got something wrong:


https://github.com/open-telemetry/semantic-conventions/issues/3019#issuecomment-3473104601

Perhaps just using the filter without the prefixed # would solve the issue.

kaylareopelle: (assuming the "that" you're referring to is the addition of "| comment" )

kaylareopelle: @arielvalentin - Your hypothesis is right, there's already a `#` in the `example`/`comment`. 

This might've been related to a weaver version change. By including the `#` where we had it in earlier versions, we were able to get some extra indentation before the examples. I can't figure out how to do that now, so the documentation will look like:
```
      # @example Sample Values
      # CERN-LineMode/2.15 libwww/2.17b3
      # Mozilla/5.0 (iPhone; CPU iPhone OS 14_7_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Mobile/15E148 Safari/604.1
      # YourApp/1.0.0 grpc-java-okhttp/1.27.2
```
I can also add a colon to the end of the `Sample Values`, if we think that'll make things look better.

semantic_conventions/lib/opentelemetry/semconv/incubating/aspnetcore/attributes.rb
arielvalentin, line 1: These seem unnecessary to me since they are language specific. 

I feel like we need to be more intentional in what conventions we add to avoid allocating memory for these unused values. 

Do these get lazy loaded? I.e. do you have to require specific semconv namespaces in your code to use these? or are they all required at startup?

semantic_conventions/lib/opentelemetry/semconv/incubating/client/attributes.rb
arielvalentin: I received a notification that you commented on this but for some reason I do not see it in the UI now. 

Did you mention something about a weaver change that added the extra octothorpe?

semantic_conventions/lib/opentelemetry/semconv/incubating/aspnetcore/attributes.rb
kaylareopelle: Yes, you have to require the specific semconv files in your code to use them for the new conventions. They won't be required automatically.

semantic_conventions/lib/opentelemetry/semconv/incubating/client/attributes.rb
kaylareopelle: @arielvalentin -- strange! This is the comment I submitted before.

This might've been related to a weaver version change. By including the # where we had it in earlier versions, we were able to get some extra indentation before the examples. I can't figure out how to do that now, so the documentation will look like:

```
      # @example Sample Values
      # CERN-LineMode/2.15 libwww/2.17b3
      # Mozilla/5.0 (iPhone; CPU iPhone OS 14_7_1 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/14.1.2 Mobile/15E148 Safari/604.1
      # YourApp/1.0.0 grpc-java-okhttp/1.27.2
```

I can also add a colon to the end of the Sample Values, if we think that'll make things look better

semantic_conventions/lib/opentelemetry/semconv/aspnetcore/attributes.rb
```diff
@@ -28,76 +28,68 @@ module ASPNETCORE
       # @note Stability Level: stable
       #
       # @example Sample Values
-      #   # handled
-      #   # unhandled
-      #
+      # handled
+      # unhandled
```
robbkidd, line 32: I'm sorry I was unavailable the past couple weeks to maybe have caught this earlier.

😬 These indents were part of the Yard markup for getting these values to appear under the `@example Sample Values` section in the rendered docs.

| Before, with inner indent) | After, without inner indent |
| -- | -- |
| Image | Image |

If showing sample values as Yard `@example` has gotten too wonky in the weaver templating, we/I can update the template to do something more straightforward.

semantic_conventions/templates/registry/ruby/attributes.j2
```diff
@@ -44,7 +44,7 @@ module OpenTelemetry
       # @example Sample Values
         {%- if attribute.examples is sequence %}
           {%- for example in attribute.examples %}
-      #   {{ example | comment }}
+          {{ example | comment }}
```
robbkidd, line 47: I think I see another fix option: don't pipe the example sample value through the templating `comment` macro (which prepends an octothorp):

```suggestion
      #   {{ example }}
```

robbkidd: Nope! It's more complicated than that. My current suggestion is for us to remove including examples. The shapes they come in are too varied.

PR #1953, opened 2025-11-02
fix: Do not log error when there are no metrics to export

wsmoak:
When collected_metrics is empty, result_code does not get set, and so report_result would log an error with a blank code.  

Since the exporters already log if they encounter problems, there is no need to repeat the error logging here.

Reported by: @utay

Related to: #1947 and #1888

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
@@ -130,9 +130,7 @@ module OpenTelemetry
           end
 
           def report_result(result_code)
-            if result_code == Export::SUCCESS
-              OpenTelemetry.logger.debug 'Successfully exported metrics'
-            end
+            OpenTelemetry.logger.debug 'Successfully exported metrics' if result_code == Export::SUCCESS
           end
 
           def lock(&block)


metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
```diff
@@ -132,6 +132,8 @@ def export(timeout: nil)
           def report_result(result_code)
             if result_code == Export::SUCCESS
               OpenTelemetry.logger.debug 'Successfully exported metrics'
+            elsif result_code.nil?
+              OpenTelemetry.logger.debug 'No metrics to export'
             else
               OpenTelemetry.handle_error(exception: ExportError.new('Unable to export metrics'))
```
arielvalentin: Would a case statement here be better and add a debug message for the `default`?

wsmoak: I was going for the simplest fix that could possibly work.  To cover all the cases, I think it would be...
```
          def report_result(result_code)
            case result_code
            when Export::SUCCESS
              OpenTelemetry.logger.debug 'Successfully exported metrics'
            when Export::FAILURE
              OpenTelemetry.handle_error(exception: ExportError.new('Unable to export metrics'))
              OpenTelemetry.logger.error("Result code: #{result_code}")
            when Export::TIMEOUT
              OpenTelemetry.handle_error(exception: ExportError.new('Export operation timed out'))
              OpenTelemetry.logger.error("Result code: #{result_code}")
            when nil
              OpenTelemetry.logger.debug 'No metrics to export'
            else
              OpenTelemetry.logger.warn "Unknown result code when exporting metrics: #{result_code}"
            end
          end
```
Would you prefer that instead?

arielvalentin: I think so yes. 

Although `handle_error` would be enough for the error cases since it logs by default. 

So I would suggest either choosing to use `handle_error` or `logger` in those cases but not both.

wsmoak: I agree that logging the result code doesn't add much value, I removed it.

wsmoak:
Tested locally.  Started the app, waited a bit and recorded a data point, then turned off the collector.  (I am not sure how to make a timeout happen.)

```
D, [2025-11-03T19:22:15.205141 #59504] DEBUG -- : No metrics to export
D, [2025-11-03T19:23:15.223879 #59504] DEBUG -- : Successfully exported metrics
W, [2025-11-03T19:24:39.281246 #59504]  WARN -- : SystemCallError in MetricsExporter#send_bytes
E, [2025-11-03T19:24:39.284574 #59504] ERROR -- : OpenTelemetry error: Unable to export metrics
```

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
```diff
@@ -130,13 +130,17 @@ def export(timeout: nil)
           end
 
           def report_result(result_code)
-            if result_code == Export::SUCCESS
+            case result_code
+            when Export::SUCCESS
               OpenTelemetry.logger.debug 'Successfully exported metrics'
-            elsif result_code.nil?
+            when Export::FAILURE
+              OpenTelemetry.handle_error(exception: ExportError.new('Unable to export metrics'))
```
arielvalentin: Something to note here is that since this was not an _exception_ then it will be missing a backtrace and that _may_ confuse someone who is trying to handle the error:

```
irb(main):001> e = StandardError.new("bad")
=> #<StandardError: bad>
irb(main):002> e.backtrace
=> nil
```

The `handle_error` message does not need to have an `exception:` parameter and in this case may only need a `message:`.

Then again when I look at the exporter there is already a set of diagnositc errors there: 

https://github.com/wsmoak/opentelemetry-ruby/blob/7419ef5888b488dcf2fbee769ae662eb7562c70b/exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb#L136

So I guess my follow up question is... why does this _also_ need to report errors in addition to the exporters. Is there some nuance I am missing?

wsmoak: All _I_ really wanted was the success message. :) 

I was following the pattern that Kayla set in the BLRP: 
https://github.com/open-telemetry/opentelemetry-ruby/blob/6f7082555ea202d06645f1eb66bafdc3ce828bf5/logs_sdk/lib/opentelemetry/sdk/logs/export/batch_log_record_processor.rb#L194-L201

I can remove the logging for errors, and assume the thing that returned the error has already logged it.

wsmoak: ... and it collapses down to only logging the success message. Since there is no error logging, the nil value doesn't accidentally fall into it. Problem solved!

kaylareopelle (approved):

PR #1956, opened 2025-11-08
fix: Move the logs-sdk and metrics-sdk requires

### LINKED ISSUE #1955: Requires in the wrong place in opentelemetry-logs-sdk and opentelemetry-metrics-sdk
**Description of the bug**

If you add the opentelemetry-exporter-otlp-metrics gem to your Gemfile, but don't reference the opentelemetry-metrics-api gem directly, then when you configure OpenTelemetry then you will get a NoMethodError when `configure` is trying to call `meter_provider=`. This is because `meter_provider=` is defined in metrics_api/lib/opentelemetry-metrics-api.rb, but that file is not required, because the require statement for opentelemetry-metrics-api is in metrics_sdk/lib/opentelemetry-metrics-sdk.rb, but the opentelemetry-exporter-otlp-metrics gem has `require 'opentelemetry/sdk/metrics'`, as seems to be the convention in the rest of the code. The require statements in metrics_sdk/lib/opentelemetry-metrics-sdk.rb should be moved to metrics_sdk/lib/opentelemetry/sdk/metrics.rb to match similar code elsewhere in the codebase.

The opentelemetry-logs-sdk and opentelemetry-logs-api gems have the same problem with the `logger_provider=` method, but the opentelemetry-exporter-otlp-logs gem has this line of code to work around it:
```ruby
require 'opentelemetry-logs-api' # the sdk isn't loading the api, but not sure why
```
The require statements being in the wrong file is why.

Operating system details: Linux, Ubuntu 22.04 LTS, also macOS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.4.7"

```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-exporter-otlp-metrics'
end

  c.use_all
end
```

BrianHawley:
This makes it so the logs and metrics will work when you require like
```ruby
require 'opentelemetry/sdk/metrics'
```
Without requiring 'opentelemetry-logs-api' or 'opentelemetry-logs-metrics' earlier.

The opentelemetry-exporter-otlp-logs gem works around this, but I'm leaving that workaround there for now, so it will work with old versions of the opentelemetry-logs-sdk gem that are still locked.

Also added a missing `require 'opentelemetry/sdk/metrics/version'`.

To verify the fixes, I added `require: false` to the Gemfile files for the internal gem overrides. If you don't include `require: false` in the gem override declarations in the Gemfile, it will require those dependency files, which will hide bad requires in the code itself. If you specify `require: false`, the actual gem code requires will apply.

[Fixes #1955]

### DIFF
--- a/api/Gemfile
+++ b/api/Gemfile
@@ -11,5 +11,5 @@ gemspec
 eval_gemfile '../contrib/Gemfile.shared'
 
 group :test, :development do
-  gem 'opentelemetry-test-helpers', path: '../test_helpers'
+  gem 'opentelemetry-test-helpers', path: '../test_helpers', require: false
 end
--- a/common/Gemfile
+++ b/common/Gemfile
@@ -12,6 +12,6 @@ eval_gemfile '../contrib/Gemfile.shared'
 
 group :development, :test do
   # Use the opentelemetry-api gem from source
-  gem 'opentelemetry-api', path: '../api'
+  gem 'opentelemetry-api', path: '../api', require: false
   gem 'pry'
 end
--- a/exporter/jaeger/Gemfile
+++ b/exporter/jaeger/Gemfile
@@ -11,10 +11,10 @@ gemspec
 eval_gemfile '../../contrib/Gemfile.shared'
 
 group :test, :development do
-  gem 'opentelemetry-api', path: '../../api'
-  gem 'opentelemetry-common', path: '../../common'
-  gem 'opentelemetry-registry', path: '../../registry'
-  gem 'opentelemetry-sdk', path: '../../sdk'
-  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions'
-  gem 'opentelemetry-test-helpers', path: '../../test_helpers'
+  gem 'opentelemetry-api', path: '../../api', require: false
+  gem 'opentelemetry-common', path: '../../common', require: false
+  gem 'opentelemetry-registry', path: '../../registry', require: false
+  gem 'opentelemetry-sdk', path: '../../sdk', require: false
+  gem 'opentelemetry-semantic_conventions', path: '../../semantic_conventions', require: false
+  gem 'opentelemetry-test-helpers', path: '../../test_helpers', require: false
 end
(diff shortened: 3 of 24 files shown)


BrianHawley:
The test failures are unrelated, an incompatibility between Truffleruby and the latest json gem. So, until that gets fixed, the test suite isn't going to pass. What do we do next?

kaylareopelle:
Hi @BrianHawley, thanks for your patience on this PR. I'm able to reproduce the bug you describe in #1955, but when I run the reproduction you have with this branch, the error still appears.

These are the contents of the Ruby file I'm using:
```ruby
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-exporter-otlp-metrics', github: 'BrianHawley/opentelemetry-ruby', branch: 'fixes_1955', glob: 'exporter/otlp-metrics/*.gemspec'
end

  c.use_all
end
```

This is the output:
```sh
Fetching https://github.com/BrianHawley/opentelemetry-ruby.git
Resolving dependencies...
Fetching gem metadata from https://rubygems.org/.../opentelemetry-api-1.7.0/lib/opentelemetry.rb:7: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add logger to your Gemfile or gemspec to silence this warning.
W, [2025-12-11T16:05:24.499438 #52700]  WARN -- : The otlp exporter cannot be configured - please add opentelemetry-exporter-otlp to your Gemfile, spans will not be exported
E, [2025-12-11T16:05:24.499588 #52700] ERROR -- : OpenTelemetry error: unexpected configuration error due to undefined method 'meter_provider=' for module OpenTelemetry - OpenTelemetry::SDK::ConfigurationError - .../opentelemetry-sdk-1.10.0/lib/opentelemetry/sdk.rb:74:in 'OpenTelemetry::SDK#configure'

```

Can you take a look? Am I misunderstanding what this PR addresses?

BrianHawley:
@kaylareopelle can you take another look, and at my above comments answering your questions?

kaylareopelle:
Hi @BrianHawley, we talked about this at the SIG and a few other folks want to review this PR this week. I'm going to hold off the merge until they get a chance to look this over.

PR #1959, opened 2025-11-10
fix: update setup-ruby action

xuan-cao-swi:
## Description

Update to the latest `setup-ruby@v1.267.0` to allow the installation of the latest TruffleRuby 25.0.0. This update resolves the issue with the failing action caused by TruffleRuby 24.1.2.

Several other actions (e.g., checkout) remain on older versions, such as v2, while the latest is v5. Additionally, I noticed that some action versions are specified using commit hashes. It's unclear whether this is intentional or a temporary setup that needs to be addressed later.

The dependabot responsible for automatically creating PRs to bump action versions is not functioning (i.e., not running).

### DIFF
--- a/exporter/jaeger/opentelemetry-exporter-jaeger.gemspec
+++ b/exporter/jaeger/opentelemetry-exporter-jaeger.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |spec|
                ::Dir.glob('*.md') +
                ['LICENSE', '.yardopts']
   spec.require_paths = ['lib']
-  spec.required_ruby_version = '>= 3.1'
+  spec.required_ruby_version = '< 3.2'
 
   spec.add_dependency 'base64', '>= 0.2.0'
   spec.add_dependency 'opentelemetry-api', '~> 1.1'


xuan-cao-swi:
Jaeger is having issue about using the `Fixnum` although `Fixnum` is not used in ruby 3+ (tstill in investigation)

exporter/jaeger/test/test_helper.rb
```diff
@@ -15,6 +15,8 @@
 require 'webmock/minitest'
 require 'rspec/mocks/minitest_integration'
 
+Fixnum = Integer if RUBY_PLATFORM == 'java'
```
arielvalentin: I don't think this is the best thing to do. It makes it look like our gem is compatible with the latest version of JRuby. I advise that we report this error and explicitly cap support at JRuby 9

xuan-cao-swi: It seems like only opentelemetry-exporter-zipkin and opentelemetry-exporter-jaeger will go for jruby test. Other exporters will be skipped by `JRuby Filter` in workflow.

Putting hard cap with `jruby-9.4` will also apply to zipkin (even though zipkin has no issue with the jruby-10).

What about remove the JRuby Filter and having following steps?
```
- name: "Test Zipkin with JRuby"
    if: "${{ matrix.os == 'ubuntu-latest' && matrix.gem == 'opentelemetry-exporter-zipkin' }}"
    uses: ./.github/actions/test_gem
    with:
       gem: "${{ matrix.gem }}"
       ruby: "jruby"
- name: "Test Jaeger with JRuby"
    if: "${{ matrix.os == 'ubuntu-latest' && matrix.gem == 'opentelemetry-exporter-jaeger' }}"
    uses: ./.github/actions/test_gem
    with:
       gem: "${{ matrix.gem }}"
       ruby: "jruby-9.4"
```

arielvalentin: JRuby 9 goes as high as Ruby 3.1 but I would like to end support for that version and require a minimum of Ruby 3.2

I would say depending on the _specification_ we may find ourselves at odds here. 

I am also very confused why JRuby's behavior would differ from C Ruby 3.2 where Fixnum seems to be supported. Shouldn't CRuby fail in the same was as JRuby?

xuan-cao-swi: > I am also very confused why JRuby's behavior would differ from C Ruby 3.2 where Fixnum seems to be supported. Shouldn't CRuby fail in the same was as JRuby?

I think Fixnum is not valid in both JRuby and CRuby. I think the issue comes down to how JRuby and CRuby behave differently within the specific test case.

The problematic test case is `exports a span from a tracer` from agent_collector_test.rb (try following test with jruby 10.0.2.0 will reproduce the issue)

```
bundle exec ruby -I test test/opentelemetry/exporter/jaeger/agent_exporter_test.rb -n /tracer/
```

I set breakpoint in [write_type](https://github.com/apache/thrift/blob/v0.22.0/lib/rb/lib/thrift/protocol/base_protocol.rb#L253) from thrift base_protocol.rb, only JRuby 10.0.2.0 will invoke this function from test case; and since there is the `Fixnum`, then it cause the issue.

This is the backtrace from write_field (with JRuby, write_field calls write_type)
```
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/protocol/base_protocol.rb:242:in 'write_field'
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/struct.rb:108:in 'block in write'
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/struct_union.rb:47:in 'block in each_field'
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/struct_union.rb:45:in 'each'
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/struct_union.rb:45:in 'each_field'
/root/.rbenv/versions/jruby-10.0.2.0/lib/ruby/gems/shared/gems/thrift-0.22.0/lib/thrift/struct.rb:98:in 'write'
/app/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb:131:in 'encoded_span_size'
```

I believe somehow this [encoded_span.write(@sizing_protocol)](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb#L130C11-L130C47) directly invoke the [write](https://github.com/apache/thrift/blob/v0.22.0/lib/rb/lib/thrift/struct.rb#L95) from JRuby; while CRuby calls Transport.[write](https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/jaeger/lib/opentelemetry/exporter/jaeger/transport.rb#L18) instead.

arielvalentin: 🤔 How could that be? They are the same implementation. 

The encoded_span is using the SizingTransport to track how large the buffer is but it has to convert the in memory span to a thrift binary: https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb#L130

The Transport.write is not going to be used _until_ the `emitBatch` method is invoked https://github.com/open-telemetry/opentelemetry-ruby/blob/main/exporter/jaeger/lib/opentelemetry/exporter/jaeger/agent_exporter.rb#L45

Then at that point it starts the flow into the transport https://github.com/apache/thrift/blob/e720e6fecb75f1242154d1847b3b2954ae9d0965/lib/rb/lib/thrift/client.rb#L33

The backtrace also seems suspicious to me since it does not exactly match a location I would have expected https://github.com/apache/thrift/blob/v0.22.0/lib/rb/lib/thrift/protocol/base_protocol.rb#L242

I do see a Patch here introducing `Fixnum` for versions of Ruby prior to 1.8.7; when the class was first introduced.  https://github.com/apache/thrift/blob/v0.22.0/lib/rb/lib/thrift/core_ext/fixnum.rb

But of course now that it was deprecated _and removed_ from Ruby Lang... this patch is incorrect as it is assuming all _future_ versions of Ruby will have `Fixnum`; so my assertion is that it fails here because the const does not exist: https://github.com/apache/thrift/blob/v0.22.0/lib/rb/lib/thrift/protocol/base_protocol.rb#L256

That still leaves me utterly confused as to why it is a non-issue for CRuby.

xuan-cao-swi: > The backtrace also seems suspicious to me since it does not exactly match a location I would have expected

I agree, although I haven't figure out why it jumped from `encoded_span_size` to thrift/struct.rb `write`.

> my assertion is that it fails here because the const does not exist

Yes, JRuby > 10 directly call thrift/struct.rb `write` make it reach to `write_type` function.

I tried `jruby-9.4.14.0` (last jruby 9 version, before 10), it behaved the same as any CRuby (reached to the `Jaeger::Transport.write` function). I also tried `jruby-10.0.0.0`, and it caused the issue same as 10.0.2.0.

I can open an issue in thrift to ask them to resolve the case that `Fixnum` is not longer used so maybe remove it, but still JRuby 10+ make the call sequence different is bigger puzzle and problem.

arielvalentin:
Yes, I'm aware. I think that's a trade off we have to make. 

My assertion is that it shouldn't work with CRuby either. 

If people want to take the risk they should pin to an older version of the gem.

Issue #1667, opened 2024-07-31
Create separate gems for http, grpc export and a common gem they can rely on


kaylareopelle:
The `exporter/otlp-http`, `exporter/otlp-grpc` and `exporter/otlp-common` directories in this repository reference tests to break out the existing `opentelemetry-exporter-otlp` gem into smaller chunks. Ruby doesn't currently have support for a gRPC exporter, and the behavior that would be shared between an HTTP and gRPC exporter would live in a common gem they both rely on.

Initial PRs related to this work: 
* https://github.com/open-telemetry/opentelemetry-ruby/pull/1179 
* https://github.com/open-telemetry/opentelemetry-ruby/pull/1180 

These exporters are not production ready as they currently exist. (See: https://github.com/open-telemetry/opentelemetry-ruby/issues/1337) 

Contributions welcome to add test coverage, error handling, and other work to make this production-ready.

PR #1960, opened 2025-11-13
Add retry and status code handling tests to trace exporter

zFlux:
Adds unit tests verifying that the OTLP HTTP trace exporter retries on 5xx and 429 responses, and does not retry on client errors (4xx). Confirms exponential backoff and export result statuses are mapped correctly according to the spec.

Adding a few tests at a time to ultimately fully harden otlp-http, otlp-grpc, and otlp-common

Refs: open-telemetry/opentelemetry-ruby#1667


zFlux:
I'm going to close this and tackle something flagged with "good first issue".

Issue #2000, opened 2025-12-15
Exponential Histogram scale for data points can become erroneously low when using multiple sets of attributes


hibachrach:
When recording values for a histogram to which an `ExponentialBucketAggregation` has been added as a aggregation, a normal sequence of measurements can lead to the histogram scale getting very low, often going below the minimum of -10. This seems to mostly occur when recording for more than one set of attributes.

**Share details about your runtime**

Operating system details: macOS Sequoia 15.7.2
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.2.7"
RUBY_DESCRIPTION: "ruby 3.2.7 (2025-02-04 revision 02ec315244) [arm64-darwin24]"

**Share a simplified reproduction if possible**

```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-metrics-sdk'
end

require 'opentelemetry-metrics-sdk'

OpenTelemetry::SDK.configure do |c|
  c.add_metric_reader(
    OpenTelemetry::SDK::Metrics::Export::PeriodicMetricReader.new(
      exporter: OpenTelemetry::SDK::Metrics::Export::ConsoleMetricPullExporter.new
    )
  )
end

OpenTelemetry.meter_provider.add_view(
  'example_hist',
  aggregation: OpenTelemetry::SDK::Metrics::Aggregation::ExponentialBucketHistogram.new
)
meter = OpenTelemetry.meter_provider.meter('exp-hist-bug')

hist = meter.create_histogram('example_hist')
attr1 = {foo: 1}
attr2 = {foo: 2}

hist.record(0.010359524076171637, attributes: attr1)
hist.record(0.006354922335082751, attributes: attr2)
hist.record(0.009615690848606532, attributes: attr1)
hist.record(0.00007659464778730196, attributes: attr1)
hist.record(0.0101281877195143937, attributes: attr2)

OpenTelemetry.tracer_provider.shutdown
```

This generated the log line
```
[...]

PR #2005, opened 2025-12-24
fix!: fix the issue of mixed scale with multiple attributes

xuan-cao-swi:
## Description

Fix the issue that multiple attributes mix up the scale.
The fix is aiming to store the mapping for each data point with unique attributes instead of share the same `@mapping` for all data point.

Resolve https://github.com/open-telemetry/opentelemetry-ruby/issues/2000

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
@@ -44,8 +44,6 @@ module OpenTelemetry
             @size           = validate_size(max_size)
             @scale          = validate_scale(max_scale)
 
-            @mapping = new_mapping(@scale)
-
             # Previous state for cumulative aggregation
             @previous_positive = {} # nil
             @previous_negative = {} # nil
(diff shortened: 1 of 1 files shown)


metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb
kaylareopelle, line 1: When I ran this branch with the example provided in #2000, the error message about scale was still reported: 

```
E, [2026-01-06T14:01:16.819311 #43327] ERROR -- : OpenTelemetry error: Scale -12 is smaller than minimum scale -10 - /opentelemetry-metrics-sdk-0.11.2/lib/opentelemetry/sdk/metrics/aggregation/exponential_bucket_histogram.rb:344:in 'OpenTelemetry::SDK::Metrics::Aggregation::ExponentialBucketHistogram#validate_scale'
```

I see that the PR description just says this "relates" to #2000, rather than resolves it. What other work do you envision to resolve the bug?

xuan-cao-swi: It should be `resolve`; changed to `resolve`.

kaylareopelle (approved):
@xuan-cao-swi - Thank you for fixing this and for your patience with my review. 

To fix things for our users, I'm comfortable getting this out now, but I think we should make a few changes very soon:
1. Add a regression test to prevent this fix from being refactored out.
2. For maintainability, refactor the `update` method into smaller chunks. It's getting pretty complex and difficult to track.

hibachrach:
thank you!!

hibachrach:
Happy to report that this change did indeed resolve the problem for us. Thanks again for all your work!

PR #2006, opened 2026-01-07
chore: Upgrade release system to the toys-release gem

dazuma:
This is a long-awaited upgrade to the release system. It was rewritten pretty much from the ground up to be more testable and maintainable. That said, the OTel-Ruby repos exercise the system more than any other customer, so I'll be around to monitor the next releases closely to ensure things go smoothly.

Some key changes:

* The release system is now distributed as a gem, [toys-release](https://rubygems.org/gems/toys-release). So loading it is less sketchy than depending on a tag in another repository as we've been doing before. This change currently pins the toys-release gem version at 0.4.0 (see the file `.toys/release.rb`).
* Most of the interfaces are mostly the same, but backward compatibility was not a goal here. I updated the release configs (`.toys/.data/releases.yml`) and the tool invocations in the GitHub Actions workflows accordingly.
* I set the GitHub Actions workflows to run on Ruby 4.0 and Toys 0.19.1. In particular, toys-release requires at least Toys 0.19.0. Ruby 4.0 isn't really necessary, but toys-release uses it by default. If people are more comfortable using 3.4, I can change it to that version.

Some of the config has also been updated to work with the toys-release gem. Specifically:

* The `main_branch` key defaults to `main`, so `main_branch: main` is redundant and was removed.
* The `docs_builder_tool` key was dropped. Toys-release now just uses `yard doc` directly to build docs (and indeed, we don't use any of it here but the build pipeline is now a lot more modular and customizable). That said, this repo is currently not configured to build and publish documentation during releases, so that key wasn't really used anyway. I think we disabled it because we had trouble with it a few years ago. I can try to set it up again if we want, as a second step after doing this initial migration.
* The `commit_lint` key was dropped. The commit linter in general was removed from toys-release, as it was not really maintained and no one was using it. And it has been disabled in our repos anyway. There are better dedicated conventional commit linter tools available if we want to use one.
* I removed the trailing slash hack for the `opentelemetry-exporter-otlp` gem configuration. That bug has been fixed, and the hack is no longer necessary.
* I removed `version_rb_path` settings for gems whose configurations did not need it.

There is quite a bit of new functionality and fixed issues in toys-release, that we might choose to take advantage of (again, as a second step after this initial migration). For example:

* There is now a more flexible way to customize which paths signal that a gem has changed, supporting shared files that are used by multiple gems. This should make hacks like https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1765 unnecessary.
* There is now a way to customize how conventional commit tags (including scopes) map to releases and semver, which Ariel requested in https://github.com/dazuma/toys/issues/215.
* Toys-release is no longer limited to releasing from the main branch. You should be able to release from any branch, by choosing it when triggering the "open release request" workflow.

Finally, I (@dazuma) expect to be able to engage a lot more now on OTel-Ruby. I quit my job at Google a few months ago, in part because it was stifling me and I was not able to make space for open source and community work. Now that I'm out, I've had a chance to, among other things, finish this long-overdue update to the release system. I also should be able to attend the weekly SIGs more often now. So anyone working on this repo, if you have questions, issues, or feature requests regarding the release system and process, please let me know.

I'm keeping this PR in draft until I've had the chance to discuss this update with the other maintainers, and to do some more testing.

Note there is a corresponding pull request in the opentelemetry-ruby-contrib repository: https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1923.

### DIFF
--- a/.toys/.data/releases.yml
+++ b/.toys/.data/releases.yml
@@ -24,11 +24,6 @@ git_user_email: 197425009+otelbot@users.noreply.github.com
 #     (Required only if it does not match the gem name. e.g. The gem
 #     "opentelemetry-sdk" would default to "lib/opentelemetry/sdk/version.rb",
 #     so this field is required only if the actual path is different.)
-#  *  version_constant: The fully-qualified version constant as an array.
-#     (Required if the actual constant name, including its capitalization,
-#     differs from what would be inferred from the gem name. This means it's
-#     always required in this repo, because the capitalization of the
-#     "OpenTelemetry" namespace does not match the gem name "opentelemetry".)
 #  *  changelog_path: Path to CHANGELOG.md relative to the gem directory.
 #     (Required only if it is not in the expected location, which is the file
 #     "CHANGELOG.md" at the root of the gem directory.)


.github/workflows/release-hook-on-closed.yml
```diff
@@ -19,17 +19,16 @@ jobs:
       - name: Install Ruby
         uses: ruby/setup-ruby@675dd7ba1b06c8786a1480d89c384f5620a42647 # v1.281.0
         with:
-          ruby-version: 3.1
+          ruby-version: "4.0"
       - name: Checkout repo
         uses: actions/checkout@v6
       - name: Install Toys
-        run: "gem install --no-document toys -v 0.15.5"
+        run: "gem install --no-document toys -v 0.19.1"
```
thompson-tomo: @dazuma so that we can have renovate manage this dependency could we do something like:

```suggestion
        run: "gem install --no-document --file .toys/GemFile"
```

And then we create the corresponding gemfile. This way we can ensure a consistent version of toys is being used etc.

dazuma: @thompson-tomo Yeah, @arielvalentin was asking for something similar in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/1923, but for different reasons: to enable setup-ruby to install (and cache) that bundle. I think your renovate argument is an even better argument. Lemme look into making this change. (I'll put the Gemfile in a slightly different location because `.toys/Gemfile` is an override location that could interfere with other tools that might be added later.)

dazuma: Updated with this modification.

.toys/.data/releases.yml
```diff
@@ -1,114 +1,92 @@
...
+#     the word "opentelemetry" whereas the directory names do not.)
 #  *  version_rb_path: Path to version.rb relative to the gem directory.
 #     (Required only if it does not match the gem name. e.g. The gem
-#     "opentelemetry-sdk" would default to lib/opentelemetry/sdk/version.rb)
+#     "opentelemetry-sdk" would default to "lib/opentelemetry/sdk/version.rb",
+#     so this field is required only if the actual path is different.)
 #  *  version_constant: The fully-qualified version constant as an array.
-#     (Required because the OpenTelemetry namespace does not match the gem
-#     name "opentelemetry".)
+#     (Required if the actual constant name, including its capitalization,
```
kaylareopelle: This comment makes it sound like the version constant should always be here, but it looks like it was removed below. Did something change?

kaylareopelle (approved):
I think this looks good. I'm a little confused by some of the changes in releases.yml and left a comment on that.

.toys/.data/releases.yml
dazuma: Oh yeah, `version_constant` is now deprecated and unused, and I forgot to remove it from the comment here. Removed now.

PR #2016, opened 2026-01-12
test: remove ostruct dependency by updating rake to 13 for ruby 4

xuan-cao-swi:
Added cgi and ostruct gem since they are removed from ruby 4.
Also fixed some warning that is about constants reinitialization.

### DIFF
--- a/exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
+++ b/exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
@@ -273,25 +273,25 @@ module OpenTelemetry
             Opentelemetry::Proto::Collector::Logs::V1::ExportLogsServiceRequest.encode(
               Opentelemetry::Proto::Collector::Logs::V1::ExportLogsServiceRequest.new(
                 resource_logs: log_record_data
-                  .group_by(&:resource)
-                  .map do |resource, log_record_datas|
-                    Opentelemetry::Proto::Logs::V1::ResourceLogs.new(
-                      resource: Opentelemetry::Proto::Resource::V1::Resource.new(
-                        attributes: resource.attribute_enumerator.map { |key, value| as_otlp_key_value(key, value) }
-                      ),
-                      scope_logs: log_record_datas
-                        .group_by(&:instrumentation_scope)
-                        .map do |il, lrd|
-                          Opentelemetry::Proto::Logs::V1::ScopeLogs.new(
-                            scope: Opentelemetry::Proto::Common::V1::InstrumentationScope.new(
-                              name: il.name,
-                              version: il.version
-                            ),
-                            log_records: lrd.map { |lr| as_otlp_log_record(lr) }
-                          )
-                        end
-                    )
-                  end
+                               .group_by(&:resource)
+                               .map do |resource, log_record_datas|
+                                 Opentelemetry::Proto::Logs::V1::ResourceLogs.new(
+                                   resource: Opentelemetry::Proto::Resource::V1::Resource.new(
+                                     attributes: resource.attribute_enumerator.map { |key, value| as_otlp_key_value(key, value) }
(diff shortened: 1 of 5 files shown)


arielvalentin:
ostruct on the other hand is a test only dependency: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/b8f887a3318fb7168c5a811462303d1762e12a93/metrics_sdk/test/opentelemetry/sdk/metrics/fork_hooks_test.rb#L79

If we can find an alternative for OpenStruct I think that we should go for it.

xuan-cao-swi:
For replacement of cgi escape and unescape maybe using `URI.decode_www_form_component/encode_www_form_component`?

I just did test and found out if rake is 13.3, then no OpenStruct is required. I think old rake (13.1.0 and before) use OpenStruct. Currently most gemspec using rake ~> 12.0 (except log_sdk), maybe they can change to newer rake.

xuan-cao-swi:
It seems `URI encode/decode` is for newer URL encoding rules (RFC 3986), and CGI is for legacy rules.

Decode seems aligned for asterisk.
```
irb(main):011> URI.decode_www_form_component("%2A")
=> "*"
irb(main):014> CGI.unescape("%2A")
=> "*"
```

If we are ok with adding `cgi` in dependency then I can revert back the changes.
I don't see any other stdlib have uri encode and decode support.
We can also copy over the function from cgi library and integrate them in sdk/exporter codebase.

xuan-cao-swi:
Based on the spec, baggage propagator and exporter protocol both use w3s baggage specification. Jaeger propagator will be deprecated. RFC 3986 is compatible with W3C Baggage specification. I think the most important consideration is to following [Character Encoding](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/entities/entity-propagation.md#character-encoding) guide

Difference about space
```
# Ruby encode_uri_component/encode_www_form_component
URI.encode_uri_component(" ")       # Space Encoding %20
URI.encode_www_form_component(" ")  # Space Encoding +
CGI.escape(" ")                     # Space Encoding +

# Python is using unquote/quote and unquote_plus/quote_plus (RFC 3986)
urllib.parse.quote(" ")        # Space Encoding	%20
urllib.parse.quote_plus(" ")   # Space Encoding	+

# JS uses default encodeURIComponent
encodeURIComponent(" ")        # Space Encoding	%20
```

Difference about * (W3C Baggage allows both * and %2A for encoding *)
```
python quote/quote_plus encode to %2A
js encodeURIComponent doesn't encode to %2A
ruby encode_uri_component/encode_www_form_component doesn't encode %2A
```

exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
```diff
@@ -10,6 +10,7 @@
 require 'opentelemetry/sdk/logs'
 require 'net/http'
 require 'zlib'
+require 'cgi/escape'
```
xuan-cao-swi: exporter used to rely on otel api to load `cgi/escape`, which is not ideal. Explicitly add the dependency here for clarity.

xuan-cao-swi: leave this for next update to avoid create new exporter release

logs_api/opentelemetry-logs-api.gemspec
```diff
@@ -26,6 +26,8 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
 
+  spec.add_development_dependency 'bundler', '>= 1.17'
```
kaylareopelle: Do we need to have bundler as a dependency?

xuan-cao-swi: I guess it's left there by merge conflict. Removed 👍

PR #2022, opened 2026-01-13
chore: Remove Explicit Bundler Dependency

arielvalentin:
Declaring bundler as a dependency is therefore redundant and our pessemistic versioning constraints blocks Ruby 4 upgrades. Bundler is included with RubyGems by default so declaring it separately seems unecessary.

### DIFF
--- a/logs_api/opentelemetry-logs-api.gemspec
+++ b/logs_api/opentelemetry-logs-api.gemspec
@@ -26,7 +26,6 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
 
-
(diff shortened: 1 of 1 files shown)


logs_api/opentelemetry-logs-api.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'opentelemetry-api', '~> 1.0'
 
-  spec.add_development_dependency 'bundler', '>= 1.17'
+
```
kaylareopelle: I believe the logs API suite is failing because of the extra line here.
```suggestion
```

kaylareopelle (approved):
Thank you!

PR #2028, opened 2026-01-15
feat: replace cgi with uri for encode and decode

xuan-cao-swi:
This PR tries to use `URI.encode/decode_uri_component` to replace `CGI.escape/unescape`.

Although `cgi/escape` is present in [Ruby 4.0.0](https://github.com/ruby/ruby/blob/v4.0.0/lib/cgi/escape.rb), it is not available in Ruby versions less than 4.0.0. We can use cgi/escape once all Ruby 3 versions are deprecated, if required.

The difference between these two implementation is the treatment for asterisk (*), which CGI encode to `%2A`, but URI preserve * (see more in https://github.com/open-telemetry/opentelemetry-ruby/pull/2016).

CGI will silence the invalid char encode/decode, while URI will raise the error.

### DIFF
--- a/propagator/jaeger/test/text_map_propagator_test.rb
+++ b/propagator/jaeger/test/text_map_propagator_test.rb
@@ -259,7 +259,7 @@ describe OpenTelemetry::Propagator::Jaeger::TextMapPropagator do
       end
       carrier = {}
       propagator.inject(carrier, context: context)
-      _(carrier['uberctx-key1']).must_equal('value+1+%2F+blah')
+      _(carrier['uberctx-key1']).must_equal('value%201%20%2F%20blah')
     end
 
     it 'injects to rack keys' do


propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_propagator.rb
```diff
@@ -110,7 +110,7 @@ def context_with_extracted_baggage(carrier, context, getter)
               next unless baggage_key
 
               raw_value = getter.get(carrier, carrier_key)
-              value = CGI.unescape(raw_value)
+              value = URI.decode_uri_component(raw_value)
```
arielvalentin, line 113: Do the same changes need to happen for B3 headers?

xuan-cao-swi: It seems B3 propagator headers should only contain trace ID, span ID, and sampling decision, which have well-defined headers that don't require decoding.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
```diff
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'cgi'
```
kaylareopelle, line 7: Are the CGI methods we're calling in the propagators the same in 3.x and 4.x, but the only difference is what needs to be required? 

To help me better understand the issue, would something like this also solve the problem? 
```ruby
if RUBY_VERSION >= '4.0'
  require 'cgi/escape'
else
  require 'cgi'
end
```

kaylareopelle: ah, I just saw the conversation on https://github.com/open-telemetry/opentelemetry-ruby/pull/2016. It sounds like URI is compatible with the spec.

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_propagator.rb
arielvalentin: Am I missing something? Then where these baggage headers encoded incorrectly when using the b3 propagators?

xuan-cao-swi: It's not decode before, also I don't see any other language decode these key and value
I can create an issue to discuss about decode b3 baggage headers

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
xuan-cao-swi: Yes, other language use their stdlib to decode

PR #2034, opened 2026-02-03
resolve Ruby 4 warnings, closes #2008

### LINKED ISSUE #2008: Ruby 4 compatability
🎉 https://www.ruby-lang.org/en/news/2025/12/25/ruby-4-0-0-released/ 🎉 

There are several warnings that appear when running specs warning us about upcoming changes that we will need to address when we upgrade to Ruby 4.  

E.g. 

```console
opentelemetry-api-1.7.0/lib/opentelemetry/baggage/propagation/text_map_propagator.rb:7: warning: CGI library is removed from Ruby 4.0. Please use cgi/escape instead for CGI.escape and CGI.unescape features.
```

ajwann:
Hey @arielvalentin! I saw #2008 and thought it was a good opportunity for a first time contributor. 

Here's a summary of changes:

- Fixed Ruby-4 related warnings related to the following gems:
  - Ostruct: 
    - Warning: `warning: ostruct used to be loaded from the standard library, but is not part of the default gems since Ruby 4.0.0.`.
    - Resolution: added to the gem spec of any gem which relied on `ostruct`.
  - CGI: 
    - Warning: `warning: CGI library is removed from Ruby 4.0. Please use cgi/escape instead for CGI.escape and CGI.unescape features.`
    - Resolution: confirmed we’re only using escape and unescape, changed require statement in several gems `text_map_propagator.rb` file from `cgi` to `cgi/escape`.
  - Logger:
    - Warning: warning: logger used to be loaded from the standard library, but is not part of the default gems since Ruby 4.0.0
    - Resolution: add `logger` to gemspec for several gems.
- Tested with Ruby 4.0.1 and also with 3.3.5 for backwards compatibility

### DIFF
--- a/common/opentelemetry-common.gemspec
+++ b/common/opentelemetry-common.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-test-helpers'
+  spec.add_development_dependency 'ostruct'
   spec.add_development_dependency 'rake', '~> 12.0'
   spec.add_development_dependency 'rubocop', '~> 1.65'
   spec.add_development_dependency 'simplecov', '~> 0.17'
--- a/api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
+++ b/api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'cgi'
+require 'cgi/escape'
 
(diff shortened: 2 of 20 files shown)


common/opentelemetry-common.gemspec
```diff
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency 'minitest', '~> 5.0'
   spec.add_development_dependency 'opentelemetry-test-helpers'
+  spec.add_development_dependency 'ostruct'
```
ajwann, line 32: Not every gemspec needed the `ostruct` dependency added, but for the ones that did, I placed it in alphabetical order to keep things tidy.

PR #2037, opened 2026-02-04
feat: instrumentation scope stable feature parity

robertlaurin:


sdk/lib/opentelemetry/sdk/trace/tracer.rb
fbogsany, line 12: We have this. Does this give us confidence that we're not going to break anyone by inserting two new params into the param list (before the end)?

fbogsany, line 17: should document the new params here.

sdk/lib/opentelemetry/sdk/trace/tracer.rb
```diff
@@ -18,7 +18,7 @@ class Tracer < OpenTelemetry::Trace::Tracer
         # @param [TracerProvider] tracer_provider TracerProvider that initialized the tracer
         #
         # @return [Tracer]
-        def initialize(name, version, tracer_provider)
+        def initialize(name, version, schema_url, attributes, tracer_provider)
```
fbogsany, line 21: we're not yet storing the new params anywhere?

sdk/lib/opentelemetry/sdk/trace/tracer.rb
robertlaurin: I'm not sure we can safely assume no one consuming this package is consuming this private api directly. However with other libraries that assumption is that if you choose to consume a private API you cannot assume it doesn't experience "breaking" changes.

PR #2052, opened 2026-03-13
chore: Remove faraday from dev dependencies

kaylareopelle:
Faraday isn't being used. 

PR #2045 sought to update the version, but when I looked into it, we aren't using it in our gems.


Agibbs1111 (changes requested):
Delete

PR #2056, opened 2026-03-16
ci: Use renovate to update semantic conventions gem

kaylareopelle:
This is a new action using Renovate to help us update the Semantic Conventions gem when new versions are released:
* We use renovate to check to see if a new version has been released
* Then, we use a GHA workflow to run the rake task to make the update
* The GH version of renovate doesn't have access to Docker, so we leverage GHA's access to run weaver
* Renovate's original version change PR gets updated with the rest of the changes

~~We should release versions 1.38 and 1.39 before we merge this PR. It will run on only the latest version (which at this time is 1.40)~~

This PR now includes code that should allow renovate to open up PRs for older versions (ex. 1.38).

PR #2058, opened 2026-03-16
feat: bump semantic conventions to 1.38.0

kaylareopelle:
After this, we also need to manually bump 1.39.0. Then, renovate can take over for future version bumps.


kaylareopelle:
Closing in favor of #2056 with the idea that the renovate manager can make a PR for this version too

PR #2059, opened 2026-03-17
fix(semconv): correct & simplify codegen templates

robbkidd:
* Update rendering of deprecation notes for deprecated attributes/metrics

The data under this key is now a dict. `.note` is informative when present, but isn't always, so fallback to `.reason`.

* Remove `@example` blocks 

Yard `@example` content needs to be indented within the code comment in a tricksy way. Combine that with the variety of shapes the data under `.examples` is provided and the template becomes complicated and brittle.

WWJD: What Would Java Do? - They don't include `.examples` in their generated semantic conventions code. So let's not also.

### DIFF
--- a/semantic_conventions/README.md
+++ b/semantic_conventions/README.md
@@ -87,7 +87,10 @@ upstream [OpenTelemetry Semantic Conventions version][semconv].
 
 ## How do I rebuild the conventions?
 
-Bump the version number in the Rakefile, and then run `rake generate`.
+To build the library against a new version of the semantic conventions, update SPEC_VERSION in the Rakefile, and then run `rake generate`.
+
+Do not update the library's VERSION in version.rb.
+That will be handled by the automation for releasing the gem.
 
 ## How can I get involved?
 
--- a/semantic_conventions/Rakefile
+++ b/semantic_conventions/Rakefile
@@ -126,7 +126,8 @@ end
 
 desc 'Bump the semantic_conventions gem version to match the spec'
 task :update_spec_version_constant do
-  puts "\n+++ Updating spec version constant to #{SPEC_VERSION}\n"
+  puts "\n+++ Updating library SPEC_VERSION constant to #{SPEC_VERSION}.\n"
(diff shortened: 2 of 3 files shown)


semantic_conventions/lib/opentelemetry/semconv/incubating/aws/attributes.rb
```diff
@@ -23,358 +23,228 @@ module SemConv
...
         AWS_LAMBDA_INVOKED_ARN = 'aws.lambda.invoked_arn'
-
-        # The UUID of the [AWS Lambda EvenSource Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html). An event source is mapped to a lambda function. It's contents are read by Lambda and used to trigger a function. This isn't available in the lambda execution context or the lambda runtime environment. This is going to be populated by the AWS SDK for each language when that UUID is present. Some of these operations are Create/Delete/Get/List/Update EventSourceMapping.
+    
+        # The UUID of the [AWS Lambda EvenSource Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html). An event source is mapped to a lambda function. It's contents are read by Lambda and used to trigger a function. This isn't available in the lambda execution context or the lambda runtime environtment. This is going to be populated by the AWS SDK for each language when that UUID is present. Some of these operations are Create/Delete/Get/List/Update EventSourceMapping.
```
robbkidd:
```suggestion
        # The UUID of the [AWS Lambda EvenSource Mapping](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-lambda-eventsourcemapping.html). An event source is mapped to a lambda function. It's contents are read by Lambda and used to trigger a function. This isn't available in the lambda execution context or the lambda runtime environment. This is going to be populated by the AWS SDK for each language when that UUID is present. Some of these operations are Create/Delete/Get/List/Update EventSourceMapping.
```

PR #2071, opened 2026-03-28
fix: tweak weaver generation to include new line

thompson-tomo:
This goes through and ensures that all weaver generated source files contain an empty new line at the end of the files like all other source files.

This change has been added to the jinja template and for confirming of the fix rubocop is now running on all os.

> [!NOTE]
> to review this PR I recommend hiding whitespace https://github.com/open-telemetry/opentelemetry-ruby/pull/2071/changes which significantly reduces the pr size

### DIFF
--- a/semantic_conventions/Rakefile
+++ b/semantic_conventions/Rakefile
@@ -48,7 +48,7 @@ default_tasks =
 task default: default_tasks
 
 desc 'update semantic conventions'
-task generate: %i[update_spec_version_constant generate_semconv generate_require_rollups rubocop_autocorrect]
+task generate: %i[update_spec_version_constant generate_require_rollups]
 
 SPEC_VERSION = '1.37.0'
 OTEL_WEAVER_VERSION = 'v0.22.1'


semantic_conventions/Rakefile
```diff
@@ -97,7 +97,7 @@ task generate_semconv: %i[check_out_semconv_version clean_generated_code] do
   DOCKER_COMMAND
 end
 
-task generate_require_rollups: %i[generate_semconv] do
+task generate_require_rollups: %i[generate_semconv rubocop_autocorrect] do
```
arielvalentin, line 100: ✨

kaylareopelle, line 100: Great call to make sure Rubocop autocorrect is running. 

We're also calling this task in `generate`: 

https://github.com/open-telemetry/opentelemetry-ruby/blob/8a4baa3497e26db028ceca16d9d04a848db0659d/semantic_conventions/Rakefile#L51

Does it do something different when called here?

thompson-tomo: In my case I didn't want to change the version and Ideally we would also add a ci check to ensure that the checked in files are up to date hence useful to be able to run it without changing the version file.

What we could do is simplify line 51 to just depend on `update_spec_version_constant generate_require_rollups` given `generate_semconv` will be called by the rollup etc.

kaylareopelle: That sounds good to me. Thank you!

thompson-tomo: All done @kaylareopelle

kaylareopelle (approved):

PR #2076, opened 2026-04-06
feat(otlp-grpc): add retry logic and comprehensive error handling to …

### LINKED ISSUE #1667: Create separate gems for http, grpc export and a common gem they can rely on

arjun-rajappa:
Closes #1667

Changes
- Implement exponential backoff retry mechanism for transient gRPC errors
- Add retry support for Unavailable, DeadlineExceeded, Cancelled, ResourceExhausted, Aborted, Internal, and DataLoss errors
- Improve certificate handling by reading file contents instead of passing file paths
- Add timeout support with deadline tracking across retry attempts
- Implement metrics reporting for export failures
- Add comprehensive test coverage for error scenarios, retry logic, and edge cases


exporter/otlp-grpc/lib/opentelemetry/exporter/otlp/grpc/trace_exporter.rb
```diff
@@ -83,6 +89,81 @@ def shutdown(timeout: nil)
...
+          def send_spans(span_data, timeout:) # rubocop:disable Metrics/MethodLength
+            retry_count = 0
+            timeout ||= @timeout
+            start_time = OpenTelemetry::Common::Utilities.timeout_timestamp
+
+            loop do
+              remaining_timeout = OpenTelemetry::Common::Utilities.maybe_timeout(timeout, start_time)
+              return FAILURE if remaining_timeout.zero?
+
+              request = OpenTelemetry::Exporter::OTLP::Common.as_etsr(span_data)
```
xuan-cao-swi: This can be moved outside the loop since neither `span_data` nor `request` will change.

arjun-rajappa: Suggested change has been made!

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
```diff
@@ -7,14 +7,730 @@
...
+    end
+
+    it 'initializes with custom endpoint' do
+      exporter = OpenTelemetry::Exporter::OTLP::GRPC::TraceExporter.new(endpoint: 'http://localhost:4317')
+      _(exporter).wont_be_nil
+    end
+
+    it 'initializes with custom timeout' do
+      exporter = OpenTelemetry::Exporter::OTLP::GRPC::TraceExporter.new(timeout: 5)
+      _(exporter).wont_be_nil
```
kaylareopelle: I notice the assertions for a lot of these tests are `wont_be_nil`. I worry we're missing out on an opportunity to validate behavior more fully, like by making sure that the endpoint/timeout/etc. was correctly assigned on the exporter. 

What are the benefits of `wont_be_nil` in these scenarios?

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
```diff
@@ -7,14 +7,730 @@
...
+    it 'exports spans with dropped attributes, events, and links' do
+      skip unless ENV['TRACING_INTEGRATION_TEST']
+      span_data = OpenTelemetry::TestHelpers.create_span_data(
+        total_recorded_attributes: 10,
+        attributes: { 'a' => 1, 'b' => 2 },
+        total_recorded_events: 5,
+        events: [OpenTelemetry::SDK::Trace::Event.new(name: 'event', timestamp: Time.now.to_i * 1_000_000_000)],
+        total_recorded_links: 3,
+        links: []
+      )
```
kaylareopelle, line 252: how does this indicate attributes/events/links were dropped?

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
```diff
@@ -7,14 +7,730 @@
...
+        end
+      ensure
+        cert_file.unlink
+      end
+    end
+  end
+
+  describe '#export' do
+    it 'exports span data successfully' do
+      skip unless ENV['TRACING_INTEGRATION_TEST']
```
kaylareopelle: I see this environment variable is used in other exporter tests, but it looks like the other exporters only use it once. Why use it so frequently here?

Also, I don't see `TRACING_INTEGRATION_TEST` being explicitly set anywhere. Will these tests be skipped most of the time?

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
```diff
@@ -7,14 +7,730 @@
...
+      span_data = OpenTelemetry::TestHelpers.create_span_data
+      exporter = OpenTelemetry::Exporter::OTLP::GRPC::TraceExporter.new(endpoint: 'http://localhost:4317')
+      result = exporter.export([span_data])
+      _(result).must_equal(success)
+    end
+
+    it 'exports multiple spans' do
```
kaylareopelle, line 172: Are these tests intended to replace the `it 'translates all the things'` tests from the other exporters? 

Ex: https://github.com/open-telemetry/opentelemetry-ruby/blob/51cf44d87b68b704d545ee4ed562bd018a78230a/exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb#L719-L911

kaylareopelle (commented):
Thank you for adding all the defensive error handling to the gRPC exporter! I have a few questions about the tests.

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
arjun-rajappa: You are correct — the use of wont_be_nil does not provide meaningful signal in this context, as the subsequent assertion would already raise a NoMethodError if the object were nil. Accordingly, these weaker assertions have been replaced with explicit equality checks:
• The custom endpoint initialization test now directly asserts that @client.@host == 'localhost:4317', eliminating both the intermediate variable and the redundant wont_be_nil check.
• The certificate-based initialization tests (certificate file, as well as client certificate with key) have been updated in the same manner.
• All #initialize environment variable tests already validate specific attributes such as @timeout, @shutdown, or @host.
As a result, each #initialize test now explicitly verifies the precise value being configured, ensuring stronger and more meaningful assertions.

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
arjun-rajappa: Good catch — the original test was only asserting that export returned SUCCESS, which says nothing about the dropped-count encoding. The test has been rewritten to intercept the actual protobuf request and assert directly on the proto fields:

_(span_proto.dropped_attributes_count).must_equal 8  # 10 recorded - 2 present
_(span_proto.dropped_events_count).must_equal 4      # 5 recorded - 1 present
_(span_proto.dropped_links_count).must_equal 3       # 3 recorded - 0 present

This verifies that as_etsr correctly computes and encodes the three dropped_*_count fields in the span proto.

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
arjun-rajappa: The original draft had the skip guard on many individual tests, which was excessive. It's now consolidated into two describe blocks — 'integration with tracer provider' and 'performance' — each with a single before { skip unless ENV['TRACING_INTEGRATION_TEST'] }. This matches the convention from the other exporters.

To answer your second question: yes, these tests are skipped by default and are only meant to run against a live OpenTelemetry Collector (e.g. in a local dev environment or a dedicated CI job). The guard is the same mechanism used by the HTTP exporter. All the behavioural assertions (encoding, error handling, retries, timeouts) are covered by unit tests that run unconditionally using a stubbed gRPC client.

exporter/otlp-grpc/test/opentelemetry/exporter/otlp/grpc/trace_exporter_test.rb
arjun-rajappa: No, the integration tests are not a replacement — they're there to confirm end-to-end connectivity against a live Collector and are skipped by default. They are intentionally light on assertions because the network path is what's being validated.

The encoding fidelity concern you're pointing at is now covered by the new it 'translates all the things' unit test (added to the #export describe block), which runs without a live server. It encodes a root and child span and directly asserts every significant proto field: trace/span IDs, parent span ID, span name, kind, status code, all attribute value types (bool, double, int, string, array), events with attributes, links with attributes, and the instrumentation scope name and version. This mirrors the pattern from the HTTP exporter test you linked.

kaylareopelle:
@arjun-rajappa - Can you take a look at the merge conflicts in this PR?

kaylareopelle:
Hi @arjun-rajappa, thanks for fixing the merge conflicts. We still have a few failing checks. Can you take a look at those?

PR #2080, opened 2026-04-08
fix: Limit length of response body read to 4mb

### LINKED ISSUE #2079: Exporters reading response body have no size limit
Our OTLP exporters read the full HTTP response body into memory without a size cap. 

Limiting the read size may help prevent memory exhaustion exploits when the configured collector endpoint is attacker-controlled (or a network attacker can mitm the exporter connection).

kaylareopelle:
Our exporters read HTTP response bodies without any limit. A misconfigured or misbehaving server could send an arbitrarily large response, causing the exporter to read it all into memory.

This implements a 4 MB response body limit also implemented by: 
* https://github.com/open-telemetry/opentelemetry-proto/pull/781
* https://github.com/open-telemetry/opentelemetry-js/pull/6552 

Based on https://cwe.mitre.org/data/definitions/789.html

Fixes #2079 

Combined with https://github.com/kaylareopelle/opentelemetry-ruby/pull/15 by @robbkidd


exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb
```diff
@@ -216,14 +218,42 @@ def handle_redirect(location)
...
+          end
+
+          def read_response_body(response)
+            return '' if response.nil?
+
+            body = +''
+            truncated = false
+
+            response.read_body do |chunk|
+              if body.bytesize + chunk.bytesize <= RESPONSE_BODY_LIMIT
```
arielvalentin, line 241: Before starting to read the response body, should we not first read the content-length or transfer-encoding and short circuit if the body exceeds the limit?

IIUC the Go collector will switch to using chunked responses that exceed the  2KiB default buffer size. In cases where the response is less than 2KiB or the backend supports larger buffer sizes we may get the content-length responses and exit early. 


https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Length

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Transfer-Encoding

kaylareopelle: I like the idea of using the headers to make this more performant!

Just to make sure I'm on the same page, are you proposing:
* Use the transfer-encoding / content-length headers to determine body size
* Skip chunking when the body is less than the limit and just return the body
* Truncate the response using chunking for bodies that exceed the limit

Or are you suggesting something else?

arielvalentin: My understanding is that it is either or. 

The response will either have a Content-Length or Transfer-Encoding Chunked

In the case the Content-Length is present and it exceeds the size then discard the response body and exit immediately.

If it's a chunked response use the code you've introduced here.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
robbkidd, line 178: `@http.request(request)` apparently **reads the body into memory under the hood**. 😱 Subsequent calls to `read_body()`, as in the new `read_response_body()` method, throw an `IOError: read_body called twice` exception (swallowed by the `rescue` in our new method).

We might need to restructure our `send_bytes()` method here a bit. I think we could wrap our case statement in `@http.request(request) do |response|` and choose how we're dealing with the `body` within the cases. Can even do @arielvalentin's header-checking suggestion within the block and before the cases.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -178,21 +179,22 @@ def send_bytes(bytes, timeout:) # rubocop:disable Metrics/CyclomaticComplexity,
 
             case response
             when Net::HTTPSuccess
-              response.body # Read and discard body
+              response.read_body(nil) # Discard without reading into memory
```
robbkidd, line 182: For our happier-path cases where the body isn't too large, `read_body(nil)` apparently won't drain the socket which will break keep-alive connections. Maybe for the cases where we don't care about the body content, we could `response.read_body { |_| }` for a not-entirely-awful drain-and-discard?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -240,14 +242,42 @@ def handle_redirect(location)
...
+          truncated = false
+
+          response.read_body do |chunk|
+            if body.bytesize + chunk.bytesize <= RESPONSE_BODY_LIMIT
+              body << chunk
+            else
+              remaining = RESPONSE_BODY_LIMIT - body.bytesize
+              body << chunk.byteslice(0, remaining) if remaining > 0
+              truncated = true
+              break
```
robbkidd, line 271: If `read_body()` hadn't thrown an exception and we got into this loop, the break would _still_ result in nearly all of the response being read into memory because [Net::HTTP reads the rest of the body after this block ends](https://github.com/ruby/net-http/blob/v0.9.1/lib/net/http/response.rb#L322). ಠ_ಠ

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -240,14 +242,42 @@ def handle_redirect(location)
         end
 
         def log_status(body)
+          truncation_note = @body_truncated ? ' (body truncated due to size limit)' : ''
```
robbkidd, line 245: Unrelated to the body-read conundrum, I worry a little about an instance variable on the exporter here. Maybe a second return value from `read_response_body()` that could be passed to a second param for `log_status()`?

robbkidd (changes requested):
In a hacky reproduction of the the problem (some malicious or misbehaving OTLP receiver responds with a large body), I still saw the whole large response body getting loaded into memory. The exporter's `@http.request(request)` call reads the full body before `read_response_body` runs, so the 4 MB cap never activates against a real HTTP server. The existing tests pass because WebMock stubs don't have real socket behavior. 😭

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -272,31 +272,34 @@ def log_status(body, truncated: false)
         def read_response_body(response)
           return ['', false] if response.nil?
 
+          body, truncated = collect_response_chunks(response)
+          body.force_encoding('UTF-8')
```
arielvalentin, line 291: If the response is truncated and the code points will be incomplete and could cause an invalid byte sequence when trying to translate it from binary to a String unicode sequence. 
e.g. if the last 8 bytes is multibyte but was cut off it will fail. E.g. take the number 1 in Chinese character U+4E00...  

```
 11100100 10111000 10000000
```

the first byte in that sequence expects to be followed by two bytes but when truncated it is  

```
 11100100 10111000
```

IIUC the `force_encoding` method only changes encoding attribute in the String and does not actually attempt to validate it. It will fail once someone tries to _read_ the string ... effectively to false advertising. The user of the body will then have to guard and ensure the encoding is valid or not or handle the invalid byte sequences. 

```
# U+4E00 (一, yī) is encoded in UTF-8 as 3 bytes: E4 B8 80
# Here we only have the first 2 bytes — the sequence is incomplete.
truncated = "\xE4\xB8".b
truncated.force_encoding("UTF-8")

puts truncated.encoding        #=> UTF-8
puts truncated.valid_encoding?  #=> false

begin
  puts truncated.encode("UTF-8")
rescue Encoding::InvalidByteSequenceError => e
  puts "Error: #{e.message}"
  #=> Error: "\xE4\xB8" on UTF-8
end
```

## What should we do? 

I am not sure but I am not a fan of flow and control by exception. If the body is truncated perhaps we should return the string with placeholders for invalid byte sequences or a blank string?

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -272,31 +272,34 @@ def log_status(body, truncated: false)
...
+          [body, truncated]
+        rescue StandardError => e
+          OpenTelemetry.handle_error(exception: e, message: 'error reading response body')
+          ['', false]
```
arielvalentin, line 296: This does not make sense to me. Does returning an empty string effectively mean we are truncating the string in its entirety?

Won't the protobuf parser be unhappy with an empty string?

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -1083,6 +1066,30 @@ def with_fake_server(response_body_size:, status: 400)
       server_thread&.join(2)
     end
 
+    def handle_fake_request(server, body, status)
+      client = server.accept
+      content_length = read_content_length(client)
+      client.read(content_length) if content_length > 0
+
+      client.print "HTTP/1.1 #{status} Bad Request\r\n"
+      client.print "Content-Type: application/octet-stream\r\n"
```
arielvalentin: IIUC, this is incorrect per the contract of the otel collector. It will respond with the same content type it received, which in our case is `application/x-protobuf`.

The exporter should probably check the response header content-type to ensure it is getting back when it requested. 

If for some reason an intermediate response with say `text/plain` then we know something went wrong and should not attempt to parse the body as protobuf anyway.

exporter/otlp/test/opentelemetry/exporter/otlp/exporter_test.rb
```diff
@@ -1083,6 +1066,30 @@ def with_fake_server(response_body_size:, status: 400)
...
+      client.print "Content-Length: #{body.bytesize}\r\n"
```
arielvalentin, line 1076: Can you add a test case for `Transfer-Encoding: chunked` without the `Content-Length` header which will likely be the most common response from the collecter?

PR #2094, opened 2026-04-16
feat: add attributes support to instrumentation scope for tracers

robertlaurin:
The OpenTelemetry spec (stable as of v1.24.0) defines an `attributes` parameter on `TracerProvider#tracer` as part of the instrumentation scope. This enables instrumentations to attach metadata to the scope that produced their spans.

The `tracer` method now accepts both legacy positional arguments and a new keyword-based interface. When both are provided for the same parameter, keyword arguments take precedence.

    # Legacy positional (unchanged, fully supported)
    tracer('name', '1.0')

    # New keyword interface
    tracer(name: 'name', version: '1.0', attributes: { 'key' => 'value' })

Backwards compatibility across independent gem releases is preserved:

- New API + Old SDK: ProxyTracerProvider introspects the delegate's method signature at assignment time and omits the attributes keyword when delegating to providers that do not support it.
- Old API + New SDK: The SDK's tracer method accepts positional arguments via splat, so old callers continue to work unchanged.

API changes:
- TracerProvider#tracer signature: (*args, name:, version:, attributes:)
- ProxyTracerProvider stores attributes in its registry key and delegates them when the underlying provider supports it

SDK changes:
- TracerProvider#tracer passes attributes through to Tracer
- Tracer#initialize accepts attributes and attaches them to InstrumentationLibrary
- InstrumentationLibrary struct gains an attributes field
- Attributes are normalized (nil -> frozen empty hash) and frozen to prevent mutation of registry keys

### DIFF
--- a/api/lib/opentelemetry/internal/proxy_tracer_provider.rb
+++ b/api/lib/opentelemetry/internal/proxy_tracer_provider.rb
@@ -13,7 +13,7 @@ module OpenTelemetry
     # It delegates to a "real" TracerProvider after the global tracer provider is registered.
     # It returns {ProxyTracer} instances until the delegate is installed.
     class ProxyTracerProvider < Trace::TracerProvider
-      Key = Struct.new(:name, :version)
+      Key = Struct.new(:name, :version, :attributes)
       private_constant(:Key)
 
       # Returns a new {ProxyTracerProvider} instance.
@@ -37,21 +37,50 @@ module OpenTelemetry
 
         @mutex.synchronize do
           @delegate = provider
-          @registry.each { |key, tracer| tracer.delegate = provider.tracer(key.name, key.version) }
+          @delegate_supports_attributes = supports_attributes?(provider)
+          @registry.each { |key, tracer| tracer.delegate = build_tracer(key.name, key.version, key.attributes) }
         end
       end
 
       # Returns a {Tracer} instance.
       #
-      # @param [optional String] name Instrumentation package name
-      # @param [optional String] version Instrumentation package version
+      # Supports both positional arguments (legacy) and keyword arguments:
+      #   tracer('name', '1.0')                                    # legacy positional
+      #   tracer(name: 'name', version: '1.0', attributes: {...})  # keyword
+      #
+      # When both positional and keyword arguments are provided for the same
+      # parameter, the keyword argument takes precedence.
+      #
+      # @param [String] name Instrumentation scope name
+      # @param [String] version Instrumentation scope version
+      # @param [Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}] attributes
+      #   Instrumentation scope attributes
       #
       # @return [Tracer]
-      def tracer(name = nil, version = nil)
+      def tracer(deprecated_name = nil, deprecated_version = nil, name: nil, version: nil, attributes: nil)
+        name ||= deprecated_name
+        version ||= deprecated_version
         @mutex.synchronize do
-          return @delegate.tracer(name, version) unless @delegate.nil?
+          return build_tracer(name, version, attributes) unless @delegate.nil?
 
-          @registry[Key.new(name, version)] ||= ProxyTracer.new
+          @registry[Key.new(name, version, attributes)] ||= ProxyTracer.new
+        end
+      end
+
+      private
+
+      def supports_attributes?(provider)
+        provider.respond_to?(:tracer) &&
+          provider.method(:tracer).parameters.any? { |_, n| n == :attributes }
+      end
+
+      # Delegates to the provider's tracer method, gracefully handling
+      # older SDK implementations that do not accept the attributes keyword.
+      def build_tracer(name, version, attributes)
+        if @delegate_supports_attributes
+          @delegate.tracer(name, version, attributes: attributes)
+        else
+          @delegate.tracer(name, version)
         end
       end
     end
--- a/api/lib/opentelemetry/internal/proxy_tracer_provider.rb
+++ b/api/lib/opentelemetry/internal/proxy_tracer_provider.rb
@@ -16,6 +16,19 @@ module OpenTelemetry
       Key = Struct.new(:name, :version, :attributes)
       private_constant(:Key)
 
+      # Wraps a legacy TracerProvider whose `#tracer` method does not accept
+      # the `attributes:` keyword argument, dropping the argument on delegation.
+      class LegacyProviderWrapper
+        def initialize(legacy_provider)
+          @legacy_provider = legacy_provider
+        end
+
+        def tracer(name = nil, version = nil, attributes: nil)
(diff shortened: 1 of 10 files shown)


api/lib/opentelemetry/internal/proxy_tracer_provider.rb
```diff
@@ -37,21 +37,50 @@ def delegate=(provider)
...
+      # @param [String] name Instrumentation scope name
+      # @param [String] version Instrumentation scope version
+      # @param [Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}] attributes
+      #   Instrumentation scope attributes
       #
       # @return [Tracer]
-      def tracer(name = nil, version = nil)
+      def tracer(*args, name: nil, version: nil, attributes: nil)
+        name ||= args[0]
+        version ||= args[1]
```
fbogsany: I'd prefer to avoid accepting (and ignoring) arbitrary positional args with a splat. Is this a reasonable alternative?

```suggestion
      def tracer(pos_name = nil, pos_version = nil, name: nil, version: nil, attributes: nil)
        name ||= pos_name
        version ||= pos_version
```

api/lib/opentelemetry/trace/tracer_provider.rb
```diff
@@ -10,11 +10,20 @@ module Trace
...
+      # parameter, the keyword argument takes precedence.
+      #
+      # @param [String] name Instrumentation scope name
+      # @param [String] version Instrumentation scope version
+      # @param [Hash{String => String, Numeric, Boolean, Array<String, Numeric, Boolean>}] attributes
+      #   Instrumentation scope attributes
       #
       # @return [Tracer]
-      def tracer(name = nil, version = nil)
+      def tracer(*args, name: nil, version: nil, attributes: nil)
```
fbogsany:
```suggestion
      def tracer(pos_name = nil, pos_version = nil, name: nil, version: nil, attributes: nil)
```

Alternative is to explicitly call these out as "legacy" or "deprecated". E.g.
```suggestion
      def tracer(deprecated_name = nil, deprecated_version = nil, name: nil, version: nil, attributes: nil)
```

robbkidd:
This is on my list to review today/this week. I had just started planning a proposal to implement `schema_url` in API and SDK for all the signals. I _think_ that will overlap with this PR in that both will update the method signature for `TracerProvider.tracer()`.  (🤔 … though, in the shift to using keyword arguments, we don't really need to decide on order in signature now.)

Is "scope-level attributes" and "scope schema_url" separate enough concerns to implement in separate PRs or should we combine forces?

sdk/test/opentelemetry/sdk/trace/tracer_provider_test.rb
```diff
@@ -168,11 +174,68 @@
...
+
+    # Mixed positional + keyword
+    it 'accepts positional name with keyword version' do
+      tracer = tracer_provider.tracer('component', version: '1.0')
+      _(tracer).wont_be_nil
+    end
+
+    it 'prefers keyword arguments over positional arguments' do
+      tracer1 = tracer_provider.tracer('positional', '1.0')
+      tracer2 = tracer_provider.tracer('keyword', '2.0', name: 'keyword', version: '2.0')
```
robbkidd, line 204: 🤔 

```suggestion
      tracer2 = tracer_provider.tracer('positional', '1.0', name: 'keyword', version: '2.0')
```
I _think_ these suggested `tracer1` positional params given to `tracer2` will properly exhibit the kw arg precedence. When all arguments are entirely different, `tracer2` will always be different than `tracer1`.

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
```diff
@@ -44,15 +46,28 @@ def initialize(sampler: sampler_from_environment(Samplers.parent_based(root: Sam
...
+        #   Instrumentation scope attributes
         #
         # @return [Tracer]
-        def tracer(name = nil, version = nil)
-          name ||= ''
-          version ||= ''
+        def tracer(deprecated_name = nil, deprecated_version = nil, name: nil, version: nil, attributes: nil) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+          name ||= deprecated_name || ''
+          version ||= deprecated_version || ''
+          attributes = attributes&.dup&.freeze || EMPTY_ATTRIBUTES
```
robbkidd, line 65: Hmm. We should probably do attribute validation here [like we do on spans](https://github.com/open-telemetry/opentelemetry-ruby/blob/0b94ef6086facf3c7ad584485bb3825b0ab90e39/sdk/lib/opentelemetry/sdk/trace/span.rb#L277).

robbkidd:
I'm a fan of all of this. Scope attributes!

Are these intentionally out-of-scope for this PR?

* updating exporters to use the scope-level attributes when serializing for send
* adding scope-level attributes to meter and logger

api/lib/opentelemetry/internal/proxy_tracer_provider.rb
```diff
@@ -37,21 +37,50 @@ def delegate=(provider)
 
         @mutex.synchronize do
           @delegate = provider
```
fbogsany, line 54: Rather than have conditional code executed on every `provider.tracer(...)` call, could we conditionally wrap the `provider` before the mutex with a legacy provider that just drops the `attributes:` arg? I.e. (roughly)
```ruby
class LegacyProviderWrapper
  def initialize(legacy_provider)
    @legacy_provider = legacy_provider
  end

  def tracer(name, version, attributes:)
    @legacy_provider.tracer(name, version)
  end
end

# then above

provider = LegacyProviderWrapper.new(provider) unless provider.respond_to?(:tracer) && provider.method(:tracer).parameters.any? { |_, n| n == :attributes }
```

This would avoid penalizing providers implementing the new signature (which should be the common case, assuming most people use the SDK provider) and avoids the conditional dispatch in the legacy case.

fbogsany (approved):

sdk/lib/opentelemetry/sdk/trace/tracer_provider.rb
kaylareopelle: The spec doesn't say anything about attribute validation on tracers, but this sounds helpful as a safety measure.

kaylareopelle:
Hi @robertlaurin, thanks for opening this PR. There's a few Rubocop errors for the API library. Can you take a look?

robertlaurin:
> I'm a fan of all of this. Scope attributes!
> 
> Are these intentionally out-of-scope for this PR?
> 
> * updating exporters to use the scope-level attributes when serializing for send
> * adding scope-level attributes to meter and logger

Yeah the intent is to keep it scoped almost entirely to the api. The sdk changes as useful for validating that this doesn't break with the api change. 

Meter and logger can be a separate effort, and this PR makes for a useful template to follow.

PR #2096, opened 2026-04-18
chore: cleanup usages of benchmark-ipsa

thompson-tomo:
This removes unused benchmark-ipsa dev depency.

Note it is used in the api gem.


xuan-cao-swi:
Maybe need to update the api/benchmarks with gem benchmark https://github.com/ruby/benchmark

xuan-cao-swi (approved):

kaylareopelle (approved):

PR #2095, opened 2026-04-18
chore: use gemfile for dev dependencies

thompson-tomo:
Closes: #2021

This switches dev dependencies to gemfile when they are commonly used. This allows renovate to manage them and allows us to progress with removing rubocop suppression.

PR #2101, opened 2026-04-20
feat: Add git tag to gemspec

thompson-tomo:
This adds the git tag to the source code Uri of the gems as done in contrib.

To reduce churn, this PR should land in the same release as https://github.com/open-telemetry/opentelemetry-ruby/pull/2095


arielvalentin (approved):

kaylareopelle:
@thompson-tomo Would you mind changing the conventional commit on this one to "feat" so we have something to put in the changelog for the gem changes leftover from the logs/metrics SDK requires?

thompson-tomo:
@kaylareopelle done

kaylareopelle (approved):

PR #2102, opened 2026-04-21
ci: Replicate contrib renovate setup

thompson-tomo:
This sync's the renovate setup from contrib which has introduced custom managers for managing dependencies not natively supported. It also introduces the concept of min ruby which is managed via package rules. This ensures examples use the lowest minor supported ruby as opposed to the latest.

Summary of additions:
- OS updating in matrix
- excluding of ruby updates from patch groups
- Min ruby sdk bumping in gemspecs & dockerfile only when previous min is eol
- identification/processing of dependencies in Appraisals & Gemspecs
- Priority given to updating setup actions


arielvalentin (approved):

.github/renovate.json5
```diff
@@ -1,21 +1,162 @@
...
+      matchStrings: [
+        "FROM\\s+(?<packageName>.+):(?<depVersion>[0-9.]+)-alpine(?<alpineVersion>[0-9.]+)@(?<currentDigest>sha256:[a-f0-9]{40,64})\\s",
+        "FROM\\s+(?<packageName>.+):(?<depVersion>[0-9.]+)-alpine(?<alpineVersion>[0-9.]+)\\s",
+      ],
+      datasourceTemplate: "docker",
+      versioningTemplate: "docker",
+      currentValueTemplate: "{{alpineVersion}}",
+      extractVersionTemplate: "^{{depVersion}}-alpine(?<version>[0-9.]+)$",
+      depNameTemplate: "{{packageName}}-alpine",
+      depTypeTemplate: "final",
```
kaylareopelle, line 102: I notice this isn't in the contrib equivalent of this manager. What does it give us here?

thompson-tomo: This just adds a type so that it matches the builtin manager which is purely informational. It's not in contrib as I didn't want to raise a pr which made such a small trivial change. I will set it when I have the next round of adjustments.

kaylareopelle (approved):

PR #2104, opened 2026-04-22
fix(metrics): return false if it is noop_exemplar_reservoir

xuan-cao-swi:
## Description
The reservoir operation remains active for noop_exemplar_reservoir, even though it doesn't store anything; however, it still slows down the operation.


robbkidd (approved):
Yup. And the `reset()` here for `NoopExemplarReservoir` fixes [a performance hit demonstrated by the benchmarks](https://github.com/open-telemetry/opentelemetry-ruby/pull/2106/changes#r3139526446) in #2106.

kaylareopelle (approved):

PR #2105, opened 2026-04-22
ci: (workflows & actions) interpolate into env vars

robbkidd:
Move ${{ ... }} interpolations from 'run:' scripts to env vars to avoid shell injection shenanigans. YAML string handling works as hoped for the variables whose values are strings with spaces in them, like flags and named-things-to-be-released.

### DIFF
--- a/.github/actions/test_gem/action.yml
+++ b/.github/actions/test_gem/action.yml
@@ -31,9 +31,12 @@ runs:
     - name: Setup
       id: setup
       shell: bash
+      env:
+        GEM: ${{ inputs.gem }}
+        RUBY_VERSION: ${{ inputs.ruby }}
       run: |
         # 🛠️ Setup 🛠️
-        dir=$(find . -iname '${{ inputs.gem }}.gemspec' -exec dirname {} \;)
+        dir=$(find . -iname "${GEM}.gemspec" -exec dirname {} \;)
         echo "gem_dir=${dir}" >> $GITHUB_OUTPUT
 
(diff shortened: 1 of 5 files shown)


robbkidd:
I exercised these changes locally as much as possible.

xuan-cao-swi (approved):

.github/actions/test_gem/action.yml
```diff
@@ -119,7 +122,9 @@ runs:
     - name: Build Gem
       shell: bash
       if: "${{ inputs.build == 'true' }}"
```
kaylareopelle, line 124: Is this an input we need to be concerned about?

kaylareopelle (commented):
Thank you for doing this, @robbkidd! One question before I approve.

.github/actions/test_gem/action.yml
robbkidd: No, but it's good to be suspicious and ask why!

The `if:` field is evaluated by whatever YAML processing GitHub Actions is doing. The value of `inputs.build` is not passed to a shell. The equality check is performed by Action YAMLing, so `sh hax0rs.activate shenanigans == 'true'` would result in `false` as the value for `if:`.

As I understand the vulnerability, it only presents when `${{ … }}` appears directly within specifically a `run:` block since that's passed on for a shell to execute.

kaylareopelle: Thank you! I appreciate your explanation! 🎉

kaylareopelle (approved):

PR #2107, opened 2026-04-22
chore(api): update benchmarks

xuan-cao-swi:
## Description

`benchmarks` was broken: added the opentelemetry-sdk as test dependency.
I don't see the need to keep using `benchmark-ipsa` for the additional `memory_profiler.report` (correct me if I am wrong)

### DIFF
--- a/api/benchmarks/context_bench.rb
+++ b/api/benchmarks/context_bench.rb
@@ -4,7 +4,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-require 'benchmark/ipsa'
+require 'benchmark/ips'
 require 'concurrent-ruby'
 require 'opentelemetry'
 
@@ -784,7 +784,7 @@ class FiberLocalImmutableArrayContext
   ROOT = empty.freeze
 end
 
-Benchmark.ipsa do |x|
+Benchmark.ips do |x|
   x.report 'FiberAttributeContext.with_value' do
     FiberAttributeContext.with_value('key', 'value') { |ctx, _| ctx }
   end
@@ -820,7 +820,7 @@ Benchmark.ipsa do |x|
   x.compare!
 end
 
-Benchmark.ipsa do |x| # rubocop:disable Metrics/BlockLength
+Benchmark.ips do |x| # rubocop:disable Metrics/BlockLength
   x.report 'LinkedListContext.with_value recursive' do
     LinkedListContext.with_value('key', 'value') do
       LinkedListContext.with_value('key', 'value') do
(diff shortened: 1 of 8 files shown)


api/opentelemetry-api.gemspec
```diff
@@ -26,7 +26,7 @@ Gem::Specification.new do |spec|
   spec.required_ruby_version = '>= 3.3'
   spec.add_dependency 'logger'
 
-  spec.add_development_dependency 'benchmark-ipsa', '~> 0.2.0'
+  spec.add_development_dependency 'benchmark-ips', '~> 2.14.0'
```
thompson-tomo, line 29: Because we are switching tools why don't we move this to the gemfile?

api/benchmarks/context_bench.rb
```diff
@@ -784,7 +784,7 @@ def set_values(values) # rubocop:disable Naming/AccessorMethodName:
   ROOT = empty.freeze
 end
 
-Benchmark.ipsa do |x|
+Benchmark.ips do |x|
```
thompson-tomo, line 787: Are you sure that this isn't changing the measurements/data being captured/recorded?

xuan-cao-swi: ipsa only have additional allocations stats from `MemoryProfiler.report()`
e.g.
```
Allocations -------------------------------------
               name=       0/0  alloc/ret        0/0  strings/ret
       set_attribute       0/0  alloc/ret        0/0  strings/ret
           add_event       2/2  alloc/ret        0/0  strings/ret
``` 
I'm not sure this information is still valuable or relevant, as it doesn't provide meaningful comparisons like `ips` does (e.g., `xxx is 1.45x faster than xxx`). `ipsa` is also quite old and conflicts with using the latest version of `ips`. If allocation data is truly needed, we could simply inline/copy the relevant `MemoryProfiler.report` logic rather than depending on an unmaintained gem.

api/opentelemetry-api.gemspec
xuan-cao-swi: I am ok with either put here or in gemfile

thompson-tomo: I would rather gem so that we can move towards eliminating the disabling of the corresponding rubocop cop & the updating process is natively handled by renovate.

api/benchmarks/context_bench.rb
thompson-tomo: I am All for switching from an unmaintained gem, my question steemed from reading the pr description.

kaylareopelle (changes requested):
As a follow-up to our discussion in the SIG, marking this as "Request Changes" to update documentation to include benchmark results.

kaylareopelle (approved):
Thank you!

PR #2108, opened 2026-04-22
fix(sdk): consistent labels for otel.bsp.dropped_spans metric

### LINKED ISSUE #2040: `OpenTelemetry::SDK::Trace::Export.report_dropped_spans` does not always produce the same set of labels, which the prometheus library considers invalid
**Description of the bug**

`OpenTelemetry::SDK::Trace::Export.report_dropped_spans` does not always produce the same set of labels, which the prometheus library considers invalid

`labels must have the same signature: (expected keys: [:reason], got: [:code.function, :reason])`.

I believe the issue is [this line of code](https://github.com/open-telemetry/opentelemetry-ruby/blob/740a5ff2b6793ddfa58d858136c03fc97c164b3d/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb#L209), which calls `compact` on the labels hash, and uses `nil` for the `function` param.

For example:

```ruby
irb(main):007> {'reason' => 'foo', 'code.function' => 'bar'}.compact
=> {"reason" => "foo", "code.function" => "bar"}
irb(main):008> {'reason' => 'foo', 'code.function' => nil}.compact
=> {"reason" => "foo"}
```

The prometheus library will detect that the label set has changed and throws an exception (see above)

Operating system details: Linux, Ubuntu 20.04 LTS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.4.5"
PROM_VERSION: prometheus-client-mmap (~> 1.1.1)

robbkidd:
Fixes #2040 

`report_dropped_spans()` is used in several of places and not all of them include a `function` name in the call. The `.compact` on the labels hash within `report_dropped_spans()` was removing `nil`s. This produces inconsistent label sets for the `'otel.bsp.dropped_spans'` metric. And that makes metrics reporters like prometheus-client unhappy because they require a stable set of label names.

This change leaves how to handle `nil`-value labels to concrete implementations of `MetricsReporter` instead.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
+++ b/sdk/lib/opentelemetry/sdk/trace/export/batch_span_processor.rb
@@ -206,7 +206,7 @@ module OpenTelemetry
           end
 
           def report_dropped_spans(dropped_spans, reason:, function: nil)
-            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: dropped_spans.size, labels: { 'reason' => reason, OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => function }.compact)
+            @metrics_reporter.add_to_counter('otel.bsp.dropped_spans', increment: dropped_spans.size, labels: { 'reason' => reason, OpenTelemetry::SemanticConventions::Trace::CODE_FUNCTION => function })
           end
 
           def lock
--- a/sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
+++ b/sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb
@@ -387,6 +387,27 @@ describe OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor do
     end
   end
 
+  describe 'dropped spans metrics' do
+    let(:metrics_reporter) { TestMetricsReporter.new }
+
+    it 'reports consistent label keys' do
+      exporter = TestExporter.new(status_codes: [FAILURE])
+      bsp = BatchSpanProcessor.new(exporter,
+                                   metrics_reporter: metrics_reporter,
+                                   max_queue_size: 1,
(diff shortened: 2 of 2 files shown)


xuan-cao-swi (approved):

arielvalentin:
What is nil? Code function? 

Should the default be a blank string instead of nil in the method signature? 

Should the method be defensive and ensure all labels are valid? 

Otherwise we are relying on the metrics reporter to have to be defensive as well.

robbkidd:
> What is nil? Code function?

Yes. There are some call-sites that do not pass in `function:`. The current default is `nil` and compacting a `nil` value in the hash removed the function label which caused grief for prometheus.

> Should the default be a blank string instead of nil in the method signature?

I don't know. Our metrics reporter is off-spec, so the handling of nil vs blank string has historically been left to the implementer of a concrete reporter.

> Should the method be defensive and ensure all labels are valid?

Valid according to whose rules? Because our current metrics reporter interface is off-spec, someone wanting to use it decides what to do about labels and validity depend on where those labels are getting sent.

> Otherwise we are relying on the metrics reporter to have to be defensive as well.

Yes, this change places the onus of defense on the concrete implementation of a metrics reporter that the SDK does not (to my knowledge) yet supply.

--- 

I'm currently thinking that these questions could be addressed in followup issues and PRs so that the poor experience for concrete implementations choosing to send to prometheus today is improved.

kaylareopelle (approved):

kaylareopelle:
> I'm currently thinking that these questions could be addressed in followup issues and PRs so that the poor experience for concrete implementations choosing to send to prometheus today is improved.

@robbkidd - I think that's a good idea. Could you create issues for the followups?

thompson-tomo:
@robbkidd / @kaylareopelle it appears that the new test is flaky on jruby as sometimes it is passing but in others it is failing. See the pr's I have just updated.

PR #2112, opened 2026-04-29
feat!: use trace_based exemplar filter by default

bdewater-thatch:
My coworkers complained about the `OTEL_METRICS_EXEMPLAR_FILTER  is not part of the provided exemplar filters. Exemplar is off` warning that started popping up after we added the metrics SDK to our app. The ENV read also triggered before one can set `OpenTelemetry.meter_provider.enable_exemplar_filter` so that was a little bit annoying.

According to https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk-environment-variables.md#exemplar trace_based is the default so that's what I implemented and it makes the aforementioned warning go away when the ENV var isn't explicitly set.


xuan-cao-swi (approved):
The intent was to ensure that metrics users do not experience any unanticipated overhead.
This is a breaking change so probably should use `feat!`

bdewater-thatch:
@xuan-cao-swi thanks for the feedback - fixed!

kaylareopelle (approved):

PR #2114, opened 2026-05-02
chore: remove redundant script

thompson-tomo:
This removes the redundant script for generating registry docs. This is redundant on multiple fronts with the big one being it is looking for gemspecs in the instrumentation folder, however that folder doesn't exist.


kaylareopelle (approved):
Good catch!

PR #2117, opened 2026-05-06
ci: Replicate & update simplecov based on contrib

### LINKED ISSUE #2245: Updates for simplecov v1
With the update to simplecov we are now getting the followingin our logs:

> [DEPRECATION] Calling `SimpleCov.start` from `.simplecov` is deprecated and will be removed in a future release. `.simplecov` should contain configuration only; move the `SimpleCov.start` call into your `spec_helper.rb` / `test_helper.rb`. Coverage tracking still begins for backward compatibility, but a future release will require the explicit `SimpleCov.start` from a test helper. See https://github.com/simplecov-ruby/simplecov/issues/581.

As such we should switch to calling it from our test helper.

While as it we should review how we are setting the command name as i suspect we don't need the overrides. Key thing to be checked is that each appraisal generates a differently named coverage report.

Also the coverage target percentage can likely come out of ci and purely env variables.

thompson-tomo:
Closes: #2245 

This cleans up simplecov configuration to be consistent with contrib, ensure coverage analysis is calculated correctly and that it is compliant with v1 of simplecov.

The .env files have been added as we are not meeting coverage requirements and was previously un-noticed due to incorrect calculation.


simi:
CI looks unrelated, but if I remember well, same issue popped recently few times. Maybe worth to investigate.

thompson-tomo:
Have merged in main & checks all pass.

thompson-tomo:
@open-telemetry/ruby-approvers could you take a look at this? Thanks

mwear (approved):

kaylareopelle (approved):

PR #2120, opened 2026-05-07
chore: Enable latest rubocop rules

thompson-tomo:
Closes: #2023

This sets rubocop rules as enabled by default which introduced a number of violations. To Address violation either rules have been disabled at the gem/global level or violation addressed by fixing source in only cases where it is a very straight forward fix.

Note a number of overrides have been removed as they were duplicate of base or repeated in most projects hence have been added to base rules ie file naming.


api/lib/opentelemetry/context/propagation/rack_env_getter.rb
```diff
@@ -33,7 +33,7 @@ def keys(carrier)
 
         def to_rack_key(key)
           # Use + for mutable string interpolation in pre-Ruby 3.0.
```
kaylareopelle: Since we're removing the `+` we don't need the comment.
```suggestion
```

kaylareopelle (approved):
Thank you for carrying this work forward. This makes a lot of progress toward bringing Rubocop up to date across the repo. In the future, it would be nice to update the code to enable all Performance cops.

thompson-tomo:
I agree, i just wanted to avoid making this pr too big by tackling too much at once.

PR #2122, opened 2026-05-09
ci: Automated check of PR titles

thompson-tomo:
This adds in spell checking of pr titles as done in contrib.

While updating this workflow the conventional commit linter has been switched to a more widely used option as done in contrib.


.github/workflows/conventional-commits.yml
```diff
@@ -1,4 +1,4 @@
-name: Conventional Commits Validation
+name: Review / PR
```
kaylareopelle: Thank you for your continued efforts to align this repo with contrib!

I prefer the old name, it's easier to identify and is closer to the name of the file. It also would be more aligned with contrib.


```suggestion
name: Conventional Commits Validation
```

.github/workflows/conventional-commits.yml
```diff
@@ -8,25 +8,55 @@ on:
...
-  group: ${{ github.workflow }}-${{ github.event.pull_request.number }}  # Ensure that only one instance of this workflow is running per Pull Request
-  cancel-in-progress: true  # Cancel any previous runs of this workflow
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.event.pull_request.number || github.ref }} # Ensure that only one instance of this workflow is running per Pull Request or ref
+  cancel-in-progress: true # Cancel any previous runs of this workflow
 
 jobs:
   validate-commits:
     if: ${{ github.repository == 'open-telemetry/opentelemetry-ruby' }}
-    name: Conventional Commits Validation
+    name: Title
```
kaylareopelle:
```suggestion
    name: Conventional Commits Validation
```

kaylareopelle (changes requested):

.github/workflows/conventional-commits.yml
thompson-tomo: The reason i changed it is because the check now also does spell checking of the pr title hence the name wasn't accurate due to the new tasks.

.github/workflows/conventional-commits.yml
thompson-tomo: As above the check now does more than conventional commit validation.

.github/workflows/conventional-commits.yml
kaylareopelle: The new name still doesn't quite make sense to me. If it feels like the old name is no longer accurate, what about this: 

```suggestion
name: Conventional Commits PR Title Validation
```

I want to keep conventional commits somewhere in there, because that is the purpose -- we want to communicate to users their PR titles need to be conventional commits.

.github/workflows/conventional-commits.yml
kaylareopelle: I see. Title still feels too vague. How is this?
```suggestion
    name: Validate PR Title
```

.github/workflows/conventional-commits.yml
```diff
@@ -8,25 +8,55 @@ on:
...
     runs-on: ubuntu-24.04
     steps:
-      - uses: dev-build-deploy/commit-me@27d14296ec218488c543616d9128d7e75ab7463a # v1.5.3
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      # restrict checks to pull requests to avoid issues blocking the merge queue
+      - name: Validate PR Title
```
kaylareopelle:
```suggestion
      - name: Validate conventional commit style title
```

.github/workflows/conventional-commits.yml
kaylareopelle: Or, to leverage the `/` that displays in the GitHub Action preview on the PR, consider: 
```suggestion
name: PR Title Validation
```

.github/workflows/conventional-commits.yml
kaylareopelle: Continuing the second idea, to leverage the `/`:

```suggestion
    name: Conventional Commits
```

kaylareopelle (approved):
Thanks for the updates, @thompson-tomo.

Issue #1435, opened 2023-03-14
API reference "latest" pages should yield HTTP status 301 not 404


chalin:
Visit one of the API reference pages and you'll initially get a 404, for example [opentelemetry-sdk/latest](https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/latest):

```console
$ curl -I https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/latest
HTTP/2 404 
```

If you [open the _same_ link in a browser](https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-sdk/latest), then **the page redirects** because it contains JS code that manipulates the window `location` to refer to the latest version of the appropriate Gem.

Since **API reference pages get _redirected_** the proper HTTP status returned should be 301, not 404. Otherwise, that's bad for SEO and for link checkers -- e.g., see https://github.com/open-telemetry/opentelemetry.io/pull/2505/files#r1135889989.

/cc @cartermp @svrnm

arielvalentin:
I am not sure why it is hosted on `github.io` since it is available when we publish libraries here: https://www.rubydoc.info/gems/opentelemetry-sdk

cartermp:
TIL! Let's update the docs site to point there for the API link too

cartermp:
Submitted https://github.com/open-telemetry/opentelemetry.io/pull/2508

cartermp:
I think we can consider this closed since we'll use the rubydoc link now.

chalin:
We might have more to discuss, see:

- https://github.com/open-telemetry/opentelemetry.io/issues/2512

Also, having the same content in two places, isn't a good idea for SEO etc. My guess is that GH pages is used for dev versions of the API reference? If not, and it's the same as rubydoc.info, then maybe it's time to drop the GH pages?

At a minimum, the GH pages should all probably be marked [noindex, nofollow](https://developers.google.com/search/docs/crawling-indexing/block-indexing), and in addition have an appropriate `robots.txt` file.

ahayworth:
The one nice thing that the GH pages version gives us is index page with links to all the docs: https://open-telemetry.github.io/opentelemetry-ruby/

It would be kinda nice to keep that somehow. But otherwise yeah - I don't really see a huge need to keep hosting these on GH pages. Does anyone else even do that?

[...]

PR #2123, opened 2026-05-09
fix: correct documentation links

thompson-tomo:
Ensure docs point to rubydoc.info which is where the docs are hosted rather than github pages.

I have no strong opinion on if links should be rubydocs or gh pages but went with rubydocs as that is what contrib uses & it is current.


kaylareopelle (approved):
This reminds me of #1435. We may want to retire GitHub pages altogether.

thompson-tomo:
I would be all for that @kaylareopelle given it appears not to be maintained etc.

PR #2128, opened 2026-05-11
chore: simplify yard rake task

thompson-tomo:
This ensures yardopts is the authorities source of yard options as done in contrib https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2292.

Further rubocop is analyzing method documentation coverage.


kaylareopelle (approved):
This seems like a more effective way to track our documentation gaps. Thank you!

PR #2132, opened 2026-05-12
feat: Add in generation of event names

thompson-tomo:


semantic_conventions/templates/registry/ruby/weaver.yaml
```diff
@@ -45,6 +45,12 @@ templates:
     file_name: "incubating/{{ctx.root_namespace | snake_case}}/metrics.rb"
     params:
       incubating: true
+  - pattern: events.j2
```
kaylareopelle, line 48: Are all events incubating at this time? Do we need to add any to the stable group?

semantic_conventions/templates/registry/ruby/weaver.yaml
```diff
@@ -45,6 +45,12 @@ templates:
...
+    application_mode: each
+    filter: '.groups | map(select(.type == "event")) | group_by(.name | split(".") | .[0]) | map({id: .[0].id | split(".") | .[1], groups: .})'
```
kaylareopelle, line 50: I'm surprised weaver doesn't have a `semconv_grouped_events` filter. I checked the same [spot where `semconv_grouped_metrics`](https://github.com/open-telemetry/weaver/blob/main/defaults/jq/semconv.jq) is defined and didn't see anything. This seems to capture the events well.

kaylareopelle (commented):
I think this looks good, @thompson-tomo. Thanks for opening the PR.

semantic_conventions/templates/registry/ruby/weaver.yaml
thompson-tomo: It is in the process of being added as it was in the docs but didnt work as wasn't implemented.

semantic_conventions/templates/registry/ruby/weaver.yaml
thompson-tomo: Yes I was going to add stable but first want to resolve/confirm namespacing situation.

thompson-tomo:
Awaiting on guidance on how to address potential naming conflicts arising from not splitting based on type.

thompson-tomo:
Awaiting guidance on the naming issue ie event name potentially being the same as attribute, entity, metric etc.

thompson-tomo:
Still concerned about the naming conflicts etc

PR #2133, opened 2026-05-12
chore: Fix Ruby version format in fossa.yml

thompson-tomo:
Currently renovate is having trouble updating the ruby version, looking at what is updating and that this instance is being misidentified as 4 not 4.0. I think this is the cause. This is backed up by contrib having no issue updating from 4.0 to 4.0.3 when it was quoted.


thompson-tomo:
@kaylareopelle could you merge this in to hopefully resolve the renovate error we are currently getting.

kaylareopelle (approved):

PR #2145, opened 2026-05-17
ci: Change pull_request to pull_request_target for semconv generation

thompson-tomo:
This will hopefully ensure that the renovate semconv pr's undergo ci checks after the second commit which is the content generation.


thompson-tomo:
@kaylareopelle another fix for the auto generation of semconv.

kaylareopelle (approved):
Good catch. Thank you!

PR #2158, opened 2026-05-20
fix: `URI.join` dropping path prefix in logs and metrics exporters

### LINKED ISSUE #2157: `OTEL_EXPORTER_OTLP_ENDPOINT` path prefix is dropped in logs and metrics exporters
**Description of the bug**

When `OTEL_EXPORTER_OTLP_ENDPOINT` contains a path prefix without a trailing slash (e.g., `http://collector.example.com/api`), the OTLP logs and metrics exporters silently drop the path prefix (`/api`) when appending the signal path (`v1/logs` or `v1/metrics`). 

This happens because they use `URI.join` directly without normalizing the trailing slash first. `URI.join` performs RFC 3986 relative-reference resolution, which treats the last path segment as a "file" and replaces it if there is no trailing slash. The traces exporter (`exporter.rb`) already handles this correctly by appending a `/` before joining.

Spec compliant behavior: https://opentelemetry.io/docs/specs/otel/protocol/exporter/#endpoint-urls-for-otlphttp

Operating system details: MacOS
RUBY_ENGINE: "ruby"
RUBY_VERSION: "3.4.5"
RUBY_DESCRIPTION: "ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [arm64-darwin24]"

```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-exporter-otlp-logs'
  gem 'opentelemetry-exporter-otlp-metrics'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'
require 'opentelemetry-exporter-otlp-logs'

ENV['OTEL_EXPORTER_OTLP_ENDPOINT'] = 'http://collector.example.com/api'

logs_exporter = OpenTelemetry::Exporter::OTLP::Logs::LogsExporter.new
metrics_exporter = OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporter.new

# Expected: /api/v1/logs
# Actual (before fix): /v1/logs
puts "Logs Exporter Path: #{logs_exporter.instance_variable_get(:@path)}"

# Expected: /api/v1/metrics
# Actual (before fix): /v1/metrics
puts "Metrics Exporter Path: #{metrics_exporter.instance_variable_get(:@path)}"
```

vivekmiyani:
Fixes #2157.

### DIFF
--- a/exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
+++ b/exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
@@ -60,6 +60,7 @@ module OpenTelemetry
             raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
             @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+                     endpoint += '/' unless endpoint.end_with?('/')
                      URI.join(endpoint, 'v1/logs')
                    else
                      URI(endpoint)
--- a/exporter/otlp-logs/test/opentelemetry/exporter/otlp/logs_exporter_test.rb
+++ b/exporter/otlp-logs/test/opentelemetry/exporter/otlp/logs_exporter_test.rb
@@ -159,6 +159,24 @@ describe OpenTelemetry::Exporter::OTLP::Logs::LogsExporter do
       _(exp.instance_variable_get(:@path)).must_equal '/v1/logs'
     end
 
+    it 'appends the correct path if OTEL_EXPORTER_OTLP_ENDPOINT does have a path without a trailing slash' do
+      exp = OpenTelemetry::TestHelpers.with_env(
+        'OTEL_EXPORTER_OTLP_ENDPOINT' => 'https://localhost:1234/api/v2/otlp'
+      ) do
+        OpenTelemetry::Exporter::OTLP::Logs::LogsExporter.new
+      end
+      _(exp.instance_variable_get(:@path)).must_equal '/api/v2/otlp/v1/logs'
+    end
+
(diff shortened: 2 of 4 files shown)


thompson-tomo (approved):

exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb
```diff
@@ -60,6 +60,7 @@ def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPOR
             raise ArgumentError, "unsupported compression key #{compression}" unless compression.nil? || %w[gzip none].include?(compression)
 
             @uri = if endpoint == ENV['OTEL_EXPORTER_OTLP_ENDPOINT']
+                     endpoint += '/' unless endpoint.end_with?('/')
```
xuan-cao-swi, line 63: why not just use something similar from trace_exporter (e.g. `URI("#{endpoint}/v1/traces")`)

vivekmiyani: I avoided that because [this older PR](https://github.com/open-telemetry/opentelemetry-ruby/pull/1339) specifically moved away from string interpolation to fix trailing slash issue.

xuan-cao-swi (approved):

mwear (approved):
LGTM, it matches what we have for traces:

https://github.com/open-telemetry/opentelemetry-ruby/blob/370a8ed0a7308eec74be904604659c1fd515d0fc/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L424-L425

Issue #2127, opened 2026-05-11
Support Declarative configuration


thompson-tomo:
In early 2026 Declarative configuration initial stable version was released and as such we should be looking to add support for it.

This is more than just reading just config from a file but a new approach to managing the configuration layer which is designed to be changing over time.

The support status is documented at https://github.com/open-telemetry/opentelemetry-configuration/blob/main/language-support-status.md

The steps I foresee:

- generate types based on json schema
- config loader using types
- use types in providers
- add a diff engine to determine what has changed
- reload only affected components

mwear:
We have two open PRs adding declarative config in different places:

- #2161: a separate `otelconfig` gem. Working implementation for configuring tracing.
- #2196: `parse`/`create` inside `SDK::Configurator`, plus constructor injection on `TracerProvider`.

I think the end state is likely a combination of the two approaches.

On placement I lean toward the separate gem. Java and Go both keep declarative config out of the core SDK, and the spec is silent on packaging.

Alongside that we need the internal restructuring from #2196, so each provider manages its children and we can pass config down through the top level components.

The third piece is our environment variable handling. The spec says that when `OTEL_CONFIG_FILE` is set, all other env vars MUST be ignored, and I don't think we can get there today, because our components read env themselves at construction. We should read env at a higher level and pass values down the same way.

For a user I'd see it looking something like:

```ruby
# common case
OpenTelemetry::Config.configure

# equivalent to
OpenTelemetry::Config.install(
  OpenTelemetry::Config.create(
    OpenTelemetry::Config.parse('otel.yaml')))
```

`install` will handle setting the globals for a user making the API more ergonomic.

We should discuss further and try to align on an approach so we can move this work forward.

thompson-tomo:
I agree we are needing dedicated gem to handle defining the config types & another gem to handle the parsing layer.

> The third piece is our environment variable handling. The spec says that when OTEL_CONFIG_FILE is set, all other env vars MUST be ignored, and I don't think we can get there today, because our components read env themselves at construction. We should read env at a higher level and pass values down the same way.

I think we need to keep the components reading their own env variables at construction however we should only trigger the reading if we pass in a nil config object as that is an indicator for no config file. It would be too hard to load the config of custom components ahead of time and force it into the right shape.

PR #2161, opened 2026-05-25
feat(chore): declarative configuration for trace_provider

xuan-cao-swi:
## Description
This PR introduces an initial declarative configuration implementation for OpenTelemetry Ruby under otelconfig, intentionally scoped to tracing only.

I followed the Go otelconfig convention for config structure and behavior where it maps well to Ruby.

Minimalist approach: no hard dependency on optional components, which optional propagators and resource detectors are resolved only if the corresponding gem has been required by the user. If an optional component is not available, configuration continues safely with warnings instead of failing startup.

Scope in this PR
- Trace configuration only:
  - tracer provider setup
  - samplers
  - span processors
  - trace exporters
  - span limits
- Declarative resource and propagation wiring needed for trace setup
- Instrumentation mapping/installation flow for declarative config
- Tests and example configuration for the implemented behavior

Try `/example`

Updated (2026-06-13):
- The in-memory struct of opentelemetry configuration now is based on https://github.com/open-telemetry/opentelemetry-configuration/tree/v1.0.0-rc.3/schema. `bundle exec rake generate:constants` will auto-generate the struct file.
- configuration will return RubySDK struct that contain providers, which user need to set the global provider point to the providers
  ```ruby
  sdk = OpenTelemetry::OtelConfig.configure
  OpenTelemetry.tracer_provider = sdk.tracer_provider
  ```


otelconfig/Rakefile
```diff
@@ -0,0 +1,31 @@
...
+
+Rake::TestTask.new :test do |t|
+  t.libs << 'test'
+  t.libs << 'lib'
+  t.test_files = FileList['test/**/*_test.rb']
+end
+
+YARD::Rake::YardocTask.new do |t|
+  t.stats_options = ['--list-undoc']
+end
```
thompson-tomo:
```suggestion
YARD::Rake::YardocTask.new
```

config/lib/opentelemetry/config/trace.rb
```diff
@@ -0,0 +1,180 @@
...
+        if exp_cfg['otlp_http']
+          configured += 1
+          exporter = build_otlp_http_span_exporter(exp_cfg['otlp_http'])
+        end
+
+        raise ArgumentError, 'must not specify multiple exporters' if configured > 1
+        raise ArgumentError, 'no valid span exporter'              if exporter.nil?
+
+        exporter
+      end
```
thompson-tomo, line 89: How will this scale as exporter is an extension point meaning additional exporters can be added?

config/lib/opentelemetry/config/trace.rb
```diff
@@ -0,0 +1,180 @@
...
+        raise ArgumentError, 'must not specify multiple span processor type' if proc_cfg['batch'] && proc_cfg['simple']
+
+        if proc_cfg['batch']
+          build_batch_span_processor(proc_cfg['batch'])
+        elsif proc_cfg['simple']
+          build_simple_span_processor(proc_cfg['simple'])
+        else
+          raise ArgumentError, 'unsupported span processor type, must be one of simple or batch'
+        end
+      end
```
thompson-tomo, line 47: How will this as scale to support additional processors the as argument error doesn't match spec.

config/lib/opentelemetry/config/trace.rb
```diff
@@ -0,0 +1,180 @@
...
+          s::ALWAYS_ON
+        elsif sampler_cfg.key?('always_off')
+          s::ALWAYS_OFF
+        elsif sampler_cfg['trace_id_ratio_based']
+          ratio = sampler_cfg['trace_id_ratio_based']['ratio'] || 1.0
+          s.trace_id_ratio_based(ratio.to_f)
+        else
+          s.parent_based(root: s::ALWAYS_ON)
+        end
+      end
```
thompson-tomo, line 123: How will this scale to support custom/additional samplers?

otelconfig/lib/opentelemetry/otelconfig/instrumentation.rb
```diff
@@ -0,0 +1,54 @@
...
+        OpenTelemetry.logger.warn('opentelemetry-instrumentation-all not available; skipping instrumentation install.')
+      end
+
+      # Transforms the YAML instrumentation config into the flat hash that
+      # install_all expects: { 'OpenTelemetry::Instrumentation::Foo' => { opt: val } }
+      def build_instrumentation_config_map(instrumentation_cfg)
+        return {} unless instrumentation_cfg.is_a?(Hash)
+
+        general = instrumentation_cfg['general']
+        return {} unless general.is_a?(Hash)
```
thompson-tomo: We should be returning the default

config/lib/opentelemetry/config/instrumentation.rb
```diff
@@ -0,0 +1,54 @@
...
+        name_map = build_instrumentation_name_map
+        general.each_with_object({}) do |(short_name, options), result|
+          full_name = name_map[short_name.to_s]
+          unless full_name
+            OpenTelemetry.logger.warn("Declarative config: unknown instrumentation short name '#{short_name}' — skipping.")
+            next
+          end
+          result[full_name] = options.is_a?(Hash) ? options.transform_keys(&:to_sym) : {}
+        end
+      end
```
thompson-tomo: What about the ruby instrumentation config options?

otelconfig/lib/opentelemetry/otelconfig/propagation.rb
```diff
@@ -0,0 +1,72 @@
...
+        composite = cfg['composite']
+        if composite.is_a?(Array)
+          return composite.flat_map { |entry| entry.is_a?(Hash) ? entry.keys : entry.to_s }
+        end
+
+        list = cfg['composite_list']
+        return list.split(',').map(&:strip) if list.is_a?(String)
+
+        []
+      end
```
thompson-tomo: This differs to the spec. The 2 properties, should be combined & duplicates removed.

config/lib/opentelemetry/config/propagation.rb
```diff
@@ -0,0 +1,72 @@
...
+          const_get_propagator('OpenTelemetry::Propagator::OTTrace')
+        when 'xray'
+          const_get_propagator('OpenTelemetry::Propagator::XRay')
+        when 'google_cloud_trace_context'
+          const_get_propagator('OpenTelemetry::Propagator::GoogleCloudTraceContext')
+        else
+          OpenTelemetry.logger.warn("Unknown propagator: #{name}")
+          nil
+        end
+      end
```
thompson-tomo, line 67: How will this scale to support custom propagators as per spec?

config/lib/opentelemetry/config/trace.rb
```diff
@@ -0,0 +1,180 @@
...
+            key, value = pair.strip.split('=', 2)
+            headers[key] = value if key && value
+          end
+        end
+
+        headers
+      end
+    end
+  end
+end
```
thompson-tomo, line 180: We seem to be missing the tracerConfig.

config/lib/opentelemetry/config/resource.rb
```diff
@@ -0,0 +1,113 @@
...
+          # Run all AWS sub-detectors; each returns an empty resource if not on that platform.
+          detect_resource('OpenTelemetry::Resource::Detector::AWS', %i[ec2 ecs eks lambda])
+        when 'azure'
+          detect_resource('OpenTelemetry::Resource::Detector::Azure')
+        when 'google_cloud_platform'
+          detect_resource('OpenTelemetry::Resource::Detector::GoogleCloudPlatform')
+        else
+          OpenTelemetry.logger.warn("OtelConfig: unknown resource detector '#{name}'; skipping.")
+          OpenTelemetry::SDK::Resources::Resource.create({})
+        end
```
thompson-tomo, line 115: How will this scale as mentioned in the spec.

otelconfig/opentelemetry-otelconfig.gemspec
```diff
@@ -0,0 +1,47 @@
...
+  spec.add_development_dependency 'opentelemetry-exporter-otlp', '~> 0.34.0'
+  spec.add_development_dependency 'opentelemetry-instrumentation-all', '~> 0.91.0'
+  spec.add_development_dependency 'opentelemetry-propagator-google_cloud_trace_context', '~> 0.4.0'
+  spec.add_development_dependency 'opentelemetry-propagator-ottrace', '~> 0.25.0'
+  spec.add_development_dependency 'opentelemetry-propagator-xray', '~> 0.27.0'
+  spec.add_development_dependency 'opentelemetry-resource-detector-aws', '~> 0.5.0'
+  spec.add_development_dependency 'opentelemetry-resource-detector-azure', '~> 0.3.0'
+  spec.add_development_dependency 'opentelemetry-resource-detector-container', '~> 0.3.0'
+  spec.add_development_dependency 'opentelemetry-resource-detector-google_cloud_platform', '~> 0.4.0'
+  spec.add_development_dependency 'opentelemetry-sdk', '~> 1.12'
```
thompson-tomo: Move to gemfile.

otelconfig/opentelemetry-otelconfig.gemspec
```diff
@@ -0,0 +1,47 @@
...
+
+  if spec.respond_to?(:metadata)
+    spec.metadata['changelog_uri'] = "https://open-telemetry.github.io/opentelemetry-ruby/opentelemetry-logs-sdk/v#{OpenTelemetry::OtelConfig::VERSION}/file.CHANGELOG.html"
```
thompson-tomo: Update as done in https://github.com/open-telemetry/opentelemetry-ruby/pull/2123

otelconfig/opentelemetry-otelconfig.gemspec
```diff
@@ -0,0 +1,47 @@
...
+    spec.metadata['source_code_uri'] = "https://github.com/open-telemetry/opentelemetry-ruby/tree/#{spec.name}/v#{spec.version}/logs_sdk"
+    spec.metadata['bug_tracker_uri'] = 'https://github.com/open-telemetry/opentelemetry-ruby/issues'
+    spec.metadata['documentation_uri'] =
```
thompson-tomo: Update as done in https://github.com/open-telemetry/opentelemetry-ruby/pull/2123

thompson-tomo (commented):
One other thing, how are we ensuring the loading of the the env variables are disabled when covered by declarative config?

Will the current approach support dynamic reloading of the config which is also a part of the spec?

config/lib/opentelemetry/config/trace.rb
xuan-cao-swi: It doesn't. Let's keep this simple and comply with what the current Ruby trace SDK can do.

config/lib/opentelemetry/config/trace.rb
xuan-cao-swi: It's better to include a link or reference to the spec so people can check it quickly. If you are referring to [SDK extension components](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#sdk-extension-components), then it's an SDK-level feature. The Ruby trace SDK doesn't have this feature.

To answer your questions about custom components:
1. It's an SDK-level feature (e.g., [ComponentProvider.java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure-spi/src/main/java/io/opentelemetry/sdk/autoconfigure/spi/internal/ComponentProvider.java), [PHP ComponentProvider](https://github.com/open-telemetry/opentelemetry-php/tree/main/src/Config/SDK/ComponentProvider)). It also requires language-level support for the Service Provider Interface, which Ruby (and Go) do not provide.
> Given the inherent differences across languages, the details of extension
component mechanisms are likely to vary to a greater degree than is the case
with other APIs defined by OpenTelemetry. This is to be expected and is
acceptable so long as the implementation results in the defined behaviors.
2. Even if someone found a way to support it, to keep this PR minimal (already 27 files), it's better to keep it simple.

otelconfig/lib/opentelemetry/otelconfig/instrumentation.rb
xuan-cao-swi: `OpenTelemetry::Instrumentation.registry.install_all({})` is the default

config/lib/opentelemetry/config/instrumentation.rb
xuan-cao-swi: `build_instrumentation_config_map` method turns:
```yaml
instrumentation/development:
  general:
    net_http:
      untraced_hosts:
        - google.ca
```
into:
```
{ 'OpenTelemetry::Instrumentation::Net::HTTP' => { :untraced_hosts => 'google.ca' } }
```

config/lib/opentelemetry/config/trace.rb
xuan-cao-swi: What is `tracerConfig`? Could you be more specific?

thompson-tomo: It is one of the types defined at https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentaltracerconfigurator-

config/lib/opentelemetry/config/instrumentation.rb
thompson-tomo: Yes but what about

```yaml
instrumentation/development:
  ruby:
    net_http:
      untraced_hosts:
        - google.ca
```

otelconfig/lib/opentelemetry/otelconfig/instrumentation.rb
thompson-tomo: Yes that installs all instrumentation, I am more talking about the default options as defined in declarative config aka within https://github.com/open-telemetry/opentelemetry-configuration/blob/main/schema-docs.md#experimentalgeneralinstrumentation-.

config/lib/opentelemetry/config/trace.rb
thompson-tomo: I am all for keeping it simple, small and manageable while also ensuring that we are doing something which scales and will serve us long term.

For instance could we not load declarative config into the global options after parsing/validating the file, inject an options node into the providers which uses them accordingly. The process of inject & then using them is repeated.

In effect each component is responsible for getting their own options, hence it avoids all of the switch options and decouples config from implementations. It also means we can follow the same path regardless of how it is configured ie env or file.

xuan-cao-swi: #### "Could we not load declarative config into the global options after parsing/validating the file, inject an options node into the providers which uses them accordingly."

I think what you're describing is something like:

```ruby
# 1. Parse into global store
GlobalOptions.config = parse_config_file(path)

# 2. Each provider pulls its own slice
TracerProvider.configure(GlobalOptions.config['tracer_provider'])
MeterProvider.configure(GlobalOptions.config['meter_provider'])
```

This isn't actually less coupling — it just moves it. Each provider still needs to know its own key name and schema. That knowledge gets scattered across multiple files instead of being in one place (`apply`). If the YAML schema changes (e.g. a key is renamed), you'd have to track down every component that reads it rather than fixing one place.

Java actually tried a global config store and then [deleted it](https://github.com/open-telemetry/opentelemetry-java/commit/d22413ee8e02b271c44b3eda8363672a90518a9e). The comment in `AutoConfiguredOpenTelemetrySdk.java` reflects the resulting separation:
> *"Returns the `ConfigProperties` used for auto-configuration, or `null` if declarative configuration was used."*

It also contradicts the [Configuration SDK spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk/#sdk-operations), which defines exactly two operations — `Parse` (file → in-memory model) and `Create` (model → all SDK components at once) — with `Create` explicitly centralized:

```
OpenTelemetry openTelemetry = OpenTelemetry.noop();
try {
    // Parse configuration file to configuration model
    OpenTelemetryConfiguration configurationModel = parse(new File("/app/sdk-config.yaml"));
    // Create SDK components from configuration model
    openTelemetry = create(configurationModel);
} catch (Throwable e) {
    log.error("Error initializing SDK from configuration file", e);
}

// Access SDK components and install instrumentation
TracerProvider tracerProvider = openTelemetry.getTracerProvider();
MeterProvider meterProvider = openTelemetry.getMeterProvider();
LoggerProvider loggerProvider = openTelemetry.getLogsBridge();
ContextPropagators propagators = openTelemetry.getPropagators();
ConfigProvider configProvider = openTelemetry.getConfigProvider();
```

Both [Java](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/src/main/java/io/opentelemetry/sdk/autoconfigure/AutoConfiguredOpenTelemetrySdkBuilder.java#L515) and [Go](https://github.com/open-telemetry/opentelemetry-go-contrib/blob/main/otelconf/config.go#L98) follow the same centralized `parse → create` pattern for this reason.

#### "In effect each component is responsible for getting their own options, hence it avoids all of the switch options and decouples config from implementations. It also means we can follow the same path regardless of how it is configured ie env or file."

There are no switch statements in the current `apply` — it's sequential method calls. The branching logic lives inside each component's builder (e.g. `Trace.build_tracer_provider` choosing which exporter to create). That logic has to live somewhere; moving it doesn't remove it.

> Decouples config from implementations

It inverts the coupling rather than removing it. Currently OtelConfig is the single place that knows about component names and config keys. Under the proposed pattern, that knowledge spreads into every component. Centralised is easier to maintain.

> Follow the same path regardless of env or file

The spec explicitly rejects this. From the [env var spec](https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/#declarative-configuration):

> *"When `OTEL_CONFIG_FILE` is set, all other environment variables **MUST be ignored** … there is no intuitive way to merge the flat environment variable scheme with the structured file configuration scheme."*

The two paths are intentionally separate. Env vars are flat strings; file config is a typed, structured, nested object. A provider trying to read from "wherever config lives" would need to handle both formats — that's more complexity inside the component, not less.

config/lib/opentelemetry/config/trace.rb
xuan-cao-swi: maybe focus on stable first? since `Maintainers are not obligated to implement support for experimental properties and types. `

config/lib/opentelemetry/config/instrumentation.rb
xuan-cao-swi: mistake, `general:` should be `ruby:` for language-specific lib instrumentation.

otelconfig/lib/opentelemetry/otelconfig/instrumentation.rb
xuan-cao-swi: I won't worry about it until the contrib semantic conventions is stable

config/lib/opentelemetry/config/trace.rb
thompson-tomo: Thanks for the detailed reply I however am still very worried that the current approach will not scale and makes it harder for us to be fully compliant with the spec without refactoring the newly implemented functionality.

I have started sketching my thoughts with an emphasis on spec & progressively adding declarative config. See https://github.com/thompson-tomo/opentelemetry-ruby/pull/44 as a starting point.

Key thing is the configure method calls parse with the intention of this returning a typed in-memory configuration model as per the spec.

The configuration model is then passed to the create call as per the spec. Within the create, the providers are then created using the scoped config section see propagators as an example. If no config then load environment variables. The create method then returns the created providers.

Further looking at https://opentelemetry.io/docs/specs/otel/configuration/sdk/#register-plugincomponentprovider the providers should then create components ie processors passing in the config https://opentelemetry.io/docs/specs/otel/configuration/sdk/#create-component

config/lib/opentelemetry/config/instrumentation.rb
thompson-tomo: Reading the spec more, shouldn't this be in fact handled by ConfigProvider https://opentelemetry.io/docs/specs/otel/configuration/api/

config/lib/opentelemetry/config/trace.rb
thompson-tomo: That is fine but not in other spots we have implemented experimental.

config/lib/opentelemetry/config/trace.rb
xuan-cao-swi: 1. It's not the best idea to touch SDK(s) when implementing otelconfig. It should be a completely noninvasive implementation (check Go, Java, PHP implementations).
2.  `configure method calls parse with the intention of this returning a typed in-memory`; this is true, parse should return a typed in-memory structure rather than a simple hash (to avoid schema changes causing unexpected issues); also apply shouldn't set the global trace_provider directly, it should just return the constructed providers (will fix that)
3. `providers should then create components ie processors passing in the config`; again, this plugincomponentprovider is only for declarative config, and without Java's SPI or PHP's Composer autoloading for automatic registration, it's not one-prompt AI work to do.
4. I want to keep the initial otelconf implementation simple and work on what already exists. If you are truly concerned, please submit your work for review. I will keep this approach and not worry about extension.

thompson-tomo: > It should be a completely noninvasive implementation (check Go, Java, PHP implementations).

Yes and this is based on https://opentelemetry.io/docs/specs/otel/configuration/

> The SDK MUST provide a programmatic interface for all configuration..... All other configuration mechanisms SHOULD be built on top of this interface.

This is what the java implementation does by reading the config and then passing scoped blocks to the sdk builders. In ruby we don't have builders but instead have initializers.

> plugincomponentprovider is only for declarative config

Where do you see this as the otel spec defines it as:

> responsible for interpreting configuration and returning an implementation of a particular type of SDK plugin component.

And based on above quote programmatic configuration is the basis for all other providers, hence I see plugincomponentprovider as a configuration concept & not declarative config.

> I want to keep the initial otelconf implementation simple and work on what already exists.

I am not saying this PR needs to support extensions but rather the approach needs to be a building block which can support adding extensions without requiring a re-write.

It also needs to be considered that declarative has only recent reached stable hence that also informs why some of the previous decisions were. A key thing which has been pointed out to me is the fields are based on their inclusion in the api.

config/lib/tasks/generate_constants.rake
```diff
@@ -0,0 +1,248 @@
...
+    assignments = props.map do |key, prop|
+      "      #{field_name(key)}: #{value_expr(prop, defs, key)}"
+    end
+    if capture_additional
+      known = props.keys.map { |k| "'#{k}'" }.join(', ')
+      assignments << "      additional_properties: h.reject { |k, _| [#{known}].include?(k) }"
+    end
+
+    <<~RUBY
+      #{name} = Struct.new(
```
mwear, line 175: This generates constants in the top-level namespace. This could cause conflicts in a host application if they have constants of the same name. We should namespace these under `OpenTelemetry::OtelConfig::Model` or something similar.

config/opentelemetry-config.gemspec
```diff
@@ -0,0 +1,33 @@
...
+
+  spec.summary     = 'Declare Config implementation for OpenTelemetry'
+  spec.homepage    = 'https://github.com/open-telemetry/opentelemetry-ruby'
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 3.3'
```
mwear, line 25: We should have dependencies on the required gems here. At a minimum we need: `opentelemetry-api`, `opentelemetry-sdk`, and `opentelemetry-common`, and possibly `opentelemetry-exporter-otlp`.

otelconfig/lib/opentelemetry/otelconfig.rb
```diff
@@ -0,0 +1,81 @@
...
+
+      # Configure directly from a file path (for testing or explicit setup).
+      def configure_from_file(path)
+        config = parse_config_file(path)
+        apply(config)
+      end
+
+      private
+
+      def apply(config)
```
mwear: This only returns an SDK on the happy path. Other paths vary in what they return. We should define what happens on these other paths? Do we want to always return an SDK backed by no-op components or something else?

config/lib/opentelemetry-config.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
```
mwear, line 1: We should rename this file `opentelemetry-otelconfig.rb` so that it matches the convention used in the rest of our gems.

otelconfig/README.md
```diff
@@ -0,0 +1,174 @@
...
+
+Set `OTEL_CONFIG_FILE` to the path of your YAML config file. Call `OpenTelemetry::OtelConfig.configure` early in your application; it returns a `RubySDK` value that you wire into the global OpenTelemetry state yourself.
+
+```sh
+OTEL_CONFIG_FILE=/path/to/otel-config.yaml bundle exec ruby app.rb
+```
+
+```ruby
+require 'opentelemetry-sdk'
+require 'opentelemetry-otelconfig'
```
mwear: This require will currently fail. We need the rename suggested previously.

mwear (commented):
I like the approach overall. I have a few comments based on a first pass through.

mwear (commented):
Sorry, I have been slow to review the declarative config PRs. I notice we also have #21960 proposing a slightly different shape for declarative config. We should discuss further in #2127 and see if we can align on an approach.

config/lib/tasks/generate_constants.rake
xuan-cao-swi: Updated to use `OpenTelemetry::OtelConfig::Model`

config/opentelemetry-config.gemspec
xuan-cao-swi: Added back

otelconfig/lib/opentelemetry/otelconfig.rb
xuan-cao-swi: Now it will return noop sdk (e.g. noop provider and propagator) if parsing failed

config/lib/opentelemetry-config.rb
xuan-cao-swi: Updated

otelconfig/lib/opentelemetry/otelconfig.rb
xuan-cao-swi: Now it will also init the otel sdk for user by replacing the global provider and propagator.

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,101 @@
...
+
+module OpenTelemetry
+  # OtelConfig module handles declarative configuration of OpenTelemetry components
+  # from YAML files.
+  module OtelConfig
+    ENV_CONFIG_FILE = 'OTEL_CONFIG_FILE'
+
+    class << self
+      # Entry point
+      def configure
```
mwear, line 26: The [specification](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#sdk-operations) says that there should be `parse` and `create` operations. Right now we only expose `configure` and `confgure_from_file`. `configure` has the right behavior, but I would prefer that we build it as suggested in #2127.

```ruby
sdk = OpenTelemetry::Config.configure
# ^^ Equivalent to vv
sdk = OpenTelemetry::Config.install(
        OpenTelemetry::Config.create(
          OpenTelemetry::Config.parse(ENV['OTEL_CONFIG_FILE'])))
          
# the SDK handle ideally should be usable for shutdown
at_exit { sdk.shutdown }
```

- `parse` should return the model, currently this is `parse_config_file`
- `create` should return a configured `RubySDK`
- `install` should take a `RubySDK` and assign it the proper globals
- `configure` is a convenience method that wraps all three operations, should return a `RubySDK`

The current `apply` implementation is `create` + `install` fused. It should be split.

Not spec'd but would be nice and mirrors what Go does: the `RubySDK` should expose a `shutdown` method that fans out to the configured providers.

otelconfig/opentelemetry-otelconfig.gemspec
```diff
@@ -0,0 +1,38 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+lib = File.expand_path('lib', __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'opentelemetry/otelconfig/version'
+
+Gem::Specification.new do |spec|
+  spec.name        = 'opentelemetry-otelconfig'
```
mwear: I would prefer to see this named: `opentelemetry-config` or `opentelemetry-configuration`.

mwear (commented):
If the comment regarding `parse`, `create`, `install` is addressed, then I think this has the overall shape and behavior that we want and we can consider this as our MVP and handle additional functionality incrementally.

The constructor injection that is proposed in #2196 can be handled as an internal refactor. However, we should not allow the configuration model to leak into the SDK.

There were comments suggesting that we should have support for custom components, which I agree with in spirit. There is likely a path forward using [`Class#inherited`](https://ruby-doc.org/3.4.1/Class.html#method-i-inherited) to build a registry of components. However there will be startup, initialization, load-order complications that need to be worked through. The spec currently lists [extensions](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/configuration/sdk.md#sdk-extension-components) as a SHOULD, so this does not need to be included in the MVP.

Environment variable handling is another thing we're going to have to work on, but it can also be handled separately.

config/lib/opentelemetry/config.rb
xuan-cao-swi: Thanks, configure now is separated to parse, create and install.

xuan-cao-swi:
> Environment variable handling is another thing we're going to have to work on, but it can also be handled separately.

I think eventually env var will be abandoned if user specifically choose to use declarative config. There are migration from env var ([otel-sdk-migration-config.yaml](https://github.com/open-telemetry/opentelemetry-configuration/blob/main/examples/otel-sdk-migration-config.yaml))

> There were comments suggesting that we should have support for custom components

The only concern I have is that user have to install and load their custom components first (before declarative config) then the declarative config can recognize them, which means extra steps for user.

otelconfig/opentelemetry-otelconfig.gemspec
xuan-cao-swi: change the folder name from `otelconfig` to `config` as well?

otelconfig/lib/opentelemetry/otelconfig.rb
```diff
@@ -0,0 +1,118 @@
...
+        end
+
+        resource = build_resource(config.resource)
+
+        # tracer_provider will be noop if opentelemetry-sdk is not installed
+        tracer_provider = Trace.build_tracer_provider(config.tracer_provider, resource)
+
+        propagators = configure_propagation(config.propagator)
+
+        configure_instrumentation(config.instrumentation_development)
```
mwear: `create` should only construct the SDK components. Instrumentation installation should happen in `install` instead.

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,118 @@
...
+          resource: resource
+        )
+      end
+
+      # Assigns the components of a RubySDK to the global OpenTelemetry state.
+      #
+      # @param ruby_sdk [RubySDK]
+      # @return [RubySDK] the same SDK handle
+      def install(ruby_sdk)
+        OpenTelemetry.tracer_provider = ruby_sdk.tracer_provider if ruby_sdk.tracer_provider
```
mwear, line 101: `install` should skip the no-op SDK. As written, it will set the globals to the no-op components since the returned no-op providers are truthy. This is a change in behavior from before the `apply` split, which left the globals as-is. 

The cleanest fix is a frozen `NOOP_SDK` constant returned from every no-op path, with install short circuiting when `ruby_sdk.equal?(NOOP_SDK)`.

otelconfig/lib/opentelemetry/otelconfig.rb
xuan-cao-swi: updated: `create` will only construct SDK conponents and instrumentation installation will only happened in `install`

otelconfig/opentelemetry-otelconfig.gemspec
xuan-cao-swi: changed folder name to `config`

config/lib/opentelemetry/config.rb
xuan-cao-swi: Updated: every no-op path will return `NOOP_SDK` constant and will be skipped from `install`

config/lib/opentelemetry/config/instrumentation.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # Config module — instrumentation configuration helpers.
+  module Config
+    class << self
```
mwear: These three methods should be private. I didn't add a suggestion because tests use these methods and will need to be updated.

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,110 @@
...
+require_relative 'config/instrumentation'
+require_relative 'config/propagation'
+require_relative 'config/resource'
+require_relative 'constants/constants'
+
+module OpenTelemetry
+  # Config module handles declarative configuration of OpenTelemetry components
+  # from YAML files.
+  module Config
+    ENV_CONFIG_FILE = 'OTEL_CONFIG_FILE'
```
mwear, line 18:
```suggestion
    ENV_CONFIG_FILE = 'OTEL_CONFIG_FILE'
    private_constant :ENV_CONFIG_FILE
```

config/lib/opentelemetry/components/trace.rb
```diff
@@ -0,0 +1,178 @@
...
+        if headers.empty? && cfg.headers_list.is_a?(String)
+          cfg.headers_list.split(',').each do |entry|
+            key, value = entry.strip.split('=', 2)
+            headers[key] = value if key && value
+          end
+        end
+
+        headers
+      end
+    end
```
mwear, line 176:
```suggestion
    end

    private_constant :Trace
```

config/lib/opentelemetry/config/resource.rb
```diff
@@ -0,0 +1,127 @@
...
+
+        OpenTelemetry.logger.warn('Config: schema_url is supported; ignoring.') if resource_cfg.schema_url
+
+        attrs = detected.merge(explicit)
+        custom = OpenTelemetry::SDK::Resources::Resource.create(attrs)
+        base.merge(custom)
+      end
+
+      private
```
mwear: `build_resources` should be private

```suggestion
      private

      # Priority: attributes > attribute_list > detected > base
      def build_resource(resource_cfg)
        base = OpenTelemetry::SDK::Resources::Resource.default

        return base unless resource_cfg

        detected = build_detected_attributes(resource_cfg.detection_development)

        explicit = {}
        Array(resource_cfg.attributes).each do |attr|
          next unless attr.name && !attr.value.nil?

          explicit[attr.name] = coerce_attribute_value(attr.value, attr.type)
        end

        if resource_cfg.attributes_list.is_a?(String)
          resource_cfg.attributes_list.split(',').each do |pair|
            key, value = pair.strip.split('=', 2)
            explicit[key] ||= value if key && value
          end
        end

        OpenTelemetry.logger.warn('Config: schema_url is supported; ignoring.') if resource_cfg.schema_url

        attrs = detected.merge(explicit)
        custom = OpenTelemetry::SDK::Resources::Resource.create(attrs)
        base.merge(custom)
      end

```

config/lib/opentelemetry/config/propagation.rb
```diff
@@ -0,0 +1,76 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
+module OpenTelemetry
+  # Config module — propagation configuration helpers.
+  module Config
+    class << self
```
mwear, line 9:
```suggestion
    class << self
      private
```

mwear (commented):
Thanks for the changes so far. Going over this with a fine-toothed comb, I have a few more changes to ask.

Can we move three files so that their layout matches their namespace. This would align with our conventions and prevent potential require / loadpath issues in the future.

```
lib/opentelemetry/components/trace.rb              -> lib/opentelemetry/config/trace.rb
lib/opentelemetry/constants/constants.rb           -> lib/opentelemetry/config/constants.rb
lib/opentelemetry/constants/generated_constants.rb -> lib/opentelemetry/config/generated_constants.rb
```

I also want to lock down the public API. I made line comments below with suggestions where applicable.

The last thing, which would be fine as a followup, is to wire up this gem for CI and releases.

config/lib/opentelemetry/config/instrumentation.rb
xuan-cao-swi: Made it private

xuan-cao-swi:
> Can we move three files so that their layout matches their namespace

Updated.

> wire up this gem for CI and releases.

Once it's good to go, I can create another pr to add this gem into release workflow

mwear (approved):
Thanks @xuan-cao-swi!

config/example/app.rb
```diff
@@ -0,0 +1,62 @@
...
+require 'net/http'
+require 'opentelemetry-sdk'
+require 'opentelemetry-instrumentation-all'
+require 'opentelemetry-config'
+
+sdk = OpenTelemetry::Config.install(
+  OpenTelemetry::Config.create(
+    OpenTelemetry::Config.parse(ENV.fetch('OTEL_CONFIG_FILE', nil))
+  )
+)
```
kaylareopelle: So does this have to be added in lieu of the SDK.configure block?

config/README.md
```diff
@@ -0,0 +1,192 @@
...
+
+## How does this gem fit in?
+
+The `opentelemetry-config` gem sits on top of the OpenTelemetry Ruby SDK. Instead of calling `OpenTelemetry::SDK.configure` with a block of Ruby code, you describe your desired configuration in a YAML file and let `opentelemetry-config` wire up all the opentelemetry components for you.
+
+It works with:
+
+- `opentelemetry-sdk` — tracing
+- `opentelemetry-exporter-otlp` — OTLP HTTP exporters
+- `opentelemetry-instrumentation-all` — auto-instrumentation
```
kaylareopelle:
```suggestion
- `opentelemetry-instrumentation-all` — all instrumentation gems bundled into a metagem
```

config/README.md
```diff
@@ -0,0 +1,192 @@
...
+A runnable example application is available in the [`example/`][example-dir] directory. It demonstrates traces configured from YAML with console output.
+
+```sh
+cd config/example
+bundle exec ruby app.rb
+```
+
+## How can I get involved?
+
+The `opentelemetry-config` gem source is [on github][repo-github], along with related gems including `opentelemetry-sdk`.
```
kaylareopelle:
```suggestion
The `opentelemetry-config` gem source is [on GitHub][repo-github], along with related gems including `opentelemetry-sdk`.
```

config/opentelemetry-config.gemspec
```diff
@@ -0,0 +1,38 @@
...
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'opentelemetry/config/version'
+
+Gem::Specification.new do |spec|
+  spec.name        = 'opentelemetry-config'
+  spec.version     = OpenTelemetry::Config::VERSION
+  spec.authors     = ['OpenTelemetry Authors']
+  spec.email       = ['cncf-opentelemetry-contributors@lists.cncf.io']
+
+  spec.summary     = 'Declare Config implementation for OpenTelemetry'
```
kaylareopelle:
```suggestion
  spec.summary     = 'Declarative Config implementation for OpenTelemetry'
```

config/opentelemetry-config.gemspec
```diff
@@ -0,0 +1,38 @@
...
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 3.3'
+
+  spec.add_dependency 'opentelemetry-api', '~> 1.11.0'
+  spec.add_dependency 'opentelemetry-common', '~> 0.25.1'
+  spec.add_dependency 'opentelemetry-exporter-otlp', '~> 0.34.1'
+  spec.add_dependency 'opentelemetry-sdk', '~> 1.13.0'
```
kaylareopelle, line 30: I think it's good to list these as dependencies based on the code in this gem currently, but eventually, I think we'll want to remove them. 

* `opentelemetry-exporter-otlp` - even though most users are using this for export, there may be some environments where they do not want to, but will not be required to if they use declarative config
* `opentelemetry-sdk` - if/when we implement the Instrumentation Config API, we do not want to make those calls dependent on including an implementation of the SDK

We also don't have requires right now for either of these gems. This can cause an error if the SDK is required after the `Config.install` call. There's a `NOOP_SDK` guard in the setup to protect against when the SDK isn't installed. It may not be relevant anymore.

config/CHANGELOG.md
kaylareopelle, line 1: I think Toys usually expects the header for the gem in this file before we have any releases, but it might generate it entirely. I'm not sure.

config/README.md
kaylareopelle:
```suggestion
- `opentelemetry-sdk` — Traces
- `opentelemetry-exporter-otlp` — OTLP HTTP exporter
- `opentelemetry-instrumentation-all` — Instrumentation for gems
```

config/README.md
```diff
@@ -0,0 +1,192 @@
...
+## How does this gem fit in?
+
+The `opentelemetry-config` gem sits on top of the OpenTelemetry Ruby SDK. Instead of calling `OpenTelemetry::SDK.configure` with a block of Ruby code, you describe your desired configuration in a YAML file and let `opentelemetry-config` wire up all the opentelemetry components for you.
+
+It works with:
+
+- `opentelemetry-sdk` — tracing
+- `opentelemetry-exporter-otlp` — OTLP HTTP exporters
+- `opentelemetry-instrumentation-all` — auto-instrumentation
+
```
kaylareopelle, line 20:
```suggestion

This code is still under development and is not a complete implementation of the declarative configuration specification. Until the code becomes stable, declarative configuration functionality will live outside stable OpenTelemetry libraries.

Some features we still need to implement include:
* Configuration file precedence over environment variables
* Using declarative configuration without a call in your code
* Declarative configuration instrumentation API and SDK

This is not an exhaustive list.
```

config/lib/opentelemetry/config/instrumentation.rb
kaylareopelle, line 1: Based on our discussion during the SIG today, please remove the instrumentation-related code from this PR so we can discuss how we want to implement it/if we want to propose any changes to the spec.

config/lib/opentelemetry/config/resource.rb
```diff
@@ -0,0 +1,127 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
```
kaylareopelle: I think a missing require error for 'opentelemetry-sdk' is being swallowed here by requiring it in our tests and our example along with the config.


```suggestion

require 'opentelemetry-sdk'

```

config/lib/opentelemetry/config/trace.rb
```diff
@@ -0,0 +1,180 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+# SPDX-License-Identifier: Apache-2.0
+
```
kaylareopelle: I think we're missing a require here for the OTLP exporter.
```suggestion

require 'opentelemetry-exporter-oltp'

```

config/example/README.md
```diff
@@ -0,0 +1,45 @@
...
+file using the `opentelemetry-config` gem — no programmatic
+`OpenTelemetry::SDK.configure` block required.
+
+## Files
+
+| File | Purpose |
+| ---- | ------- |
+| `app.rb` | Example application — emits spans |
+| `otel-config-console.yaml` | console-only exporter, works without a collector |
+| `otel-config.yaml` | Include otlp_http exporter, need working collector |
```
kaylareopelle:
```suggestion
| `otel-config.yaml` | Includes otlp_http exporter, need a working collector |
```

config/example/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Quick start (console output, no collector needed)
```
kaylareopelle, line 15: Should we add collector setup instructions too for OTLP HTTP export?

config/example/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+1. Set the `OTEL_CONFIG_FILE` environment variable to the path of your YAML file.
+2. `require 'opentelemetry-config'` reads the file, parses it, and wires
+   up `TracerProvider`, propagators, and instrumentation — all in one step.
+3. Use the standard OpenTelemetry API (`OpenTelemetry.tracer_provider`) as normal.
+
+If `OTEL_CONFIG_FILE` is not set, call `OpenTelemetry::Config.configure`
+manually with a config hash, or configure programmatically using the SDK.
+
+## YAML config key reference
```
kaylareopelle: Is there public documentation we can link to with similar information? I like having the table here for handy reference. Mostly curious as a jumping-off point to look at more configuration.

config/lib/tasks/generate_constants.rake
```diff
@@ -0,0 +1,268 @@
...
+
+require 'fileutils'
+require 'open-uri'
+require 'tmpdir'
+require 'zlib'
+require 'stringio'
+require 'yaml'
+require 'rubygems/package'
+
+SCHEMA_VERSION = ENV.fetch('SCHEMA_VERSION', 'v1.0.0-rc.3')
```
kaylareopelle: Do you think we'll eventually want to connect this to a github workflow, like we have for semantic conventions to have the CI manage updates?

config/opentelemetry-config.gemspec
kaylareopelle: We also don't have any requires right now for `opentelemetry-exporter-otlp` or `opentelemetry-sdk`. I think we should add requires before merging this PR. The issues were silently swallowed by requiring the SDK in our tests and our example app.

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,111 @@
...
+        end
+
+        if config.disabled
+          OpenTelemetry.logger.info('OpenTelemetry SDK disabled by configuration.')
+          return NOOP_SDK
+        end
+
+        resource = build_resource(config.resource)
+
+        # tracer_provider will be noop if opentelemetry-sdk is not installed
```
kaylareopelle, line 81: Is this still accurate if we're installing `opentelemetry-sdk` as a dependency?

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,111 @@
...
+      # @return [RubySDK] {NOOP_SDK} when the configuration is absent, invalid,
+      #   disabled, or opentelemetry-sdk isn't loaded
+      def create(config)
+        return NOOP_SDK if config.nil?
+
+        unless defined?(OpenTelemetry::SDK)
+          warn '[opentelemetry-config] opentelemetry-sdk is not loaded. ' \
+               'Add `gem "opentelemetry-sdk"` to your Gemfile.'
+          return NOOP_SDK
+        end
```
kaylareopelle: Do we need this if `opentelemetry-sdk` is a dependency? Do we want to remove the dependency on the SDK since we have this protection?

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,111 @@
...
+      # modifying global OpenTelemetry state.
+      #
+      # @param config [Model::OpenTelemetryConfiguration, nil]
+      # @return [RubySDK] {NOOP_SDK} when the configuration is absent, invalid,
+      #   disabled, or opentelemetry-sdk isn't loaded
+      def create(config)
+        return NOOP_SDK if config.nil?
+
+        unless defined?(OpenTelemetry::SDK)
+          warn '[opentelemetry-config] opentelemetry-sdk is not loaded. ' \
```
kaylareopelle: Would using `OpenTelemetry.logger.warn` make sense here?

config/lib/opentelemetry/config.rb
```diff
@@ -0,0 +1,111 @@
...
+        end
+
+        resource = build_resource(config.resource)
+
+        # tracer_provider will be noop if opentelemetry-sdk is not installed
+        tracer_provider = Trace.build_tracer_provider(config.tracer_provider, resource)
+
+        propagators = configure_propagation(config.propagator)
+
+        RubySDK.new(
```
kaylareopelle: For whatever reason the RubySDK name is throwing me off. Would calling it SDKConfig work? Or some other name that doesn't include Ruby? Happy to be overruled here too if this is the best name @xuan-cao-swi @mwear

config/README.md
```diff
@@ -0,0 +1,192 @@
...
+sdk = OpenTelemetry::Config.install(
+  OpenTelemetry::Config.create(
+    OpenTelemetry::Config.parse(ENV.fetch('OTEL_CONFIG_FILE', nil))
+  )
+)
+```
+
+## YAML configuration reference
+
+See full configuration reference in [declarative-configuration](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/)
```
kaylareopelle:
```suggestion
See full configuration reference in [declarative-configuration](https://opentelemetry.io/docs/languages/sdk-configuration/declarative-configuration/).
```

kaylareopelle (changes requested):
I think this is a great first step toward setting up declarative configuration for this repository.

I'm going to mark this review as "Request changes" based on our discussion in the SIG today to remove the instrumentation code from this implementation. Most of the comments/suggestions are minor text/wording things.

I also have a few overlapping comments/questions about requiring the SDK and the OTLP Exporter.

config/CHANGELOG.md
xuan-cao-swi: Added the header for CHANGELOG.md

config/lib/opentelemetry/config/instrumentation.rb
xuan-cao-swi: Removed

config/example/README.md
xuan-cao-swi: I think they can get the detailed instruction for custom collector from https://opentelemetry.io/docs/collector/quick-start/

config/opentelemetry-config.gemspec
xuan-cao-swi: Originally they are not there, but someone required to have these dependencies in gemspec.
I am ok with either way but favour not having these dependencies for cleaner setup.

kaylareopelle: Thank you for the context. Let's keep this in mind and revisit it in a separate issue.

kaylareopelle (approved):

kaylareopelle:
@xuan-cao-swi - Just thought of one other thing, I don't think the tests for config are running in the CI. Could you update the CI?

kaylareopelle:
I ran the tests locally and it looks like we're around 75% line coverage, but otherwise things are passing. let's add the CI in a separate PR and get this initial work merged.

PR #2167, opened 2026-05-29
Add proxy_address to OTLP Exporter

### LINKED ISSUE #2166: Feature Request: Allow setting an HTTP proxy on OTLP Exporters
https://github.com/open-telemetry/opentelemetry-ruby/blob/245aea57aec3fb362ffeaf355fef76abefa3da31/exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb#L127

We need to be able to configure different exporters to use different proxy configurations. Environment variables are not sufficient.

SpamapS:
This allows more fine-graned control over proxy configurations of various exporters.

Fixes #2166


SpamapS:
I will hopefully get IBM's CLA linked to my account shortly.

exporter/otlp/lib/opentelemetry/exporter/otlp/exporter.rb
```diff
@@ -53,12 +53,14 @@ def initialize(endpoint: nil,
                        headers: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_HEADERS', 'OTEL_EXPORTER_OTLP_HEADERS', default: {}),
                        compression: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_COMPRESSION', 'OTEL_EXPORTER_OTLP_COMPRESSION', default: 'gzip'),
                        timeout: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_TIMEOUT', 'OTEL_EXPORTER_OTLP_TIMEOUT', default: 10),
-                       metrics_reporter: nil)
+                       metrics_reporter: nil,
+                       proxy_address: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_PROXY_ADDRESS', 'OTEL_EXPORTER_OTLP_PROXY_ADDRESS'),
+                       proxy_port: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPORTER_OTLP_TRACES_PROXY_PORT', 'OTEL_EXPORTER_OTLP_PROXY_PORT'))
```
thompson-tomo, line 60: This is based on https://opentelemetry.io/docs/specs/otel/protocol/exporter/ so unfortunately I don't think it is that simple to maintain compatability with the spec.

thompson-tomo, line 60: I don't see either of these environment variables listed at https://opentelemetry.io/docs/specs/otel/configuration/sdk-environment-variables/

thompson-tomo (changes requested):
Thanks for the suggestion however unfortunately we are governed by the spec in terms of api design and I can't see any references to supporting proxy config there. What I would suggest is first testing the native `http_proxy` env variable which based on some searching should work as per below:

> ENV['http_proxy'] = 'http://172.16.3.160:4226' # your http://address:port here

And if that fails, raising a spec issue to discuss adding the ability to proxy address.

SpamapS:
Thanks so much for the reply!

The problem is that we need a way to explicitly configure whether or not to
use which proxy, rather than the implicit, all or nothingmethod those
environment variables offer. We have several different competing reasons
for this unfortunately.

I'll take a look at the spec closely and see where we can plumb it in via a
compliant method.

xuan-cao-swi:
If you prefer not to use the implicit HTTP proxy support provided by the standard `net/http` library (through env [http_proxy](https://ruby-doc.org/stdlib-2.4.3/libdoc/net/http/rdoc/Net/HTTP.html#class-Net::HTTP-label-Proxies)), maybe you can try to replace the [`@http`](https://github.com/open-telemetry/opentelemetry-ruby/blob/opentelemetry-exporter-otlp/v0.34.0/exporter/otlp-http/lib/opentelemetry/exporter/otlp/http/trace_exporter.rb#L51) instance variable as an alternative approach, which is similar to what python exporter supports custom sessions as an exporter parameter, the Ruby exporter does not currently support this feature (nor is it required by the OpenTelemetry specification).

```ruby
e.g. hacky way
exporter = OpenTelemetry::Exporter::OTLP::Exporter.new
exporter.instance_variable_set(:@http, Net::HTTP.new(...))
```

kaylareopelle:
Hi @SpamapS, thanks for opening this PR! I wanted to check in on the CLA progress. How are things going?

PR #2185, opened 2026-06-10
ci: Automate instant semconv release

thompson-tomo:
This adds in an instant release pr when semconv spec version has been bumped.

this avoids potential scenarios where we have merged a new semconv version and then a new semconv version is released before our weekly pr has been merged.

note it could be a separate workflow but given it is almost purely additive the addition path was chosen to reduce maintenance effort going forward.

### DIFF
--- a/semantic_conventions/semconv.yaml
+++ b/semantic_conventions/semconv.yaml
@@ -1,0 +1,1 @@
+spec_version: '1.40.0'
--- a/semantic_conventions/semconv.yaml
+++ b/semantic_conventions/semconv.yaml
@@ -1,1 +1,1 @@
-spec_version: '1.40.0'
+spec_version: '1.42.0'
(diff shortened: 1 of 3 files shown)


semantic_conventions/semconv.yaml
```diff
@@ -0,0 +1 @@
+spec_version: '1.40.0'
```
kaylareopelle: Does this need to be the current version before we merge?

kaylareopelle (commented):
Thanks for opening this! I think we should use this for the 1.43.0 release. I'm going to send 1.42.0 out today. One question about the semconv.yaml file before approving.

semantic_conventions/semconv.yaml
thompson-tomo: It gets set by renovate when bumping the version so shouldn't be an issue, it just might mean potentially an additional pr raised which bumps this file.

kaylareopelle (approved):

mwear (approved):

PR #2191, opened 2026-06-11
chore: remove naming suppressions

thompson-tomo:
This removes the naming cop suppressions.


metrics_sdk/lib/opentelemetry/sdk/metrics/view/registered_view.rb
```diff
@@ -32,9 +32,11 @@ def match_instrument?(metric_stream)
             true
           end
 
+          # rubocop:disable Naming/PredicateMethod
```
simi: Any reason to not use inline def comment here instead of disable/enable? 🤔

thompson-tomo: Updated accordingly.

mwear (approved):

kaylareopelle (approved):

PR #2197, opened 2026-06-14
chore(deps): Upgrade dependency minitest to v6

thompson-tomo:
See https://github.com/open-telemetry/opentelemetry-ruby/pull/2180 for details.

This PR is a subset of #2180, scoped to the gems which can be updated without changes.

This reduces qty of gems out of date by ensuring they can be updated going forward and avoids the chance changes could be added which would block a v6 update.

### DIFF
--- a/propagator/jaeger/Gemfile
+++ b/propagator/jaeger/Gemfile
@@ -9,7 +9,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :test, :development do
-  gem 'minitest', '~> 5.27.0'
+  gem 'minitest', '~> 6.0.0'
   gem 'rake', '~> 13.4.0'
   gem 'rubocop', '~> 1.87.0'
   gem 'rubocop-minitest', '~> 0.39.0'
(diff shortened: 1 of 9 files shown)


simi (approved):

mwear (approved):

propagator/jaeger/Gemfile
```diff
@@ -9,7 +9,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :test, :development do
-  gem 'minitest', '~> 5.27.0'
+  gem 'minitest', '~> 6.0.0'
```
kaylareopelle, line 12: I'm hesitant to continue updating the dependencies on Jaeger since the library itself is deprecated. Is there a way to exclude it from Renovate?

kaylareopelle (approved):

propagator/jaeger/Gemfile
kaylareopelle: I think it's fine for this change, but I'll open an issue.

thompson-tomo: Actually the deprecated gem is the jaeger exporter not propagator. I have opened #2205 to exclude the deprecated from renovate & rubocop as per the instructions in https://github.com/open-telemetry/opentelemetry-ruby-contrib/pull/2219

PR #2207, opened 2026-06-28
chore: docker housekeeping

thompson-tomo:
This sets an explicit docker image tag rather than latest. It also bumps to the latest version.


mwear (approved):

kaylareopelle (approved):
Thank you! Will this be update by renovate now that it has specific versions?

thompson-tomo:
Correct they will be updated by renovate.

PR #2215, opened 2026-07-03
fix: Update attribute type conformance validations to be in line with current OpenTelemetry Semantic Conventions

0robustus1:
OpenTelemetry v1.51.0 introduced a wider definition of [AnyValue](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.51.0/specification/common/README.md). This PR aims to validate attribute values against that newer spec rather than the older one.

### DIFF
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -28,21 +28,31 @@ module OpenTelemetry
         value.instance_of?(String) || boolean?(value) || numeric?(value) || value.nil?
       end
 
-      def valid_array_value?(value)
-        return false unless value.is_a?(Array)
-
-        value.all? { |v| valid_value?(v) }
-      end
-
-      def valid_mapping_value?(value)
-        return false unless value.is_a?(Hash)
-        return false unless value.keys.all? { |k| valid_key?(k) }
-
-        value.values.all? { |v| valid_value?(v) }
-      end
-
       def valid_value?(value)
-        valid_simple_value?(value) || valid_array_value?(value) || valid_mapping_value?(value)
+        to_check = [value]
+        seen = Set.new
+        until to_check.empty?
+          current = to_check.pop
+          next if valid_simple_value?(current)
+
+          return false if seen.include?(current.object_id)
+
+          seen << current.object_id
+
+          case current
+          when Array
+            current.each { |v| to_check << v }
+          when Hash
+            current.each do |k, v|
+              return false unless valid_key?(k)
+
+              to_check << v
+            end
+          else
+            return false
+          end
+        end
+        true
       end
 
       def valid_attributes?(owner, kind, attrs)
--- a/sdk/lib/opentelemetry/sdk/internal.rb
+++ b/sdk/lib/opentelemetry/sdk/internal.rb
@@ -25,27 +25,24 @@ module OpenTelemetry
       end
 
       def valid_simple_value?(value)
-        value.instance_of?(String) || boolean?(value) || numeric?(value)
+        value.instance_of?(String) || boolean?(value) || numeric?(value) || value.nil?
       end
 
       def valid_array_value?(value)
         return false unless value.is_a?(Array)
-        return true if value.empty?
 
-        case value.first
-        when String
-          value.all? { |v| v.instance_of?(String) }
-        when TrueClass, FalseClass
-          value.all? { |v| boolean?(v) }
-        when Numeric
-          value.all? { |v| numeric?(v) }
-        else
-          false
-        end
+        value.all? { |v| valid_value?(v) }
+      end
+
+      def valid_mapping_value?(value)
+        return false unless value.is_a?(Hash)
+        return false unless value.keys.all? { |k| valid_key?(k) }
+
+        value.values.all? { |v| valid_value?(v) }
       end
 
       def valid_value?(value)
-        valid_simple_value?(value) || valid_array_value?(value)
+        valid_simple_value?(value) || valid_array_value?(value) || valid_mapping_value?(value)
       end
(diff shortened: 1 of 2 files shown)


sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -25,27 +25,32 @@ def numeric?(value)
...
+        value.instance_of?(String) || boolean?(value) || numeric?(value) || value.nil?
       end
 
-      def valid_array_value?(value)
-        return false unless value.is_a?(Array)
-        return true if value.empty?
+      def valid_value?(value)
+        to_check = [value]
+        seen = Set.new
+        until to_check.empty?
```
xuan-cao-swi, line 34: Thanks for updating this to comply latest spec.
Have you considered using a recursive approach instead?
```ruby
def valid_value?(value, depth = 0)
  # ...
  case value
  when String, TrueClass, FalseClass, Integer, Float, NilClass
    true
  when Array
    value.all? { |v| valid_value?(v, depth + 1) }
  when Hash
    value.all? { |k, v| valid_key?(k) && valid_value?(v, depth + 1) }
  else
    false
  end
  # ...
end
```

0robustus1: Yes, I originally used the recursive approach, but I was a bit concerned about larger more complex JSON style attributes, since there isn't a natural limit on nesting.

mwear (changes requested):
Thanks for putting this together. I don't think we can move forward with this at this stage, for the reasons below:

The spec did widen `AnyValue` to include complex values, but in v1.51.0 those additions are all marked `Status: Development`, not stable. Only the primitives and homogeneous primitive arrays are stable. So this is really adopting development-stability spec features as the default validation behavior, and I don't think we should do that without a way to turn it off.

It also isn't contained to the validator. Anything that consumes attributes has to understand the wider shape, and our exporters don't yet. Concretely: `as_otlp_any_value` in the OTLP exporter has no `Hash` branch, so a map attribute passes validation, produces no error, then serializes to an empty `AnyValue` and disappears on export. `truncate_attribute_value` doesn't descend into hashes either. As it stands, the change lets people set values that silently vanish downstream.

Given the development-level stability, this needs to be opt-in behind a flag, with the default staying on the current stable behavior. No other SDK accepts maps for standard attributes by default as far as I know.

Separately, I don't think the set-based cycle tracking is correct. `seen` is never unwound, so it flags any object that appears more than once, not just an ancestor on the current path. A shared but acyclic value gets falsely rejected. For example `shared = [1, 2]; span.set_attribute('foo', [shared, shared])` is valid, but comes back invalid. Cycle detection needs to track the current path, not everything visited.

Happy to talk through what an opt-in version would look like if you want to take it that direction.

mwear:
I also wanted to note that there is [a proposal ](https://github.com/open-telemetry/opentelemetry-specification/pull/5186) to limit the depth for array and map attributes. I think that would make the recursive solution the most straightforward and preferred option.

0robustus1:
@mwear, the AnyValue extension is no longer marked as `Development` since [v1.53.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/common/README.md) (with v1.58.0 being current).

So in general I would say that it is valid and encouraged for SDKs to develop against that spec, especially when the spec allows something that the SDK currently does not.

I'll take a closer look at the other changes that would be necessary to not drop these attributes.

Regarding the [depth proposal](https://github.com/open-telemetry/opentelemetry-specification/pull/5186): That is only a proposal at this stage, isn't it?

Regarding the `seen` handling: Yes, this would treat the same object (Array, Hash) repeated as siblings the same, which isn't a cycle. I'll take a look at addressing that as well.

mwear:
Sorry for the confusion about stability. You're right that it's been stable since [1.53.0](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.53.0/specification/common/README.md). I followed the link in the description and saw development without comparing to current spec (my bad). We don't need this to be gated by a flag.

The depth proposal is only a proposal at this stage, however, there's no disagreement that limits on array and map values are needed. The only debate is what shape that should take. Given that we limit all other attribute types I think we can safely assume there will be a limit for maps and arrays.

One last thing that I would like to point out is that attribute validation is a hot code path. We're very sensitive to any unnecessary object allocations on hot paths. One unfortunate reality about the `Set` based approach is the `Set` allocation per validation. For this reason, and the likely need to track depth make the recursive solution more attractive.

0robustus1:
> One last thing that I would like to point out is that attribute validation is a hot code path. We're very sensitive to any unnecessary object allocations on hot paths. One unfortunate reality about the Set based approach is the Set allocation per validation. For this reason, and the likely need to track depth make the recursive solution more attractive.

That's a good point. I had considered the possibility of StackLevelToDeep errors as more problematic, but you are right, the `Set` negatively impacts all validations, while the `SystemStackError` scenarios would really only hit the problematic cases. I'll adjust it back to recursion and ensure that we handle the `SystemStackError` issue gracefully.

sdk/lib/opentelemetry/sdk/internal.rb
```diff
@@ -25,27 +25,34 @@ def numeric?(value)
       end
 
       def valid_simple_value?(value)
-        value.instance_of?(String) || boolean?(value) || numeric?(value)
+        value.instance_of?(String) || boolean?(value) || numeric?(value) || value.nil?
```
jprosevear, line 28: Could the string check be softened, either to allow descendants:

```
Rails.env.instance_of?(String) => false
Rails.env.is_a?(String) => true
```

or duck type

```
value.responds_to?(:to_s)
```

the latter needs to be checked after other simple types

PR #2216, opened 2026-07-03
fix: do not report disabled instrumentations as failures

0robustus1:
It can be very confusing to see a `Instrumentation: OpenTelemetry::Instrumentation::Sinatra failed to install` log line when that instrumentation is in fact just disabled (via `enabled: false`) or `OTEL_RUBY_INSTRUMENTATION_SINATRA_ENABLED=false`. This instead reports it as a not-enabled instrumentation.

### DIFF
--- a/registry/lib/opentelemetry/instrumentation/registry.rb
+++ b/registry/lib/opentelemetry/instrumentation/registry.rb
@@ -81,6 +81,8 @@ module OpenTelemetry
           OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was successfully installed with the following options #{instrumentation.config}"
         elsif !instrumentation.enabled?(config)
           OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was not installed, because it is not enabled"
+        elsif !instrumentation.compatible?(config)
+          OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install: compatibility issue"
         else
           OpenTelemetry.logger.warn "Instrumentation: #{instrumentation.name} failed to install"
         end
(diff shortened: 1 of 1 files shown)


xuan-cao-swi:
Thanks for bringing this up. Since you're already working on this, could you also include a [compatibility](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/main/instrumentation/base/lib/opentelemetry/instrumentation/base.rb#L136) check (e.g., `Instrumentation: #{instrumentation.name} failed to install: compatibility issue`)?

0robustus1:
@xuan-cao-swi good point. Added it.

0robustus1:
Test failure seems to report on other PRs as well:

```
1) Failure:
OpenTelemetry::SDK::Trace::Export::BatchSpanProcessor::dropped spans metrics#test_0001_reports consistent label keys [test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb:407]:
Expected: nil
  Actual: ["code.function", "reason"]
```

So it might be broken on main?

mwear (approved):

registry/lib/opentelemetry/instrumentation/registry.rb
```diff
@@ -79,6 +79,10 @@ def install_instrumentation(instrumentation, config)
           OpenTelemetry.logger.debug "Instrumentation: #{instrumentation.name} skipping install given corresponding dependency not found"
         elsif instrumentation.install(config)
           OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was successfully installed with the following options #{instrumentation.config}"
+        elsif !instrumentation.enabled?(config)
+          OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was not installed, because it is not enabled"
```
kaylareopelle: nit: 
```suggestion
          OpenTelemetry.logger.info "Instrumentation: #{instrumentation.name} was not installed because it is not enabled"
```

kaylareopelle (commented):
What a great catch! Thank you for opening this PR. One small grammar nit.

kaylareopelle (approved):

PR #2217, opened 2026-07-03
chore: Update README.md with additional badges

thompson-tomo:
Added badges for GitHub Discussions, OpenSSF Scorecard, and FOSSA status. As well as the contributor wall.


xuan-cao-swi (approved):

mwear (approved):
I noticed the fossa license scan is failing. Anyone want to take a look into that?

thompson-tomo:
@mwear in relation to fossa @kaylareopelle and i have discussed it in the past, issue is the packages(rdoc, yard & diff-lcs) are being mis-classified due to bad definition on package side ie using comma instead of OR to describe license.

As such i have an issue in the otel community repo to correct it.

PR #2228, opened 2026-07-08
chore: decrease rubocop metric rule overrides

thompson-tomo:
This seeks to reduce the volume of metric cop overrides used. Instead project/global settings are used.

No source code was changed hence the adjusting of max values when it was needed on a per gem basis to avoid violations.


xuan-cao-swi:
Is there any guideline that set to these specific number?

thompson-tomo:
If the global defaults could be used instead, the overrides were removed. Otherwise, The values chosen were the lowest value which didn't result in violations and obviously greater than the inherited value (global/project) defaults.

Note If there were guidelines, they would be used to set the projecf defaults rather than overrides.

xuan-cao-swi (approved):

kaylareopelle (approved):

Issue #2170, opened 2026-06-03
Consolidating proto/ and common functions into opentelemetry-common/


xuan-cao-swi:
Propose to move proto/ and common functions into opentelemetry-common/ so that otlp, otlp-metrics, and otlp-logs exporters
can use the centralized utilities.
1. Current opentelemetry-proto is [v1.10](https://github.com/open-telemetry/opentelemetry-proto/releases/tag/v1.10.0), while our proto is still v1.8.0. Centralizing proto to common avoids managing three separate proto/ folders.
2. There are common utilities such as `prepare_endpoint`, `prepare_headers`, etc. Consolidating them together prevents issues like https://github.com/open-telemetry/opentelemetry-ruby/pull/2158.

Other things to consider:
otlp and otlp-http are confusing—people may check otlp-http rather than otlp, and otlp-http still needs to be updated if other otlp exporters change.
1. Rename otlp → otlp-trace (similar to otlp-metrics and otlp-logs); if gRPC support is added in the future, consider adding the gRPC exporter inside otlp-trace/.
2. Add a notice that otlp-http is not maintained and redirect users to otlp.

thompson-tomo:
I would suggest rather than centralising proto into common we introduce a types-otlp gem which contains just the generated types which is versioned as per the proto used in the generation process. We could then replicate what we do with semconv and raise a pr when a new proto release occurs.

In relation to packages I wouldn't deprecate any but instead focus on having a clear hierarchy. The way I see it is:

```
otlp # meta package containing all otlp packages
  otlp-http # meta package containing all otlp http packages
    otlp-http-log # implementation of http log exporter
    otlp-http-metric # implementation of http metric exporter
    otlp-http-trace # implementation of http trace exporter
  otlp-grpc # meta package containing all otlp grpc packages
```

thompson-tomo:
Interesting finding, it appears that otlp-http & otlp-grpc are not even published.

hiSandog:
A dedicated generated-types gem seems cleaner than putting protocol types into a broad `common` package, but its versioning contract should be explicit: record the upstream proto tag/commit used for generation and test every exporter against the same types release. Keeping endpoint/header helpers in a separate transport utility layer would avoid making non-OTLP users of `common` depend on protobuf. A dependency graph test could also ensure trace, metric, and log exporters cannot silently drift to different protocol versions.

thompson-tomo:
Still in progress

[...]

PR #2236, opened 2026-07-11
feat: use shared otlp types

thompson-tomo:
Progresses: #2170

This ensures all otlp exporters use the otlp-common package which has a dependency on the new types-otlp gem which is generated based on proto.

It was considered to put the types in common however that makes for difficult versioning. But more significantly by separating them, it provides an architecture which will enable us to address the current jruby limitation.

Follow up steps will be:

- introduce otel-types-java and have common switch to this on jruby if necessary to get exporters to work on jruby
- ~~automating the types to be auto updated just like we do with semconv~~


types/otlp/Gemfile
```diff
@@ -0,0 +1,25 @@
...
+
+group :test, :development do
+  gem 'minitest', '~> 5.27.0'
+  gem 'google-protobuf', '~> 3.19.6'
+  gem 'rake', '~> 13.4.0'
+  gem 'rubocop', '~> 1.88.0'
+  gem 'rubocop-minitest', '~> 0.39.0'
+  gem 'rubocop-performance', '~> 1.26.0'
+  gem 'rubocop-rake', '~> 0.7.1'
+  gem 'rubocop-rspec', '~> 3.10.0'
```
thompson-tomo:
```suggestion
```

types/otlp/Gemfile
```diff
@@ -0,0 +1,25 @@
...
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
+
+gemspec
+
+group :test, :development do
+  gem 'minitest', '~> 5.27.0'
```
thompson-tomo:
```suggestion
  gem 'minitest', '~> 6.0.0'
```

types/otlp/Gemfile
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Copyright The OpenTelemetry Authors
+#
+# SPDX-License-Identifier: Apache-2.0
+
+source 'https://rubygems.org'
```
thompson-tomo: Add eval

thompson-tomo:
I managed to identify the cause of the generation differences. Ubuntu was installing a much older version compared to alpine. Have switched from apt-get to using mise to install explicit version which is managed by renovate.

.github/workflows/ci-semconv.yml
```diff
@@ -1,26 +1,55 @@
...
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+
+      - name: Select directories
+        id: set_dirs
+        shell: bash
+        run: |
+          # Static list (replace with dynamic detection if needed)
+          DIRS='["semantic_conventions", "types/otlp"]'
```
xuan-cao-swi: I don't see the need of having this step. 
Why not just use `matrix: dir: ["semantic_conventions", "types/otlp"]`

thompson-tomo: The intention was to have it dynamic based on the changed files so we could fail in the case of multiple changing & a diff exists. This would be to preserve accurate changelogs.

Note there is now 5+ semconv repos each versioned  separately as this is the new approach to federate the conventions out.

thompson-tomo: Anyway i have gone and made it dynamic based on presence of gemspec in directory

types/otlp/opentelemetry-types-otlp.gemspec
```diff
@@ -0,0 +1,37 @@
...
+  spec.license     = 'Apache-2.0'
+
+  spec.files = ::Dir.glob('lib/**/*.rb') +
+               ::Dir.glob('*.md') +
+               ['LICENSE', '.yardopts']
+  spec.require_paths = ['lib']
+  spec.required_ruby_version = '>= 3.3'
+
+  spec.add_dependency 'googleapis-common-protos-types', '~> 1.3'
+  spec.add_dependency 'google-protobuf', '~> 3.19'
```
mwear: This should match the dep the exporters declare.
```suggestion
  spec.add_dependency 'google-protobuf', '>= 3.19'
```

types/otlp/Rakefile
```diff
@@ -0,0 +1,60 @@
...
+  'collector/trace/v1/trace_service.proto',
+  'common/v1/common.proto',
+  'logs/v1/logs.proto',
+  'metrics/v1/metrics.proto',
+  'resource/v1/resource.proto',
+  'trace/v1/trace.proto'
+].freeze
+
+desc 'Generate probobuf'
+task :generate do
```
mwear, line 46: Can we clean first for regeneration?

semantic_conventions/otel-spec.yaml
```diff
@@ -0,0 +1 @@
+semantic-conventions: '1.44.0' # semantic_conventions
```
mwear, line 1: Does anything read this?

thompson-tomo: Renovate uses it as the trigger for when an update is needed and the ci also uses it to trigger an automated release/regeneration.

types/otlp/Rakefile
thompson-tomo: Done have added it.

PR #2240, opened 2026-07-12
feat: Move otlp exporter uri builder to common

thompson-tomo:
This moves uri builder out of each exporter into the otlp-common gem.


mwear (approved):

kaylareopelle (changes requested):
I think we should hold off on merging this PR into the exporters until we have the first `opentelemetry-exporter-otlp-common` release. Marking this as "request changes" until that's complete.

thompson-tomo:
@kaylareopelle now that common has been released, could you take another look at this.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common/utilities.rb
```diff
@@ -0,0 +1,34 @@
...
+      module Common
+        # Contains common utilities used by the otlp exporters
+        module Utilities
+          extend self
+
+          # Builds a url using the endpoint defined and if not present uses the configured sources
+          def build_uri(endpoint, path = '', primary_src = '', secondary_src = '', default = nil)
+            env_endpoint = OpenTelemetry::Common::Utilities.config_opt(primary_src, secondary_src, default: default) unless endpoint
+            endpoint ||= env_endpoint
+            raise ArgumentError, "invalid url for OTLPExporter #{endpoint}" unless OpenTelemetry::Common::Utilities.valid_url?(endpoint)
```
kaylareopelle, line 21: I'd like to see the exporter class that called this method in this message (ex. `OTLP::Logs::LogsExporter`) to assist with troubleshooting. With the current message, I worry debugging efforts may incorrectly start with `opentelemetry-exporter-otlp`.

kaylareopelle (changes requested):
Thanks for the ping, @thompson-tomo. I have one request with the error message. Other than that, this looks good to me.

exporter/otlp-common/lib/opentelemetry/exporter/otlp/common/utilities.rb
thompson-tomo: The difficulty with that is how can the common lib know that info without adding a param just to be used for logging? The message already has the url so wouldn't people to debug look at the configuration of that value?

thompson-tomo: @kaylareopelle i have come up with a mid ground solution where we log the name of the env variable so hopefully addresses your request.

PR #2241, opened 2026-07-12
fix: update total_recorded_attributes when LogRecord#attributes= is called

### LINKED ISSUE #2194: LogRecord#attributes= setter does not update @total_recorded_attributes
**Description of the bug**

When mutating a LogRecord's attributes after initialization (e.g. in a custom LogRecordProcessor#on_emit), the attributes= setter updates the attributes hash but does not recalculate @total_recorded_attributes. This means to_log_record_data reports a stale count from initialization time.

There is no public API to update @total_recorded_attributes — the only workaround is instance_variable_set(:@total_recorded_attributes, attrs.size), which is fragile and depends on SDK internals.

Expected behavior: Either attributes= should automatically update @total_recorded_attributes, or a public method should be provided to recalculate it.

Relevant code: logs/lib/opentelemetry/sdk/logs/log_record.rb (https://github.com/open-telemetry/opentelemetry-ruby/blob/main/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb) — @total_recorded_attributes is set once in initialize and never updated when attributes is reassigned via the attr_accessor.

Here's the merge request/project where I encountered this issue: https://gitlab.com/gitlab-org/developer-relations/contributor-success/contributors-gitlab-com/-/merge_requests/2325

- Operating system details: Linux, Debian 12 (bookworm), aarch64
- RUBY_ENGINE: "ruby"
- RUBY_VERSION: "4.0.5"
- RUBY_DESCRIPTION: "ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +PRISM aarch64-linux"

```rb
require 'bundler/inline'

gemfile(true) do
  source 'https://rubygems.org'

  gem 'opentelemetry-api'
  gem 'opentelemetry-sdk'
  gem 'opentelemetry-logs-sdk'
end

require 'opentelemetry-api'
require 'opentelemetry-sdk'

log_record = OpenTelemetry::SDK::Logs::LogRecord.new(
  body: 'test message',
  attributes: { 'key1' => 'value1' }
)

puts "Initial attributes count: #{log_record.instance_variable_get(:@total_recorded_attributes)}"
# => 1

log_record.attributes = { 'key1' => 'value1', 'key2' => 'value2', 'key3' => 'value3' }

puts "Attributes after reassignment: #{log_record.attributes.size}"
# => 3
puts "total_recorded_attributes (stale): #{log_record.instance_variable_get(:@total_recorded_attributes)}"
# => 1 (BUG: should be 3)

data = log_record.to_log_record_data
puts "Reported in log_record_data: #{data.total_recorded_attributes}"
# => 1 (stale)
```

ltickett:
Closes https://github.com/open-telemetry/opentelemetry-ruby/issues/2194

The attr_accessor-generated attributes= setter reassigned the attributes hash but left @total_recorded_attributes at its initialization value, so to_log_record_data reported a stale count after mutation (e.g. in a custom LogRecordProcessor#on_emit).

Replace the generated setter with a custom attributes= that recalculates @total_recorded_attributes and reapplies the configured attribute limits, and reuse it from the constructor.

### DIFF
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
@@ -18,13 +18,14 @@ module OpenTelemetry
                       :severity_text,
                       :severity_number,
                       :body,
-                      :attributes,
                       :event_name,
                       :trace_id,
                       :span_id,
                       :trace_flags,
                       :resource,
                       :instrumentation_scope
+
+        attr_reader :attributes
 
         # Creates a new {LogRecord}.
         #
@@ -81,7 +82,6 @@ module OpenTelemetry
           @severity_text = severity_text
           @severity_number = severity_number
           @body = body
-          @attributes = attributes&.to_h # We need a mutable copy of attributes
           @event_name = event_name
           @trace_id = trace_id
           @span_id = span_id
@@ -89,6 +89,18 @@ module OpenTelemetry
           @resource = resource
           @instrumentation_scope = instrumentation_scope
           @log_record_limits = log_record_limits || LogRecordLimits::DEFAULT
+          self.attributes = attributes
+        end
+
+        # Sets the attributes for this {LogRecord}, recalculating the total
+        # recorded attribute count and reapplying the configured attribute
+        # limits.
+        #
+        # @param [optional Hash{String => String, Numeric, Boolean,
+        #   Array<String, Numeric, Boolean>}] new_attributes Attributes to
+        #   associate with the {LogRecord}.
+        def attributes=(new_attributes)
+          @attributes = new_attributes&.to_h # We need a mutable copy of attributes
           @total_recorded_attributes = @attributes&.size || 0
 
           trim_attributes(@attributes)
@@ -115,9 +127,7 @@ module OpenTelemetry
         private
 
         def to_integer_nanoseconds(timestamp)
-          return unless timestamp.is_a?(Time)
-
-          (timestamp.to_r * (10**9)).to_i
+          (timestamp.to_r * (10**9)).to_i if timestamp.is_a?(Time)
         end
 
         def trim_attributes(attributes)
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
@@ -89,21 +89,18 @@ module OpenTelemetry
           @resource = resource
           @instrumentation_scope = instrumentation_scope
           @log_record_limits = log_record_limits || LogRecordLimits::DEFAULT
+          @dropped_attributes_count = 0
           self.attributes = attributes
         end
 
-        # Sets the attributes for this {LogRecord}, recalculating the total
-        # recorded attribute count and reapplying the configured attribute
-        # limits.
+        # Sets the attributes for this {LogRecord} and reapplies the
+        # configured attribute limits.
         #
         # @param [optional Hash{String => String, Numeric, Boolean,
         #   Array<String, Numeric, Boolean>}] new_attributes Attributes to
         #   associate with the {LogRecord}.
         def attributes=(new_attributes)
-          @attributes = new_attributes&.to_h # We need a mutable copy of attributes
-          @total_recorded_attributes = @attributes&.size || 0
-
-          trim_attributes(@attributes)
+          trim_attributes(@attributes = new_attributes&.to_h) # We need a mutable copy of attributes
         end
 
         def to_log_record_data
@@ -120,7 +117,7 @@ module OpenTelemetry
             @trace_flags,
             @resource,
             @instrumentation_scope,
-            @total_recorded_attributes
+            @dropped_attributes_count
           )
         end
 
@@ -130,24 +127,25 @@ module OpenTelemetry
           (timestamp.to_r * (10**9)).to_i if timestamp.is_a?(Time)
         end
 
-        def trim_attributes(attributes)
-          return if attributes.nil?
+        def trim_attributes(attrs)
+          return if attrs.nil?
 
           # truncate total attributes
-          truncate_attributes(attributes, @log_record_limits.attribute_count_limit)
+          truncate_attributes(attrs, @log_record_limits.attribute_count_limit)
 
           # truncate attribute values
-          truncate_attribute_values(attributes, @log_record_limits.attribute_length_limit)
+          truncate_attribute_values(attrs, @log_record_limits.attribute_length_limit)
 
           # validate attributes
-          validate_attributes(attributes)
-
-          nil
+          validate_attributes(attrs)
         end
 
-        def truncate_attributes(attributes, attribute_limit)
-          excess = attributes.size - attribute_limit
-          excess.times { attributes.shift } if excess.positive?
+        def truncate_attributes(attrs, attribute_limit)
+          excess = attrs.size - attribute_limit
+          return unless excess.positive?
+
+          excess.times { attrs.shift }
+          @dropped_attributes_count += excess
         end
 
         def validate_attributes(attrs)
--- a/logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
+++ b/logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
@@ -154,6 +154,36 @@ describe OpenTelemetry::SDK::Logs::LogRecord do
       end
     end
 
+    describe '#attributes=' do
+      it 'updates total_recorded_attributes when reassigned' do
+        log_record = Logs::LogRecord.new(attributes: { 'key1' => 'value1' })
+        assert_equal(1, log_record.instance_variable_get(:@total_recorded_attributes))
+
+        log_record.attributes = { 'key1' => 'value1', 'key2' => 'value2', 'key3' => 'value3' }
+
+        assert_equal(3, log_record.instance_variable_get(:@total_recorded_attributes))
+        assert_equal(3, log_record.to_log_record_data.total_recorded_attributes)
+      end
+
+      it 'sets total_recorded_attributes to 0 when set to nil' do
+        log_record = Logs::LogRecord.new(attributes: { 'key1' => 'value1' })
+
+        log_record.attributes = nil
+
+        assert_equal(0, log_record.instance_variable_get(:@total_recorded_attributes))
+        assert_nil(log_record.attributes)
+      end
+
+      it 'applies attribute count limits when reassigned' do
+        limits = Logs::LogRecordLimits.new(attribute_count_limit: 1)
+        log_record = Logs::LogRecord.new(log_record_limits: limits, attributes: { 'a' => 'a' })
+
+        log_record.attributes = { 'old' => 'old', 'new' => 'new' }
+
+        assert_equal({ 'new' => 'new' }, log_record.attributes)
+      end
+    end
+
     describe 'attribute value limit' do
       it 'truncates the values that are too long' do
         length_limit = 32
--- a/logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
+++ b/logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
@@ -155,22 +155,26 @@ describe OpenTelemetry::SDK::Logs::LogRecord do
     end
 
     describe '#attributes=' do
-      it 'updates total_recorded_attributes when reassigned' do
-        log_record = Logs::LogRecord.new(attributes: { 'key1' => 'value1' })
-        assert_equal(1, log_record.instance_variable_get(:@total_recorded_attributes))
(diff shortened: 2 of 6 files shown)


thompson-tomo:
Looking at this change, it improves the situation however it doesn't appear to address adding/removing attributes which is supported in the spec.

Chatting with copilot, it came up with:

Added a trackedAttributes class

```ruby
class TrackedAttributes
  attr_reader :dropped_count

  def initialize(initial = {})
    @attributes = initial.dup
    @dropped_count = 0
  end

  def delete(key)
    existed = @attributes.key?(key)
    result = @attributes.delete(key)
    @dropped_count += 1 if existed
    result
  end

  def delete_if(&block)
    @attributes.delete_if do |k, v|
      should_drop = block.call(k, v)
      @dropped_count += 1 if should_drop
      should_drop
    end
  end

  # Forward everything else to the underlying hash
  def method_missing(name, *args, &block)
    @attributes.public_send(name, *args, &block)
  end

  def respond_to_missing?(name, include_private = false)
    @attributes.respond_to?(name, include_private)
  end
end
```

And we change our logrecord to be

```ruby
class LogRecordData
  attr_accessor :attributes

  def initialize(...)
    @attributes = TrackedAttributes.new({})
  end

  def dropped_attributes_count
    @attributes.dropped_count
  end
end
```

That solution can be scaled to support adding/updating of attributes including having the limits enforced on those additional operations. We can also reuse it for spans etc

kaylareopelle:
> Looking at this change, it improves the situation however it doesn't appear to address adding/removing attributes which is supported in the spec...

@thompson-tomo, thanks for reviewing this with missing features in mind! This PR does a great job resolving the `total_recorded_attributes` bug. Adding/removing attributes feels like a new feature to me that I'd rather see addressed in a separate PR.

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -81,14 +82,25 @@ def initialize(
...
           @body = body
-          @attributes = attributes&.to_h # We need a mutable copy of attributes
           @event_name = event_name
           @trace_id = trace_id
           @span_id = span_id
           @trace_flags = trace_flags
           @resource = resource
           @instrumentation_scope = instrumentation_scope
           @log_record_limits = log_record_limits || LogRecordLimits::DEFAULT
+          self.attributes = attributes
```
kaylareopelle, line 93: Ah, interesting! Why go with `self.attributes` here? Just want to make sure I'm following the flow.

ltickett:
Thanks @thompson-tomo - I agree with @kaylareopelle - let's look at this in a follow-up and keep this PR scope small.

> Ah, interesting! Why go with self.attributes here? Just want to make sure I'm following the flow.

@kaylareopelle calling self.attributes = attributes in the constructor (log_record.rb:92) routes through the new setter (log_record.rb:102) so that @total_recorded_attributes calculation and trim_attributes happen in exactly one place. Without it you'd duplicate the count + trim logic in both the constructor and the setter, which is what caused issue #2194 in the first place (setter diverged from constructor). So reusing the setter seems like the logical DRY choice.

Wdyt?

thompson-tomo:
I have no problem with doing it in small stages however if we do so, we should avoid saying that this pr fixes the issue when it doesn't and ensure that the bug is kept open.

In fact on thinking about it, this solution might lead to scenarios which were previously correct now being wrong. For instance create a log record with 5 attributes, then update the attributes to contain 4 using set.  With using total when attributes is set you would no longer know about the dropped attribute.

kaylareopelle:
> I have no problem with doing it in small stages however if we do so, we should avoid saying that this pr fixes the issue when it doesn't and ensure that the bug is kept open.
> 
> In fact on thinking about it, this solution might lead to scenarios which were previously correct now being wrong. For instance create a log record with 5 attributes, then update the attributes to contain 4 using set. With using total when attributes is set you would no longer know about the dropped attribute.

@thompson-tomo, I'm not sure I'm following. Could you provide a code snippet that creates this situation or write a test that fails with the current code?

thompson-tomo:
@kaylareopelle Here is the test

```

it 'updates total_recorded_attributes when reassigned' do
        log_record = Logs::LogRecord.new(attributes: { 'key1' => 'value1', 'key2' => 'value2', 'key3' => 'value3' })
        assert_equal(3, log_record.instance_variable_get(:@total_recorded_attributes))

        log_record.attributes = { 'key1' => 'value1' }

        assert_equal(3, log_record.instance_variable_get(:@total_recorded_attributes))
        assert_equal(3, log_record.to_log_record_data.total_recorded_attributes)
        assert_equal(2, log_record.to_log_record_data.total_dropped_attributes)
      end
```

In effect it is the opposite to the test which now passes

kaylareopelle:
@thompson-tomo - Thank you, I see what you're saying. I think in this case though, shouldn't `total_recorded_attributes` be 4, because four attributes have been seen, even though one remains? The new hash is effectively overwriting the old hash, so even though it's the same key/value pair content, I think we should count it as a new attribute.

This would return the expected 3 for `dropped_attributes_count` in the OTLP logs exporter: https://github.com/open-telemetry/opentelemetry-ruby/blob/85b5248fa852988fcfe342ef4d89a1f294f31287/exporter/otlp-logs/lib/opentelemetry/exporter/otlp/logs/logs_exporter.rb#L312

Does this logic make sense to you, @ltickett? I'll add a code comment with a way we could adjust things if we think the value should resolve to four in @thompson-tomo's [previous comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/2241#issuecomment-4998290803).

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle, line 93: Part 1 (of 2) for the new suggestion based on [this comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/2241#issuecomment-5028195642): 
```suggestion
          @total_recorded_attributes = 0
          self.attributes = attributes
```

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
```diff
@@ -81,14 +82,25 @@ def initialize(
...
+        # Sets the attributes for this {LogRecord}, recalculating the total
+        # recorded attribute count and reapplying the configured attribute
+        # limits.
+        #
+        # @param [optional Hash{String => String, Numeric, Boolean,
+        #   Array<String, Numeric, Boolean>}] new_attributes Attributes to
+        #   associate with the {LogRecord}.
+        def attributes=(new_attributes)
+          @attributes = new_attributes&.to_h # We need a mutable copy of attributes
           @total_recorded_attributes = @attributes&.size || 0
```
kaylareopelle: Part 2 (of 2) based on [this comment](https://github.com/open-telemetry/opentelemetry-ruby/pull/2241#issuecomment-5028195642): 
```suggestion
          @total_recorded_attributes += @attributes&.size || 0
```

kaylareopelle:
I just realized we've strayed from the spec. `@total_recorded_attributes` should only change if an attribute is dropped due to log limits, otherwise, it remains identical to the attributes size of the payload that gets sent up. That would likely move this logic to `trim_attributes`. 

Here's the definition for `dropped_attributes_count` in the Traces proto (the definition isn't [in the Logs proto](https://github.com/open-telemetry/opentelemetry-proto/blob/main/opentelemetry/proto/logs/v1/logs.proto#L179)).
```
  // The number of attributes that were discarded. Attributes
  // can be discarded because their keys are too long or because there are too many
  // attributes. If this value is 0, then no attributes were dropped.
```
https://github.com/open-telemetry/opentelemetry-proto/blob/46f1e4ce1d903009f3e38e3052ce8eab5c4e4d0c/opentelemetry/proto/trace/v1/trace.proto#L215-L218

Changing attributes during the life of a Log Record isn't enough to make that attribute "dropped" and change the `total_attributes_count`.

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: Resolving, as I realized the intention for dropped attribute is related to changes based on limits.

logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
kaylareopelle: Resolving, as I realized the intention for dropped attribute is related to changes based on limits.

thompson-tomo:
I agree in my scenario the total could be considered 4 but it is under defined.

Looking at https://opentelemetry.io/docs/specs/otel/logs/sdk/#readwritelogrecord is it even supported to be replacing the entire collection?

Would it make more sense to track the dropped count directly that we could track all removals.

I have asked questions in the [spec channel](https://cloud-native.slack.com/archives/C01N7PP1THC/p1784599985851789) as area feels under defined.

ltickett:
@thompson-tomo @kaylareopelle did we land on a decision regarding the way forward here please?

kaylareopelle:
Hi @ltickett, apologies for the delayed response. We didn't get any feedback in Slack. I'm going to bring this to the SIG meeting on Tuesday to try to get a consensus and we'll have more for you after that.

kaylareopelle:
During the SIG meeting, it was suggested that I look at other implementations to see if there's a consistent approach for handling this value. 

After reviewing Python, JS, and Java, I think it's safe to say that `dropped_attributes_count` relates to attributes dropped because they exceeded the total attribute limit (`OTEL_LOGRECORD_ATTRIBUTE_COUNT_LIMIT` or `OTEL_ATTRIBUTE_COUNT_LIMIT`). 

* [Python test](https://github.com/open-telemetry/opentelemetry-python/blob/0110bb1d273dbe848312d429326b844da9e86d93/opentelemetry-sdk/tests/logs/test_handler.py#L466-L494)
* [JS test](https://github.com/open-telemetry/opentelemetry-js/blob/b4ef7a518f878f342f93f11b45df1ae1740a48ec/experimental/packages/sdk-logs/test/common/LogRecord.test.ts#L304-L324)
* [Java test](https://github.com/open-telemetry/opentelemetry-java/blob/db1d6bee36537b4f356a6b7d616e587217938177/sdk/common/src/test/java/io/opentelemetry/sdk/common/internal/AttributesMapTest.java#L103-L113) - In Java, the AttributesMap is used for both logs and traces, so the traces prototype definition seems to apply to both cases. 

@thompson-tomo, what do you think about the spec interpretation given these examples? 

cc @mwear @xuan-cao-swi

thompson-tomo:
I have no issue with dropped being scoped to counting those dropped due to limits. With that being said, i still don't see this change as resolving the issue and not even progressing it.

If we were to instead simply track how many attributes have been dropped due to limits, we can then calculate the total.

Some test cases i can think of would be:

- create log record with limit + 2 attributes, add 1 attribute, remove 1 attribute, add 2 attributes. The dropped count should ideally be 4 but 3 is explainable. 

If we only use the size prior to last limiting we would end up with 1.

kaylareopelle:
I see what you're saying, @thompson-tomo. I would love to get this merged in and out to our users. In that spirit, I created a branch off of this PR that should address these concerns: https://github.com/ltickett/opentelemetry-ruby/pull/1

@thompson-tomo, @ltickett - please review that branch when you have a chance. If you like its approach, we can merge it into this PR.

ltickett:
@kaylareopelle I just left one question- otherwise LGTM, thank you

thompson-tomo (approved):
LGTM as it increases predictability of dropped count value as it will not now always be the qty of attributes dropped from a log record due to it being above the limit.

logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
```diff
@@ -154,6 +154,74 @@
...
+        log_record = nil
+
+        OpenTelemetry::TestHelpers.with_test_logger do
+          log_record = Logs::LogRecord.new(attributes: { 'a' => 'a', 'b' => Class.new })
+        end
+
+        assert_equal(0, log_record.instance_variable_get(:@dropped_attributes_count))
+      end
+
+      it 'accumulates dropped_attributes_count across repeated reassignment' do
```
kaylareopelle, line 210: @ltickett - how's this for documentation?
```suggestion

      # our choice to accumulate across reassignment is based on 
      # a test in the JS implementation:
      # https://github.com/open-telemetry/opentelemetry-js/blob/7b155b84c86be06b2b6d309c28e61e617af6e779/experimental/packages/sdk-logs/test/common/LogRecord.test.ts#L313-L324
      it 'accumulates dropped_attributes_count across repeated reassignment' do
```

kaylareopelle (approved):

logs_sdk/test/opentelemetry/sdk/logs/log_record_test.rb
ltickett: LGTM- thanks @kaylareopelle

PR #2259, opened 2026-07-20
fix: Export binary-tagged UTF-8 attributes over OTLP

bdewater-thatch:
Rack exposes request header values as ASCII-8BIT strings. A Cloudflare `cf-ipcity` header containing `Montréal` therefore reaches the exporter (if Rack instrumentation is [so configured](https://github.com/open-telemetry/opentelemetry-ruby-contrib/blob/ccec56fc57e3c28614dfb7724a198951664aadce/instrumentation/rack/lib/opentelemetry/instrumentation/rack/instrumentation.rb#L24)) with valid UTF-8 bytes but a binary encoding tag. Protobuf attempts an ASCII-8BIT to UTF-8 conversion, raises Encoding::UndefinedConversionError, and the exporter sends the literal value `Encoding Error`.

The exporter now duplicates binary-tagged strings, relabels the duplicate as UTF-8, and uses it only when valid_encoding? succeeds. It does not mutate the instrumentation-owned value or silently scrub malformed data. This corresponds with https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-type-mapping.md#string-values

### DIFF
--- a/common/lib/opentelemetry/common/utilities.rb
+++ b/common/lib/opentelemetry/common/utilities.rb
@@ -54,9 +54,9 @@ module OpenTelemetry
       #
       # @param [String] string The string to be utf8 encoded
       # @param [optional boolean] binary This option is for displaying binary data
-      # @param [optional String] placeholder The fallback string to be used if encoding fails
+      # @param [String, nil] placeholder The fallback value to be used if encoding fails
       #
-      # @return [String]
+      # @return [String, nil]
       def utf8_encode(string, binary: false, placeholder: STRING_PLACEHOLDER)
         string = string.to_s
 
@@ -66,6 +66,11 @@ module OpenTelemetry
           string.encode('UTF-8', 'binary', invalid: :replace, undef: :replace, replace: '')
         elsif string.encoding == ::Encoding::UTF_8
           string
+        elsif string.encoding == ::Encoding::ASCII_8BIT
+          utf8_string = string.dup.force_encoding(::Encoding::UTF_8)
+          raise Encoding::InvalidByteSequenceError, 'binary string is not valid UTF-8' unless utf8_string.valid_encoding?
+
+          utf8_string
         else
           string.encode(::Encoding::UTF_8)
         end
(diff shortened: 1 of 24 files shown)


thompson-tomo (commented):
Do we not need to also make the same changes in both metrics & log exporter?

xuan-cao-swi:
Both Python (which uses native Unicode strings) and JS (via TextEncoder) inherently enforce valid text encoding, so they don't have this kind of issue. I think this is a valid fix for safely relabeling incorrectly tagged binary strings back to UTF-8.

kaylareopelle (changes requested):
Thanks for your patience, @bdewater-thatch! 

After thinking this through a bit more, I'd like to see the scope of this fix narrower than what's currently in this PR.

The original problem came from OTLP export. I recommend we focus on that problem as a first step (including the changes to `Common::Utilities#utf8_encode`) and consider adding more checking/normalizing to other parts of the project later on.

Adding normalization to attributes in the SDKs, Resources, etc. does make it seem safer to avoid this error from ever reaching the exporter, but this seems like a lot of changes to some stable parts of code that I'd rather consider more closely, in a separate PR, before making changes. 

There are a few sections in the [Common spec concepts](https://github.com/open-telemetry/opentelemetry-specification/tree/main/specification/common) that make it seem like we may not want do this normalization, but it requires a bit of deciphering to know for sure. Fully supporting this part of the spec may include adding some new features (like the ability to [convert strings into `bytes_value`](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/common/attribute-type-mapping.md?plain=1#L109-L113). 

IIRC, the SDK changes came from our SIG meeting and if that's the case, I'm sorry to go back on what we discussed there. Since OTel switched to the new LFX Zoom platform, I haven't been able to find our recordings so I can't go back and check.

My request:
* Keep the changes in common/utilities
* Keep the changes in the OTLP exporters
* Remove the changes from the SDKs

bdewater-thatch:
Dusted off the code and addressed the feedback :)

kaylareopelle (approved):
Thank you for the updates!

thompson-tomo (approved):

PR #2264, opened 2026-07-24
docs: document custom error handlers

RedZapdos123:
## Description:

Issue #1715 reported that OpenTelemetry Ruby supports custom error handlers, but the usage was not documented in an easy to find place.

This PR documents custom error handler configuration in the root README and adds YARD documentation for both `OpenTelemetry.error_handler=` and `OpenTelemetry::SDK::Configurator#error_handler=`.

It also adds a focused SDK regression test that verifies `config.error_handler=` is installed during SDK configuration.

Closes #1715.

## Validations Run:

- `docker compose run --rm api bundle exec rake test`
- `docker compose run --rm sdk bundle exec rake test`
- `docker compose run --rm api bundle exec rubocop lib/opentelemetry.rb`
- `docker compose run --rm sdk bundle exec rubocop lib/opentelemetry/sdk/configurator.rb test/opentelemetry/sdk_test.rb`

## Notes:
- The focused SDK RuboCop command reports 5 pre-existing `Style/DocumentationMethod` informational offenses in `sdk/lib/opentelemetry/sdk/configurator.rb`.
- `bundle exec rake yard` in both `api` and `sdk` still reports pre-existing repository YARD warnings unrelated to this change.

PR #2265, opened 2026-07-24
chore: add configurator method docs

RedZapdos123:
`Style/DocumentationMethod` reports 5 informational offenses in `sdk/lib/opentelemetry/sdk/configurator.rb` for undocumented methods in `NoopTextMapPropagator` and the `Configurator` getters.

This PR adds minimal method documentation comments for those methods and keeps the change scoped to a single file. It does not change behavior.


simi (approved):

thompson-tomo (approved):

sdk/lib/opentelemetry/sdk/configurator.rb
```diff
@@ -56,6 +60,7 @@ def logger=(new_logger)
         @logger = ForwardingLogger.new(new_logger, level: ENV['OTEL_LOG_LEVEL'] || Logger::INFO)
       end
 
+      # Returns the configured error handler or the global OpenTelemetry error handler.
       def error_handler
```
kaylareopelle, line 64: Interesting, this showed up where I expected the changes in #2264 to show up.

kaylareopelle (commented):
One concern related to error_handler. Everything else looks good to go! Thank you!

sdk/lib/opentelemetry/sdk/configurator.rb
RedZapdos123: This PR documents the `error_handler` getter so that is why this text showed up here. I updated #2264 to use explicit `error_handler=` setter methods. Those setter docs now render in YARD too, now.

kaylareopelle (approved):

RedZapdos123:
Looked into the failing `opentelemetry-sdk / ubuntu-24.04` check. It seems to be a separate flaky test in `sdk/test/opentelemetry/sdk/trace/export/batch_span_processor_test.rb`, not a regression from this PR.

Reproduced it under JRuby. The `reports consistent label keys` example can race with the worker thread so the queue overflow path is sometimes skipped and only one `otel.bsp.dropped_spans` metric entry is recorded. I opened #2321 to make that example deterministic by setting `start_thread_on_boot: false`.

PR #2266, opened 2026-07-25
chore(api): clean up YARD param tags

RedZapdos123:
YARD currently reports `@param` tag warnings in the API docs in `Context`, `Trace.context_with_span`, and `Span#add_link`. This cleans up those stale parameter tags and adds the small missing comment for `Context.empty` so the touched files stay lint clean.


simi (approved):

thompson-tomo (approved):

api/lib/opentelemetry/context.rb
```diff
@@ -79,7 +79,6 @@ def with_current(ctx)
 
       # @param [String] key The lookup key
       # @param [Object] value The object stored under key
-      # @param [Callable] Block to execute in a new context
```
kaylareopelle, line 82: Hmm... I wonder if YARD has a different way to denote blocks that we're not using? 

Even though we don't have an explicit argument for a block, a block is accepted and used with this method call.

https://github.com/open-telemetry/opentelemetry-ruby/blob/cf8a18b359ebcea7905a28a26120289cfa2ad834/api/lib/opentelemetry/trace.rb#L62-L71

api/lib/opentelemetry/context.rb
```diff
@@ -93,10 +92,8 @@ def with_value(key, value)
       # Execute a block in a new context where its values are merged with the
       # incoming values. Restores the previous context after the block executes.
 
-      # @param [String] key The lookup key
       # @param [Hash] values Will be merged with values of the current context
       #  and returned in a new context
-      # @param [Callable] Block to execute in a new context
```
kaylareopelle, line 99: Same question about Block for this method too, though I don't have an example of the method being called.

api/lib/opentelemetry/context.rb
RedZapdos123: `@yieldparam` is a better YARD way to document this block than a fake `Block` parameter. I opened #2318 to switch the `OpenTelemetry::Context` block docs to `@yieldparam` so the rendered docs match the actual block behavior.

api/lib/opentelemetry/context.rb
RedZapdos123: Included `with_values` in #2318 and switched its block docs to `@yieldparam` for the yielded context and values.

kaylareopelle (approved):

PR #2273, opened 2026-07-29
feat: Add attributes to Meter InstrumentationScope

kaylareopelle:
Following the pattern in #2094, add attributes to the MeterProvider#meter method and the instrumentation scope associated with a Meter.

This will hopefully allow us to quiet the unnecessary release with something that has a related feature.

### DIFF
--- a/metrics_api/lib/opentelemetry/internal/proxy_meter_provider.rb
+++ b/metrics_api/lib/opentelemetry/internal/proxy_meter_provider.rb
@@ -13,7 +13,7 @@ module OpenTelemetry
     # It delegates to a "real" MeterProvider after the global meter provider is registered.
     # It returns {ProxyMeter} instances until the delegate is installed.
     class ProxyMeterProvider < Metrics::MeterProvider
-      Key = Struct.new(:name, :version)
+      Key = Struct.new(:name, :version, :attributes)
       private_constant(:Key)
 
       # Returns a new {ProxyMeterProvider} instance.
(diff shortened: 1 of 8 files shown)


mwear (commented):
Two things. The ProxyMeterProvider needs to be updated to accept attributes:

https://github.com/open-telemetry/opentelemetry-ruby/blob/912748bbb2308186e0f33bf79bdafe2be56bf268/metrics_api/lib/opentelemetry/internal/proxy_meter_provider.rb#L50

The gemspec needs to depend on opentelemetry-sdk v1.13.0 since that is where the attributes parameter was added to `InstrumentationScope`.

xuan-cao-swi (approved):
LGTM

kaylareopelle:
> Two things. The ProxyMeterProvider needs to be updated to accept attributes:
> 
> https://github.com/open-telemetry/opentelemetry-ruby/blob/912748bbb2308186e0f33bf79bdafe2be56bf268/metrics_api/lib/opentelemetry/internal/proxy_meter_provider.rb#L50
> 
> The gemspec needs to depend on opentelemetry-sdk v1.13.0 since that is where the attributes parameter was added to `InstrumentationScope`.

Thanks, @mwear. Great catches. I believe I've added these changes:
* ProxyMeterProvider: 013b7a1
* SDK bump: f39f29c

I also realized this doesn't add attributes to the OTLP exporter's instrumentation scope. We're also missing that from the Traces OTLP exporter, so I can add support for both in a separate PR. I may submit the logs equivalent of this PR first though.

mwear (approved):

PR #2298, opened 2026-08-17
fix: Enforce W3C Baggage limits on the extract path

### LINKED ISSUE #2163: W3C Baggage extract does not enforce the MAX_ENTRIES / MAX_ENTRY_LENGTH / MAX_TOTAL_LENGTH constants the inject path enforces

serhiy-bzhezytskyy:
The `baggage` extract path decodes the inbound header without applying the `MAX_ENTRIES` (180), `MAX_ENTRY_LENGTH` (4096) and `MAX_TOTAL_LENGTH` (8192) limits that the inject path (`#encode`) enforces, so an unbounded inbound header is parsed in full. This is the class the Java runtime fixed as CVE-2026-45292 (unbounded memory/CPU via W3C Baggage parsing, fixed in opentelemetry-java 1.62.0). Fixes #2163.

`extract` now enforces the same three limits through a small private helper mirroring `#encode`: over-limit entries are dropped where the limit is reached, and the entries decoded before that are kept. Three extract-side tests mirror the existing inject-side limit tests; each fails on `main` and passes with the change.

Assisted-By: Claude Fable 5


simi (approved):

simi:
notice this still reads full header into memory

serhiy-bzhezytskyy:
Thanks, you're right. Fixed and covered it with two tests: extract now walks the header lazily and stops at 180 entries, so nothing past the 180th is scanned or copied. The tests assert the property, that allocations do not grow with the header length; the numbers I measured separately are 180 strings instead of 500,000 and 0.04 ms instead of 53 ms on a 7.8 MB header with 500k entries.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
```diff
@@ -55,17 +55,10 @@ def extract(carrier, context: Context.current, getter: Context::Propagation.text
           header = getter.get(carrier, BAGGAGE_KEY)
           return context if header.nil? || header.empty?
 
-          entries = header.gsub(/\s/, '').split(',')
+          entries = header.each_line(',', chomp: true).lazy.take(MAX_ENTRIES)
```
kaylareopelle: I think this works really well when we have a header playing by the rules, with lots of different keys of different lengths. What if we have one giant header without a separator? 

Is it worth comparing the header's bytesize with `MAX_TOTAL_LENGTH` before diving into the entries?

kaylareopelle:
Thanks for this PR, @serhiy-bzhezytskyy! One question on the PR itself. I also noticed some JRuby tests are failing related to this PR.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
serhiy-bzhezytskyy: Thanks, you're right on both.

Fixed the giant header — bounded the separator search instead of scanning the whole thing: 1 GB entry with no comma measured 1s+ before, ~0.0002s now. Also switched the length checks to bytesize (spec is byte-denominated).

Fixed JRuby too — GC.stat[:total_allocated_objects] is nil there, skipped those two tests on non-MRI engines.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
```diff
@@ -82,6 +75,46 @@ def fields
 
         private
 
+        # Like String#each_line(','), but never scans more than 2x MAX_ENTRY_LENGTH
+        # ahead for the next separator.
+        def bounded_entries(header)
```
mwear: I think we can simplify this considerably by bounding the header once, up front. Truncate when `bytesize` exceeds the max and cut back to the last separator. Something like the following:

```ruby
def split_entries(header)
  return header.split(',') if header.bytesize <= MAX_TOTAL_LENGTH

  window = header.byteslice(0, MAX_TOTAL_LENGTH + 1)
  last_separator = window.byterindex(',')
  return [] unless last_separator

  window.byteslice(0, last_separator).split(',')
end
```

Python, Go, and Rust all use the same up-front check ([python](https://github.com/open-telemetry/opentelemetry-python/blob/60b6749a/opentelemetry-api/src/opentelemetry/baggage/propagation/__init__.py#L101-L106), [go](https://github.com/open-telemetry/opentelemetry-go/blob/3e1f0653c/baggage/baggage.go#L517-L518), [rust](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/propagation/baggage.rs#L107-L122)), though they all drop the whole header rather than keeping the prefix. Keeping the prefix seems better to me given the spec's "SHOULD propagate as much baggage information as is reasonable".

mwear (changes requested):

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
serhiy-bzhezytskyy: Applied, thanks.

One behaviour change: a >8192-byte header whose prefix is one oversized entry now yields nothing, where the old walk skipped it and kept the small entries behind it. Closer to go/python/rust, and that header is already over budget.

serhiy-bzhezytskyy:
Could someone approve the workflow run? Same on #2299.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
```diff
@@ -55,17 +55,10 @@ def extract(carrier, context: Context.current, getter: Context::Propagation.text
           header = getter.get(carrier, BAGGAGE_KEY)
           return context if header.nil? || header.empty?
 
-          entries = header.gsub(/\s/, '').split(',')
+          entries = split_entries(header).take(MAX_ENTRIES)
```
mwear: This counts candidates rather than entries that are actually kept. Can we move the counting to `decode_entries` and stop when we reach the max?

mwear (changes requested):
Thanks for the changes @serhiy-bzhezytskyy. I have just one more ask (see below). Also there is a test that appears to be legitimately failing on Truffle Ruby.

api/lib/opentelemetry/baggage/propagation/text_map_propagator.rb
serhiy-bzhezytskyy: Moved. `decode_entries` now breaks on the kept count, matching `#encode`. Added a test: 200 empty candidates before valid entries kept 0 before, 180 now.

serhiy-bzhezytskyy:
The truffleruby failure was my test, not the fix — it timed the first pass, which costs ~0.1s of JIT there. It now warms up and times the second call: 0.000068s on truffleruby. An unbounded scan still measures ~0.1s warmed.

api/test/opentelemetry/baggage/propagation/text_map_propagator_test.rb
```diff
@@ -70,6 +70,87 @@
...
+
+        propagator.extract(carrier, context: OpenTelemetry::Context.empty)
+
+        start = Process.clock_gettime(Process::CLOCK_MONOTONIC)
+        propagator.extract(carrier, context: OpenTelemetry::Context.empty)
+        elapsed = Process.clock_gettime(Process::CLOCK_MONOTONIC) - start
+
+        _(elapsed).must_be(:<, 0.02)
+      end
+
```
mwear: While these tests are somewhat interesting, I don't think we need to keep them.
```suggestion
```

mwear (changes requested):

api/test/opentelemetry/baggage/propagation/text_map_propagator_test.rb
serhiy-bzhezytskyy: Removed. With `split_entries` reverted to a plain `header.split(',')` the remaining 19 tests still pass — the suite covers the limits now, not the bound.

mwear (approved):

robbkidd (approved):
oh, it'll fit

kaylareopelle (approved):

PR #2299, opened 2026-08-18
fix: Enforce baggage limits on the Jaeger extract path

serhiy-bzhezytskyy:
The Jaeger propagator extracts every `uberctx-` header into baggage with no bound: `context_with_extracted_baggage` walks all carrier keys and sets each one, so an inbound carrier decides how much baggage enters the context.

The Jaeger format defines no limits, so this borrows the W3C Baggage spec limits — 180 entries, 4096 bytes per entry, 8192 bytes total, byte-denominated — dropping members once a limit is reached and keeping the earlier ones. `opentelemetry-java` caps its own Jaeger propagator the same way and for the same reason (CVE-2026-45292); the W3C propagator in this repo is being brought to parity in #2298.

Adds four extract tests: entry count, per-entry bytes, per-entry bytes measured in bytes rather than characters, and the total-bytes cap. Each fails without the change.

Assisted-By: Claude Fable 5


simi (approved):
not sure we need all those comments

serhiy-bzhezytskyy:
Thanks — dropped the explanatory comments and kept one line where the byte-vs-character distinction is not obvious from the code. @simi

propagator/jaeger/lib/opentelemetry/propagator/jaeger/text_map_propagator.rb
```diff
@@ -103,15 +108,24 @@ def build_span(match, sampling_flags)
...
+            count = 0
+            total_bytes = 0
             getter.keys(carrier).each do |carrier_key|
-              baggage_key = carrier_key.start_with?(baggage_key_prefix) && carrier_key[baggage_key_prefix.length..]
+              break unless count < MAX_BAGGAGE_ENTRIES
+
+              baggage_key = carrier_key.start_with?(BAGGAGE_KEY_PREFIX) && carrier_key[BAGGAGE_KEY_PREFIX.length..]
               next unless baggage_key
 
               raw_value = getter.get(carrier, carrier_key)
```
xuan-cao-swi, line 119: Maybe add `next unless raw_value` to ignore the case when `raw_value=nil`? 
This can also prevent the `undefined method 'bytesize'` issue when raw_value is nil

serhiy-bzhezytskyy: Thanks, you're right. Fixed and covered it with a test.

Rubocop then hit 8/7 on Metrics/CyclomaticComplexity and 101/100 on Metrics/ClassLength, so the limit check moved into `within_baggage_limits?` and the `private_constant` list is reflowed. `bundle exec rake` is green.

mwear (commented):
Overall LGTM. The Jaeger propagator is deprecated, so we should make an issue to remove it, but I'm fine with there being a version of it that enforces these limits. My only question is: should we apply the same limits to the inject path while we're here? Java does. I'm inclined to say yes, but open to other opinions.

serhiy-bzhezytskyy:
Done — inject enforces the same three limits, same shape as extract: break on the count, skip an entry over the per-entry or total byte budget. Java caps count and total there but not per-entry, and it measures `key.length() + value.length()`, so this one is in bytes.

The class was sitting exactly on Metrics/ClassLength 100, so I collapsed three multi-line argument lists to pay for the new method. It's back at 100, with no headroom.

Jaeger gem: 32 runs, 0 failures, rubocop clean.

mwear (commented):
The code changes from the last commit look good, but can you undo the formatting changes? You can increase the classlength in `/propagator/jaeger/.rubocop.yml` e.g.:

```
Metrics/ClassLength:
  Max: 120
```

serhiy-bzhezytskyy:
Done - formatting reverted.

mwear (approved):

kaylareopelle (approved):

PR #2301, opened 2026-08-20
fix!: align with the spec for metrics sdk and metrics otlp exporter

xuan-cao-swi:
## Description
1. periodic metrics reader with console exporter default 10s export interval - [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/sdk_exporters/stdout.md)
2. follow the python metrics otlp implementation that shutdown should stop export - [py](https://github.com/open-telemetry/opentelemetry-python/blob/v1.44.0/exporter/opentelemetry-exporter-otlp-proto-http/src/opentelemetry/exporter/otlp/proto/http/metric_exporter/__init__.py#L342)


mwear (commented):
lgtm, but there are some rubocop failures that need to be addressed.

mwear (approved):

simi:
Not sure if related directly to this change, but shouldn't we synchronize shutdown with the export mutex or re-check `@shutdown` inside the synchronized section? It seems there's lifecycle race now.

kaylareopelle (approved):
This looks good to me! Do we want any tests to make sure we don't drift away from these changes?

kaylareopelle:
> Not sure if related directly to this change, but shouldn't we synchronize shutdown with the export mutex or re-check `@shutdown` inside the synchronized section? It seems there's lifecycle race now.

That's a good question, @simi. It might be a separate change to think through. If I'm understanding your question, we don't synchronize shutdown checks in our other exporters. The main OTLP (Traces) exporter doesn't use a mutex at all. I think we should look into this more. If we find a solution, it should be applied to traces, metrics, and logs.

PR #2302, opened 2026-08-20
feat: add flags to metrics data model

xuan-cao-swi:
## Description

As per [spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/data-model.md#metric-points), flags should be in every data points.


mwear (commented):
This also lgtm, but also has some rubocop offenses to address.

mwear (approved):

simi (approved):

kaylareopelle (commented):
This is great, thank you! What do you think about adding some tests for these changes?

kaylareopelle (approved):

PR #2305, opened 2026-08-21
chore: Renovate updatable version array

thompson-tomo:
This switches the appraisal syntax to one which renovate could update going forward. No appraisals have been removed but some added.

### DIFF
--- a/exporter/otlp-logs/Appraisals
+++ b/exporter/otlp-logs/Appraisals
@@ -4,18 +4,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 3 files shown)


simi:
just to make it clear, this needs https://github.com/open-telemetry/opentelemetry-ruby/pull/2304 per my understanding

thompson-tomo:
When that is merged the versions are updated going forward but without this is just a cleanup of the existing. Hence no issue with this landing before the other.

mwear (approved):

kaylareopelle (approved):

PR #2309, opened 2026-08-24
feat!: update api for spec compliance

xuan-cao-swi:
## Description

This is purely for specification compliance without actual sdk implementation
1. add the bind interface ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#bind)).
2. add the advisory interface ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#instrument-advisory-parameters))
3. add the `enabled` interface ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#instrument-advisory-parameters))
4. change callable to callback because callback is more compliance in terms of naming of the callable parameters (this need to review since it's a big parameter change)
5. allow no callback passed when create the asynchronous instrument (`Users can provide callback functions, but it is up to their discretion. Therefore, this API MUST be structured to accept a variable number of callback functions, including none.`)
6. for interface, each individual instrument can register callback by themselves rather than relying on meter only (`The API SHOULD support registration of callback functions associated with asynchronous instruments after they are created.`)
7. add more test case to cover api (originally less than 20 assertions, now it's around 90 assertions)


metrics_api/lib/opentelemetry/metrics/instrument/up_down_counter.rb
```diff
@@ -17,6 +17,28 @@ class UpDownCounter
         #   Array values must not contain nil elements and all elements must be of
         #   the same basic type (string, numeric, boolean).
         def add(amount, attributes: {}); end
+
+        # Returns whether this UpDownCounter is enabled for recording measurements.
+        #
+        # @return [Boolean] default to true in noop implementation
+        def enabled?
+          true
+        end
```
kaylareopelle: I think the [spec also wants a line like this in the documentation](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md?plain=1#L498-L500): 
```suggestion
        # Returns whether this UpDownCounter is enabled for recording 
        #  measurements.
        #
        # Instrumentation authors need to call this API each time 
        #  they record a measurement to ensure they have the most
        #  up-to-date value.
        # 
        # @return [Boolean] default to true in noop implementation
        def enabled?
          true
        end
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -55,10 +55,14 @@ def initialize
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::SimpleFixedSizeExemplarReservoir` - uniformly-weighted sampling
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::AlignedHistogramBucketExemplarReservoir` - one exemplar per histogram bucket (recommended for histograms)
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::NoopExemplarReservoir` - no-op, disables exemplar collection
+      # @param advisory [optional Hash] optional advisory parameters: recommendations from the
+      #   instrument author that implementations MAY ignore. The API does not validate them.
+      #   - `:attributes` [Array<String>] recommended set of attribute keys for the resulting metrics.
+      #   `:explicit_bucket_boundaries` applies to histograms only and has no effect here.
```
kaylareopelle: There's another way we can document this in YARD that might be easier to read:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -86,10 +90,15 @@ def create_counter(name, unit: nil, description: nil, exemplar_filter: nil, exem
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::SimpleFixedSizeExemplarReservoir` - uniformly-weighted sampling
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::AlignedHistogramBucketExemplarReservoir` - one exemplar per histogram bucket (recommended for histograms)
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::NoopExemplarReservoir` - no-op, disables exemplar collection
+      # @param advisory [optional Hash] optional advisory parameters: recommendations from the
+      #   instrument author that implementations MAY ignore. The API does not validate them.
+      #   - `:explicit_bucket_boundaries` [Array<Numeric>] recommended bucket boundaries when this
+      #     instrument is aggregated to an explicit bucket histogram.
+      #   - `:attributes` [Array<String>] recommended set of attribute keys for the resulting metrics.
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Recommended bucket boundaries when
      #     this instrument is aggregated to an explicit bucket histogram.
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -115,10 +124,14 @@ def create_histogram(name, unit: nil, description: nil, exemplar_filter: nil, ex
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::SimpleFixedSizeExemplarReservoir` - uniformly-weighted sampling
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::AlignedHistogramBucketExemplarReservoir` - one exemplar per histogram bucket (recommended for histograms)
       #   - `OpenTelemetry::SDK::Metrics::Exemplar::NoopExemplarReservoir` - no-op, disables exemplar collection
+      # @param advisory [optional Hash] optional advisory parameters: recommendations from the
+      #   instrument author that implementations MAY ignore. The API does not validate them.
+      #   - `:attributes` [Array<String>] recommended set of attribute keys for the resulting metrics.
+      #   `:explicit_bucket_boundaries` applies to histograms only and has no effect here.
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -144,10 +157,14 @@ def create_gauge(name, unit: nil, description: nil, exemplar_filter: nil, exempl
...
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -177,10 +194,14 @@ def create_up_down_counter(name, unit: nil, description: nil, exemplar_filter: n
...
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -211,10 +232,14 @@ def create_observable_counter(name, callback:, unit: nil, description: nil, exem
...
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -245,15 +270,39 @@ def create_observable_gauge(name, callback:, unit: nil, description: nil, exempl
...
```
kaylareopelle:
```suggestion
      #   @option advisory [Array<String>] :attributes Recommended set of attribute keys for the resulting metrics.
      #   @option advisory [Array<Numeric>] :explicit_bucket_boundaries Applies to histograms only and has no effect here.
```

metrics_api/lib/opentelemetry/metrics/instrument/counter.rb
```diff
@@ -17,6 +17,28 @@ class Counter
         #   Array values must not contain nil elements and all elements must be of
         #   the same basic type (string, numeric, boolean).
         def add(increment, attributes: {}); end
+
+        # Returns whether this Counter is enabled for recording measurements.
+        #
+        # @return [Boolean] default to true in noop implementation
```
kaylareopelle, line 27: I think the [spec also wants a line like this in the documentation](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md?plain=1#L498-L500): 

```suggestion
        # Returns whether this Counter is enabled for recording measurements.
        #
        # Instrumentation authors need to call this API each time 
        #  they record a measurement to ensure they have the most
        #  up-to-date value.
        #
        # @return [Boolean] default to true in noop implementation
```

metrics_api/lib/opentelemetry/metrics/instrument/gauge.rb
```diff
@@ -17,6 +17,29 @@ class Gauge
         #   Array values must not contain nil elements and all elements must be of
         #   the same basic type (string, numeric, boolean).
         def record(amount, attributes: {}); end
+
+        # Returns whether this Gauge is enabled for recording measurements.
+        #
+        # @return [Boolean] default to true in noop implementation
+        def enabled?
```
kaylareopelle, line 28:
```suggestion
        # Returns whether this Gauge is enabled for recording measurements.
        #
        # Instrumentation authors need to call this API each time 
        #  they record a measurement to ensure they have the most
        #  up-to-date value.
        #
        # @return [Boolean] default to true in noop implementation
        def enabled?
```

metrics_api/lib/opentelemetry/metrics/instrument/histogram.rb
```diff
@@ -17,6 +17,29 @@ class Histogram
         #   Array values must not contain nil elements and all elements must be of
         #   the same basic type (string, numeric, boolean).
         def record(amount, attributes: {}); end
+
+        # Returns whether this Histogram is enabled for recording measurements.
+        #
+        # @return [Boolean] default to true in noop implementation
+        def enabled?
```
kaylareopelle, line 28:
```suggestion
        # Returns whether this Histogram is enabled for recording measurements.
        #
        # Instrumentation authors need to call this API each time 
        #  they record a measurement to ensure they have the most
        #  up-to-date value.
        #
        # @return [Boolean] default to true in noop implementation
        def enabled?
```

kaylareopelle (approved):
Thanks for pulling this together, @xuan-cao-swi! I want to make sure our changelog captures this correctly. Since this contains breaking changes to the API, would `feat!` be a better fit than `chore!` here, or did you have a specific workflow in mind?

I have comments for two docs things: one related to YARD formatting and the other related to the spec recommendations.

Also, I don't see any uses of `callable` anymore, so I think this should be a safe migration.

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
```diff
@@ -8,25 +8,28 @@ module OpenTelemetry
...
+        @exemplar_filter      = exemplar_filter
+        @exemplar_reservoir   = exemplar_reservoir
+        @advisory             = advisory
+        @registered_callbacks = []
         @delegate = nil
       end
 
       # Replaces the no-op delegate with a real instrument from the given meter.
       def upgrade_with(meter)
         @delegate = case @kind
```
mwear, line 26: This is existing, but I thought I would mention that we're missing a case for `gauge`. We could fix it here or do it as a follow up.

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
```diff
@@ -8,25 +8,28 @@ module OpenTelemetry
...
                     when :counter, :histogram, :up_down_counter
-                      meter.send("create_#{@kind}", @name, unit: @unit, description: @desc, exemplar_filter: @exemplar_filter, exemplar_reservoir: @exemplar_reservoir)
+                      meter.send("create_#{@kind}", @name, unit: @unit, description: @desc, exemplar_filter: @exemplar_filter, exemplar_reservoir: @exemplar_reservoir, advisory: @advisory)
                     when :observable_counter, :observable_gauge, :observable_up_down_counter
-                      meter.send("create_#{@kind}", @name, unit: @unit, description: @desc, exemplar_filter: @exemplar_filter, exemplar_reservoir: @exemplar_reservoir, callback: @callback)
+                      meter.send("create_#{@kind}", @name, unit: @unit, description: @desc, exemplar_filter: @exemplar_filter, exemplar_reservoir: @exemplar_reservoir, callback: @callback, advisory: @advisory)
                     end
+        @registered_callbacks.each { |callback| @delegate.register_callback(callback) }
```
mwear, line 32: We need ensure this only runs on delegates that have a `register_callback` method, e.g. asynchronous instruments.

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
```diff
@@ -38,6 +41,29 @@ def add(amount, attributes: nil)
...
+      end
+
+      # Delegates to the underlying instrument, or removes the callback from the queue.
+      def unregister(callback)
+        if @delegate
+          @delegate.unregister(callback)
+        else
+          @registered_callbacks.delete(callback)
+        end
+      end
```
mwear, line 66: I notice we add `bind` methods to instruments below, but not to the proxy. If we call `bind` on any instrument before the SDK is installed, we'll get a `NoMethodError`.

metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -245,15 +270,39 @@ def create_observable_gauge(name, callback:, unit: nil, description: nil, exempl
...
+
+      # Registers a callback function against a declared set of asynchronous instruments from this Meter.
+      #
+      # @param [Array] instruments the asynchronous instruments the callback function reports Measurements for
+      # @param [Proc] callback the callback function
+      #
+      # @return [nil]
+      def register_callback(instruments, callback)
+        instruments.each { |instrument| instrument.register_callback(callback) }
+      end
```
mwear, line 291: I don't think this matches the spec for multiple-instrument callbacks. ([api.md, Asynchronous Instrument API](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#asynchronous-instrument-api)):

> Idiomatic APIs for multiple-instrument Callbacks MUST distinguish the instrument associated with each observed `Measurement` value.

The spec gives an example of how this should look [here](https://github.com/open-telemetry/opentelemetry-specification/blob/v1.60.0/specification/metrics/api.md#multiple-instrument-callbacks). This is how it'd translate to Ruby:

```ruby
registration = meter.register_callback([usage, pressure]) do |result|
  usage_value, pressure_value = expensive_system_call
  result.observe(usage,    usage_value,    attributes: { 'device' => id })
  result.observe(pressure, pressure_value, attributes: { 'device' => id })
end

registration.unregister
```

The idea is that we can fanout to n instruments, but only call the `callback` once. As this is currently implemented, we call the callback n times for n instruments.

The example above matches what Go and JS do. They pass an observer to the callback. That way you can perform one expensive call and the values can be routed to the instrument via the call to observe.

Fixing this properly means changing the SDK callback contract, so it should probably be it's own PR, but open to however is easiest to handle it.

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
xuan-cao-swi: Added the `gauge`

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
xuan-cao-swi: Added the guard `if @delegate.respond_to?(:register_callback)`

metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
xuan-cao-swi: Added

metrics_api/lib/opentelemetry/metrics/meter.rb
xuan-cao-swi: yeah, I agree we should work on this in another PR since it will be a big change

mwear (approved):
LGTM so long as we follow-up with a PR to bring our async instrument API inline with the spec.

PR #2310, opened 2026-08-25
ci: migrate Scorecard to shared workflow

opentelemetrybot:
Migrate OSSF Scorecard to the shared workflow. This limits code scanning alerts from Scorecard to `BinaryArtifactsID`, `DangerousWorkflowID`, `PinnedDependenciesID`, and `TokenPermissionsID`. See open-telemetry/sig-security#309 for details.

The upstream-only job guard is also removed so the same exact workflow can be qualified on the controlled bot fork.


thompson-tomo (approved):

simi (approved):

kaylareopelle (approved):
Our previous workflow scoped runs to only the main `open-telemetry` owned repo so it wouldn't run on forks. I believe the runs used to fail. After running this workflow on my fork of semantic-conventions, there weren't any failures, so I'm fine leaving off the repo scope.

mwear (approved):

PR #2312, opened 2026-08-25
ci: Add opentelemetry-exporter-otlp-common to releases

thompson-tomo:
This adds exporter-otlp-common to the release process which is necessary due to exporter-otlp listing this as a dependency.

This moves the grpc services out of the generic common as they are only applicable for the grpc exporter hence shouldn't be in common.


simi (approved):

kaylareopelle (approved):
Thank you!

mwear (changes requested):
`google-protobuf` can be a problematic dependency if the host application also uses it. We need to ensure that we use an open version for it in `opentelemetry-exporter-otlp-common` as we do in `opentelemetry-exporter-oltp` (e.g. `spec.add_dependency 'google-protobuf', '>= 3.18'`) to minimize any compatibility issues.

thompson-tomo:
@mwear have changed it.

mwear (approved):

PR #2320, opened 2026-08-29
ci: Replicate contrib codeql

thompson-tomo:
This adds codeql as was done for contrib.


thompson-tomo:
@kaylareopelle it looks like this repo needs the same change as contrib.

PR #2342, opened 2026-09-08
chore: Replicate simplecov changes to config gems

thompson-tomo:
This replicates the changes in https://github.com/open-telemetry/opentelemetry-ruby/pull/2117 to the config gem as it was missed

### DIFF
--- a/config/Gemfile
+++ b/config/Gemfile
@@ -10,6 +10,7 @@ eval_gemfile File.expand_path('../Gemfile', __dir__)
 gemspec
 
 group :test, :development do
+  gem 'dotenv', '3.2.0'
   gem 'minitest', '6.0.6'
   gem 'simplecov', '1.0.0'
   gem 'yard', '0.9.45'
--- a/config/test/test_helper.rb
+++ b/config/test/test_helper.rb
@@ -3,6 +3,9 @@
 # Copyright The OpenTelemetry Authors
 # SPDX-License-Identifier: Apache-2.0
 
+require 'dotenv'
+Dotenv.load(File.expand_path('.env', __dir__))
+
 require 'simplecov'
 SimpleCov.start
 


config/Gemfile
```diff
@@ -10,6 +10,7 @@ eval_gemfile File.expand_path('../Gemfile', __dir__)
 gemspec
 
 group :test, :development do
+  gem 'dotenv', '3.2.0'
```
simi, line 13: This is in already evaled root Gemfile 🤔. https://github.com/open-telemetry/opentelemetry-ruby/blob/ab29f8582ec55ed19276e6b4289fb40683541b58/Gemfile#L9

thompson-tomo: You are correct but I would rather have it as it is to be consistent with all the other gems & we can tackle further optimisations in a future pr.

simi: Doesn't that emit warning (or failure) in bundler?

thompson-tomo: Neither, it is just an info

```log
2026-09-08T14:53:01.2947238Z Your Gemfile lists the gem dotenv (~> 3.2.0) more than once.
2026-09-08T14:53:01.2948161Z You should probably keep only one of them.
2026-09-08T14:53:01.2948774Z Remove any duplicate entries and specify the gem only once.
2026-09-08T14:53:01.2949671Z While it's not a problem now, it could cause errors if you change the version of one of them later.
```

We have renovate which will update both of them hence we would avoid the error scenario

kaylareopelle (approved):

config/Gemfile
kaylareopelle: I hear @simi's concern, but agree with @thompson-tomo that we should address it in a different PR. @thompson-tomo would you mind opening an issue to address this?

thompson-tomo: Done https://github.com/open-telemetry/opentelemetry-ruby/issues/2391

mwear (approved):

PR #2375, opened 2026-09-09
fix(logs-api)!: align LoggerProvider#logger with the SDK signature

mwear:
Obtaining a logger before `SDK.configure` breaks SDK configuration. See below for a repro:

```ruby
OpenTelemetry.logger_provider.logger('component', '1.0')  # registers a ProxyLogger
OpenTelemetry::SDK.configure
# ERROR -- : OpenTelemetry error: unexpected configuration error due to
#            wrong number of arguments (given 2, expected 0; required keyword: name)
```

This is because `ProxyLoggerProvider#delegate=` calls `provider.logger(key.name, key.version)` positionally, but the logs SDK takes a required `name:` keyword, so upgrading any registered proxy logger raises. This PR aligns `LoggerProvider#logger` with the SDK signature.

This is technically a breaking change because API only users could previously write `logger('name', '1.0')`, but they now have to change to `logger(name: 'name', version: '1.0')`.

### DIFF
--- a/logs_api/lib/opentelemetry/internal/proxy_logger_provider.rb
+++ b/logs_api/lib/opentelemetry/internal/proxy_logger_provider.rb
@@ -38,19 +38,19 @@ module OpenTelemetry
 
         @mutex.synchronize do
           @delegate = provider
-          @registry.each { |key, logger| logger.delegate = provider.logger(key.name, key.version) }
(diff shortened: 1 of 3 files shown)


simi (approved):

kaylareopelle (approved):
Thank you!

PR #2376, opened 2026-09-09
fix(logs-api): forward on_emit arguments to the delegate logger

mwear:
Log records emitted through an upgraded proxy are empty. The proxy accepts the log record fields, but it passes `nil`s for all of them to the delegate. This affects any logger obtained before the logs SDK is configured.

This PR fixes that by forwarding the actual values through the delegate.


kaylareopelle (approved):
Another great fix. Thank you!

PR #2377, opened 2026-09-09
fix(metrics-api): pass the callback through when upgrading a proxy

mwear:
`ProxyInstrument#initialize` stores the callback in `@callable`, but `upgrade_with` reads `@callback`, which is always nil. So when the metrics SDK is installed, observable instruments created against the proxy are rebuilt with `callback: nil` and never report. This affects `observable_counter`, `observable_gauge` and `observable_up_down_counter` created before the SDK is configured. 

This PR renames the parameter to `callback` and adds test coverage.

### DIFF
--- a/metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
+++ b/metrics_api/lib/opentelemetry/internal/proxy_instrument.rb
@@ -8,12 +8,12 @@ module OpenTelemetry
   module Internal
     # @api private
     class ProxyInstrument
-      def initialize(kind, name, unit, desc, callable, exemplar_filter, exemplar_reservoir)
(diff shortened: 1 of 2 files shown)


kaylareopelle (commented):
Thanks for submitting this fix. This looks like it's also part of #2309, but has a slightly different testing approach. 

I don't have a strong preference for which approach we take. Since #2309 is larger, I'm somewhat inclined to merge this PR first and then fix merge conflicts in #2309 so we might theoretically get a fix out to users sooner. Though, they could both be merged/ready at about the same time.

@xuan-cao-swi - thoughts?

xuan-cao-swi:
@kaylareopelle Yes, I'd like this PR get merged first

xuan-cao-swi (approved):

PR #2384, opened 2026-09-09
refactor!: make metrics and logs instrumentation-only

mwear:
Draft for discussion. Core half of making the unstable metrics and logs APIs usable by instrumentation without exposing them to users. The contrib half of this work is here: open-telemetry/opentelemetry-ruby-contrib/pull/2569.

## Problem

Requiring `opentelemetry-metrics-api` or `opentelemetry-logs-api` immediately defines `OpenTelemetry.meter_provider` and `OpenTelemetry.logger_provider`. Those are the public, user-facing surface of unstable APIs. So instrumentation cannot depend on either API without exposing them to users via the globals on the `OpenTelemetry` module.

Separating the two lets us exercise and evolve these APIs against our instrumentation without committing users to them, and gives us much more confidence before we stabilize.

## Change

Requiring the API does not need to set the global on the `OpenTelemetry` module. In this PR that responsibility is moved to the SDK. This allows us to consume the API in instrumentation without exposing it to users via the global. The changes are as follows:

- the provider slot moves to `OpenTelemetry::Internal`
- the public accessors move to `opentelemetry/metrics/global.rb` and `opentelemetry/logs/global.rb`
- each SDK requires its own `global.rb` (the API does not)

| you require | instrumentation gets | user gets |
| --- | --- | --- |
| `opentelemetry-metrics-api` | a meter, via `Internal.meter_provider` | no `OpenTelemetry.meter_provider` |
| `opentelemetry-metrics-sdk` | the same meter, backed by the SDK | `OpenTelemetry.meter_provider` |

Requiring the SDK is the opt-in. Logs work the same way. Nothing changes for anyone already using the SDKs.

## Breaking

Only for code that requires an `-api` gem directly and reads the top-level accessor (without an SDK). Requiring the SDK restores it.

## Notes

- Release sequencing needs care. This will require coordinating changes across contrib, the API, and the SDK.

### DIFF
--- a/logs_api/lib/opentelemetry-logs-api.rb
+++ b/logs_api/lib/opentelemetry-logs-api.rb
@@ -9,33 +9,11 @@ require 'opentelemetry/logs'
 require 'opentelemetry/logs/version'
 require 'opentelemetry/internal/proxy_logger_provider'
 require 'opentelemetry/internal/proxy_logger'
+require 'opentelemetry/internal/global_logger_provider'
 
 # OpenTelemetry is an open source observability framework, providing a
 # general-purpose API, SDK, and related tools required for the instrumentation
 # of cloud-native software, frameworks, and libraries.
 #
-# The OpenTelemetry module in the Logs API gem provides global accessors
-# for logs-related objects.
-module OpenTelemetry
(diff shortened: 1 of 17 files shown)


mwear:
Superseded by #2415 and #2416

PR #2389, opened 2026-09-10
feat(metrics_sdk): support OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION env var

### LINKED ISSUE #2372: Read `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION`
## Spec requirement
The `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` env var (`explicit_bucket_histogram`/`base2_exponential_bucket_histogram`, default `explicit_bucket_histogram`) SHOULD be supported ([sdk_exporters/otlp.md:43-70](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/sdk_exporters/otlp.md#L43-L70)).

## Current behavior
The default (`ExplicitBucketHistogram`, `instrument/histogram.rb` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb)) default aggregation) is correct, but the env var itself is never read in `exporter/otlp-metrics` or `metrics_sdk`. A user can't switch the default to exponential histograms via configuration, only by explicitly registering a `View`.

## Suggested fix
Read `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` in the OTLP exporter or `metrics_sdk` configurator, and use it to select the default histogram aggregation when no View overrides it.

OTLP-3

Abhayanthk:
## Summary

Adds support for `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` and, per the review discussion, makes the **exporter/reader** the source of truth for default aggregation as a function of instrument kind. The `MeterProvider` and `Meter` only coordinate registration — they pass the reader's resolved configuration down to the metric stream, which builds the aggregation. Readers and the OTLP exporter also accept an explicit `default_aggregation:` parameter, so a preference can be set without the env var.

Fixes #2372

## For reviewers

The commits are meant to be read in order, since the squash hides the structure:
1. **fix** (`8cff36b`) — the deadlock fix, self-contained and independent of the rest.
2. **refactor** (`64e7518`) — resolve default aggregation on the reader, build it in the stream.
3. **feat** (`5660bd1`) — add the explicit `default_aggregation:` option.

## What changed

**Exporter drives the env var**
- The OTLP `MetricsExporter` reads and validates `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` (case-insensitive): `base2_exponential_bucket_histogram` → `ExponentialBucketHistogram`, anything else → `ExplicitBucketHistogram` with a logged warning on unrecognized values. It turns the env var into a preference in its constructor.

**Resolved once, on the reader**
- `MetricReader::DEFAULT_AGGREGATION` holds the spec's instrument-kind → aggregation-class table. Any reader or exporter preference is merged over it when the reader is constructed, not looked up per instrument. `PeriodicMetricReader` defers to its wrapped exporter.

**Provider and meter only coordinate**
- `MeterProvider` and `Meter` pass the reader's resolved hash along; `MetricStream` builds the aggregation, since it owns the instrument kind and the `exemplar_reservoir` each aggregation instance needs. The old `resolve_aggregation` path in `MeterProvider` is gone. The preference is stored as aggregation **classes**, not configured instances (the difference from Python): Ruby aggregation objects hold per-stream state, so each stream instantiates its own.
- `Meter#add_metric_reader` re-registers existing instruments with the newly added reader's preference, so an exporter attached *after* an instrument is created still gets its aggregation.

**New optional `default_aggregation:`**
- `MetricReader` and the OTLP exporter now accept an optional `default_aggregation:` — a hash of instrument kind to aggregation class, applied on top of the spec default. Precedence: explicit parameter, then env var, then spec default; a matching View still overrides all of them. This gives pull readers (which never see the env var) a way to express a preference.

**Temporality (flagged for confirmation — see comment below)**
- With the per-instrument defaults gone, `determine_temporality` now handles UpDownCounter / ObservableUpDownCounter directly, returning Cumulative regardless of the delta/low-memory preference, per the table in that file. This was previously masked by UpDownCounter hardcoding cumulative in its own `default_aggregation`. Flagged in a comment for a sanity-check since it touches pre-existing behavior.

**Deadlock fix**
- `Meter#add_metric_reader` no longer takes the Meter mutex. Instrument creation takes the Meter mutex and then the MeterProvider mutex; `add_metric_reader` runs while already holding the provider mutex, so taking the Meter mutex there inverted the lock order and could hang the process. The lock order is now documented so it isn't reintroduced. (The unsynchronized registry iteration is pre-existing on `main` and left unchanged.)

## Design notes

Two choices I made deliberately — happy to change either if you'd prefer:
- **Values are classes, not callables.** `default_aggregation` maps to aggregation classes, so custom bucket boundaries still go through a View rather than a configured instance here.
- **No validation of entries yet.** An unknown instrument kind in the hash is currently ignored rather than raising (Python raises on an unknown instrument class). I can add validation if you'd like it.

## Tests
- New `metric_reader_test.rb` covering the reader's resolved default aggregation across both registration paths — a reader added before *and* after the instrument is created — plus the fallback when a reader doesn't implement `default_aggregation`. It also asserts the frozen `DEFAULT_AGGREGATION` table isn't mutated by a caller's hash.
- Integration test (`periodic_metric_reader_test.rb`) with two readers requesting different histogram aggregations, verifying they stay isolated while sharing one instrument (the multiple-readers case raised in review).
- OTLP exporter tests for the `default_aggregation:` parameter and the parameter → env var → spec-default precedence.
- Deadlock regression test that fails when the Meter mutex is present in `add_metric_reader`.
- `metrics_sdk` and `otlp-metrics` suites pass locally (CI awaiting maintainer approval).

## Docs
- `exporter/otlp-metrics/README.md` now documents:
  - the env var and its two accepted values (default `explicit_bucket_histogram`), matched case-insensitively, with an unrecognized value logging a warning and falling back to explicit buckets;
  - the `default_aggregation:` parameter and its precedence over the env var, with a runnable usage example;
  - that `default_aggregation:` is also available on `MetricReader` and its subclasses, so pull readers can set a preference.

### DIFF
--- a/exporter/otlp-metrics/README.md
+++ b/exporter/otlp-metrics/README.md
@@ -109,9 +109,12 @@ The collector exporter can be configured explicitly in code, or via environment 
 | `headers:` | `OTEL_EXPORTER_OTLP_HEADERS` | |
 | `compression:` | `OTEL_EXPORTER_OTLP_COMPRESSION` | `"gzip"` |
 | `timeout:` | `OTEL_EXPORTER_OTLP_TIMEOUT` | `10` |
+| | `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` | `"explicit_bucket_histogram"` |
 | `ssl_verify_mode:` | `OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_PEER` or `OTEL_RUBY_EXPORTER_OTLP_SSL_VERIFY_NONE` | `OpenSSL::SSL:VERIFY_PEER` |
 
 `ssl_verify_mode:` parameter values should be flags for server certificate verification: `OpenSSL::SSL:VERIFY_PEER` and `OpenSSL::SSL:VERIFY_NONE` are acceptable. These values can also be set using the appropriately named environment variables as shown where `VERIFY_PEER` will take precedence over `VERIFY_NONE`.  Please see [the Net::HTTP docs](https://ruby-doc.org/stdlib-2.7.6/libdoc/net/http/rdoc/Net/HTTP.html#verify_mode) for more information about these flags.
+
+`OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` parameter values can be either `"explicit_bucket_histogram"` (the default) or `"base2_exponential_bucket_histogram"`.
 
 ## How can I get involved?
 
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
@@ -35,9 +35,13 @@ module OpenTelemetry
           private
 
           def default_aggregation
-            # TODO: at this point, histogram always take the default DEFAULT_BOUNDARIES. In future, the histogram should be able to
-            # define the custom boundaries so for exemplar_reservoir can also takes the bounaries as parameter
-            OpenTelemetry::SDK::Metrics::Aggregation::ExplicitBucketHistogram.new(exemplar_reservoir: @exemplar_reservoir)
+            if ENV['OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION'] == 'base2_exponential_bucket_histogram'
+              OpenTelemetry::SDK::Metrics::Aggregation::ExponentialBucketHistogram.new(exemplar_reservoir: @exemplar_reservoir)
+            else
+              # TODO: at this point, histogram always take the default DEFAULT_BOUNDARIES. In future, the histogram should be able to
+              # define the custom boundaries so for exemplar_reservoir can also takes the bounaries as parameter
+              OpenTelemetry::SDK::Metrics::Aggregation::ExplicitBucketHistogram.new(exemplar_reservoir: @exemplar_reservoir)
+            end
           end
         end
       end
--- a/metrics_sdk/test/opentelemetry/sdk/metrics/instrument/histogram_test.rb
+++ b/metrics_sdk/test/opentelemetry/sdk/metrics/instrument/histogram_test.rb
@@ -35,4 +35,23 @@ describe OpenTelemetry::SDK::Metrics::Instrument::Histogram do
     _(last_snapshot[0].data_points[0].attributes).must_equal('foo' => 'bar')
     _(last_snapshot[0].aggregation_temporality).must_equal(:cumulative)
   end
+
+  describe 'with exponential buckets configured' do
+    before do
+      @original_env = ENV['OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION']
+      ENV['OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION'] = 'base2_exponential_bucket_histogram'
+    end
+
+    after do
+      ENV['OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION'] = @original_env
+    end
+
+    it 'uses exponential bucket aggregation' do
+      histogram.record(5, attributes: { 'foo' => 'bar' })
+      metric_exporter.pull
+      last_snapshot = metric_exporter.metric_snapshots
+      
+      _(last_snapshot[0].data_points[0]).must_be_instance_of(OpenTelemetry::SDK::Metrics::Aggregation::ExponentialHistogramDataPoint)
+    end
+  end
 end


Abhayanthk:
@kaylareopelle, thanks for the clear write-up on this one!

I traced through the code and it looks like the `Base2ExponentialBucketHistogram` aggregation is already implemented — the missing piece is just reading `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` and using it to pick the default when no View overrides it.

I noticed the issue mentioned checking this in the OTLP exporter or the `metrics_sdk` configurator. However, because the Ruby SDK builds and shares instrument aggregations globally across all exporters, parsing the env var strictly inside the OTLP exporter wouldn't cleanly propagate back up to the instrument without a large rewrite of the pipeline.

To keep things pragmatic and robust, I've added the env var check directly inside the `default_aggregation` method in `metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb`. This swaps the default to exponential buckets exactly when the user requests it, without breaking the pipeline for other exporters.

Tests are written, and I've pushed the commit. Let me know what you think!

Abhayanthk:
I also updated `exporter/otlp-metrics/README.md` to document both options. I put the docs there rather than in `metrics_sdk` because users will naturally look in the exporter README when configuring an `OTEL_EXPORTER_OTLP_*` variable.

xuan-cao-swi:
`OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` is supposed to used in exporter, not the actual aggregation class level.

See [python](https://github.com/open-telemetry/opentelemetry-python/blob/v1.44.0/exporter/opentelemetry-exporter-otlp-common/src/opentelemetry/exporter/otlp/common/_aggregation.py#L10) and [js](https://github.com/open-telemetry/opentelemetry-js/blob/v2.11.0/experimental/packages/configuration/src/EnvironmentConfigFactory.ts#L506) implementation

Abhayanthk:
Thanks @xuan-cao-swi, you're totally right that conceptually this belongs in the Exporter (like the Python and JS implementations).

However, looking at the current situation in the Ruby SDK, `MetricReader` doesn't currently have the plumbing to pass a `preferred_aggregation` dictionary back up to the `MeterProvider`. When a `MetricReader` is added, the `MeterProvider` just pushes it to an array, and when an instrument is registered via `register_with_new_metric_store`, it relies entirely on the instrument's own `default_aggregation` method.

So, if I were to implement it in the Exporter here, this is what I would have to do: I'd need to refactor `MetricReader` to expose a `default_aggregation(instrument_type)` method, and update `register_synchronous_instrument` to query the reader for its preference and pass it down the chain.

I’m happy to take this on and do it properly — just let me know if you’d like me to go ahead with the refactor!

thompson-tomo (changes requested):
While this will work, after reading https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/ i don't think what we are doing here is spec compliant and would make declarative config very difficult.

> OTLP Metrics Exporter MUST provide configuration to influence the MetricReader default aggregation as a function of instrument kind. This option MAY be named default_aggregation, and MUST set the default aggregation for all instrument kinds by default.

So my understanding is that metric exporter needs to set a function/proc on the metric reader which can they be called when creating the instrument.

The difficulty is a metric provider may have multiple metric readers.

Abhayanthk:
@thompson-tomo @xuan-cao-swi

Thanks for the excellent review and for catching that! You're totally right — putting the ENV check directly in the Histogram bypassed the Exporter entirely, which violated the spec and would have made the upcoming declarative configuration work much harder.

I've pushed a refactor that moves the decision up into the exporter → reader → provider pipeline, as you suggested.

What changed in this push:

1. **Reverted `histogram.rb`** — it no longer reads any environment variables.
2. **Built the delegation pipeline** — added a `default_aggregation(instrument_kind)` method to the `MetricReader` base class (returns `nil` by default). `PeriodicMetricReader` delegates this call to its wrapped exporter.
3. **OTLP exporter drives the config** — `MetricsExporter` now reads the ENV var and is the source of truth, returning the `ExponentialBucketHistogram` class when queried for `:histogram`.
4. **MeterProvider queries the readers** — in `register_synchronous_instrument`, it loops through the readers, asks each for its preferred aggregation, and passes that down into the new metric store (preserving the user's custom `exemplar_reservoir`).

**On the multiple-readers case you flagged:** I added an integration test (`test/integration/periodic_metric_reader_test.rb`) that spins up two readers — one wanting explicit buckets, one wanting exponential — and verifies the MeterProvider isolates them, feeding the correct bucket type to each exporter from a single shared instrument.

I've run the full `metrics_sdk` and `otlp-metrics` test suites locally and everything passes, including the new coverage for this change.

exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
```diff
@@ -80,6 +80,16 @@ def initialize(endpoint: OpenTelemetry::Common::Utilities.config_opt('OTEL_EXPOR
...
 
+          # Returns the default aggregation class for the given instrument kind, or nil if none.
+          def default_aggregation(instrument_kind)
+            return nil unless instrument_kind == :histogram
+            if ENV['OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION'] == 'base2_exponential_bucket_histogram'
+              OpenTelemetry::SDK::Metrics::Aggregation::ExponentialBucketHistogram
+            else
+              nil
+            end
+          end
```
thompson-tomo, line 100: As per https://opentelemetry.io/docs/specs/otel/metrics/sdk_exporters/otlp/#additional-environment-variable-configuration we should support handling `explicit_bucket_histogram` when it is set.

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
```diff
@@ -40,6 +40,11 @@ def initialize(export_interval_millis: Float(ENV.fetch('OTEL_METRIC_EXPORT_INTER
             start
           end
 
+          # Returns the default aggregation class for the given instrument kind, or nil if none.
+          def default_aggregation(instrument_kind)
+            @exporter.respond_to?(:default_aggregation) ? @exporter.default_aggregation(instrument_kind) : nil
```
thompson-tomo: We should return the default as defined in We should also support https://opentelemetry.io/docs/specs/otel/metrics/sdk/#default-aggregation

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
```diff
@@ -35,8 +35,6 @@ def record(amount, attributes: {})
           private
 
           def default_aggregation
-            # TODO: at this point, histogram always take the default DEFAULT_BOUNDARIES. In future, the histogram should be able to
-            # define the custom boundaries so for exemplar_reservoir can also takes the bounaries as parameter
             OpenTelemetry::SDK::Metrics::Aggregation::ExplicitBucketHistogram.new(exemplar_reservoir: @exemplar_reservoir)
           end
```
thompson-tomo: Is this function even needed

metrics_sdk/test/opentelemetry/sdk/metrics/instrument/histogram_test.rb
```diff
@@ -35,4 +35,6 @@
     _(last_snapshot[0].data_points[0].attributes).must_equal('foo' => 'bar')
     _(last_snapshot[0].aggregation_temporality).must_equal(:cumulative)
   end
+
+
```
thompson-tomo: Revert

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -130,7 +130,14 @@ def add_metric_reader(metric_reader)
         def register_synchronous_instrument(instrument)
           @mutex.synchronize do
             @metric_readers.each do |mr|
-              instrument.register_with_new_metric_store(mr.metric_store)
+              preferred_agg_class = mr.respond_to?(:default_aggregation) ? mr.default_aggregation(instrument.instrument_kind) : nil
```
thompson-tomo: We should use the default as defined in https://opentelemetry.io/docs/specs/otel/metrics/sdk/#default-aggregation

exporter/otlp-metrics/lib/opentelemetry/exporter/otlp/metrics/metrics_exporter.rb
Abhayanthk: You're right, I missed handling the explicit setting. I've pushed a fix: MetricsExporter now parses `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` and handles both `base2_exponential_bucket_histogram` and `explicit_bucket_histogram`. On an unrecognized value it logs a warning via `OpenTelemetry.logger.warn` and falls back to `explicit_bucket_histogram`, per the sdk_exporters/otlp spec.

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
Abhayanthk: Good point. Now that the MetricReader base class provides the default mappings for all instrument kinds, the pipeline relies entirely on the Reader/Exporter. I left ‎`default_aggregation` here only as a backward-compat fallback for custom third-party readers that don’t implement a ‎`default_aggregation` method yet

metrics_sdk/test/opentelemetry/sdk/metrics/instrument/histogram_test.rb
Abhayanthk: fixed

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
Abhayanthk: You're right. In my previous commit, I hadn't populated the default mappings, so `MetricReader#default_aggregation` returned nil and this fell back to the instrument defaults. I've pushed a fix: the MetricReader base class now implements the full default-aggregation table from the spec (e.g. Sum for Counters, Last Value for Gauges). PeriodicMetricReader delegates to the Exporter but uses `super` to fall back to this base mapping, so the MeterProvider always receives the correct spec-defined default class here (and passes `monotonic: true` for Counters).

metrics_sdk/lib/opentelemetry/sdk/metrics/instrument/histogram.rb
thompson-tomo: But isn't the fallback handled by https://github.com/Abhayanthk/opentelemetry-ruby/blob/7dd87c5b4db5d632fad8c00fd511ef71cf71b035/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb#L45

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
thompson-tomo: We are still setting it to nil if metric reader doesn't implement it. Shouldn't we be able to optimise this and the following block.

thompson-tomo (commented):
See my replies to existing threads.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
Abhayanthk: Good call. Since the MetricReader base class now implements the full fallback table, `preferred_agg_class` never needs to be nil. I've pushed a change based on your comment:

1. MeterProvider now falls back to the MetricReader base `default_aggregation` when a custom duck-typed reader doesn't implement the method.
2. That lets the `else` (nil) branch be removed.
3. Since we no longer fall back to the instruments, the `default_aggregation` methods in the instrument classes were dead code — I've removed them from all 7 instrument classes (including histogram.rb).

Thanks for pushing on this; the pipeline is simpler now.

Abhayanthk:
One bug I found while doing this optimization, worth flagging:

While removing the dead `default_aggregation` methods from the instruments, a test failed in `Meter#add_metric_reader`. It turned out that `add_metric_reader` was calling `register_with_new_metric_store` without passing an aggregation argument, so it silently relied on the instrument's default. That meant if a user created an instrument first and attached an exporter second, the exporter's aggregation preference (e.g. exponential buckets) was ignored.

Removing the instrument fallback is what surfaced this. I've updated `Meter#add_metric_reader` to query the metric reader for its preferred aggregation and retrofit it, matching what `MeterProvider` already does. Let me know if you'd handle it differently.

Abhayanthk:
<p>Flagging a temporality edge case I ran into during this refactor — I want to make sure I’m reading the spec right rather than assuming.</p>
<p>While moving the fallback out of the instruments and routing it through <code>AggregationTemporality.determine_temporality</code>, I think this change surfaced a pre-existing gap. Right now, when the temporality preference is <code>delta</code>, <code>determine_temporality</code> returns <code>:delta</code> for every instrument kind. But the table in that same file (and the OTLP spec) says UpDownCounter and Async UpDownCounter should stay Cumulative even under a Delta preference:</p>

Preference | Counter | Async Counter | Histogram | UpDownCounter | Async UpDownCounter
-- | -- | -- | -- | -- | --
Delta | Delta | Delta | Delta | Cumulative | Cumulative


<p>I believe this was previously masked because UpDownCounter hardcoded <code>aggregation_temporality: :cumulative</code> in its own <code>default_aggregation</code>. Since this PR removes the per-instrument defaults, that path now falls through to <code>determine_temporality</code>, which returns delta.</p>
<p>I’ve adjusted <code>determine_temporality</code> to route <code>up_down_counter</code> and <code>observable_up_down_counter</code> to cumulative so it matches the table (this also applies to the <code>LowMemory</code> preference, which had the same gap). Please sanity-check my reading — if I’ve misunderstood the intended behavior, I’m happy to revert or adjust.</p>

Abhayanthk:
@xuan-cao-swi — pushed a small follow-up (`e55e201`) fixing the RuboCop offenses from the last CI run. Green locally now. Whenever you have a moment, could you approve the workflow run? Thanks!

metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb
```diff
@@ -40,6 +40,11 @@ def initialize(export_interval_millis: Float(ENV.fetch('OTEL_METRIC_EXPORT_INTER
             start
           end
 
+          # Returns the default aggregation class for the given instrument kind.
+          def default_aggregation(instrument_kind)
+            @exporter.respond_to?(:default_aggregation) ? @exporter.default_aggregation(instrument_kind) : super
```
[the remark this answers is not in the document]
Abhayanthk: Added a fallback to the parent default so third-party exporters that don't override every instrument kind won't hit a nil. Done in e944f05.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -126,11 +126,25 @@ def add_metric_reader(metric_reader)
           end
         end
 
+        # @api private
+        def self.resolve_aggregation(metric_reader, instrument_kind, exemplar_reservoir)
+          preferred_agg_class = metric_reader.respond_to?(:default_aggregation) ? metric_reader.default_aggregation(instrument_kind) : OpenTelemetry::SDK::Metrics::Export::MetricReader.default_aggregation(instrument_kind)
+
+          kwargs = { exemplar_reservoir: exemplar_reservoir }
+          if preferred_agg_class == OpenTelemetry::SDK::Metrics::Aggregation::Sum
+            kwargs[:monotonic] = true if %i[counter observable_counter].include?(instrument_kind)
```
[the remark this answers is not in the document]
Abhayanthk: Set `monotonic` explicitly (true for counters, false for up/down counters) to preserve the prior semantics rather than relying on Sum's default. Done in 324121f.

metrics_sdk/lib/opentelemetry/sdk/metrics/meter_provider.rb
```diff
@@ -126,11 +126,25 @@ def add_metric_reader(metric_reader)
...
+            kwargs[:instrument_kind] = instrument_kind
+          end
+
+          preferred_agg_class.new(**kwargs)
+        end
```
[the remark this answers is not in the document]
Abhayanthk: Added a secondary fallback here so that a custom reader returning nil falls back to the default aggregation instead of raising. Done in 0387c1a.

metrics_sdk/lib/opentelemetry/sdk/metrics/aggregation/aggregation_temporality.rb
```diff
@@ -35,6 +35,9 @@ def cumulative
             # | **Delta**        | Delta      | Delta            | Delta      | Cumulative    | Cumulative          |
             # | **LowMemory**    | Delta      | Cumulative       | Delta      | Cumulative    | Cumulative          |
             def determine_temporality(aggregation_temporality: nil, instrument_kind: nil, default: nil)
+              # UpDownCounter and ObservableUpDownCounter MUST always be Cumulative
+              return cumulative if %i[up_down_counter observable_up_down_counter].include?(instrument_kind)
```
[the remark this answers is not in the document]
Abhayanthk: Added two focused tests confirming UpDownCounter and ObservableUpDownCounter always resolve to cumulative, even under a delta preference. Done in 022285a.

metrics_sdk/test/integration/periodic_metric_reader_test.rb
```diff
@@ -200,6 +200,65 @@
...
+        OpenTelemetry.meter_provider.add_metric_reader(mock_exporter)
+        meter = OpenTelemetry.meter_provider.meter('test')
+        histogram = meter.create_histogram('histogram')
+
+        histogram.record(5, attributes: { 'foo' => 'bar' })
+
+        mock_exporter.pull
+        last_snapshot = mock_exporter.metric_snapshots
+
+        _(last_snapshot[0].data_points[0]).must_be_instance_of(OpenTelemetry::SDK::Metrics::Aggregation::ExponentialHistogramDataPoint)
```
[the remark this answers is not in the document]
Abhayanthk: Switched the assertions to look up snapshots by metric name instead of hardcoded indices, so they're not order-dependent. Done in 1630953.

exporter/otlp-metrics/README.md
```diff
@@ -109,10 +109,13 @@ The collector exporter can be configured explicitly in code, or via environment
 | `headers:` | `OTEL_EXPORTER_OTLP_HEADERS` | |
 | `compression:` | `OTEL_EXPORTER_OTLP_COMPRESSION` | `"gzip"` |
 | `timeout:` | `OTEL_EXPORTER_OTLP_TIMEOUT` | `10` |
+| | `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` | `"explicit_bucket_histogram"` |
```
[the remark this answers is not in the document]
Abhayanthk: Fixed in 0cf6435.

Abhayanthk:
@xuan-cao-swi thanks again for the earlier review. I've addressed all six of Copilot's comments — nil fallbacks in the reader and provider, explicit monotonic flag, the UpDownCounter temporality tests, order-independent test assertions, and the README fix — and resolved the threads.

The new commits are waiting on workflow approval, so CI hasn't re-run yet. Would appreciate another look whenever you have time.

Abhayanthk:
Quick update: I've merged the latest `main` into the branch to keep it current, so the workflows are back in the "awaiting approval" state. @xuan-cao-swi, whenever you have a moment, could you approve the workflow run so CI can re-verify? All previous review comments are addressed and threads resolved.

@thompson-tomo — I believe I've covered everything from your earlier review; I would appreciate a re-review to clear the pending change request whenever you get a chance. Thanks, both!

xuan-cao-swi:
Thanks for the contribution.

The following points look good to me:

1. Define the aggregation using the `OTEL_EXPORTER_OTLP_METRICS_DEFAULT_HISTOGRAM_AGGREGATION` environment variable in the OTLP metrics exporter.
2. Remove `default_aggregation` from each instrument class.
3. Allow `MetricReader` to define the `preferred_aggregation`.

However, I'd like to discuss the following:

1. `MeterProvider` should only coordinate registration and pass the reader’s configuration to the metric stream; therefore, it might not be the right place to handle `default_aggregation`.
2. `MetricReader` / OTLP Metrics Exporter can accept aggregation as an optional parameter.

See reference from [python](https://github.com/open-telemetry/opentelemetry-python/blob/main/opentelemetry-sdk/src/opentelemetry/sdk/metrics/_internal/export/__init__.py#L171) and [js](https://github.com/open-telemetry/opentelemetry-js/blob/main/packages/sdk-metrics/src/export/MetricReader.ts)

Abhayanthk:
@xuan-cao-swi

After checking the py and js implementations, I agree that decoupling the meter provider from aggregation selection gives a clearer separation of concerns and a single source of truth for each layer. I plan to implement it by having the meter provider pass the aggregation hash to the metric stream, which will then create the aggregation object.

image</parameter>

Adding optional aggregation to MetricReader / OTLP Metrics Exporter, I'll add this in MetricReader and
MetricsExporter, as in the references. 

Can I go with this?

Abhayanthk:
@xuan-cao-swi, thanks for the review — I've pushed the changes and rewritten the PR description to match. Short version: the provider no longer decides or builds the aggregation; the default is resolved once on the reader from its exporter, and `default_aggregation:` is now an optional parameter on both `MetricReader` and the OTLP exporter. That follows the spec's own MetricReader wording:

> The default output aggregation (optional), a function of instrument kind. This function SHOULD be obtained from the exporter.

([SDK spec — MetricReader](https://opentelemetry.io/docs/specs/otel/metrics/sdk/#metricreader).) The [reader defers to its exporter](https://github.com/Abhayanthk/opentelemetry-ruby/blob/5660bd110e018f73103d6c23c63721140073fed1/metrics_sdk/lib/opentelemetry/sdk/metrics/export/periodic_metric_reader.rb#L29-L31) for exactly this.

Usage is a single optional parameter:

```ruby
OpenTelemetry::Exporter::OTLP::Metrics::MetricsExporter.new(
  default_aggregation: {
    histogram: OpenTelemetry::SDK::Metrics::Aggregation::ExponentialBucketHistogram
  }
)
```

The two choices I wasn't sure about (aggregation classes vs callables, and no validation of unknown instrument kinds yet) are under Design notes in the description if you'd like to weigh in.

One ask: CI hasn't run on any of this yet (only EasyCLA), so could you approve the workflow run when you get a chance? And @thompson-tomo — I believe everything from your Sept 12 review is addressed now; a re-review when you have a moment would be appreciated.

PR #2390, opened 2026-09-10
chore: all api gem’s are in api folder

thompson-tomo:
restructure files so that all api gem's are in the api folder. This repeats the pattern from exporters etc & most other locations.

Future pr’s this enables:

- provide an api-common so that logs/metrics do not need a dependency on api or sdk
- when logs/metrics is stabilised add it to the api gem so the api gem provides all stable api's

Potential pr's

- extract trace/context into a dedicated gem like logs and added it as a dependency to api hence no breakage. This enables plugins to only include the api they need.


kaylareopelle:
Hi @thompson-tomo, we talked about this during the SIG today. Though we may eventually want to work toward a refactor like this, now isn't the right time. We'd like to table this discussion until after the Metrics and Logs implementations reach stability.

thompson-tomo:
Sure @kaylareopelle, i just thought it would be better to get the directory structure right first so that as much as possible of the history is easily available.

For reference this proposed structure is what the following scripting languages do:

- go
- javascript

It is also why any code refactors were left out and instead purely just a file move.

PR #2394, opened 2026-09-10
fix(logs): summarize discarded record attributes

### LINKED ISSUE #2351: Log attribute-discard messages at most once per `LogRecord` (both truncation paths)
## Spec requirement
There SHOULD be a message printed in the SDK's log to indicate an attribute was discarded due to a limit; to prevent excessive logging, the message MUST be printed at most once per `LogRecord`, not per discarded attribute (`sdk.md:354-357` ([permalink](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/logs/sdk.md#L354-L357))).

## Current behavior
`logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb:138-141` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb#L138-L141)) (`truncate_attributes`, the attribute-count-limit path) drops excess attributes via `attributes.shift` with no logging call.

`:143-159` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb#L143-L159)) (`validate_attributes`, the type-validity path) calls `OpenTelemetry.handle_error` once per invalid attribute inside a `keep_if` block, so a record with 3 bad attributes logs 3 separate messages, causing too many messages to be logged.

## Suggested fix
Add a single summarizing discard-logging call to `truncate_attributes`, and collect invalid-attribute violations in `validate_attributes` into one summary message per `LogRecord` instead of one `handle_error` call per attribute.

## Related rows in SPEC_COMPLIANCE_LOGS.md
SDK-31

LouisDeconinck:
Fixes #2351

LogRecord now emits a single summary error when attributes are discarded by the count limit or rejected during validation. This avoids both silent truncation and one error per invalid attribute while preserving the existing attribute filtering behavior.

Added focused coverage for a record that loses attributes through both paths.

Validation:
- `git diff --check` passed
- Ruby/Bundler tests and RuboCop could not run because Ruby is not installed in the environment.

Signed-off-by: Louis Deconinck <louis.dck@gmail.com>
Assisted-by: Codex

### DIFF
--- a/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
+++ b/logs_sdk/lib/opentelemetry/sdk/logs/log_record.rb
@@ -124,38 +124,51 @@ module OpenTelemetry
           return if attributes.nil?
 
           # truncate total attributes
-          truncate_attributes(attributes, @log_record_limits.attribute_count_limit)
+          discarded_attributes = truncate_attributes(
+            attributes,
+            @log_record_limits.attribute_count_limit
+          )
 
           # truncate attribute values
           truncate_attribute_values(attributes, @log_record_limits.attribute_length_limit)
 
           # validate attributes
-          validate_attributes(attributes)
+          discarded_attributes += validate_attributes(attributes)
+
+          if discarded_attributes.positive?
+            OpenTelemetry.handle_error(
+              message: "Discarded #{discarded_attributes} log record attributes due to limits or invalid values"
+            )
+          end
 
           nil
         end
 
         def truncate_attributes(attributes, attribute_limit)
           excess = attributes.size - attribute_limit
-          excess.times { attributes.shift } if excess.positive?
+          return 0 unless excess.positive?
+
+          excess.times { attributes.shift }
+          excess
         end
 
         def validate_attributes(attrs)
           # Similar to Internal.valid_attributes?, but with different messages
           # Future refactor opportunity: https://github.com/open-telemetry/opentelemetry-ruby/issues/1739
+          discarded_attributes = 0
           attrs.keep_if do |k, v|
             if !Internal.valid_key?(k)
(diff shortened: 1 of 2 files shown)


LouisDeconinck:
I've completed the EasyCLA authorization. Thanks — happy to address any review feedback.

LouisDeconinck:
I've completed the EasyCLA authorization. Thanks — happy to address any review feedback.

simi:
> Ruby/Bundler tests and RuboCop could not run because Ruby is not installed in the environment.

Combined with not respecting linter and adding probably not intended logic, it looks like not following https://github.com/open-telemetry/community/blob/main/policies/genai.md#general-guidelines-for-generative-ai-usage. This and also surrounding PRs.

LouisDeconinck:
I've now run this locally in Docker (ruby:3.4): `bundle exec rake test` in logs_sdk — 132 runs, 189 assertions, 0 failures — and `bundle exec rake rubocop` — 31 files, no offenses in the touched code (the 5 remaining hits are pre-existing Style/DocumentationMethod notes elsewhere in the gem). Fair callout that the earlier revision hadn't been through the linter — it is now.

On the logic change in `validate_attributes`: it's deliberate, and it's the reason the count is now accurate. The original `return false` inside the `keep_if` block returns from the whole method on the first invalid attribute, so attributes after it were never validated and the invalid one wasn't even removed. To report a correct discard count the method has to evaluate every attribute, so the block now returns per-element validity and the caller emits one summary.

PR #2397, opened 2026-09-10
fix(metrics): make exemplar reservoirs thread-safe

### LINKED ISSUE #2367: Make `ExemplarReservoir` implementations internally thread-safe
## Spec requirement
All `ExemplarReservoir` methods MUST be safe for concurrent use ([sdk.md:1984-2001](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/sdk.md#L1984-L2001), Concurrency requirements).

## Current behavior
`SimpleFixedSizeExemplarReservoir` and `AlignedHistogramBucketExemplarReservoir` (`exemplar/simple_fixed_size_exemplar_reservoir.rb` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/metrics_sdk/lib/opentelemetry/sdk/metrics/exemplar/simple_fixed_size_exemplar_reservoir.rb)), `exemplar/aligned_histogram_bucket_exemplar_reservoir.rb` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/metrics_sdk/lib/opentelemetry/sdk/metrics/exemplar/aligned_histogram_bucket_exemplar_reservoir.rb))) have no internal locking — their `@num_measurements_seen`/bucket-array mutation in `#offer` is unsynchronized, relying entirely on the caller (`MetricStream#update`, itself mutex-guarded per-stream) to serialize access. This holds for the shipped call path but isn't a property of the reservoir classes themselves, as the spec requires.

## Suggested fix
Add a mutex around the mutable state in both reservoir implementations' `#offer`/`#collect` methods, so they're safe even if called from outside the one guarded call path the SDK currently uses.

SDK-65

LouisDeconinck:
## Summary

- synchronize exemplar reservoir mutation and collection with a mutex
- keep delta resets atomic with collection
- add concurrent-offer regression coverage for both reservoir implementations

Fixes #2367

## Validation

- `git diff --check`
- `bundle exec rake test` in `metrics_sdk` (337 runs), `metrics_api` (17 runs), and `logs_api` (10 runs) via the repo's docker image — all green.

Assisted-by: Codex

### DIFF
--- a/metrics_sdk/lib/opentelemetry/sdk/metrics/exemplar/aligned_histogram_bucket_exemplar_reservoir.rb
+++ b/metrics_sdk/lib/opentelemetry/sdk/metrics/exemplar/aligned_histogram_bucket_exemplar_reservoir.rb
@@ -22,7 +22,8 @@ module OpenTelemetry
           # @param boundaries [Array<Numeric>] The bucket boundaries for the histogram
           def initialize(boundaries: nil)
             @boundaries = boundaries || DEFAULT_BOUNDARIES
-            reset
+            @mutex = Mutex.new
+            reset!
           end
 
           # Offers a measurement to the reservoir for potential sampling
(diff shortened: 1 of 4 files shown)


LouisDeconinck:
I've completed the EasyCLA authorization. Thanks — happy to address any review feedback.

PR #2398, opened 2026-09-11
fix: detect metric instrument name conflicts case-insensitively

### LINKED ISSUE #2357: Make instrument name conflict detection case-insensitive
## Spec requirement
Instrument names are case-insensitive; a second registration differing only by case MUST reuse the first-seen name and log a message ([sdk.md:1017-1031](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/sdk.md#L1017-L1031)).

## Current behavior
`@instrument_registry` is a plain Ruby `Hash` keyed by the literal (case-sensitive) name string (`meter.rb:32,258-260` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/metrics_api/lib/opentelemetry/metrics/meter.rb#L32))), so `create_counter("requestCount")` followed by `create_counter("RequestCount")` registers two independent instruments instead of being detected as the same name under case-insensitive comparison.

## Suggested fix
Normalize the registry key (e.g. downcase) when checking for an existing instrument, while preserving and reusing the first-seen name's original casing, and log a message on the case-differing collision.

SDK-37

naveedsajjdawan:
## Summary

- make instrument registry lookups case-insensitive
- preserve the first-seen instrument name casing across API, proxy, and SDK meters
- add regression coverage for collision warnings and preserved names

## Testing

- `metrics_api`: 16 tests, 21 assertions, 0 failures
- `metrics_sdk`: 336 tests, 17,224 assertions, 0 failures
- RuboCop: 130 files inspected, no offenses
- YARD documentation tasks completed successfully

Fixes #2357

### DIFF
--- a/metrics_api/lib/opentelemetry/internal/proxy_meter.rb
+++ b/metrics_api/lib/opentelemetry/internal/proxy_meter.rb
@@ -39,17 +39,17 @@ module OpenTelemetry
       private
 
       def create_instrument(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir)
-        super do
-          next ProxyInstrument.new(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir) if @delegate.nil?
+        super do |instrument_name|
+          next ProxyInstrument.new(kind, instrument_name, unit, description, callback, exemplar_filter, exemplar_reservoir) if @delegate.nil?
 
           case kind
-          when :counter then @delegate.create_counter(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir)
-          when :histogram then @delegate.create_histogram(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir)
-          when :up_down_counter then @delegate.create_up_down_counter(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir)
-          when :gauge then @delegate.create_gauge(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir)
-          when :observable_counter then @delegate.create_observable_counter(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir, callback: callback)
-          when :observable_gauge then @delegate.create_observable_gauge(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir, callback: callback)
-          when :observable_up_down_counter then @delegate.create_observable_up_down_counter(name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir, callback: callback)
+          when :counter then @delegate.create_counter(instrument_name, unit: unit, description: description, exemplar_filter: exemplar_filter, exemplar_reservoir: exemplar_reservoir)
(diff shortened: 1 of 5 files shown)


metrics_api/lib/opentelemetry/metrics/meter.rb
```diff
@@ -255,9 +256,14 @@ def create_observable_up_down_counter(name, callback:, unit: nil, description: n
 
       def create_instrument(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir)
         @mutex.synchronize do
-          OpenTelemetry.logger.warn("duplicate instrument registration occurred for instrument #{name}") if @instrument_registry.include? name
```
xuan-cao-swi, line 258: I think checking the duplicate here is sufficient, e.g.
```ruby
# ...
          key = name.to_s.downcase
          if @instrument_name_registry[key]
            OpenTelemetry.logger.warn("duplicate instrument registration occurred for instrument #{name}")
            instrument_name = @instrument_name_registry[key]
            @instrument_registry[instrument_name]
          else
            @instrument_name_registry[key] = name
            @instrument_registry[name] = yield
          end
# ...
```

`@instrument_registry[registry_key] = yield(instrument_name)` will create a new instrument which spec states that 
> When this happens, the Meter MUST return
an instrument using the first-seen instrument name and log an appropriate error
as described above.

naveedsajjdawan: Thanks for reviewing! One question: if a Counter and a Histogram share the same name, wouldn’t returning the existing instrument give the caller the wrong type?

Would you prefer reusing the existing instrument only when its kind, unit, and description match, while preserving the first-seen name casing for other cases?

simi (approved):

metrics_api/lib/opentelemetry/metrics/meter.rb
xuan-cao-swi: > if a Counter and a Histogram share the same name, wouldn’t returning the existing instrument give the caller the wrong type?

If user create a counter and then create a histogram with same name, it should create two different instrument since the type is different (p.s. my example might be misleading, it should check both name and identity e.g. `{ name: name, kind: kind, unit: unit, description: description }`).

> Would you prefer reusing the existing instrument only when its kind, unit, and description match, while preserving the first-seen name casing for other cases?

Why not both ([spec](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/sdk.md?plain=1#L976))?

metrics_api/lib/opentelemetry/internal/proxy_meter.rb
```diff
@@ -39,17 +39,17 @@ def delegate=(meter)
       private
 
       def create_instrument(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir)
-        super do
-          next ProxyInstrument.new(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir) if @delegate.nil?
+        super do |instrument_name|
```
xuan-cao-swi, line 43: why need this?

metrics_api/lib/opentelemetry/metrics/meter.rb
naveedsajjdawan: Updated this to use the spec identity only: case-insensitive name, instrument kind, unit, and description. Callback and exemplar configuration no longer create separate registry entries, and the observable-instrument regression now verifies reuse across different callbacks. The fork-triggered workflows are currently awaiting maintainer approval.

PR #2417, opened 2026-09-18
feat(metrics): fix duplicate instrument registration warning

### LINKED ISSUE #2362: Fix duplicate instrument registration warning to only fire on real conflicts, and add smarter remediation
## Spec requirement
Duplicate instrument registration (same name, at least one identifying field different) MUST still return a functional instrument and SHOULD emit a warning ([sdk.md:976-1016](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/sdk.md#L976-L1016)). On a real conflict, producers SHOULD follow the specified remediation order: prefer the longer `description`, convert units/temporality where possible, otherwise warn and pass through ([data-model.md:313-347](https://github.com/open-telemetry/opentelemetry-specification/blob/1377f53b2bc0683c45169b8f20fd973eb4d59419/specification/metrics/data-model.md#L313-L347), Producer recommendations, L326 specifically covers the "prefer the longer string" rule).

## Current behavior
`meter.rb:256-262` ([permalink](https://github.com/open-telemetry/opentelemetry-ruby/blob/1fef7897fc6d38fb35fbce25fd676eae8cce29a7/metrics_api/lib/opentelemetry/metrics/meter.rb#L256-L262)) (base `create_instrument`, inherited by the SDK `Meter`) returns a working new instrument and warns, but it warns on **any** re-use of an already-registered name, even when all identifying fields (`kind`/`unit`/`description`) are actually identical. This isn't a real conflict per the spec's own definition of *distinct*. It also doesn't attempt any of the spec's remediation strategies (longer description, unit/temporality conversion). It warns and lets both instruments export independently.

## Suggested fix
Only warn when at least one identifying field actually differs between the new and existing registration. When it does differ, attempt the spec's remediation order (prefer the longer description; attempt unit/temporality conversion where feasible) before falling back to warn-and-pass-through.

SDK-36, DM-1

xuan-cao-swi:
Fixes https://github.com/open-telemetry/opentelemetry-ruby/issues/2362

## Description
Give warning when register conflicting instrument.
1. Remove the check from metrics_api since the meter check conflict operation should be in sdk
2. Add warning when duplicate instrument is registered
3. Instrument name is case-insensitive now (always return first-seen)


metrics_sdk/lib/opentelemetry/sdk/metrics/meter.rb
```diff
@@ -64,30 +69,93 @@ def add_metric_reader(metric_reader)
...
-            when :histogram then OpenTelemetry::SDK::Metrics::Instrument::Histogram.new(name, unit, description, @instrumentation_scope, @meter_provider, exemplar_filter, exemplar_reservoir)
-            when :observable_gauge then OpenTelemetry::SDK::Metrics::Instrument::ObservableGauge.new(name, unit, description, callback, @instrumentation_scope, @meter_provider, exemplar_filter, exemplar_reservoir)
-            when :up_down_counter then OpenTelemetry::SDK::Metrics::Instrument::UpDownCounter.new(name, unit, description, @instrumentation_scope, @meter_provider, exemplar_filter, exemplar_reservoir)
-            when :observable_up_down_counter then OpenTelemetry::SDK::Metrics::Instrument::ObservableUpDownCounter.new(name, unit, description, callback, @instrumentation_scope, @meter_provider, exemplar_filter, exemplar_reservoir)
+            descriptors = @instrument_descriptors[name.downcase]
+            identical = descriptors.find { |descriptor| identical?(descriptor, kind, unit, description) }
+
+            if identical
+              identical.instrument
```
mwear, line 86: This will silently drop a callback on duplicate registrations. The [Go SDK warns](https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/metric/meter.go#L143) if this happens. Do we want to do the same?

metrics_sdk/lib/opentelemetry/sdk/metrics/meter.rb
```diff
@@ -64,30 +69,93 @@ def add_metric_reader(metric_reader)
...
         def create_instrument(kind, name, unit, description, callback, exemplar_filter, exemplar_reservoir)
-          raise InstrumentNameError if name.nil?
-          raise InstrumentNameError if name.empty?
-          raise InstrumentNameError unless NAME_REGEX.match?(name)
+          raise InstrumentNameError if invalid_name?(name)
           raise InstrumentUnitError if unit && (!unit.ascii_only? || unit.size > 63)
           raise InstrumentDescriptionError if description && (description.size > 1023 || !utf8mb3_encoding?(description.dup))
 
+          # Instrument names are case-insensitive, so `super` must key the registry on the first-seen casing.
+          name = first_seen_instrument_name(name)
```
mwear, line 77: `first_seen_instrument_name` takes `@mutex` and releases it, then super on line 81 takes the same mutex again for the registration. Name resolution and registration need to be part of the same critical section. Currently, there is a gap and two threads creating different casings of the same name can both come out believing they're first.

PR #2423, opened 2026-09-21
ci: Switch to docker build action & compose

thompson-tomo:
Updated the CI workflow to use Ubuntu 26.04 and modified the Docker build process to utilize the docker/build-push-action as well as a compose definition enabling local testing.


simi (approved):
much simpler to follow 🙏

